PR #40, opened 2019-12-13
Adds a check that clients are compatible with new versions of FDB before upgrading.

### LINKED ISSUE #5: Checking client compatibility during upgrades
When we do upgrades, we need to check that the clients are compatible with the new version before pushing out the new server version. We should build this logic into the operator.

brownleej:
Resolves #5


pkg/controller/foundationdbcluster/check_client_compatibility.go
```diff
@@ -0,0 +1,93 @@
...
+				clientsSupported[description] = false
+			}
+		}
+	}
+	unsupportedClients := make([]string, 0, len(clientsSupported))
+	for client, supported := range clientsSupported {
+		if !supported {
+			unsupportedClients = append(unsupportedClients, client)
+		}
+	}
```
JoshuaMcManus: You can probably skip making the map and combine these two parts together to just get the unsupported clients.

brownleej: I think it would be harder to try and construct the array in a single pass. We might see a client on an unsupported version and then see it under a supported version later. If we were building the array as we went, we'd have to remove it from the array when we found an unsupported version, which Go doesn't have a simple method for. I think it would also be slower, since removing an item from an array is O(n).

PR #59, opened 2019-12-24
This PR adds support for having helm chart of foundationDB

### LINKED ISSUE #49: Add helm chart to deploy fdb operator

viveksinghggits:
Fixes #49 

To test the chart please follow below commands 

```
# create namespace 
kubectl create ns fdb-test

# create operator
helm install fdb-operator helm/fdb-operator -n fdb-test
```
Once we have have the operator running we can go ahead and create the foundation db instance using below command
```
helm install fdb-instance helm/fdb-instance -n fdb-test
```
and you will have foundationDB pod running in the `fdb-test` namespace with double redundancy mode.

**Note**
we will have to push the operator image to a public repository so that the image name can be used in the `values.yaml` for `fdb-operator` chart.


brownleej:
Sorry for the delay. Apple's on holiday from the 24th through the 1st, so I haven't been looking at PRs. I'll try to take a look at this when I get back next week.

helm/fdb-instance/templates/local_cluster.yaml
```diff
@@ -0,0 +1,57 @@
+apiVersion: apps.foundationdb.org/v1beta1
```
brownleej: The local_cluster file is designed for local development, so I don't think we want to use it as part of a helm chart. Most of the things that are defined in this sample are things that aren't as likely to be good fits for real clusters. In general, I'm not sure how much value there is having a helm chart for the CRD, rather than just having the customization go through the CRD.

helm/fdb-operator/templates/crds/foundationdbcluster.yaml
```diff
@@ -0,0 +1,477 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
brownleej: Will this file updated automatically when we make changes to the CRD?

helm/fdb-operator/templates/manager/manager.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    control-plane: controller-manager
+    controller-tools.k8s.io: "1.0"
+  name: system
```
brownleej: Is it the best practice to create a namespace when deploying things through Helm, or is this an artifact of the defaults from kubebuilder?

helm/fdb-operator/templates/manager/secret.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ .Values.webhook.secret.name }}
+  namespace: {{ .Release.Namespace }}
```
brownleej: Are we doing anything with this webhook secret? I think it came from the defaults in kubebuilder, but we don't have any webhooks right now. We may add some in the future, though.

helm/fdb-operator/templates/manager/statefulset.yaml
```diff
@@ -0,0 +1,66 @@
...
+    metadata:
+      labels:
+        {{- include "chart.labels" . | nindent 8 }}
+    spec:
+      volumes:
+        - name: cert
+          secret:
+            defaultMode: 420
+            secretName: {{ .Values.webhook.secret.name }}
+        - name: fdb-certs
```
brownleej: The `fdb-certs` secret is part of the local development environment, but I don't think we want it in the helm chart.

helm/fdb-operator/templates/manager/statefulset.yaml
```diff
@@ -0,0 +1,66 @@
...
+            name: fdb-certs
+            readOnly: true
+        env:
+          - name: POD_NAMESPACE
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.namespace
+          - name: SECRET_NAME
+            value: {{ .Values.secret.name }}
+          - name: DISABLE_SIDECAR_TLS_CHECK
```
brownleej: This shouldn't be necessary outside of local development environments.

helm/fdb-operator/templates/rbac/rbac_role.yaml
```diff
@@ -0,0 +1,77 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
brownleej: I think it's better to use a role rather than a  cluster role, to support using the operator in environments where the person running the operator doesn't have cluster-wide access.

helm/fdb-instance/templates/local_cluster.yaml
viveksinghggits: Hi @brownleej 
It totally depends on us, is we want to get the CR,which in our case is a FoundationDBCluster, created as part of helm or not. By in my experience some database use this approach where they install the operator through one helm installation and actual database (CR) through another.
What I suggest is to install the CR using second helm deployment and improve that for real use. WDYT

helm/fdb-operator/templates/crds/foundationdbcluster.yaml
viveksinghggits: I am also not an expert here but I think we will have to make the changes manually.

helm/fdb-operator/templates/manager/manager.yaml
viveksinghggits: If we talk specifically this namespace creation its not required.

helm/fdb-operator/templates/manager/secret.yaml
viveksinghggits: I agree, I just tried to make helm chart of whatever was already present, if we dont use this, I think it can be removed.

helm/fdb-operator/templates/manager/statefulset.yaml
viveksinghggits: if you see the statefulset that gets created for the operator, manifest of that statefulset has this secret mounted, so I just made that more manageable but having the name in the values.yaml file, but it was already in the statefulset's manifest.

helm/fdb-operator/templates/manager/statefulset.yaml
viveksinghggits: a comment above addresses this.

helm/fdb-operator/templates/rbac/rbac_role.yaml
viveksinghggits: I agree, we should have `role` in place of ClusterRole here.

viveksinghggits:
Hi @brownleej 
can you please look into/address the comments.

viveksinghggits:
Hi @brownleej 
I have made the changes that were about removing the `system` namespace  and the fdb-instance chart.
However I suggest that we should have instance chart here as well.

helm/fdb-operator/templates/crds/foundationdbcluster.yaml
brownleej: In all the other places we use it, we generate the CRD from the code. Is there a reason we can't do that with Helm? I'd rather not introduce any way for things to get out of sync if we can avoid it.

helm/fdb-operator/templates/manager/statefulset.yaml
brownleej: We should base this on the latest sample deployment YAML (https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/config/samples/deployment.yaml) after the Kubebuilder 2 upgrade.

brownleej:
What problem does the instance chart solve?

viveksinghggits:
> What problem does the instance chart solve?

Hi @brownleej 
So, the operator-chart creates the CRD and the instance chart actually creates a CR that run the foundationdb cluster in double redundancy mode. So if we don't deploy the instance chart we wont have foundationdb cluster up and running. we will anyway have to deploy the instance chart separately if we dont include that in this repo with operator chart.

brownleej:
What is the advantage of creating the cluster through Helm rather than directly?

viveksinghggits:
> What is the advantage of creating the cluster through Helm rather than directly?

Hi @brownleej, 
Here is one of the many reasons why we use helm or any other package manager to deploy the resources on Kubernetes-
Right now if we see the [resource manifest](https://raw.githubusercontent.com/foundationdb/fdb-kubernetes-operator/master/config/samples/cluster_local.yaml) for the CR `FoundationDBCluster` that spins up the FoundationDB cluster we dont actually know which redundancy mode the cluster will be started in. Even if we know, and want to change that we will have to checkout that file and then make some modifications in that file to get the redundancy mode reflected.
Now if we had created the same resource using helm, we will read the `redundancy_mode` from `values.yaml` file and we can change any value in `values.yaml` file using `--set` flag while installing helm chart.

viveksinghggits:
Hi @brownleej 
Did you get time to look into the comments.?

brownleej:
I see you've removed the instance chart; I had missed that before. That clears up my biggest area of concern. There are still some open questions and suggestions in the PR that I would like to see resolved. A lot of that comes down to bringing this in line with other things that have changed since you opened the PR, such as moving to a deployment rather than a statefulset for the operator and removing defaults that Kubebuilder added that we're not using now. I also have a question about how the CRD in the helm chart gets defined.

viveksinghggits:
Hi @brownleej,
I have addressed the comment, can you please have another look.
This is how we can install the chart

```
~   helm install fdb helm/fdb-operator/    

NAME: fdb
LAST DEPLOYED: Wed Feb 12 18:25:11 2020
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
FoundationDB operator has been installed successfully.

To see the logs of the operator you can use below command

kubectl logs deployment/fdb-kubernetes-operator-controller-manager -n <namespace-name> -f

```

brownleej:
@viveksinghggits What are your thoughts on generating the CRD in the helm chart from the code so it automatically updates?

viveksinghggits:
Hey @brownleej,
If I understood you correctly, you want the CRD manifest that is there as part of helm chart to be updated automatically according to the code change. I am not very familiar how we can achieve that, per my understanding as of now, we can configure the helm chart manifest using only the `Values.yaml` files. But let me ask around my team once and update.

brownleej:
Why can't we have a step in the Makefile that updates the copy of the CRD in the Helm chart?

viveksinghggits:
Hi @brownleej,
The reason why we want to introduce helm chart is to not have anything other than 
```
helm install fdb helm/fdb-operator/
```
if we are including Makfile in between, we are not actually resolving the problem that we are trying to resolve by introducing Helm installation.

On the other note, can you please explain what do you actually mean by update the copy of CRD, because in that also we will have to have the CRD somewhere right?

brownleej:
What I mean is that when the CRD changes, like when we add a new field, we should update the copy in the helm chart in addition to updating the copy in `config/crd`.

viveksinghggits:
Hey @brownleej,
What I suggest to do is store that CRD at only one place, inside `helm/fdb-operator/templates/crds`. Will that be problem in any way?

brownleej:
I suspect it will be a problem, since it means removing the CRD from the standard location in config/crd.

viveksinghggits:
Hey @brownleej,
I have created a soft link to the CRD file that is in the config dir, now even if we change the content of the file in the cofig dir, helm would be updated.

We will have to discuss about two things now, how do we host this helm chart, one option I know id github pages.
How are we going to support the CR installation.

brownleej:
This looks good. I think the best course is to merge this in as it is and file smaller-scoped issues for the things we don't have in this PR.

viveksinghggits:
Hey @brownleej,
The first thing I think we should focus on host this chart. We can easily use github pages to host this chart.

PR #60, opened 2020-01-02
Change the way we generate instance IDs to remove the need to track the next instance ID in the spec.

### LINKED ISSUE #51: New strategy for generating pod names and instance IDs
We are currently generating instance IDs using an incrementing counter, which is stored in the cluster spec. This means that when we do replacements or delete stateless pods, we are moving to a higher set of instance IDs. I have an idea for an alternative strategy that will keep the pod names bounded to a smaller set and remove the need to track instance IDs in the spec.

I propose that we make the instance IDs take the form `${process_class}-${num}`. The corresponding pod name would be `${cluster_name}-${process_class}-${num}`. When adding new pods, we would first look for any PVCs that match the cluster and process class and do not have a matching pod, and we would try to re-use them. If we have no more re-usable PVCs, we would look for the smallest number `num` greater than 0 that does not have a corresponding pod for this process class, and would create a pod for that number.

When we replace instances by adding them to the removal list in the spec, this strategy will create a gap in the instance IDs. This means that unlike with StatefulSets, there is not a direct correspondence between the highest pod name and the number of instances. I think this will not be a problem for us, and it's necessary to preserve the ability to do live replacements. The next time we need to expand or replace instances, however, we will fill in these gaps, so there is a bound on how high the instance IDs will get based on the size of the cluster.

This strategy should be resilient against stale results from the cache in the operator. If we see a gap in instance IDs where it doesn't exist, then our attempt to create the pod will fail. If we fail to see a gap in instance IDs where it does exist, then a subsequent operation will eventually see the gap and fill it.

Another benefit of this naming convention is that it is clear from the pod name what role a pod is filling.

brownleej:
Resolves #51

### DIFF
--- a/config/crds/apps_v1beta1_foundationdbcluster.yaml
+++ b/config/crds/apps_v1beta1_foundationdbcluster.yaml
@@ -195,8 +195,8 @@ spec:
                   type: array
               type: object
             nextInstanceID:
-              description: NextInstanceID defines the ID to use when creating the
-                next instance.
+              description: 'NextInstanceID defines the ID to use when creating the
+                next instance.  Deprecated: This is no longer used.'
               format: int64
               type: integer
             pendingRemovals:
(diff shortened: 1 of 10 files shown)


brownleej:
I found an issue with the handling of the cluster_controller process type, so this isn't ready for merge yet.

PR #105, opened 2020-01-29
Auto-generate API docs

### LINKED ISSUE #101: Generating better documentation on the cluster resource
We have documentation in the code for all of the fields in the cluster resource, but this documentation isn't easy for people to discover or browse. We should look for a way to automatically generate documentation on the cluster resource in a more human-readable format, similar to what the Kubernetes core resources have in their API docs.

brownleej:
Resolves #101

### DIFF
--- a/docs/cluster_spec.md
+++ b/docs/cluster_spec.md
@@ -68,10 +68,10 @@ ContainerOverrides provides options for customizing a container created by the o
 | ----- | ----------- | ------ | -------- |
 | enableTls | EnableTLS controls whether we should be listening on a TLS connection. | bool | false |
 | peerVerificationRules | PeerVerificationRules provides the rules for what client certificates the process should accept. | string | false |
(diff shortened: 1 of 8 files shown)


docs/cluster_spec.md
```diff
@@ -0,0 +1,444 @@
...
+
+## FdbVersion
+
+FdbVersion represents a version of FoundationDB.  This provides convenience methods for checking features available in different versions.
+
+| Field | Description | Scheme | Required |
+| ----- | ----------- | ------ | -------- |
+| Major | Major is the major version | int | false |
+| Minor | Minor is the major version | int | false |
+| Patch | Patch is the major version | int | false |
```
JoshuaMcManus: Looks like the documentation has some copy/paste errors. It might be worth fixing it in the external file

JoshuaMcManus: Otherwise everything looks good and ready to merge

Issue #82, opened 2020-01-11
Customizing the items in the config map


brownleej:
We should allow people to add custom items to the config map that the operator creates. We can use the items field in the config map in the cluster spec to specify additional items, and allow mounting them in the pod by specifying the items in the volume source for the config-map volume.

PR #148, opened 2020-02-18
adding items to config map in the cluster spec

akshaychitneni:
fix for https://github.com/FoundationDB/fdb-kubernetes-operator/issues/82

### DIFF
--- a/controllers/controller.go
+++ b/controllers/controller.go
@@ -401,6 +401,14 @@ func GetConfigMap(context ctx.Context, cluster *fdbtypes.FoundationDBCluster, ku
 	}
 	data["sidecar-conf"] = string(sidecarConfData)
 
+	if cluster.Spec.ConfigMap != nil && len(cluster.Spec.ConfigMap.Data) > 0 {
+		items, err := json.Marshal(cluster.Spec.ConfigMap.Data)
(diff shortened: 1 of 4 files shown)


brownleej:
It looks like there are some formatting issues. @akshaychitneni  Can you run `make fmt` on your branch?

controllers/controller.go
```diff
@@ -401,6 +401,14 @@ func GetConfigMap(context ctx.Context, cluster *fdbtypes.FoundationDBCluster, ku
 	}
 	data["sidecar-conf"] = string(sidecarConfData)
 
+	if cluster.Spec.ConfigMap != nil && len(cluster.Spec.ConfigMap.Data) > 0 {
+		items, err := json.Marshal(cluster.Spec.ConfigMap.Data)
+		if err != nil {
+			return nil, err
+		}
+		data["items"] = string(items)
```
brownleej: Instead of putting these in a special "items" key, can we add them directly into the ConfigMap? For instance, in the test case below, I think that the "itemVal" value should be in `configMap.Data["itemKey"]`.

akshaychitneni: Done. thanks

PR #157, opened 2020-02-29
Update controller-gen to 0.2.5

brownleej:


brownleej:
It looks like this is incompatible with older versions of Kubernetes so we may need to defer this.

PR #184, opened 2020-03-23
Starting backups

### LINKED ISSUE #117: Starting backups
As part of our work on managing backups, we should allow users to define their desired backup state as part of the backup spec.

The backup spec should have a `state` field, which can be either `Running`, or the empty string. These two options will have the same meaning for now, but we will add more states in subsequent issues. The operator should ensure that there is a backup running on the `default` tag.

We also need configuration items in the spec for the account name and the backup name. The account name is required. The backup name is optional. If the backup name is empty, we should use the name from the backup object.

The backups should always be started in continuous mode.

### LINKED ISSUE #152: Making a secure connection to MinIO
In our local examples, we are using an insecure connection to MinIO. This probably isn't a good representation of how people will be configuring backups in real environments, and may set a bad example. We should find a way to use a secure connection in the local environment.

brownleej:
Resolves  #117
Resolves #152


api/v1beta1/foundationdbcluster_types.go
```diff
@@ -1783,7 +1811,24 @@ type FoundationDBBackupSpec struct {
 	// The cluster this backup is for.
 	ClusterName string `json:"clusterName"`
 
+	// +kubebuilder:validation:Enum=Running
+	// The desired state of the backup.
+	// The default is Running.
+	BackupState string `json:"backupState,omitempty"`
```
JoshuaMcManus: Can we make this an enum instead of hardcoded strings for state? Especially once we add mode states such as stopped and paused.

brownleej: The closest we can get, I think, is to define a type called `BackupState` that is aliased to `string`, and then define string constants that represent the known states. This wouldn't restrict the possible values, though, and it would just turn the comparisons to strings we're making in the code into comparisons to constants for those strings. There's definitely some advantages there, but it's also more code, and I don't know if it's worth it. The other downside is that we would have to define the list of backup states both in that new list of constants and in the kubebuilder validations.

The number of states is going to stay small, and we're only going to be doing comparisons against them in the methods like `ShouldRun`, so I think having them stay as strings is going to be manageable.

PR #185, opened 2020-03-30
Subreconciler tests

### LINKED ISSUE #112: Adding more tests for subreconcilers
We have tests for the reconciliation flow in the operator, but there are cases that are hard to reproduce when doing an end-to-end reconciliation in these tests. I think we should add more tests for the individual subreconciliation steps that can focus more on the cases each step has to handle.

brownleej:
Resolves #112 

This is a big one, but it's mostly in tests. I've also updated some non-test code to fix problems and inconsistencies I found while expanding the tests.

### DIFF
--- a/controllers/add_pods_test.go
+++ b/controllers/add_pods_test.go
@@ -32,7 +32,7 @@ import (
 	fdbtypes "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta1"
 )
 
-var _ = Describe("add_pods_test", func() {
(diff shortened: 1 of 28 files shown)


brownleej:
I may have just gotten lucky when I tested this before posting the PR, because I can't consistently get it to pass the test suite now. I'm thinking that this is ultimately us running against the testability limits of interacting with the API server. We might want to explore having our own mock implementation at some point to improve testability, but I think this PR is way too much trouble for the value it's delivering, so I'm going to drop it for now.

PR #189, opened 2020-03-31
Do not override environment variables set in manifest.

matiascognite:


matiascognite:
Weird, it compiles and runs fine on my machine, and solves the problem. But feel free to change it

brownleej:
Can you check this against the latest master branch? I think we fixed this in a PR that got merged earlier today.

matiascognite:
@brownleej yup, it's fixed 👍 Do you know when the new release will come?

PR #209, opened 2020-04-28
Defining the backup snapshot time

### LINKED ISSUE #180: Customizing the backup snapshot time
FDB allows customizing the backup snapshot time, so that users can balance the retention requirements and the bandwidth they have available for writing range data. We should allow customizing this through the operator. The default snapshot time should be 10 days, matching the default in FDB itself. If this field is changed, we should use the `fdbbackup modify` command to update the current backup.

Extracted from #117

brownleej:
Closes #180


controllers/admin_client.go
```diff
@@ -444,6 +452,42 @@ func (client *CliAdminClient) ResumeBackups() error {
 	return err
 }
 
+// ModifyBackup updates the backup aparameters.
```
JoshuaMcManus, line 455:
```suggestion
// ModifyBackup updates the backup parameters.
```

PR #256, opened 2020-06-16
Leave the IP address empty in the choose_removals step when the pod does not have an IP.

### LINKED ISSUE #246: Reconciliation getting stuck when replacing a pod without an IP
If a pod is stuck waiting to be scheduled, we want to be able to replace it to get the cluster back to a healthy state. If we try to replace it through the operator, it will get stuck at the ChooseRemovals step. I think the reason it's getting stuck is because this step needs to associate the pod name with an IP address for the exclusions, and that's not possible until the pod is scheduled.

I think if the pod is unscheduled the best course of action is to ignore it for the purposes of exclusions. Eventually we will be able to exclude instances by instance ID, which will be safer than using IPs and will avoid this whole problem. In the mean time, we need to be able to get out of these kind of bad situations.

We may be able to add an additional check where we look in the status JSON for any pod that matches the pod name, but we'll need to make sure we can handle pods in different KCs that may have the same pod name.

brownleej:
Resolves #246

### DIFF
--- a/controllers/admin_client.go
+++ b/controllers/admin_client.go
@@ -25,6 +25,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"os"
 	"os/exec"
 	"regexp"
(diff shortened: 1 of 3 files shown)


controllers/admin_client.go
```diff
@@ -695,6 +696,10 @@ func (client *MockAdminClient) ExcludeInstances(addresses []string) error {
 	exclusionMap := make(map[string]bool, count)
 	newExclusions := make([]string, 0, count)
 	for _, address := range addresses {
+		if !isValidAddress(address) {
```
rbtcollins, line 699: I realise I'm new to this code base, but these mocks seem very heavy - using testify mock would give finer control and avoid duplicating business logic that exists within fdbcli. I'd be happy to whip up an example patch for you if you'd like.

brownleej: My personal preference is for more heavyweight mocks that give us a more realistic simulation of the service that we're stubbing out.

PR #257, opened 2020-06-16
Allow specifying custom labels on backup deployments.

### LINKED ISSUE #165: Defining custom labels for backup deployments
The operator puts labels on the backup deployment to allow the operator itself to find and manage it, but it doesn't allow user-defined labels. We should add an option for this in the backup spec.

brownleej:
Resolves #165

### DIFF
--- a/api/v1beta1/foundationdbcluster_types.go
+++ b/api/v1beta1/foundationdbcluster_types.go
@@ -2130,9 +2130,9 @@ type FoundationDBBackupSpec struct {
 	// This is measured in seconds. The default is 864,000, or 10 days.
 	SnapshotPeriodSeconds *int `json:"snapshotPeriodSeconds,omitempty"`
 
-	// DeploymentMetadata allows customizing labels and annotations on the
+	// BackupDeploymentMetadata allows customizing labels and annotations on the
 	// deployment for the backup agents.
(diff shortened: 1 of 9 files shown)


api/v1beta1/foundationdbcluster_types.go
```diff
@@ -2023,6 +2023,10 @@ type FoundationDBBackupSpec struct {
 	// This is measured in seconds. The default is 864,000, or 10 days.
 	SnapshotPeriodSeconds *int `json:"snapshotPeriodSeconds,omitempty"`
 
+	// DeploymentMetadata allows customizing labels and annotations on the
+	// deployment for the backup agents.
+	DeploymentMetadata *metav1.ObjectMeta `json:"deploymentMetadata,omitempty"`
```
JoshuaMcManus, line 2028: are we able to make this more generic to apply to other agents as well? Or at the very least can we rename it to something other than "deploymentMetadata" as it doesn't really make it clear that it's specific to the backup deployment.

JoshuaMcManus: Yeah, that's fine. I can't think of any other deployments it would apply to either. I still think it should be renamed to backupMetadata or similar just so it's more clear

PR #260, opened 2020-06-16
Ignore PVCs that are not owned by the cluster when checking for misconfigured pods.

### LINKED ISSUE #232: replace_misconfigured_pods not checking volume ownership
The replace_misconfigured_pods step in reconciliation looks for existing PVCs that do not match the latest PVC spec and replaces those instances. This step can pick up PVCs that have labels that match the cluster, but are not owned by the cluster. When this happens, it can cause the operator to hit an error that blocks reconciliation, because it cannot parse the instance IDs from the PVCs. I think that we should ignore PVCs that are not owned by the cluster resource, as we do in other steps.

brownleej:
Resolves #232

### DIFF
--- a/controllers/replace_misconfigured_pods.go
+++ b/controllers/replace_misconfigured_pods.go
@@ -49,6 +49,18 @@ func (c ReplaceMisconfiguredPods) Reconcile(r *FoundationDBClusterReconciler, co
 	}
 
 	for _, pvc := range pvcs.Items {
+		ownedByCluster := false
+		for _, ownerReference := range pvc.OwnerReferences {
+			if ownerReference.UID == cluster.UID {
+				ownedByCluster = true
+				break
+			}
+		}
+
+		if !ownedByCluster {
+			continue
+		}
+
 		instanceID := GetInstanceIDFromMeta(pvc.ObjectMeta)
 		if instancesToRemove[instanceID] {
 			continue


rbtcollins:
mmm, I'm torn on this; with deployments / replicasets / pods the rule is an IFF rule for labels: that is if the labels match for an arbitrary pod, it is managed by the deployment, and the ownerreference really only turns up for gc purposes: it isn't enough to be owned, and it is enough to have the labels. 

This permits, for instance, deleting the deployment with recursive delete *disabled*, then recreating it, and the recreated deployment takes over the pods preserved by the non-recursive delete.

Following that same approach here would give the current code, with the corner case where someone sets up a PVC with matching labels and different or missing ownerreference as a potential recovery strategy. 

OTOH I do understand the idea of just making everything hermetic and self-contained, and with the current limitations around IP addresses and coordinator updates, perhaps thats best. Certainly the code you've written looks fit for purpose with that approach; though perhaps a test to capture the new behaviour would be worth doing?

brownleej:
Yeah, maybe a better approach would be to specify matching labels that we use to determine which resources are owned by the operator. This PR is following the pattern we've used in other parts of the system, such as when adding and removing pods, so I think it's worth merging as-is. We should consider whether there's a better approach and implement that across the board.

brownleej:
I'll file another issue about @rbtcollins 's suggestion to use match labels to identify which pods the operator should manage.

PR #261, opened 2020-06-16
Add a test for replacing a pod while re-using its volume

### LINKED ISSUE #3: Add tests for replacing pod while reusing its volume
The Kubernetes operator has support for deleting a pod and recreating it with the same volume. When I initially implemented with this I hit challenges creating unit tests for this scenario. We should revisit this and add tests for this case.

brownleej:
Resolves #3

### DIFF
--- a/controllers/cluster_controller_test.go
+++ b/controllers/cluster_controller_test.go
@@ -528,6 +528,35 @@ var _ = Describe("cluster_controller", func() {
 			})
 		})
 
+		Context("with a pod that gets deleted", func() {
+			var pod corev1.Pod
+			BeforeEach(func() {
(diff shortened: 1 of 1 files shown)


rbtcollins:
I don't see anything in the test that validates that the storage *isn't* deleted : perhaps I'm missing something though. AFAICT you test that you can delete the pod directly; that the pod gets replaced, but you don't test that no other side effects occur.

brownleej:
I don't know how we can test that within the unit tests. Maybe we can check that the UID of the PVC doesn't change, though that doesn't strictly prove that the data is intact. We may also want to build out an integration test suite to check for things like data integrity.

Issue #11, opened 2019-11-19
Using non-blocking excludes in the Kubernetes operator


brownleej:
When we do excludes in the operator, we are running the exclusion command with a short timeout, and then re-running the reconciliation until we can get the exclusion command to succeed. There are risks in relying on timeouts to determine if an operation has completed, since we can get timeouts for other reasons and this process can make it harder to debug those issues.  Triggering an exclude can also restart data distribution, which can block progress on the excludes, getting us caught in a loop. Instead, I think we should be doing a non-blocking exclude for any new additions to the exclusion list, and use a separate mechanism to check if the exclusion has finished. That mechanism does not exist yet, but there's an issue for it on the FoundationDB repo: https://github.com/apple/foundationdb/issues/2159

brownleej:
As part of this, we should move the step to change coordinators to be between starting the exclusion and verifying that the exclusion has finished. This should remove the need to check the pending removals when changing coordinators, and give us better behavior when replacing instances in multiple Kubernetes clusters at the same time.

brownleej:
I think that the issue with the exclude restarting data distribution is not going to change the interface from what we have in 6.3, so we can move forward with this now. We'll want to only enable this in 6.3, so we can rely on the new output in https://github.com/apple/foundationdb/pull/2870.

PR #295, opened 2020-07-27
Support for non-blocking excludes

brownleej:
This addresses part of #11, but leaves the non-blocking exclude path disabled for now. In the course of testing it, I found a bug in the non-blocking exclude logic in FDB itself, so we're not ready to enable it yet. The code in this PR is otherwise functional, so I think we should get it merged in rather then letting it wait for a patch to FDB.


controllers/cluster_controller.go
```diff
@@ -114,8 +114,9 @@ func (r *FoundationDBClusterReconciler) Reconcile(request ctrl.Request) (ctrl.Re
 		UpdateLabels{},
 		UpdateDatabaseConfiguration{},
 		ChooseRemovals{},
-		ChangeCoordinators{},
 		ExcludeInstances{},
+		ChangeCoordinators{},
+		ConfirmExclusionCompletion{},
 		BounceProcesses{},
```
JoshuaMcManus, line 120: Is this safe to do if we're excluding a coordinator?

brownleej: Yes, it should be safe to start the exclusion while the process is still a coordinator. We just need to make sure we change coordinators before removing the process, which this sequence will accomplish.

PR #296, opened 2020-07-27
Prevent executing a bounce when the minimum process uptime is below a certain limit.

### LINKED ISSUE #2: Protections against bouncing processes repeatedly
The Kubernetes operator bounces the processes under its control when it updates the start commands for the processes. There may be situations where it does multiple bounces close together, especially when someone is deploying across Kubernetes clusters and has multiple instances of the operator working on the same FDB cluster. We should try to add protections against this, like refusing to bounce when there has been a recovery recently.

brownleej:
Resolves #2

### DIFF
--- a/api/v1beta1/foundationdbcluster_types_test.go
+++ b/api/v1beta1/foundationdbcluster_types_test.go
@@ -496,7 +496,8 @@ func TestParsingClusterStatusWithSixOneCluster(t *testing.T) {
 						"processid":   "c813e585043a7ab55a4905f465c4aa52",
 						"zoneid":      "sample-cluster-storage-3",
 					},
-					Version: "6.1.12",
+					Version:       "6.1.12",
+					UptimeSeconds: 160.009,
 				},
 				"f9efa90fc104f4e277b140baf89aab66": {
 					Address:      "10.1.38.82:4501",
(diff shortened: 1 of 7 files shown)


api/v1beta1/foundationdbcluster_types_test.go
```diff
@@ -496,7 +496,8 @@ func TestParsingClusterStatusWithSixOneCluster(t *testing.T) {
 						"processid":   "c813e585043a7ab55a4905f465c4aa52",
 						"zoneid":      "sample-cluster-storage-3",
 					},
-					Version: "6.1.12",
+					Version:       "6.1.12",
+					UptimeSeconds: 160.009,
```
JoshuaMcManus, line 500: Is this uptimeSeconds value going to change between tests? Or if we add other settings/processing in our methods? I want to make sure this isn't going to be a brittle test case.

brownleej: Much of the contents of the status can change if we take a fresh status JSON, including the process IDs. However, these test cases are based off of status JSON output from live clusters, which we keep in the repository, and we shouldn't have any need to update the JSON file for an existing test case. When we add support for new versions, or possibly some new features or configurations, we'll add new test cases and create new JSON files to support those test cases.

PR #297, opened 2020-07-28
Replace any instances that have an instance ID with the wrong prefix.

### LINKED ISSUE #242: Replacing instances when changing the instance ID prefix
When we update the instance ID prefix, we're left with older instances that don't have the correct prefix. I think that the best way to fulfill the user's intention is to replace any instances that have a prefix in their instance ID that does not match the one specified.

brownleej:
Resolves #242

### DIFF
--- a/controllers/replace_misconfigured_pods.go
+++ b/controllers/replace_misconfigured_pods.go
@@ -119,7 +119,6 @@ func (c ReplaceMisconfiguredPods) Reconcile(r *FoundationDBClusterReconciler, co
 		}
 
 		if instanceID != desiredInstanceID {
-			log.Info("JPB removing pod", "instanceID", instanceID, "desiredInstanceID", desiredInstanceID)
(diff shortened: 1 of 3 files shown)


controllers/replace_misconfigured_pods.go
```diff
@@ -90,39 +90,55 @@ func (c ReplaceMisconfiguredPods) Reconcile(r *FoundationDBClusterReconciler, co
...
+		if err != nil {
+			return false, err
+		}
 
-			_, idNum, err := ParseInstanceID(instanceID)
-			if err != nil {
-				return false, err
-			}
+		if instanceID != desiredInstanceID {
+			log.Info("JPB removing pod", "instanceID", instanceID, "desiredInstanceID", desiredInstanceID)
```
JoshuaMcManus, line 122: Looks like there's a leftover debug statement here

PR #305, opened 2020-08-19
Avoid removing instances that do not have an IP address to prevent losing volumes that are temporarily inaccessible.

### LINKED ISSUE #303: Exclusion commands should fail when IP is absent
In the exclude_instances step, we handle missing IPs by considering the exclusion complete. This will cause the rest of the removal to go forward without an exclusion happening. This means that if a pod is failing to schedule, we will delete the pod and its PVC without confirming that the data has been replicated. This can be a problem in certain scenarios. For instance, we may see a scenario where multiple pods are failing to schedule because of a temporary condition, and those pods hold all copies of a shard. This will leave the database in a state where that shard is unavailable. We can wait for the pods to eventually schedule, and it will come back with its data, allowing the database to recover. If we were to initiate a replacement, the operator would delete the PVCs, and we would never be able to recover that data.

I think the safer course of action is to ignore pods in this state in the exclude_instances action, but raise an error for these pods in the confirm_exclusion_completion action. This will allow us to start whatever exclusions we can, and change coordinators to a new set, but keep us from deleting any pods or PVCs until we know it is safe. We will need to make sure that there is a mechanism to automatically update the addresses if the pods schedule again.

If the pods are fully unrecoverable, and the user wants to allow things to move forward, I think they would have to delete the pods and the PVCs manually, and edit the cluster status to remove this entry from the pending removals. This flow should get cleaner once we can exclude instances by process ID instead of IP address.

brownleej:
Update the IP address in the pending removal state automatically when the pod gets rescheduled.
Add an option to remove an instance without excluding it to work around issues with stuck pods.

Fixes #303

### DIFF
--- a/controllers/check_instances_to_remove.go
+++ b/controllers/check_instances_to_remove.go
@@ -22,6 +22,7 @@ package controllers
 
 import (
 	ctx "context"
+	"reflect"
 	"time"
 
 	fdbtypes "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta1"
@@ -34,8 +35,6 @@ type CheckInstancesToRemove struct{}
 
 // Reconcile runs the reconciler's work.
 func (c CheckInstancesToRemove) Reconcile(r *FoundationDBClusterReconciler, context ctx.Context, cluster *fdbtypes.FoundationDBCluster) (bool, error) {
-	hasNewRemovals := false
-
 	var removals = cluster.Status.PendingRemovals
 
 	if removals == nil {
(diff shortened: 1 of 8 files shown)


controllers/check_instances_to_remove.go
```diff
@@ -34,29 +35,53 @@ type CheckInstancesToRemove struct{}
...
+		_, present := removals[instanceID]
+		if !present {
+			removals[instanceID] = fdbtypes.PendingRemovalState{}
+		}
+	}
+
+	for _, instanceID := range cluster.Spec.InstancesToRemoveWithoutExclusion {
+		_, present := removals[instanceID]
+		removalState := fdbtypes.PendingRemovalState{}
+		if !present {
```
JoshuaMcManus, line 54: You're checking if a value isn't present, then fetching it if it's not?

brownleej: You're right, this is backward. It didn't come up in my testing because the data can be reconstructed most of the time. I think it can be made much simpler now.

PR #317, opened 2020-09-03
Use slim base image for sample container

johscheuer:
This change reduces the size of the resulting container (and also reduces the attack surface) for faster startup of the container:

```bash
REPOSITORY              TAG                 IMAGE ID            CREATED             SIZE
johscheuer/sample-app   latest              b5d4c94a6dfc        2 minutes ago       120MB
johscheuer/sample-app   old                 47b9677ed2f9        3 seconds ago       889MB
```


brownleej (approved):
👍

Issue #252, opened 2020-06-12
Fail to build docker image with latest GoLang version


JoshuaMcManus:
When running the `make samples docker-build` command I'm getting the following error:

```
Step 6/27 : RUN 	curl $FDB_WEBSITE/downloads/$FDB_VERSION/ubuntu/installers/foundationdb-clients_$FDB_VERSION-1_amd64.deb -o fdb.deb && 	dpkg -i fdb.deb && rm fdb.deb && 	for version in ${FDB_VERSION} ${FDB_ADDITIONAL_VERSIONS}; do 		minor=${version%.*} && 		mkdir -p /usr/bin/fdb/$minor/tmp && 		curl $FDB_WEBSITE/downloads/$version/linux/fdb_$version.tar.gz -o /usr/bin/fdb/$minor/binaries.tar.gz && 		tar --strip-components=1 -C /usr/bin/fdb/$minor/tmp -xzf /usr/bin/fdb/$minor/binaries.tar.gz && 		rm /usr/bin/fdb/$minor/binaries.tar.gz && 		for binary in fdbcli fdbbackup fdbrestore; do 			mv /usr/bin/fdb/$minor/tmp/$binary /usr/bin/fdb/$minor/$binary; 		done && 		rm -r /usr/bin/fdb/$minor/tmp && 		chmod u+x /usr/bin/fdb/$minor/fdb*; 	done && 	mkdir -p /usr/lib/fdb && 	for version in ${FDB_ADDITIONAL_VERSIONS}; do 		curl $FDB_WEBSITE/downloads/$VERSION/linux/libfdb_c_$VERSION.so -o /usr/lib/fdb/libfdb_c_$VERSION.so; 	done
 ---> Running in 3c79e8478602
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (60) SSL certificate problem: unable to get local issuer certificate
More details here: https://curl.haxx.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
```

It seems to go away if I set the Go version to 1.13.11 explicitly. We'll want to investigate this to see if there's some sort of process change when pulling a file from the foundationdb servers.

[...]

PR #318, opened 2020-09-03
Add workaround for missing CA (see: #252)

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/252

This PR adds a workaround for the issue with the missing CA. During the build time the CA will be added and afterwards removed so the resulting image won't have the CA. This is an intermediate solution until the FDB website get another certificate.


Dockerfile
```diff
@@ -1,13 +1,18 @@
...
 ARG FDB_VERSION=6.2.22
 ARG FDB_ADDITIONAL_VERSIONS="6.1.13"
 ARG FDB_WEBSITE=https://www.foundationdb.org
 
 COPY foundationdb-kubernetes-sidecar/website/ /mnt/website/
+
+# FIXME: Workaround for (https://github.com/FoundationDB/fdb-kubernetes-operator/issues/252#issuecomment-643812649)
+# download GeoTrust_Global_CA.crt during install and remove it afterwards
 RUN set -eux && \
+    curl -sLo /usr/local/share/ca-certificates/GeoTrust_Global_CA.crt https://www.geotrust.com/resources/root_certificates/certificates/GeoTrust_Global_CA.pem && \
```
brownleej, line 14: Rather than downloading the CA at build-time, can we add it to our repo and copy it into the build? I think pinning it that way will be more secure.

brownleej (changes requested):

foundationdb-kubernetes-sidecar/files/GeoTrust_Global_CA.pem
```diff
@@ -0,0 +1,20 @@
+-----BEGIN CERTIFICATE-----
```
johscheuer, line 1: I put the file this directory so we only need to store it once (and Docker only allows to copy files from subdirectories or the same directory).

brownleej (approved):
👍

Issue #282, opened 2020-06-24
Detecting the use of deprecated fields


brownleej:
To ease the upgrade from version 0 of the operator to version 1, we should have a tool that detects when someone has cluster specs that use deprecated fields and recommends newer alternatives. We may be able to do this by adding a more general action for updating the spec to replace deprecated fields with the newer alternatives, allowing the rest of the reconciliation to refer only to the latest supported fields. We could then re-use that code to build a tool that scans the clusters, runs the code to update the specs in-place, detects any changes, and prints out a new recommended spec.

We should see if there's a way to also detect fields that have changes to the defaults from the previous version and recommend that they update their specs to explicitly set the new defaults. This will allow people to opt in to new defaults earlier and avoid changes to their spec when they upgrade the operator, or to opt in to keeping the old defaults even once they're running the new version of the operator. This may get tricky with changes to highly nested fields, like the resource requirements in the pod spec, but it could still be helpful to have this feature even if it were limited to simpler fields.

PR #323, opened 2020-09-08
Add default resource requirements for the sidecar

### LINKED ISSUE #99: Adding default resource requirements for the sidecar and init container in the operator
We have default resource requirements for the main foundationdb container, but not for the sidecar or the init container. I think having defaults for all the built-in containers will lead to a more consistent experience and prevent people from hitting operational issues due to lower defaults in the environments where they are running the operator. The user should be able to override these defaults through the pod template, like they can for the main foundationdb container.

brownleej:
Resolves #99 

This solution is a little more complex than might be required for this specific issue, but I wanted to accomplish a couple of other goals:

1. This establishes a general pattern for defining new defaults that won't take effect until the next major version. This will let us define changes to the defaults gradually, without disrupting any existing clusters immediately.
2. This lays some of the groundwork for #282, where we are going to be defining new tools to analyze out-of-date configurations in advance of a major upgrade of the operator

### DIFF
--- a/controllers/pod_models.go
+++ b/controllers/pod_models.go
@@ -876,7 +876,7 @@ func NormalizeClusterSpec(spec *fdbtypes.FoundationDBClusterSpec, defaults defau
 		}
 
 		sidecarUpdater := func(container *corev1.Container) {
-			if defaults.ApplyLatestDefaults {
+			if defaults.UseFutureDefaults {
 				if container.Resources.Requests == nil {
 					container.Resources.Requests = corev1.ResourceList{
 						"cpu":    resource.MustParse("100m"),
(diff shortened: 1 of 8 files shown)


controllers/pod_models.go
```diff
@@ -790,3 +755,139 @@ func GetHeadlessService(cluster *fdbtypes.FoundationDBCluster) (*corev1.Service,
...
+		if container.Name == name {
+			return containers, index
+		}
+	}
+
+	if insertIndex < 0 || insertIndex >= len(containers) {
+		containers = append(containers, corev1.Container{Name: name})
+		return containers, len(containers) - 1
+	} else {
+		containerCount := 1 + len(containers)
```
johscheuer, line 772: This could be simplified with https://github.com/golang/go/wiki/SliceTricks#insert e.g.

```go
containers = append(containers, corev1.Container{})
copy(containers[insertIndex+1:], containers[insertIndex:])
containers[insertIndex] = corev1.Container{Name: name}
```
this has the benefit that we don't need to allocate a new slice.

JoshuaMcManus (approved):

Issue #90, opened 2020-01-16
Doing atomic copies in the Kubernetes sidecar


brownleej:
When we copy monitor conf into the shared volume from the sidecar, it's possible that fdbmonitor could pick up an incomplete copy of the monitor conf. That could lead to misconfigured processes. We should try to do an atomic copy to reduce the risk of this.

PR #324, opened 2020-09-10
Make use of atomic copy for sidecar files

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/90


brownleej (approved):
👍

PR #328, opened 2020-09-14
Use Python slim for sidecar

johscheuer:
This PR changes the base image from Python yo Python-slim to reduce the size of the image:

```bash
johscheuer/foundationdb-kubernetes-sidecar   6.2.25-1            72f0583a0be6        9 minutes ago       1.02GB
johscheuer/foundationdb-kubernetes-sidecar   6.2.25-1-slim       65f52536de49        6 seconds ago       265MB
```

### DIFF
--- a/foundationdb-kubernetes-sidecar/Dockerfile
+++ b/foundationdb-kubernetes-sidecar/Dockerfile
@@ -17,18 +17,20 @@
 # limitations under the License.
 #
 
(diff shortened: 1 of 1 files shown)


johscheuer:
I just ran the tests locally without any issues so I would assume that the failure is a race condition.

brownleej (approved):
👍

PR #329, opened 2020-09-15
Update docs for local dev

johscheuer:
I updated the docs for local testing of the controller (e.g. the step for installing the fdb client packages was missing). I also fixed some markdown lint issues in the Readme (that's the reason why so much lines are moved around).


brownleej (approved):
👍

PR #336, opened 2020-10-06
Add initial support for golangci-lint support

johscheuer:
[golangci-lint](https://github.com/golangci/golangci-lint) is used by many golang based repositories as linter. The current configuration is inspired by the [kubebuilder](https://github.com/kubernetes-sigs/kubebuilder/blob/master/.golangci.yml) repository.


brownleej (approved):
👍

Issue #290, opened 2020-07-23
Using match labels to identify which pods are owned by the operator


brownleej:
When determining which resources are owned by the operator, we use a combination of a fixed `fdb-cluster-name` label and the owner reference in the metadata. In the discussion around #260 @rbtcollins raised concerns around using the owner reference, which is meant more for garbage collection, for something that has such an important role in correctness. It also limits flexibility with deleting and recreating clusters. As an alternative, we should look at using match labels, following the pattern used by Deployments and StatefulSets. This would give the user flexibility to define the labels that distinguish operator-owned pods from non-operator-owned pods.

In the interest of making the transition smooth, I think we want to continue using the owner references in the scope of this issue, and have a follow-up issue to remove support for it once people have had time to adjust their configurations to make sure they don't need to rely on owner reference.

brownleej:
We should probably have match labels for all of the resources that the operator manages.

PR #338, opened 2020-10-08
Simplify service reconciler

johscheuer:
I switched from `List` to `Get` since we only query one service.

### DIFF
--- a/controllers/remove_services.go
+++ b/controllers/remove_services.go
@@ -24,6 +24,8 @@ import (
 	ctx "context"
 	"time"
 
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+
 	fdbtypes "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta1"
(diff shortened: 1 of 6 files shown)


controllers/remove_services.go
```diff
@@ -34,26 +36,22 @@ type RemoveServices struct{}
...
+		}
 		return false, err
 	}
 
-	if len(existingServices.Items) > 0 {
-		err = r.Delete(context, &existingServices.Items[0])
-		if err != nil {
-			return false, err
-		}
+	err = r.Delete(context, existingService)
```
johscheuer, line 52: I think we should additionally check if the current cluster is the owner otherwise we could delete a service that is not part of the cluster (and managed by the operator).

brownleej: We might want to look at that as part of #290, which proposes using match labels instead of owner references when checking a similar relationship on pods.

brownleej (approved):
👍

PR #339, opened 2020-10-09
Add a job to check for deprecated fields

### LINKED ISSUE #282: Detecting the use of deprecated fields

brownleej:
Resolves #282


.gitignore
```diff
@@ -34,4 +34,7 @@ foundationdb-kubernetes-sidecar/website
 # Customizable templates
 
 config/default/manager_image_patch.yaml
-config/default/manager_image_patch.yaml-e
\ No newline at end of file
+config/default/manager_image_patch.yaml-e
+
+# General temp directory
+tmp/
```
johscheuer, line 40: NIT: could you add a newline here?

controllers/pod_models.go
```diff
@@ -806,41 +743,262 @@ func customizeContainerFromList(containers []corev1.Container, name string, cust
...
+		spec.Processes = make(map[string]fdbtypes.ProcessSettings)
+	}
+	generalSettings := spec.Processes["general"]
+	if generalSettings.PodTemplate == nil {
+		generalSettings.PodTemplate = &corev1.PodTemplateSpec{}
+	}
+	spec.Processes["general"] = generalSettings
+}
+
+// ensureCustomParametersPresent defines custom parameters the general process
```
johscheuer, line 759:
```suggestion
// ensureCustomParametersPresent defines custom parameters for the general process
```

docs/compatibility.md
```diff
@@ -1,8 +1,85 @@
...
-| Operator Version  | Supported Cluster Models  | Supported FDB Versions  | Supported Kubernetes Versions |
-| ----------------- | ------------------------- | ----------------------- | ----------------------------- |
-| 0.x               | v1beta1                   | 6.1.12+                 | 1.15.0+                       |
\ No newline at end of file
+* Dropping support for old versions of FDB
+* Dropping support for old versions of Kubernetes
+* Removing deprecated fields from the cluster resource
+* Changing the default values when fields are omitted in the spec
+* Removing flags and environment variables from the operator
+* Removing behavior from the operator
```
johscheuer, line 12: We should probably also mention removing deprecated metrics?

docs/compatibility.md
```diff
@@ -1,8 +1,85 @@
...
+running. You can find that release in the table above. This will ensure that you
+can opt in to new behavior and move to the latest supported fields in the spec
+in advance of the upgrade, through whatever process you need to update your
+clusters safely.
+
+At this point, you can run a job to check your cluster specs for deprecated
+fields or defaults. 
+
+		---
+		apiVersion: batch/v1
```
johscheuer, line 42: I think the job manifest needs some reformatting?

main.go
```diff
@@ -110,6 +118,19 @@ func main() {
...
+		DeprecationOptions:  deprecationOptions,
+	}
+
+	if checkDeprecations {
+		go startCache(mgr)
+		err = clusterReconciler.CheckDeprecations(context.Background())
+		if err != nil {
+			setupLog.Error(err, "unable to check deprecations")
+			os.Exit(1)
+		} else {
```
johscheuer, line 131: The `else` can be dropped since we will exit with `1` if an error occurred.

johscheuer (commented):
I just found some minor things otherwise 👍

JoshuaMcManus (approved):

PR #355, opened 2020-10-21
Adds start commands for fast restore processes

JoshuaMcManus:
This is to support the fast restore features of FDB 6.3.

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -596,6 +596,10 @@ func getStartCommandLines(cluster *fdbtypes.FoundationDBCluster, processClass st
 		fmt.Sprintf("locality_zoneid = %s", zoneVariable),
 	)
 
+	if processClass == "fast_restore" {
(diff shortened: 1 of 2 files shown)


controllers/cluster_controller.go
```diff
@@ -596,6 +596,10 @@ func getStartCommandLines(cluster *fdbtypes.FoundationDBCluster, processClass st
 		fmt.Sprintf("locality_zoneid = %s", zoneVariable),
 	)
 
+	if processClass == "fast_restore" {
+		confLines = append(confLines, "blob_credential_file = /var/secrets/blob_credentials.json")
```
brownleej, line 600: Do we want to hardcode this, or is it better to make it a configuration option? It feels weird to hardcode a path that the operator itself doesn't provide. It may also make it less clear that users need to define the volume mount for the secrets themselves.

brownleej (changes requested):

PR #356, opened 2020-10-21
Coordinate exclusions through the operator locking mechanism

### LINKED ISSUE #144: Synchronizing exclusions across multiple instances of the operator
When running an FDB cluster across multiple Kubernetes clusters, KC will have its own instance of the operator. Each instance of the operator runs independently, but the operator has some actions that it takes that affect the entire cluster. For instance, each instance of the operator is responsible for excluding instances under its control before removing them, but the exclusion can trigger a recovery which impacts the entire cluster. If each instance of the operator is acting independently, we could see multiple exclusions in a short period of time, which may cause a latency increase that impacts customers. We should try to find a way to coordinate the exclusions so that we can either do them in one large exclusion, or have a wait time between exclusions. At a high level I think we can use the FDB cluster as a coordinating mechanism, by storing additional state in the database, but this requires a more detailed design.

brownleej:
Resolves #144

### DIFF
--- a/controllers/change_coordinators.go
+++ b/controllers/change_coordinators.go
@@ -76,11 +76,11 @@ func (c ChangeCoordinators) Reconcile(r *FoundationDBClusterReconciler, context 
 			return false, err
 		}
 
-		hasLock, err := lockClient.TakeLock()
+		lock, err := lockClient.TakeLock()
 		if err != nil {
 			return false, err
 		}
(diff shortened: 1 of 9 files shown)


brownleej:
I was testing this in conjunction with another PR and found some issues where aggregation values in the lock data would stick around after the operation completes. I'm going to close this pending further discussion of how that should work.

PR #373, opened 2020-11-09
Using services for public IPs

### LINKED ISSUE #283: Using service IPs instead of pod IPs for public addresses
Using pod IPs as the public addresses for our instances can present some challenges. First, pod IPs may not be routable between Kubernetes clusters, which may cause challenges for people running multi-AZ configurations. Second, pod IPs are not retained when a pod is destroyed and recreated, so a coordinator being deleted can cause it to be unavailable as a coordinator until we recruit new coordinators. As an alternative to pod IPs, I think we should explore having an option to use service IPs. 

This would require expanding on the service config we added recently to support defining a service for every pod, with customization for the service spec to allow people to use different service types based on the constraints of their environment. We would also need options to use the service IPs for the fdbserver IPs.

I think that changing either the service spec or the IP source would require replacing all of the old processes. We'll need to be careful about how we determine the IPs in the operator to make sure we can use the correct IP based on how the process was initially configured.

This may be a shorter-term solution to #266 that doesn't require changes in FDB.

brownleej:
Resolves #283


controllers/add_pods.go
```diff
@@ -226,6 +212,55 @@ func (a AddPods) Reconcile(r *FoundationDBClusterReconciler, context ctx.Context
...
+
+	instanceID := GetInstanceIDFromMeta(pod.ObjectMeta)
+
+	if existingMeta != nil && instanceID != GetInstanceIDFromMeta(*existingMeta) {
+		return false, fmt.Errorf("Failed to create new pod to match %s", existingMeta.Name)
+	}
+
+	pod.ObjectMeta.Annotations[LastConfigMapKey] = configMapHash
+
+	services := &corev1.ServiceList{}
```
johscheuer, line 232: I would move this List request into the if statement below otherwise we query the API server and don't use the response.

controllers/cluster_controller.go
```diff
@@ -608,6 +613,10 @@ func getStartCommandLines(cluster *fdbtypes.FoundationDBCluster, processClass st
 		confLines = append(confLines, fmt.Sprintf("tls_verify_peers = %s", cluster.Spec.MainContainer.PeerVerificationRules))
 	}
 
+	if cluster.NeedsExplicitListenAddress() {
+		confLines = append(confLines, fmt.Sprintf("listen_address = %s", cluster.GetFullAddressList("$FDB_POD_IP", false)))
```
johscheuer, line 617: Is there any reason we don't want to add the flag by default?

controllers/pod_models.go
```diff
@@ -51,8 +49,31 @@ func getInstanceID(cluster *fdbtypes.FoundationDBCluster, processClass string, i
...
+	owner := buildOwnerReference(cluster.TypeMeta, cluster.ObjectMeta)
+
+	metadata := getObjectMetadata(cluster, nil, processClass, id)
+	metadata.Name = name
+	metadata.OwnerReferences = owner
+
+	return &corev1.Service{
+		ObjectMeta: metadata,
+		Spec: corev1.ServiceSpec{
+			Type: "ClusterIP",
```
johscheuer, line 65:
```suggestion
			Type: corev1.ServiceTypeClusterIP,
```

controllers/pod_models.go
```diff
@@ -280,18 +301,21 @@ func configureSidecarContainerForCluster(cluster *fdbtypes.FoundationDBCluster,
 		versionString = cluster.Spec.Version
 	}
 
-	return configureSidecarContainer(container, initMode, instanceID, versionString, cluster.Spec.SidecarVariables, cluster.Spec.FaultDomain, cluster.Spec.SidecarContainer, cluster.Spec.SidecarVersions, len(cluster.Spec.TrustedCAs) > 0)
+	publicIPSource := cluster.Spec.Services.PublicIPSource
+	usePublicIPFromService := publicIPSource != nil && *publicIPSource == fdbtypes.PublicIPSourceService
```
johscheuer, line 305: This could be simplified to `usePublicIPFromService := cluster.NeedsExplicitListenAddress()`

johscheuer: or you set it directly as argument without assigning a separate variable.

docs/user_manual.md
```diff
@@ -502,6 +507,34 @@ You can build this kind of configuration easily from the sample deployment by ch
...
+
+You can choose this option by setting `spec.services.publicIPSource=service`. This feature is new, and still experimental, but we plan to make it the default in the future.
+
+In this mode, we create one service for each pod, and use that service's IP as the public IP for the pod. The pod IP will still be used as the listen address. This ensures that IPs stay fixed even when pods get rescheduled, which reduces the need for changing coordinators and protects against some unrecoverable failure modes.
+
+Using service IPs presents its own challenges:
+
+* In some networking configurations, pods may not be able to access service IPs that route to the pod. See the section on hairpin mode in the [Kubernetes Docs](https://kubernetes.io/docs/tasks/debug-application-cluster/debug-service/#a-pod-fails-to-reach-itself-via-the-service-ip) for more information.
+* Creating one service for each pod may cause performance problems for the Kubernetes cluster
+* We currently only support services with the ClusterIP type. These IPs may not be routable from outside the Kubernetes cluster.
```
johscheuer, line 536: I would also add that the Service CIDR is normally not that big compared to the Pod CIDR which means it could happen that the cluster runs out of Service IPs.

johscheuer (commented):
Good work! Just some comments from my side. Also what happens if someone deletes a service? Will the service be recreated and the according process too?

Maybe in the future work we can reduce the number of created services by limiting the created services to the storage servers or only the coordinators. AFAIK it was discussed here: https://forums.foundationdb.org/t/proposal-dont-identify-a-coordinator-based-on-its-ip/502/19.

controllers/cluster_controller.go
brownleej: We don't want to make any introduce any changes to a cluster by default with new versions of the operator, as it would cause uncontrolled changes across all the clusters. We may want to add it across the board in the future to avoid needless variation between different configurations. I think we can tackle that in its own issue, though.

controllers/pod_models.go
brownleej: In my head, I'm distinguishing between those two things, since we likely want to move to explicit listen addresses everywhere in the future. It's a little complicated in how it's being used, though, so I'll need to take a look at that. The way we're passing information to the configureSidecar method is getting unsustainable.

johscheuer (approved):
LGTM 👍 Thanks.

PR #374, opened 2020-11-10
Use the zones and regions to infer whether locks are necessary.

### LINKED ISSUE #235: Inferring whether we need to take a lock for global operations
When we added the locking around global operations, we enabled it across the board by default. I think that we can infer this based on other aspects of the cluster spec. We only need to use locks when the cluster is running in multiple Kubernetes clusters, which we can identify by matching one of two conditions:

1) The cluster has regions defined, and has multiple regions
2) The cluster has a zoneCount greater than 1 in its replication configuration

If neither of these is met, we should not take locks unless it is explicitly enabled.

brownleej:
Resolves #235


johscheuer (approved):
LGTM 👍

PR #375, opened 2020-11-12
Take a lock when bouncing processes.

### LINKED ISSUE #237: Taking locks when bouncing processes
When we bounce processes, we should require a global lock. For instances that don't have the lock, they should submit the instances they want to bounce to allow the lock owner to aggregate and bounce them all at once.

brownleej:
Resolves #237


johscheuer (approved):
LGTM 👍

PR #376, opened 2020-11-12
Customizing lock duration

### LINKED ISSUE #240: Customizing lock durations
We should have options for users to configure the duration for which locks are held, and the duration the operator waits for aggregated operations before proceeding. This will help people get out of unforeseen situations, and allow people to tune things based on their knowledge of their environments.

brownleej:
Resolves #240


controllers/lock_client.go
```diff
@@ -227,12 +188,48 @@ func (client *MockLockClient) Disabled() bool {
...
+// subsequent calls, based on the cluster namespace and name.
+func getFDBDatabase(cluster *fdbtypes.FoundationDBCluster) (fdb.Database, error) {
+	cacheKey := fmt.Sprintf("%s/%s", cluster.ObjectMeta.Namespace, cluster.ObjectMeta.Name)
+	database, present := fdbDatabaseCache[cacheKey]
+	if present {
+		return database, nil
+	}
+
+	clusterFile, err := ioutil.TempFile("", "")
+	clusterFilePath := clusterFile.Name()
```
johscheuer, line 209: This could lead to a nil pointer exception we should check the error before using the `clusterFile`.

controllers/lock_client.go
```diff
@@ -227,12 +188,48 @@ func (client *MockLockClient) Disabled() bool {
...
+	if err != nil {
+		return fdb.Database{}, err
+	}
+
+	defer clusterFile.Close()
+	if err != nil {
```
johscheuer, line 215: Seems like a redundant check to the if statement before?

controllers/lock_client.go
```diff
@@ -227,12 +188,48 @@ func (client *MockLockClient) Disabled() bool {
...
+		return fdb.Database{}, err
+	}
+
+	_, err = clusterFile.WriteString(cluster.Status.ConnectionString)
+	if err != nil {
+		return fdb.Database{}, err
+	}
+	err = clusterFile.Close()
```
johscheuer, line 223: We close this file multiple times with the `defer` and in this statement. This shouldn't be a problem, we just should keep this in mind.

johscheuer (changes requested):
Just some smaller things otherwise LGTM

PR #380, opened 2020-11-13
Initial add of multiple storage servers per disk

johscheuer:
Implements: https://github.com/FoundationDB/fdb-kubernetes-operator/wiki/Multiple-Storage-Servers-per-Disk-Design

Currently this will require that all processes are restarted with a new `ConfigMap` to introduce the `process_id`


api/v1beta1/foundationdbcluster_types.go
```diff
@@ -328,6 +335,11 @@ type FoundationDBClusterStatus struct {
 	// sidecar conf in the config map even when the latest version should not
 	// require it.
 	NeedsSidecarConfInConfigMap bool `json:"needsSidecarConfInConfigMap,omitempty"`
+
+	// StorageServersPerDisk defines the current count of storage servers per disk
+	// this value should be StorageServersPerPod * ProcessCount for storage class.
+	// If the value differs the reconcile phase is not finished
+	StorageServersPerDisk map[string]bool `json:"storageServersPerDisk,omitempty"`
```
brownleej, line 342: What are the keys in this map? I'm guessing instance IDs? I generally prefer to make that explicit in the documentation.

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -2113,3 +2137,24 @@ const (
 	// PublicIPSourceService specifies that a pod gets its IP from a service.
 	PublicIPSourceService PublicIPSource = "service"
 )
+
+// ProcessClass models the different FDB classes
+type ProcessClass string
```
brownleej, line 2142: Is this intended to be an exhaustive list, or just a list of process classes that have specialized behavior?

controllers/bounce_processes.go
```diff
@@ -50,28 +50,35 @@ func (b BounceProcesses) Reconcile(r *FoundationDBClusterReconciler, context ctx
...
+	for process := range cluster.Status.IncorrectProcesses {
+		if addressMap[process] == "" {
+			return false, fmt.Errorf("could not find address for process: %s", process)
 		}
 
-		addresses = append(addresses, addressMap[instanceID])
+		addresses = append(addresses, addressMap[process])
+
+		instanceID := GetInstanceIDFromProcessID(process)
+		log.Info("GetPod", "instanceID", instanceID)
```
brownleej, line 74: If we're going to leave this logging statement in, can we add some more context?

controllers/cluster_controller.go
```diff
@@ -400,15 +424,41 @@ func GetConfigMap(context ctx.Context, cluster *fdbtypes.FoundationDBCluster, ku
...
+			if processClass == fdbtypes.ProcessClassStorage {
+				storageServersPerDisk, err := convertStringSetToIntSet(cluster.Status.StorageServersPerDisk)
 				if err != nil {
 					return nil, err
 				}
-				data[filename] = conf
+				// If the status field is not initialized we fallback to only the specified count
+				// in the cluster spec. This should only happen in the initial phase of a new cluster.
+				if len(cluster.Status.StorageServersPerDisk) == 0 {
+					log.Info("GetConfigMap", "fallback to", cluster.GetStorageServersPerPod(), "storageServersPerDisk", storageServersPerDisk)
```
brownleej, line 435: If we're going to leave this logging statement in, can we add some more context?

controllers/exclude_instances.go
```diff
@@ -55,7 +55,7 @@ func (e ExcludeInstances) Reconcile(r *FoundationDBClusterReconciler, context ct
 
 		for _, state := range cluster.Status.PendingRemovals {
 			if state.Address != "" {
-				address := RemoveAddressFlags(cluster.GetFullAddress(state.Address))
+				address := state.Address
```
brownleej, line 58: We should do a test to make sure the transition between exclusion methods is safe:

1. Start a replacement on the old code
2. Let the replacement get far enough to start the exclusion
3. Update the operator to use the new code
4. Let the replacement finish on the new code.
5. Confirm that exclusion checks work fine, and that the inclusion command cleans up everything in the exclusion list.

controllers/pod_models.go
```diff
@@ -207,6 +219,10 @@ func GetPodSpec(cluster *fdbtypes.FoundationDBCluster, processClass string, idNu
 		return nil, err
 	}
 
+	if processClass == fdbtypes.ProcessClassStorage {
+		sidecarContainer.Env = append(sidecarContainer.Env, corev1.EnvVar{Name: "STORAGE_SERVERS_PER_POD", Value: fmt.Sprintf("%d", cluster.GetStorageServersPerPod())})
```
brownleej, line 223: Will this cause a latent change to the environment variables for existing clusters? If so, I think we need to find a way to avoid that.

controllers/pod_models_test.go
```diff
@@ -304,8 +305,165 @@ var _ = Describe("pod_models", func() {
 					VolumeSource: corev1.VolumeSource{ConfigMap: &corev1.ConfigMapVolumeSource{
 						LocalObjectReference: corev1.LocalObjectReference{Name: fmt.Sprintf("%s-config", cluster.Name)},
 						Items: []corev1.KeyToPath{
+							{Key: "cluster-file", Path: "fdb.cluster"},
```
brownleej, line 308: Why is this changing as part of the PR? It seems like something that always should have been there, but I don't know why it would have been passing before if that were the case.

brownleej: Ah, nevermind, I see that this is just an artifact of the way the diff is shown.

controllers/update_pods.go
```diff
@@ -73,6 +74,19 @@ func (u UpdatePods) Reconcile(r *FoundationDBClusterReconciler, context ctx.Cont
...
 		if instance.Metadata.Annotations[LastSpecKey] != specHash {
+			if instance.GetProcessClass() == fdbtypes.ProcessClassStorage {
+				storageServersPerPodStr := getStorageServersPerPodForInstance(&instance)
+				storageServersPerPod, err := strconv.Atoi(storageServersPerPodStr)
+				if err != nil {
+					return false, err
+				}
+
+				// If the storage servers differ skip these instances since they will be replaced
+				if storageServersPerPod != cluster.GetStorageServersPerPod() {
```
brownleej, line 85: This check shouldn't be necessary, because we should be choosing pods for replacement before this step happens. We rely on that behavior in other areas, so that we don't have to duplicate logic between two actions.

controllers/update_status.go
```diff
@@ -118,10 +127,12 @@ func (s UpdateStatus) Reconcile(r *FoundationDBClusterReconciler, context ctx.Co
 	status.IncorrectPods = make([]string, 0)
 	status.FailingPods = make([]string, 0)
 
-	configMap, err := GetConfigMap(context, cluster, r)
+	configMap, err := GetConfigMap(cluster)
 	if err != nil {
 		return false, err
 	}
+
+	// TODO (johscheuer): should be process specific #377
```
brownleej, line 135: If we already have an issue tracking it, do we need a TODO comment in the code?

api/v1beta1/foundationdbcluster_types.go
brownleej: From context later on, it seems like the keys in this map are a string representation of the number of storage servers?

controllers/update_status.go
```diff
@@ -131,50 +142,34 @@ func (s UpdateStatus) Reconcile(r *FoundationDBClusterReconciler, context ctx.Co
...
 
-		processStatus := processMap[instanceID]
-		if len(processStatus) == 0 {
-			existingTime, exists := cluster.Status.MissingProcesses[instanceID]
-			if exists {
-				status.MissingProcesses[instanceID] = existingTime
-			} else {
-				status.MissingProcesses[instanceID] = time.Now().Unix()
+		if processClass == fdbtypes.ProcessClassStorage && cluster.GetStorageServersPerPod() > 1 {
+			storageServersPerPod := cluster.GetStorageServersPerPod()
```
brownleej, line 157: Do we want to use the cluster-level setting here, or the instance-level setting?

controllers/bounce_processes.go
brownleej: In general, I think it's best to include the cluster name and namespace in all log messages that are specific to a cluster. This helps us narrow down relevant log messages when dealing with issues for a cluster.

brownleej (changes requested):

api/v1beta1/foundationdbcluster_types.go
johscheuer: In the end I think this could be an exhaustive list for now I only added the process classes that are used (in a follow up PR I would also adjust the tests to use these strings).

controllers/bounce_processes.go
johscheuer: That makes sense, I remove these logging statements I just used them for debugging.

controllers/pod_models.go
johscheuer: Probably the simplest solution us to set this environment variable only if `storageServersPerPod` is bigger than 1 (which is the default).

controllers/update_status.go
johscheuer: Good point 👍

controllers/update_status.go
johscheuer: Actually I like a todo in the code base even if we have an issue for it this makes it easier to identify the place for the fix and I will be remembered that there is an issue when I take a look at this code path again.

api/v1beta1/foundationdbcluster_types.go
munnerz: Total drive by comment here, but map types are not advised in Kubernetes APIs and are against convention 😄 https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps

controllers/exclude_instances.go
johscheuer: In my tests everything worked as expected. I ran the tests manually since we currently don't have any mechanism to run these tests automatically.

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -2127,3 +2150,26 @@ const (
...
```
brownleej, line 2155: Where is this type being used?

brownleej (commented):
Small question about the ProcessClass type.

api/v1beta1/foundationdbcluster_types.go
johscheuer: I use it in some places like: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/380/files#diff-9e300e65cd8198dd2dcc3e8f7a252ecea4fa5d072e3bf55bb1aff38aa4927d7fR222 + https://github.com/FoundationDB/fdb-kubernetes-operator/pull/380/files#diff-9e300e65cd8198dd2dcc3e8f7a252ecea4fa5d072e3bf55bb1aff38aa4927d7fR242 + https://github.com/FoundationDB/fdb-kubernetes-operator/pull/380/files#diff-9e300e65cd8198dd2dcc3e8f7a252ecea4fa5d072e3bf55bb1aff38aa4927d7fR538

johscheuer: I initially thought it makes sense to have a general `ProcessClass` type but I think we actually don't need it and I will remove it.

brownleej (approved):
👍

PR #381, opened 2020-11-13
Add a design for automating replacements.

brownleej:


docs/design/automatic_replacements.md
```diff
@@ -0,0 +1,51 @@
+# Automatic Replacements
+
```
johscheuer, line 2: What about some metadata after the first heading e.g. https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/0031-20181017-kube-proxy-services-optional.md (not with every field) but `creation date`, `last updated` and `author` seems to me reasonable. I think the status field is not required for us or at least not in the current state.

johscheuer (approved):
LGTM 👍

PR #382, opened 2020-11-13
Ensure that the script ends if the curl fails

johscheuer:
We want that the script fails if the curl command doesn't work otherwise the script will run in some strange issues.


brownleej (approved):
👍

PR #383, opened 2020-11-16
Design for tracking process group details in the status

brownleej:


docs/design/process_group_status.md
```diff
@@ -0,0 +1,108 @@
...
+*	Allow recovering almost all process information from the running state.
+
+## Proposed Design
+
+The term "process group" will gradually replace the term "instance". This design
+will introduce the term, but will not require the removal of all uses of the
+word "instance", which has additional design challenges. There will be a 1:1
+relationship between a process group and a pod, in the healthy state, but there
+will be times where a process group does not have a pod, in which case the
+operator will create on. The term "process" will refer to a single fdbserver
```
johscheuer, line 51:
```suggestion
operator will create one. The term "process" will refer to a single fdbserver
```

johscheuer (commented):
One NIT otherwise LGTM 👍

johscheuer (approved):
LGTM 👍

docs/design/process_group_status.md
```diff
@@ -0,0 +1,108 @@
...
+add new entries to the map. After that action, we will run separate actions for
+creating new services, new PVCs, and new pods, in that order. Each of those
+actions will target a single resource and will update the status to reflect the
+new conditions.
+
+The exclusion action will use the `processGroups` as its source of truth, and
+will exclude all of the addresses for process groups marked for removal. If a
+process group has no addresses, then we will not be able to complete the
+exclusion action. The inclusion action will include all of the addresses for the
+process groups, and will then remove those entries from the map.
```
JoshuaMcManus, line 95: Can you elaborate a bit on this process? I think I understand it, but just want to be totally clear. When a process group is marked for inclusion, you will run an `include` command on FDB with the know addresses for that process group, but you'll then remove the process group entirely from the map? 

Are there not scenarios where we want to not remove a process group entirely from the cluster, and instead just exclude then re-include it?

JoshuaMcManus (commented):
Small question, otherwise looks good.

docs/design/process_group_status.md
brownleej: We have scenarios like that, but it's not currently supported by the operator. The operator only excludes / includes processes as part of removing them.

Issue #392, opened 2020-11-20
Supporting multiple storage servers in a single service IP when using 2 per disk


brownleej:
When we are configuring a cluster with two or more storage servers per disk, they will be listening on different ports. We need to make sure all these ports are supported in the service config when we are using service IPs for our public IPs.

PR #394, opened 2020-11-21
Expose all ports in a service for multi storage servers

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/392

### DIFF
--- a/controllers/add_pods.go
+++ b/controllers/add_pods.go
@@ -236,7 +236,12 @@ func createInstance(r *FoundationDBClusterReconciler, context ctx.Context, clust
 			return false, err
 		}
 
-		if len(services.Items) == 0 {
+		serviceCount := len(services.Items)
+		if serviceCount > 1 {
+			return false, fmt.Errorf("expected to find one or no service but found %d services for %s instance in cluster: %s/%s", serviceCount, instanceID, cluster.Namespace, cluster.Name)
+		}
+
+		if serviceCount == 0 {
 			service, err := GetService(cluster, processClass, idNum)
 			if err != nil {
 				return false, err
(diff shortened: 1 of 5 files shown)


controllers/add_pods.go
```diff
@@ -245,6 +250,32 @@ func createInstance(r *FoundationDBClusterReconciler, context ctx.Context, clust
...
+
+		// If we create a new Pod and reuse a service we should ensure that it matches.
+		if processClass == fdbtypes.ProcessClassStorage {
+			// Ensure that the service contains all required ports.
+			// Since a process can listen on non-TLS and TLS we multiply by 2.
+			// Since we are creating a new Pod we can use the cluster.GetStorageServersPerPod()
+			// if this value would be changed during the creation it should be reconciled again.
+			if len(services.Items[0].Spec.Ports) != (cluster.GetStorageServersPerPod() * 2) {
+				log.Info("Updating service missing ports", "got", len(services.Items[0].Spec.Ports), "expected", cluster.GetStorageServersPerPod()*2)
+				log.Info("Updating service missing ports", "namespace", cluster.Namespace, "cluster", cluster.Name, "serviceName", services.Items[0].ObjectMeta.Name)
```
brownleej, line 262: Should we combine these into a single log item? It may be harder to search through logs with two separate items.

controllers/add_pods.go
```diff
@@ -245,6 +250,32 @@ func createInstance(r *FoundationDBClusterReconciler, context ctx.Context, clust
...
 			return false, err
 		}
+
+		// If we create a new Pod and reuse a service we should ensure that it matches.
+		if processClass == fdbtypes.ProcessClassStorage {
+			// Ensure that the service contains all required ports.
+			// Since a process can listen on non-TLS and TLS we multiply by 2.
+			// Since we are creating a new Pod we can use the cluster.GetStorageServersPerPod()
+			// if this value would be changed during the creation it should be reconciled again.
+			if len(services.Items[0].Spec.Ports) != (cluster.GetStorageServersPerPod() * 2) {
```
brownleej, line 260: This seems like an odd place to update service config. We've got another issue on allowing customization of the service spec, so maybe we should think about where we want to update service config in that context, and whether updating the ports can work in the same place.

brownleej (changes requested):

controllers/add_pods.go
johscheuer: Makes sense 👍

controllers/add_pods.go
johscheuer: The idea here was that If a Pod was replaced with the same `instance_id` and there already exists an service with the given `instance_id` we need to ensure that all required ports are exposed.

johscheuer: I think for the current stage this is fine but probably once we got a more complex/richer reconciliation for services we want to move it there.

brownleej (approved):
👍

Issue #98, opened 2020-01-28
Using a distroless image for the operator


brownleej:
The default configuration for Kubebuilder 2 uses a distroless image as the base for the operator image. I left this out of the Kubebuilder 2 conversion to avoid changing more than was necessary. We should investigate this more and see if it provides a safer and lighterweight way of running the operator.

PR #395, opened 2020-11-21
Use distroless as base image for the operator

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/98 I validated that the operator still can call `fdbcli`.


brownleej (approved):
👍

Issue #131, opened 2020-02-13
Kubectl plugin for fdb operations


akshaychitneni:
Following https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/, we can extend the kubectl to support easy operations with fdb-operator,

To start with, it can be kubectl fdb scale sample-cluster <process-type> 3 or simple get operation for logs/metrics per process

brownleej:
Yeah, that sounds great!

johscheuer:
I close this issue since we got a simple plugin and some additional issues to extend the plugin. Feel free to open an issue for your use case if you can't find it. Thanks for reporting!

PR #399, opened 2020-11-22
Initial creation of kubectl plugin for fdb

johscheuer:
Initial support for a Kubectl plugin to interact with the fdb operator. This add the skeleton for a Kubectl fdb plugin to simplify working with the operator and fdb clusters on Kubernetes. Further commands can be added afterwards.

Solves: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/131


kubectl-fdb/cmd/exclude.go
```diff
@@ -0,0 +1,118 @@
...
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"log"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	ctx "context"
+	fdbtypes "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta1"
+)
+
+// excludeCmd represents the exclude command of the fdb cli
+var excludeCmd = &cobra.Command{
```
johscheuer, line 35: In an additional PR we could also add the following features:

- exclude all instances of a cluster
- exclude all instances with a specific label (e.g. all storage instances)

I will open an follow up issue for this task.

kubectl-fdb/Readme.md
```diff
@@ -0,0 +1,31 @@
...
+export PATH="$PATH:$(pwd)/bin" 
+```
+
+## Usage
+
+Run `kubectl fdb help` to get the latest help.
+
+### Supported operations
+
+- `exlcude` instances from a cluster.
```
brownleej, line 21:
```suggestion
- `exclude` instances from a cluster.
```

kubectl-fdb/cmd/exclude.go
brownleej: Wouldn't those operations fail due to excluding too much of the cluster?

brownleej: Ah, I was confused about what the `exclude` operation does. I think we need to be consistent in our use of `exclude` and `replace`. We may want to talk about this more out-of-band to get the right terminology.

kubectl-fdb/cmd/version.go
```diff
@@ -0,0 +1,76 @@
...
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/spf13/cobra"
+	"k8s.io/client-go/kubernetes"
+)
+
+// TODO: should follow the operator versioning
+var pluginVersion string = "0.0.1"
```
brownleej, line 33: Should we make this match the current version, and then add a note to the release guide to update the version here? We're already updating the version by find-and-replace.

brownleej (changes requested):

kubectl-fdb/cmd/exclude.go
johscheuer: I adjusted the name to `remove` like discussed "offline"

kubectl-fdb/cmd/remove.go
```diff
@@ -0,0 +1,159 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	ctx "context"
+
+	fdbtypes "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta1"
+)
+
+// removeCmd represents the removal of
```
brownleej, line 39: Is there meant to be more in this comment?

brownleej (changes requested):
Small question about a comment. Otherwise this looks good.

kubectl-fdb/cmd/remove.go
johscheuer: I fixed it, I probably missed this one during refactoring.

brownleej (approved):
👍

PR #403, opened 2020-11-30
Prevent bounce when we update the operator with support for multiple storage servers per disk

brownleej:
This PR also addresses an issue I found with testing this where we were not bouncing storage servers correctly when we did want to bounce them.


johscheuer (approved):
LGTM 👍

Issue #335, opened 2020-10-05
Adding documentation on customizing storage classes


brownleej:
People will likely want to customize their storage class, if they're running in an environment with more than one, or with no default storage class. We should add an example of this to the user manual.

PR #414, opened 2020-12-07
Add docs for StorageClass and multi Storage Servers per Pod

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/335


brownleej (approved):
👍

PR #425, opened 2020-12-10
Add securityContext to helm chart

funkypenguin:
Hey guys,

This PR adds a securityContext to both the manager pod, and the specific manager container. Because we enforce `readOnlyRootFilesystem` in `container.securityContext`, we also add an `emptyDir` volumeMount to allow `/tmp` to be read/write, without allowing the **rest** of the pod to be read/write.

I also adjusted the UID to 40590 in both the `pod.securityContext` and the `Dockerfile`, so that subsequent CI actions (_coming in a separate PR_) using [kube-score](_https://github.com/zegl/kube-score_) will pass the manifests (_since the UID will be > 10000_). Additionally, clusters employing PSPs enforcing containers to run as non-root will allow the container image to run, since the `USER` value is now numeric (_this PR supercedes my previous PR #424_)

Cheers!
D

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,8 +47,8 @@ RUN CGO_ENABLED=1 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager 
 
 # Create user and group here since we don't have the tools
 # in distroless
-RUN groupadd --gid 4059 fdb && \
-	useradd --gid 4059 --uid 4059 --create-home --shell /bin/bash fdb && \
+RUN groupadd --gid 40590 fdb && \
+	useradd --gid 40590 --uid 40590 --create-home --shell /bin/bash fdb && \
 	mkdir -p /var/log/fdb && \
 	touch /var/log/fdb/.keep
 
@@ -64,7 +64,7 @@ COPY --from=builder /usr/lib/libfdb_c.so /usr/lib/
 COPY --from=builder /usr/lib/fdb /usr/lib/fdb/
 COPY --chown=fdb:fdb --from=builder /var/log/fdb/.keep /var/log/fdb/.keep
 
-USER fdb
+USER 40590
 
 ENV FDB_NETWORK_OPTION_TRACE_LOG_GROUP=fdb-kubernetes-operator
 ENV FDB_NETWORK_OPTION_TRACE_ENABLE=/var/log/fdb
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,8 +47,8 @@ RUN CGO_ENABLED=1 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager 
 
 # Create user and group here since we don't have the tools
 # in distroless
-RUN groupadd --gid 40590 fdb && \
-	useradd --gid 40590 --uid 40590 --create-home --shell /bin/bash fdb && \
+RUN groupadd --gid 4059 fdb && \
+	useradd --gid 4059 --uid 4059 --create-home --shell /bin/bash fdb && \
 	mkdir -p /var/log/fdb && \
 	touch /var/log/fdb/.keep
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -47,8 +47,8 @@ RUN CGO_ENABLED=1 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager
 
 # Create user and group here since we don't have the tools
 # in distroless
-RUN groupadd --gid 4059 fdb && \
-	useradd --gid 4059 --uid 4059 --create-home --shell /bin/bash fdb && \
+RUN groupadd --gid 40590 fdb && \
+	useradd --gid 40590 --uid 40590 --create-home --shell /bin/bash fdb && \
```
johscheuer, line 51: Any reason to change the UID/GID?

johscheuer (changes requested):

Dockerfile
funkypenguin: I've reverted these, since the kube-score test which precipitated these changes is fixed in #431 now

helm/fdb-operator/templates/manager/deployment.yaml
```diff
@@ -20,6 +20,13 @@ spec:
       labels:
         {{- include "chart.labels" . | nindent 8 }}
     spec:
+      securityContext:
```
johscheuer, line 23: Could you also update https://github.com/FoundationDB/fdb-kubernetes-operator/tree/master/config/samples/deployment they should always match

johscheuer (changes requested):
Thanks for your PR, the failing CI is due to a new (broken) kustomize release. Just one comment otherwise LGTM 👍

helm/fdb-operator/templates/manager/deployment.yaml
funkypenguin: Updated :)

johscheuer:
Could you fix the CI failure you need to run `make clean all` and push the diff.

funkypenguin:
> Could you fix the CI failure you need to run `make clean all` and push the diff.

I tried, but this is the result:

```
run `controller-gen object:headerFile=./hack/boilerplate.go.txt paths=./... -w` to see all available markers, or `controller-gen object:headerFile=./hack/boilerplate.go.txt paths=./... -h` for usage
make: *** [api/v1beta1/zz_generated.deepcopy.go] Error 1
```

johscheuer:
Could you ensure that you have setup everything https://github.com/FoundationDB/fdb-kubernetes-operator#environment-set-up ?

funkypenguin:
I've run `make clean all`, and committed the resulting update the samples. However, the build CI run seems to be failing, although linting now succeeds.

config/samples/deployment.yaml
```diff
@@ -126,5 +126,13 @@ spec:
...
             cpu: 500m
             memory: 256Mi
+        securityContext:
+          allowPrivilegeEscalation: false
+          privileged: false
+          readOnlyRootFilesystem: true
+      securityContext:
+        fsGroup: 4059
+        runAsGroup: 4059
+        runAsUser: 4059
```
johscheuer, line 136: The `emptyDir` is missing here. Could you add it, I assume that otherwise the operator won't work since `readOnlyRootFilesystem` is set.

johscheuer (changes requested):
The build failure is a flaky test (we need to fix this). Could you add the missing `emptyDir`? Once the `emptyDir` is added LGTM>

funkypenguin:
I added an extra emptyDir for `/var/log/fdb` (_I noticed after a while that when I did something bad on mine, the operator would log a fault to complain that it couldn't create a trace in `/var/log/fdb` because of `readOnlyRootFilesystem`_). I also manually updated the sample deployment.

johscheuer:
Could you run `make clean all` again? The CI complains about uncommitted changes.

config/samples/deployment.yaml
funkypenguin: Done by running `make clean all` :)

johscheuer (approved):
👍 LGTM, thanks

PR #427, opened 2020-12-10
Release the kubectl plugin with prebuilt binaries

johscheuer:
I need to check how I can tests this in my branch.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,73 @@
+name: Release kubectl plugin
+on:
+  push:
+    tags:
+    - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
+
+jobs:
+  create-release:
+    runs-on: ubuntu-latest
+    outputs:
+      upload_url: ${{ steps.create_release.outputs.upload_url }}
+    steps:
+      - name: Create Release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ github.ref }}
+          # Should use body_path + auto generate release info
+          body: |
+            Changes in this Release
+            - More awesomeness !
+          release_name: Release ${{ github.ref }}
+          draft: false
+          prerelease: false
+  release-plugin:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os:
+          # - windows-latest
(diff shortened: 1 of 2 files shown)


brownleej:
I'd like to learn more about how the release process will work. Maybe we can talk about it in our next sync-up?

johscheuer:
The release works for Linux currently (see: https://github.com/johscheuer/fdb-kubernetes-operator/actions/runs/423000747) for MacOS and Windows I ran into some strange cross compile issues that I need to solve first. Maybe another solution would be to use the according Windows/MacOS GitHub runner.

scripts/install_dependency
```diff
@@ -0,0 +1,16 @@
...
+
+# Workaround for never Ubuntu/Debian releases which doesn't contain the GeTrust CA anymore
+mkdir -p /usr/local/share/ca-certificates
+cp ./foundationdb-kubernetes-sidecar/files/GeoTrust_Global_CA.pem /usr/local/share/ca-certificates/GeoTrust_Global_CA.crt
+update-ca-certificates
+
+# Required for Windows cross-compile
+apt-get update && apt-get install -y gcc-multilib gcc-mingw-w64
+
+# Install the Ubtunu packahe
```
johscheuer, line 14:
```suggestion
# Install the Ubtunu package
```

johscheuer:
Sorry for the Spam! I found a way to release a linux and a macOS binary: https://github.com/johscheuer/fdb-kubernetes-operator/actions/runs/519212388 (I didn't get Windows to work but I don't know how many people use Windows and if somebody want's that I'm totally open for a PR). We have some different possibilities here when the job creates a release: https://github.com/johscheuer/fdb-kubernetes-operator/releases/tag/v0.0.1348 that I would like to discuss with you.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,73 @@
...
+      upload_url: ${{ steps.create_release.outputs.upload_url }}
+    steps:
+      - name: Create Release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ github.ref }}
+          # Should use body_path + auto generate release info
```
johscheuer, line 20: We should raise an additional issue to provide the release notes automatically.

brownleej (approved):
👍

We just cut the 0.28 release, so this won't make it out for that release. That should give us some time to refine things like the release notes before the next release.

PR #429, opened 2020-12-10
Allow customization of pod resources in helm chart

funkypenguin:
Continuing the PR blitz, this PR adds the option for the user to customize the resource requests/limits in the helm chart, while "falling back" to the existing requests/limits by default.

Why? We like to fine-tune our resource allocations, and being able to control the allocations for the operator pod is desirable ;)

Cheers!
D


johscheuer (approved):
LGTM 👍

PR #430, opened 2020-12-10
Allow override of deployment replicas value

funkypenguin:
This PR allows the helm-chart deployer the ability to customize the amount of replicas in the deployment. A potential future enhancement in this area would be to add a PodDisruptionBudget, as well as affinity settings, to accommodate availability requirements on an HA cluster.


helm/fdb-operator/values.yaml
```diff
@@ -4,3 +4,4 @@ operator:
   tag: 0.24.0
   role: fdb-kubernetes-operator-manager-role
   rolebinding: fdb-kubernetes-operator-manager-rolebinding
+  replicas: 1
```
johscheuer, line 7: NIT: missing newline

johscheuer (changes requested):

johscheuer:
We can add a `PodDisruptionBudget` optionally but the availability of the operator doesn't directly effect the availability of the database itself.

helm/fdb-operator/values.yaml
funkypenguin: Fixed :)

funkypenguin:
> We can add a `PodDisruptionBudget` optionally but the availability of the operator doesn't directly effect the availability of the database itself.

Agreed, it's a nice-to-have in future, and I felt that adding it to this changeset was unnecessary :)

johscheuer (approved):
👍

johscheuer:
Once the conflict is solved I merge the PR. Thanks!

funkypenguin:
Conflict resolved ;)

johscheuer:
Thanks!

PR #432, opened 2020-12-10
Fix initial issues with plugin

johscheuer:
Fixes some initial issues with the plugin and auto selects the default namespace from the current context.


brownleej (approved):
👍

PR #435, opened 2020-12-11
Pin kustomize version

johscheuer:
customise released a new version which breaks our build so I pinned it to `v3.8.8` which works (on my machine 😄) I open an issue to check what breaks https://github.com/kubernetes-sigs/kustomize/releases.


johscheuer:
@brownleej I merge this into the master since otherwise our tests are not running.

Issue #441, opened 2020-12-14
StorageServerPerDisk is not set in update status


johscheuer:
We have a bug that doesn't update the status field which means the `ConfigMap` is never updated and changing the `StorageServerPerPod` can't be changed. I assume that we missed this in our tests because we run a mocked API (doesn't really create/spawn the pods.

PR #442, opened 2020-12-14
Fix bug to update storageServerPerPod

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/441 since we passed `clusterStatus FoundationDBClusterStatus` by value and not by reference the "real" status was always empty.


controllers/update_status.go
```diff
@@ -287,9 +289,8 @@ func (s UpdateStatus) Reconcile(r *FoundationDBClusterReconciler, context ctx.Co
 		status.FailingPods = nil
 	}
 
-	if len(status.StorageServersPerDisk) == 0 {
```
johscheuer, line 290: This case can never happen since we add the default value: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/442/files#diff-291a96adb6e62f693eebc44f9b612d5eff1fcf23f5479a5151340afb91dbd118R51

brownleej (approved):
👍

Issue #437, opened 2020-12-11
Checking variable shadowing


brownleej:
We had a subtle bug recently caused by shadowing variables in an assignment with the `:=` operator. There's a lint check for variable shadowing that we can enable to detect this. I tried this earlier and it found several more examples of variable shadowing, so we should add the check and clean up that shadowing.

    govet:
        enable:
            - shadow

PR #445, opened 2020-12-15
Enable golint to check for shadowing

johscheuer:
Fixes:  https://github.com/FoundationDB/fdb-kubernetes-operator/issues/437


brownleej (approved):
👍

Issue #421, opened 2020-12-09
Use constants instead of magic strings


johscheuer:
Currently we use some "magic string" in different places like `"fdb-instance-id"` or `"fdb-process-class"` these string should be defined as constants.

johscheuer:
A good place would be: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/controllers/controllers.go#L30

PR #446, opened 2020-12-17
Use constants instead of magic strings

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/421


brownleej (approved):
👍

PR #447, opened 2020-12-21
Design for unifying the sidecar and fdbmonitor into a single process and image.

brownleej:


docs/design/unified_sidecar_and_process_monitor.md
```diff
@@ -0,0 +1,118 @@
...
+
+When the new launcher is used, both the main container and the sidecar container will run the same image, but the sidecar container will have a different set of arguments that cause it to only copy the fdbserver binary, where necessary, and to open up a listener. The main container will load its configuration file, launch the configured server processes, and open a listener. That listener will be an HTTP server that provides an API for checking the hashes of files and other local configuration, similar to the API the sidecar currently provides. However, the fdb-kubernetes-monitor API will be not cover all use cases of the sidecar API. It will provide the following API calls:
```
johscheuer, line 34: Should we plan to secure the API with TLS and auth?

docs/design/unified_sidecar_and_process_monitor.md
```diff
@@ -0,0 +1,118 @@
...
+
+* `get_current_configuration` - returns the current process configuration
+* `check_hash` - check the hash of a file inside its configuration directory
+* `ready` - returns an `OK` response
+* `substitutions` - returns the values for any environment variables that are used in the process configuration
```
johscheuer, line 39: I would also add a prometheus `/metrics` endpoint.

docs/design/unified_sidecar_and_process_monitor.md
```diff
@@ -0,0 +1,118 @@
...
+Let's assume this is run with a process count of 2, that 6.3.0 is the version of the main container, and that the following environment variables are set:
+
+	FDB_PUBLIC_IP=127.0.0.1
+	FDB_POD_IP=127.0.0.2
+	FDB_ZONE_ID=zone1
+	FDB_INSTANCE_ID=storage-1
+
+This would spawn two processes, with the following configuration:
+
+	/usr/bin/fdbserver --public-address 127.0.0.1:4500:tls --listen-address 127.0.0.2:4500:tls --datadir /var/fdb/data/1 --class storage --locality-zoneid zone1 --locality_instance_id storage-1 --locality-procerss-id storage-1-1
```
johscheuer, line 116:
```suggestion
	/usr/bin/fdbserver --public-address 127.0.0.1:4500:tls --listen-address 127.0.0.2:4500:tls --datadir /var/fdb/data/1 --class storage --locality-zoneid zone1 --locality_instance_id storage-1 --locality_process_id storage-1-1
```

docs/design/unified_sidecar_and_process_monitor.md
```diff
@@ -0,0 +1,118 @@
...
+
+	/usr/bin/fdbserver --public-address 127.0.0.1:4502:tls --listen-address 127.0.0.2:4502:tls --datadir /var/fdb/data/2 --class storage --locality-zoneid zone1 --locality_instance_id storage-1 --locality-procerss-id storage-1-2
```
johscheuer, line 118:
```suggestion
	/usr/bin/fdbserver --public-address 127.0.0.1:4502:tls --listen-address 127.0.0.2:4502:tls --datadir /var/fdb/data/2 --class storage --locality-zoneid zone1 --locality_instance_id storage-1 --locality_process_id storage-1-2
```

docs/design/unified_sidecar_and_process_monitor.md
```diff
@@ -0,0 +1,118 @@
...
+				{"type": "Environment", "source": "FDB_INSTANCE_ID"},
+				{"value": "-"},
+				{"type": "ProcessNumber"}
+			},
+		]
+	}
+
+Let's assume this is run with a process count of 2, that 6.3.0 is the version of the main container, and that the following environment variables are set:
+
+	FDB_PUBLIC_IP=127.0.0.1
```
johscheuer, line 109: I would rather use here `10.0.0.1` and `192.168.0.1` as Pod IP to make it more obvious.

docs/design/unified_sidecar_and_process_monitor.md
```diff
@@ -0,0 +1,118 @@
...
+	ArgumentType enum {
+		Literal - A value provided in the argument
+		Concatenante - A combination of other arguments
+		Environment - A value from an environment variable
+		ProcessNumber - A value calculated from the process number. The processes are numbered starting from 1.
+	}
+
+Example configuration, in JSON form:
+
+	{
```
johscheuer, line 70: NIT: can we use ```json here?

docs/design/unified_sidecar_and_process_monitor.md
```diff
@@ -0,0 +1,118 @@
...
+
+fdb-kubernetes-monitor will watch the configuration file for changes and automatically reload the desired process state, but will not automatically kill processes. It will watch for its child processes to die and relaunch them with the latest configuration. It will also do appropriate signal handling for SIGKILL and SIGTERM images in accordance with Docker best practices. The configuration file will be designed such that all of the processes of a single machine class can share a configuration file, with environment variables for command-line arguments that vary between processes. fdb-kubernetes-monitor will have an option for how many fdbserver processes to spawn, and the configuration file will support defining arguments that vary between different processes in the pod based on the process number.
```
johscheuer, line 22: Could you add some more information how a change is propagated? The current design states `..., but will not automatically kill processes. It will watch for its child processes to die and relaunch them with the latest configuration.` Why should the child process die?

docs/design/unified_sidecar_and_process_monitor.md
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+Example configuration, in JSON form:
+
+	{
+		"version": "6.3.0",
+		"arguments": [
+			{"value": "--public-address"},
+			{"type": "Concatenate", "values": [
+				{"type": "EnvVar", "source": "FDB_PUBLIC_IP"},
```
johscheuer, line 75:
```suggestion
				{"type": "Environment", "source": "FDB_PUBLIC_IP"},
```

docs/design/unified_sidecar_and_process_monitor.md
```diff
@@ -0,0 +1,118 @@
...
+				{"value": ":"},
+				{"type": "ProcessNumber", "offset": 4498, "multiplier": 2},
+				{"value": ":tls"}
+			]},
+			{"value": "--listen-address"},
+			{"type": "Concatenate", "values": [
+				{"type": "EnvVar", "source": "FDB_POD_IP"},
```
johscheuer, line 82:
```suggestion
				{"type": "Environment", "source": "FDB_POD_IP"},
```

docs/design/unified_sidecar_and_process_monitor.md
```diff
@@ -0,0 +1,118 @@
...
+			{"value": "--locality-process-id"},
+			{"type": "Concatenate", "values": [
+				{"type": "Environment", "source": "FDB_INSTANCE_ID"},
+				{"value": "-"},
+				{"type": "ProcessNumber"}
+			},
+		]
+	}
+
+Let's assume this is run with a process count of 2, that 6.3.0 is the version of the main container, and that the following environment variables are set:
```
johscheuer, line 107: I assume that the process count is defined as env variable or as an argument? Or how is the value propagated?

docs/design/unified_sidecar_and_process_monitor.md
sears: Also, why watch for changes to configuration file instead of only re-reading it whenever it's time to take an action?  (Since it won't be automatically killing processes)

brownleej: That's a good point. I think we're better off not reloading the configuration automatically, so instead we would want an API to poll for (but not reload) changes, and an API to reload the changes immediately before a kill. This would also help limit the risk of an organic process failure causing it to come up on config that is different from the rest of the processes.

brownleej:
I'm gonna get back to this design this week and take a look at all the comments in detail. I made the mistake of posting this right before winter break, which meant it was out of my mind by the time I got back in January 😳

docs/design/unified_sidecar_and_process_monitor.md
johscheuer: I would prefer a solution where we don't need an API and can rely on the mechanisms from Kubernetes e.g. update the status of the object with the current version(s), config-file versions etc. so that we don't need to use an API. That should make the system more reliable if there is a network partition between the operator and some Pods.

docs/design/unified_sidecar_and_process_monitor.md
brownleej: I think we should save that for a follow-up design, since it introduces more questions about the scope of the metrics we're going to expose through this API.

docs/design/unified_sidecar_and_process_monitor.md
brownleej: We're going to need connectivity between the operator and the pods in order to do the kills, and I don't think there's anything in the Kubernetes API that could give us kills that are as reliable and synchronized as doing the kills through the CLI or client library.

docs/design/unified_sidecar_and_process_monitor.md
brownleej: It's a command-line option. I've added some language in an earlier section to clarify that.

docs/design/unified_sidecar_and_process_monitor.md
brownleej: Yes. I've added some language to cover that.

docs/design/unified_sidecar_and_process_monitor.md
johscheuer: Sounds good

docs/design/unified_sidecar_and_process_monitor.md
johscheuer: That's right but I see them as different things. Pushing/checking the "version" of the binaries and ConfigMap requires the operator to talk to all Pods (e.g. needs connectivity to all of them) whereas the `fdbcli --exec `kill ..` command only requires access to one (?) or a few Pods.

johscheuer (commented):
LGTM, I still think we should directly design everything in a way that the operator doesn't need to call an API.

docs/design/unified_sidecar_and_process_monitor.md
```diff
@@ -0,0 +1,123 @@
...
+
+fdb-kubernetes-monitor will log its activity to standard output, and will have an option to also log to a file.
+
+fdb-kubernetes-monitor will ship in an image called `foundationdb-kubernetes`. This image will be versioned with FoundationDB, and will include multiple client libraries, like the `foundationdb-kubernetes-sidecar` image currently does. The first `foundationdb-kubernetes` image we cut for a release will have the FoundationDB version as its tag. For subsequent updates to the image that do not involve a FoundationDB update, we will append a build number, of the format `-2`. For instance, we would have images in a series like `6.3.0`, `6.3.0-2`, `6.3.0-3`, and so on.
```
apkar, line 28: It makes sense to name the first one `6.3.0-1` instead of 6.3.0, to keep it consistent with following releases.

docs/design/unified_sidecar_and_process_monitor.md
```diff
@@ -0,0 +1,123 @@
...
+
+In order to keep the new image lightweight, and allow the development team for the operator to easily manage the development on the monitor, we will write the fdb-kubernetes-monitor process in go.
+
+We will have an option in the cluster spec to use the new launcher, which will be off by default. We will have an entire major release of FoundationDB where we publish both the old sidecar image and the new kubernetes image, to aid with the transition. For any version of FoundationDB in that major release, the user will have the option of which launch strategy to use. Upon the next major release, we will stop publishing the old sidecar image, which means users will need to switch to the new launcher before that release.
```
apkar, line 32: Not sure if we made a decision on major release cadence for FDB. 6.0 to 7.0 took a while. Does it make sense to be bit more aggressive here and support this only for a minor release? Operator is still relatively new, it is acceptable to be bit more aggressive.

docs/design/unified_sidecar_and_process_monitor.md
```diff
@@ -0,0 +1,123 @@
...
+	ArgumentType enum {
+		Literal - A value provided in the argument
+		Concatenante - A combination of other arguments
+		Environment - A value from an environment variable
+		ProcessNumber - A value calculated from the process number. The processes are numbered starting from 1.
+	}
+
+Example configuration, in JSON form:
+
+```json
```
apkar, line 73: As the goal/scope of this process is to spawn `fdbserver` processes only, not generic process spawning - did you consider a configuration model where only defined fields/arguments accepted? If so, what are the reasons to not go for it.

docs/design/unified_sidecar_and_process_monitor.md
apkar, line 28: How does the release process for `foundationdb-kubernetes` image work with FDB server release process pipeline?

How well these changes work with ongoing FDB server CI/CD work?

apkar, line 28: Why do we need FDB client libraries in this image?

docs/design/unified_sidecar_and_process_monitor.md
brownleej: I believe the `kill` command requires the operator to be able to reach every pod that it is trying to kill.

docs/design/unified_sidecar_and_process_monitor.md
brownleej: In practice, updates to the sidecar within a single FDB release are very rare, so there may be a benefit to having simpler version numbers for the common case where there is a single sidecar release for an FDB version.

docs/design/unified_sidecar_and_process_monitor.md
brownleej: I think we can make it the plan of record to have it be a single minor release. We can get feedback before dropping it completely.

docs/design/unified_sidecar_and_process_monitor.md
brownleej: I think that would run the risk of creating more brittleness and duplicate work if we need to add new arguments for new versions of FDB. For instance, if FDB 7.1 involves a new flag that users may want to use, we would need to update the sidecar to permit it. We would also need wildcard support for fields like `--locality_*`, and maybe `--knob_*` as well.

docs/design/unified_sidecar_and_process_monitor.md
brownleej: I think it will be basically the same as the `foundationdb-kubernetes-sidecar` image that we build today. The pipeline that produces the `foundationdb` image will also have to produce the `foundationdb-kubernetes` image, and we will need an option to produce a new build of the latter image without rebuilding the former image.

brownleej: We have a recommended pattern where people can use the current sidecar image as an init container to update their libraries to support a new version of FDB without having to rebuild their application's image. We've started using this pattern in the default deployments for the operator itself as well.

docs/design/unified_sidecar_and_process_monitor.md
brownleej: I think you're right about the core idea of reducing this interaction and relying on the Kubernetes API as a means of expressing this state. I'll work on modifying this design to accommodate that.

brownleej: I updated this to use the Kubernetes API to coordinate this state. As part of that, the design still has the launcher proactively watch for changes, since I don't think that a flow where the operator tells the launcher to reload configuration will work as well without a direct API to the launcher.

docs/design/unified_sidecar_and_process_monitor.md
apkar: If we plan to use `6.3.0-2` at some point, I assume we have code for it. I prefer to make it the standard format rather than having two formats. Having said that, I do not have strong preference, take this as a suggestion and do as it fits.

apkar (approved):
Couple of soft comments (non blockers), take them into consideration. Otherwise, no concerns with merge.

PR #453, opened 2021-01-07
Add devcontainer for GitHub codespaces

funkypenguin:
Hey guys!

I had a horrible time trying to get the dev environment dependencies working on my MacOS desktop, for my recent PRs (https://github.com/FoundationDB/fdb-kubernetes-operator/pull/425).

This PR contributes a `.devcontainer` folder, which provides the dev environment within GitHub Codespaces - I used it to run `make clean all` on my recent PR, it worked great :)

You may want to make some improvements (_specifically to the fact that golang:1 thinks that the SSL cert for https://foundationdb-origin.apple.com contains an untrusted, self-signed cert_), but I hope it'll be useful ;)

Cheers!
D

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,118 @@
+#-------------------------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
+#-------------------------------------------------------------------------------------------------------------
+
+# This Dockerfile copied from https://github.com/alitari/vscode-kubebuilder
+
+FROM golang:1
+
+# Avoid warnings by switching to noninteractive
+ENV DEBIAN_FRONTEND=noninteractive
+
+# This Dockerfile adds a non-root user with sudo access. Use the "remoteUser"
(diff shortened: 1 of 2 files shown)


johscheuer:
Could you elaborate a little bit more why you had a horrible time setting up everything? I would rather update the documentation if there are some steps missing to help everyone. Since GitHub Codespaces is in "early access" I'm not really sure it makes that much sense to add this here but maybe @brownleej has a different opinion on this. The issue with the certificate is a well-known problem that we hopefully can fix in the near time https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/Dockerfile#L10

funkypenguin:
Sure - kube-builder was the big culprit. Apparently installing it via homebrew doesn't install other binaries such as etcd, and short of manually inserting the binaries yourself, there's not an easy answer -> https://github.com/Homebrew/homebrew-core/issues/44175

funkypenguin:
Oh, and I was bitten by that recent kustomize bug on home brew too...

johscheuer:
I think we could add a hint regarding Kube-builder and hopefully there are not too many kustomize bugs anymore (we should probably document the currently used version).

johscheuer:
I'll close that since we currently don't plan to support the codespaces.

Issue #456, opened 2021-01-13
Nil pointer exception when fetching storage servers


JoshuaMcManus:
The operator entered a crash loop when trying to iterate over a list of containers in a pod. The pertinent line of code can be found here: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/bf7f8844a2a9a65dff385a527e7315fdd90cdffd/controllers/pod_models.go#L1227-L1236

Here is the entire stack trace:
```
E0113 20:55:41.125922       1 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 615 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x148e460, 0x22a7340)
	/go/pkg/mod/k8s.io/apimachinery@v0.17.0/pkg/util/runtime/runtime.go:74 +0xa3
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	/go/pkg/mod/k8s.io/apimachinery@v0.17.0/pkg/util/runtime/runtime.go:48 +0x82
panic(0x148e460, 0x22a7340)
	/usr/local/go/src/runtime/panic.go:679 +0x1b2
github.com/FoundationDB/fdb-kubernetes-operator/controllers.getStorageServersPerPodForPod(0x0, 0xc0019f2f30, 0x1647661, 0xf)
	/go/pkg/mod/github.com/!foundation!d!b/fdb-kubernetes-operator@v0.25.1/controllers/pod_models.go:1230 +0x3a
github.com/FoundationDB/fdb-kubernetes-operator/controllers.getStorageServersPerPodForInstance(...)
	/go/pkg/mod/github.com/!foundation!d!b/fdb-kubernetes-operator@v0.25.1/controllers/pod_models.go:1224
github.com/FoundationDB/fdb-kubernetes-operator/controllers.UpdateStatus.Reconcile(0xc0007f2ab0, 0x1863fa0, 0xc000106210, 0xc001d44800, 0x3000100, 0x0, 0x0)
	/go/pkg/mod/github.com/!foundation!d!b/fdb-kubernetes-operator@v0.25.1/controllers/update_status.go:147 +0x2211
[...]

PR #457, opened 2021-01-14
Check if instance contains Pod before continuing

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/456


brownleej (approved):
👍

PR #461, opened 2021-01-19
Add a timeout to the transactions for managing locks.

### LINKED ISSUE #459: Adding explicit timeouts to lock acquisition transactions in the operator
We've seen some suspicious cases where the operator hangs that may be due to the lock acquisition getting stuck. I think we should make sure the transactions for lock acquisition have explicit timeouts, and test the behavior when the database is unavailable.

We should also add more logging before acquiring a lock so we can see clear before-and-after messages to identify when we're stuck in this state.

brownleej:
Add more logging when taking locks.
Fix the priorities in the database config in the multi-DC sample.

Resolves #459

### DIFF
--- a/config/samples/multi_dc/final.yaml
+++ b/config/samples/multi_dc/final.yaml
@@ -28,15 +28,17 @@ spec:
     regions:
       - datacenters:
           - id: dc1
+            priority: 1
           - id: dc3
             satellite: 1
-        priority: 1
+            priority: 1
         satellite_logs: 3
       - datacenters:
           - id: dc2
+            priority: 0
           - id: dc3
             satellite: 1
-        priority: 0
+            priority: 1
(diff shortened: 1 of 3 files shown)


config/samples/multi_dc/final.yaml
```diff
@@ -28,15 +28,17 @@ spec:
...
-        priority: 1
+            priority: 1
         satellite_logs: 3
       - datacenters:
           - id: dc2
+            priority: 0
           - id: dc3
             satellite: 1
-        priority: 0
+            priority: 1
```
johscheuer, line 41: Is this an intended change from `priority: 0` to `priority: 1`?

johscheuer (commented):
Just one question the rest LGTM 👍

PR #462, opened 2021-01-19
Add an exec subcommand for the kubectl plugin.

### LINKED ISSUE #455: FDB exec
It will be pretty useful for debugging if we can specify `kubectl fdb exec --cluster sample-cluster` and we will exec into one of the Pods of the cluster (e.g. to `kubectl exec deployment/test`). This will reduce the steps to connect to a FDB cluster running on Kubernetes.

brownleej:
Resolves #455

### DIFF
--- a/kubectl-fdb/cmd/exec.go
+++ b/kubectl-fdb/cmd/exec.go
@@ -24,6 +24,7 @@ import (
 	ctx "context"
 	"fmt"
 	"log"
+	"math/rand"
 	"os"
 	"os/exec"
 
@@ -73,12 +74,7 @@ func newExecCmd(streams genericclioptions.IOStreams) *cobra.Command {
 				return err
 			}
 
-			context, err := cmd.Flags().GetString("context")
-			if err != nil {
-				return err
-			}
-
-			err = runExec(kubeClient, cluster, context, namespace, args)
+			err = runExec(kubeClient, cluster, *o.configFlags.Context, namespace, args)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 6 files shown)


kubectl-fdb/cmd/exec.go
```diff
@@ -0,0 +1,165 @@
...
+		client.InNamespace(namespace),
+		client.MatchingLabelsSelector{Selector: labels.NewSelector().Add(*clusterRequirement, *processClassRequirement)},
+	)
+	if err != nil {
+		return exec.Cmd{}, err
+	}
+	if len(pods.Items) == 0 {
+		return exec.Cmd{}, fmt.Errorf("No pods found for cluster %s", clusterName)
+	}
+	kubectlPath, err := exec.LookPath("kubectl")
```
johscheuer, line 130: That's a good idea to use `kubectl` the last time I wanted to use the `exec` "API it was not very well documented and a little bit "experimental". For me using `kubectl` is fine since this is a requirement for the plugin anyway.

kubectl-fdb/cmd/exec.go
```diff
@@ -0,0 +1,165 @@
...
+			if err != nil {
+				return err
+			}
+
+			namespace, err := getNamespace(*o.configFlags.Namespace)
+			if err != nil {
+				return err
+			}
+
+			context, err := cmd.Flags().GetString("context")
```
johscheuer, line 76: We already have the context in `o.configFlags.Context`

kubectl-fdb/cmd/exec.go
```diff
@@ -0,0 +1,165 @@
...
+	if err != nil {
+		return exec.Cmd{}, err
+	}
+
+	args := []string{kubectlPath}
+	if context != "" {
+		args = append(args, "--context", context)
+	}
+
+	args = append(args, "--namespace", namespace, "exec", "-it", pods.Items[0].Name)
```
johscheuer, line 140: We may want to choose a Pod randomly and not always the first one? I'm not quite sure but I would assume that the API sorts the Pods and if a Pod is in a strange state we could end up trying to `exec` to the same Pod the whole time.

kubectl-fdb/cmd/exec_test.go
```diff
@@ -0,0 +1,130 @@
...
+	tt := []struct {
+		Name          string
+		ClusterName   string
+		Context       string
+		Command       []string
+		ExpectedArgs  []string
+		ExpectedError string
+	}{
+		{
+			Name:         "Remove instance with valid pod",
```
johscheuer, line 77: The name of the test needs to be adjusted?

johscheuer (changes requested):
Some comments otherwise LGTM 👍

PR #463, opened 2021-01-22
Move tracking of removal state into the process group status

### LINKED ISSUE #413: Moving removal information to the process group list
As part of the Process Group Status work, we should move the information from the `pendingRemovals` list into the process group status. The `pendingRemovals` list will be deprecated.

brownleej:
Resolves #413


controllers/exclude_instances.go
```diff
@@ -40,9 +40,16 @@ func (e ExcludeInstances) Reconcile(r *FoundationDBClusterReconciler, context ct
 	}
 	defer adminClient.Close()
 
-	addresses := make([]string, 0, len(cluster.Status.PendingRemovals))
+	removalCount := 0
+	for _, processGroup := range cluster.Status.ProcessGroups {
```
johscheuer, line 44: Instead of using a counter here we could create a new slice only containing the processGroups marked for removal, this way we don't have to traverse the processGroups twice.

controllers/include_instances.go
```diff
@@ -40,10 +40,17 @@ func (i IncludeInstances) Reconcile(r *FoundationDBClusterReconciler, context ct
 	}
 	defer adminClient.Close()
 
-	addresses := make([]string, 0, len(cluster.Status.PendingRemovals))
-	for _, state := range cluster.Status.PendingRemovals {
-		if state.Address != "" {
-			addresses = append(addresses, state.Address)
+	addresses := make([]string, 0)
+
+	needsUpdate := false
```
johscheuer, line 45: I would prefer if we could use the same name to indicate if the status needs an update above we use `hasStatusUpdates`  we should either change it here or in `confirm_exclusion_completion.go`.

controllers/remove_pods.go
```diff
@@ -35,15 +35,25 @@ type RemovePods struct{}
 
 // Reconcile runs the reconciler's work.
 func (u RemovePods) Reconcile(r *FoundationDBClusterReconciler, context ctx.Context, cluster *fdbtypes.FoundationDBCluster) (bool, error) {
-	if len(cluster.Status.PendingRemovals) == 0 {
+
+	processesToRemove := make([]string, 0, len(cluster.Status.ProcessGroups))
```
johscheuer, line 39: NIT: shouldn't this be `processesGroupsToRemove `

controllers/update_status.go
```diff
@@ -629,3 +652,15 @@ func validateInstance(r *FoundationDBClusterReconciler, context ctx.Context, clu
 
 	return false, false, needsSidecarConfInConfigMap, nil
 }
+
+func cleanAddressList(slice []string) []string {
```
johscheuer, line 656: Could you add a short comment here that we remove duplicates and and empty addresses?

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -157,6 +157,9 @@ type FoundationDBClusterSpec struct {
 	// client compatibility when performing an upgrade.
 	IgnoreUpgradabilityChecks bool `json:"ignoreUpgradabilityChecks,omitempty"`
 
+	// Buggify defines settings for injecting faults into a cluster for testing.
+	Buggify BuggifyConfig `json:"buggify,omitempty"`
```
johscheuer, line 161: Do you expect to use this in the integration tests or more in (manual) e2e tests? If we want to use this for e2e tests I would rather look into https://chaos-mesh.org or something similar to reduce the complexity in the code.

johscheuer (changes requested):
Looks overall good just some minor comments.

controllers/exclude_instances.go
brownleej: I think we're going to have to traverse it twice no matter what, because we want to ignore addresses that have already been excluded.

api/v1beta1/foundationdbcluster_types.go
brownleej: I think we want to be able to use this in integration tests, though I added it in this PR for the purpose of manual testing.

controllers/exclude_instances.go
johscheuer: That's right but we would traverse the complete list once and a second time the new list which contains only the removed once (with the assumption that in most cases the complete list will be much smaller than the list with the ProcessGroups to remove)

johscheuer: But we can also optimise this in a next step.

johscheuer (approved):
LGTM 👍

Issue #75, opened 2020-01-10
Provide up-to-date docker image


PierreZ:
Hi!

I have started playing with this operator and was stuck on this error `quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'`

```
[...]

PR #470, opened 2021-01-29
Build and push docker images for releases and on master

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/75

You can see the successful workflow here: https://github.com/johscheuer/fdb-kubernetes-operator/actions/runs/522035788

There is a bunch of stuff we should discuss after the weekend to see if we want to use the GitHub solution.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -12,13 +12,10 @@ jobs:
       matrix:
         image:
           - fdb-kubernetes-operator
-          - foundationdb-kubernetes-sidecar
           - fdb-data-loader
         include:
           - image: fdb-kubernetes-operator
             context: ./
-          - image: foundationdb-kubernetes-sidecar
-            context: ./foundationdb-kubernetes-sidecar
           - image: fdb-data-loader
             context: ./sample-apps/data-loader
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/master.yml
```diff
@@ -0,0 +1,50 @@
...
+jobs:
+  push_images:
+    name: Push Docker images
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        image:
+          - fdb-kubernetes-operator
+          - foundationdb-kubernetes-sidecar
+          - fdb-data-loader
```
brownleej, line 16: I think the only image we want to publish here is the operator. The sidecar is normally published as part of the FDB release process, since it has to be versioned with FDB. We don't regularly publish the data loader image, but I guess there's no harm in it unless it confuses people.

.github/workflows/release.yaml
```diff
@@ -71,3 +71,53 @@ jobs:
...
+        uses: docker/login-action@v1
+        with:
+          registry: docker.pkg.github.com
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and push to registry
+        uses: docker/build-push-action@v2
+        with:
+          context: ${{ matrix.context }}
+          # todo adjust to FoundationDB
```
brownleej, line 112: What does "adjust to FoundationDB" mean here?

.github/workflows/master.yml
```diff
@@ -0,0 +1,50 @@
...
+          context: ${{ matrix.context }}
+          tags: docker.pkg.github.com/${{ github.repository }}/${{ matrix.image }}:latest
+          # push: true
+          # Workaround: otherwise we have a multi-arch image which is not supported by the deprecated
+          # GitHub packages but GHCR doesn't support the workflow credentials and is only in public beta :(
+          # https://github.com/docker/build-push-action/blob/master/TROUBLESHOOTING.md#errors-on-pushing-to-registry
+          outputs: type=docker
+      # Really? -> https://github.community/t/docker-pull-from-public-github-package-registry-fail-with-no-basic-auth-credentials-error/16358/85
+      - name: Push the image with docker
+        run: docker push docker.pkg.github.com/${{ github.repository }}/${{ matrix.image }}:latest
```
brownleej, line 50: Does this mean that the images will be available under `docker.pkg.github.com` instead of `docker.com`, or is that a proxy?

brownleej (changes requested):

.github/workflows/master.yml
johscheuer: It will be published to the GitHub packages which will be deprecated by the GitHub container registry (which doesn't support the ` ${{ secrets.GITHUB_TOKEN }}` currently and is in beta. Also the GitHub packages doesn't allow anonymous pull (so we should either use DockerHub or GitHub container registry).

.github/workflows/release.yaml
johscheuer: This was an old comment, I was not sure if I could use the context in all places.

.github/workflows/master.yml
brownleej: If possible, I think we should push to DockerHub so that we don't disrupt any existing usage.

.github/workflows/master.yml
johscheuer: I removed the sidecar but I think it makes sense to publish the data-loader as part of the release.

johscheuer:
We just need https://github.com/FoundationDB/fdb-kubernetes-operator/pull/492 before we can merge this one.

brownleej (approved):
👍

PR #475, opened 2021-02-02
Run the race detector in our test suite.

### LINKED ISSUE #406: Code not checked with the race detector
Go has an excellent race detector; this can be run various ways, for instance in this code base via 

```shell
ginkgo -race ./...
```

I don't know whether there are races in the operator itself, but the test suite is not race-clean (see e.g. #405), and without that, we cannot say anything about the operator at this stage.

brownleej:
Resolves #406 

This makes larger-scale changes to our test suite to make it more reliable. With this PR, we run the Reconcile method explicitly in the tests, rather than relying on the event stream. It also uses a custom mock Kubernetes client instead of using the standard controller-runtime client, so that we do not have to worry about nondeterministic behavior from the cache.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,14 +39,14 @@ clean:
 # Run tests
 test:
 ifneq "$(SKIP_TEST)" "1"
-	go test ./... -coverprofile cover.out
+	go test -race ./... -coverprofile cover.out
 endif
 
 test_if_changed: cover.out
 
 cover.out: ${GO_ALL} ${MANIFESTS}
 ifneq "$(SKIP_TEST)" "1"
-	go test ./... -coverprofile cover.out -tags test
+	go test -race ./... -coverprofile cover.out -tags test
 endif
 
 # Build manager binary
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,10 @@ GO_ALL=${GO_SRC} ${GENERATED_GO}
 MANIFESTS=config/crd/bases/apps.foundationdb.org_foundationdbbackups.yaml config/crd/bases/apps.foundationdb.org_foundationdbclusters.yaml config/crd/bases/apps.foundationdb.org_foundationdbrestores.yaml
 CONTROLLER_GEN=$(GOBIN)/controller-gen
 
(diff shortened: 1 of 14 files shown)


mock-kubernetes-client/client/client.go
```diff
@@ -0,0 +1,668 @@
+/*
```
johscheuer, line 1: The benefit of writing our own mock instead of using the official one (that we use in the plugin tests) is that it's more flexible?

brownleej: Is there an official mock for the controller-runtime client?

Makefile
```diff
@@ -24,6 +24,10 @@ GO_ALL=${GO_SRC} ${GENERATED_GO}
 MANIFESTS=config/crd/bases/apps.foundationdb.org_foundationdbbackups.yaml config/crd/bases/apps.foundationdb.org_foundationdbclusters.yaml config/crd/bases/apps.foundationdb.org_foundationdbrestores.yaml
 CONTROLLER_GEN=$(GOBIN)/controller-gen
 
+ifeq "$(TEST_RACE_CONDITIONS)" "1"
+	go_test_flags := $(go_test_flags) -race
```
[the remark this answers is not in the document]
brownleej: It makes the tests significantly slower.

mock-kubernetes-client/client/client.go
brownleej: I've taken a look at that mock, and there are a handful of things that our tests are relying on that the official mock doesn't implement: generations, UIDs, field selectors in list calls, service IPs. There are some further things that we could add to a custom mock to improve our testing, such as pod IPs, scheduling failures, and termination failures. I think it's worth investing in our own mock.

johscheuer: Seems reasonable.

Makefile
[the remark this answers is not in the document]
johscheuer: Okay, than it makes sense to make it optional.

johscheuer (approved):
LGM 👍

PR #477, opened 2021-02-03
Parallelize pod deletion

### LINKED ISSUE #472: Preventing stuck pod from blocking deletion of other pods
When we delete pods as part of the removal flow, we go through the pods in serial, deleting each one and then confirming that it has been fully deleted. This means that if there is one pod that is stuck in a terminating state, it can block the operator from initiating the deletion on the other pods. I think that we should make it so that we always go through the full list, even if one of the pods is stuck in terminating, and return the success or failure result based on the evaluation of all the pods.

We should also merge the inclusion logic into this loop, so that we can re-include addresses and remove the old entries from the process group list even when there is a pod stuck in terminating.

brownleej:
Resolves #472 

I was able to test this manually by setting a longer termination grace period on some of the pods, and then replacing two pods with different grace periods. I think we should make it a priority to add more unit tests for different scenarios here once we merge the other PR with test suite improvements.


controllers/remove_pods.go
```diff
@@ -168,6 +182,55 @@ func confirmPodRemoval(r *FoundationDBClusterReconciler, context ctx.Context, cl
...
+		} else {
+			processGroups = append(processGroups, processGroup)
+		}
+	}
+
+	if len(addresses) > 0 {
+		r.Recorder.Event(cluster, "Normal", "IncludingInstances", fmt.Sprintf("Including removed processes: %v", addresses))
+	}
+
+	err = adminClient.IncludeInstances(addresses)
```
johscheuer, line 210: Should we move the include call into the if statement above? I assume that the `adminClient.IncludeInstances` just will do nothing but it would be cleaner.

controllers/remove_pods.go
```diff
@@ -168,6 +182,55 @@ func confirmPodRemoval(r *FoundationDBClusterReconciler, context ctx.Context, cl
...
+	if err != nil {
+		return err
+	}
+
+	needsSpecUpdate := cluster.Spec.PendingRemovals != nil
+	if needsSpecUpdate {
```
johscheuer, line 216: We can merge these lines, since `needsSpecUpdate` is never used.

johscheuer (changes requested):
Just some minor comments.

Issue #484, opened 2021-02-05
Set the version for fdb plugin to match the tag


johscheuer:
Currently the version for the kubectl fdb plugin is not set correctly to match the released tag:

```bash
kubectl fdb version
kubectl-fdb: v0.23.1 # should be 0.27.1
```

We should ensure that we set the version to match the tag before building it. This can be done simply during the release action.

Issue #483, opened 2021-02-05
Fetch the operator version correctly


johscheuer:
I just notices that if someone is running the operator with multiple containers the `kubectl fdb version` plugin currently prints the tag of the first container (which might be a different container) e.g.:

```bash
kubectl fdb version
kubectl-fdb: v0.23.1
foundationdb-operator: 1.1.1
```

PR #485, opened 2021-02-06
Fetch version plugin

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/484
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/483

This PR ensures we fetch the operator version from the correct container and during the release process we set the tag for the plugin.


brownleej (approved):
👍

Issue #487, opened 2021-02-07
Getting status JSON through the client library rather than the CLI


brownleej:
Getting status JSON from the CLI can be fragile, since the output from the CLI is not guaranteed to be exclusively a valid JSON document. For instance, if the database is slow, the status JSON command can emit a warning message. Instead, we could try getting it through the client library from the `\xff\xff/status/json` key.

PR #488, opened 2021-02-07
Read status directly from FDB

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/487

I will open some follow ups to evaluate how we want to test the interaction with FDB and commands that we run with `fdbcli` currently we have some mock but we don't actually test "real" logic.


brownleej (approved):
👍

PR #489, opened 2021-02-08
Refactor resource creation

### LINKED ISSUE #411: Decoupling the creation of services, PVCs, and pods
As part of the Process Group Status work, we should decouple the creation of services, PVCs, and pods. This will make each action simpler and give us more reliable behavior when one resource gets deleted. We should first add process groups to the status to make the process counts match, and then add missing resources based on the process group list. We should create the resources in the order given above: service, then PVC, then pod.

### LINKED ISSUE #471: Replacing ProcessCounts in the status with the ProcessGroup details
Now that we're tracking all process groups in the process group list, the process counts are a redundant representation of the same information. I think we should deprecate the process counts field in the status and replace usage of it with the process group list.

brownleej:
Resolves #411
Resolves #471


controllers/add_pods.go
```diff
@@ -75,217 +62,46 @@ func (a AddPods) Reconcile(r *FoundationDBClusterReconciler, context ctx.Context
...
-					if err != nil {
-						return false, err
-					}
-					if len(matchingInstances) == 0 {
-						reusablePvcs[index] = true
-					}
-				}
+			pod, err := GetPod(cluster, processGroup.ProcessClass, idNum)
+			if err != nil {
+				r.Recorder.Event(cluster, "Error", "GetPod", fmt.Sprintf("failed to get the PodSpec for %s/%d with error: %s", processGroup.ProcessClass, idNum, err))
```
johscheuer, line 80: We might want to add the Pod name here (it might be contained in the error message)?

controllers/add_process_groups.go
```diff
@@ -0,0 +1,112 @@
...
+		return false, err
+	}
+	desiredCounts := desiredCountStruct.Map()
+
+	processCounts := make(map[string]int)
+	processGroupIDs := make(map[string]map[int]bool)
+	for _, processGroup := range cluster.Status.ProcessGroups {
+
+		processGroupID := processGroup.ProcessGroupID
+		_, num, err := ParseInstanceID(processGroupID)
```
johscheuer, line 47:
```suggestion
		_, id, err := ParseInstanceID(processGroupID)
```

controllers/add_process_groups.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	hasNewProcessGroups := false
+	for _, processClass := range fdbtypes.ProcessClasses {
+		desiredCount := desiredCounts[processClass]
+		if desiredCount < 0 {
+			desiredCount = 0
+		}
+		newCount := desiredCount - processCounts[processClass]
+		if newCount > 0 {
```
johscheuer, line 71: Can we flip this to be:

```golang
if newCount <= 0 {
  continue
}

```

controllers/add_process_groups.go
```diff
@@ -0,0 +1,112 @@
...
+// Reconcile runs the reconciler's work.
+func (a AddProcessGroups) Reconcile(r *FoundationDBClusterReconciler, context ctx.Context, cluster *fdbtypes.FoundationDBCluster) (bool, error) {
+	desiredCountStruct, err := cluster.GetProcessCountsWithDefaults()
+	if err != nil {
+		return false, err
+	}
+	desiredCounts := desiredCountStruct.Map()
+
+	processCounts := make(map[string]int)
+	processGroupIDs := make(map[string]map[int]bool)
```
johscheuer, line 43:
```suggestion
	processGroupIDs := make(map[string]bool)
```

Using a map as a set should be fine if we use the `processGroupID` (which already contains the processClass and the the processGroupID).

controllers/add_pvcs.go
```diff
@@ -0,0 +1,81 @@
+/*
+ * add_process_groups.go
```
johscheuer, line 2:
```suggestion
 * add_pvcs.go
```

controllers/add_pvcs.go
```diff
@@ -0,0 +1,81 @@
...
+		if err != nil {
+			return false, err
+		}
+
+		pvc, err := GetPvc(cluster, processGroup.ProcessClass, idNum)
+		if err != nil {
+			return false, err
+		}
+
+		if pvc != nil {
```
johscheuer, line 54: What about:

```golang
// process class doesn't use PVC's
if pvc == nil {
  continue
}
...
```

controllers/add_services.go
```diff
@@ -37,24 +37,53 @@ type AddServices struct{}
...
+				service, err := GetService(cluster, processGroup.ProcessClass, idNum)
+				if err != nil {
+					return false, err
+				}
+
+				err = r.Create(context, service)
+
+				if err != nil {
+					return false, err
+				}
```
johscheuer, line 84: We might want to consider to enforce the expected state in the reconcile loop? Currently we only create PVC's/services but validate in a different place if the current state == expected state.

johscheuer (changes requested):

controllers/add_pods.go
brownleej: I'm not sure the pod name will be relevant here, since the errors would come when building the pod spec, and we won't have an existing pod to reference when debugging the error.

controllers/add_process_groups.go
brownleej: What's the advantage of changing this to "id"? We generally refer to it as a number in other uses.

controllers/add_process_groups.go
brownleej: The problem is that the processGroupID contains other things too, specifically the instance ID prefix, and the instance ID prefix is not contained in the pod name. This means that using the process group ID as the key would cause this method to miss duplicate pod names when changing the instance ID prefix.

controllers/add_services.go
brownleej: Which state are you thinking of?

controllers/add_process_groups.go
brownleej: Maybe that suggests we should be using the pod name in this map instead.

controllers/add_process_groups.go
johscheuer: Right it's only the process "number" and not the whole ID

controllers/add_services.go
johscheuer: I just thought if it makes sense to have rather a "reconcile_service" rather than a "add_service" but for now this is fine 👍

johscheuer (approved):
LGTM 👍  feel free to implement https://github.com/FoundationDB/fdb-kubernetes-operator/pull/489#discussion_r573281318 if you want.

PR #490, opened 2021-02-08
Enable leader election by default.

### LINKED ISSUE #325: Enabling leader election by default
We have a flag on the command-line for the manager process that enables leader election. This is off by default. I think we want leader election to be on by default, since it is an important safeguard against conflicting operations.

brownleej:
Resolves #325


johscheuer (approved):
LGTM 👍

Issue #481, opened 2021-02-05
FDB plugin remove instance with Pod name


johscheuer:
Since we have the fdb plugin now to remove instances I would also like to add the possibility to remove instances based on the Pod name rather than the instance ID. During the remove process the plugin would fetch the instance ID information directly from the Pod. This has the benefits that I don't have to lookup the instance ID when I want to remove an instance (Pod).

johscheuer:
Once https://github.com/FoundationDB/fdb-kubernetes-operator/pull/480 is merged I start working on this otherwise I will run in a bunch of conflicts.

Issue #500, opened 2021-02-17
Remove should use patch instead update


johscheuer:
There might be the case where someone is using an old version of the `kubectl fdb remove` plugin command. In order to prevent that the `kubectl fdb remove` changes the `Cluster` spec or deletes fields that are unknown to the old version we should make use of the patch command.

PR #505, opened 2021-02-18
Allow to remove instances by pod name

johscheuer:
fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/481

I made this option the default since I think most operators will use the Pod name rather than the instance name. I also restructured the remove command a little bit so that we have now:

```bash
kubectl fdb -n default remove instances -c cluster instance-1 instance-2
```

So that we can add later on the possibility to remove also clusters, backups, etc.

### DIFF
--- a/kubectl-fdb/cmd/remove.go
+++ b/kubectl-fdb/cmd/remove.go
@@ -1,5 +1,5 @@
 /*
- * remove_instances.go
+ * remove.go
  *
  * This source file is part of the FoundationDB open source project
  *
--- a/kubectl-fdb/cmd/remove.go
+++ b/kubectl-fdb/cmd/remove.go
@@ -39,10 +39,14 @@ func newRemoveCmd(streams genericclioptions.IOStreams, rootCmd *cobra.Command) *
 		},
 		Example: `
 # Remove instances for a cluster in the current namespace
-kubectl fdb remove instances -c cluster instance-1 instance-2
+kubectl fdb remove instances -c cluster pod-1 -i pod-2
 
 # Remove instances for a cluster in the namespace default
-kubectl fdb -n default remove instances -c cluster instance-1 instance-2
+kubectl fdb -n default remove instances -c cluster pod-1 pod-2
+
+# Remove instances for a cluster with the instance ID.
+# The instance ID of a Pod can be fetched with "kubectl get po -L fdb-instance-id"
+kubectl fdb -n default remove instances --use-instance-id -c cluster storage-1 storage-2
 `,
 	}
 
--- a/kubectl-fdb/cmd/remove.go
+++ b/kubectl-fdb/cmd/remove.go
@@ -1,5 +1,5 @@
 /*
- * remove.go
+ * remove_instances.go
  *
  * This source file is part of the FoundationDB open source project
  *
@@ -21,22 +21,9 @@
 package cmd
 
 import (
-	"fmt"
-	"log"
-
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 
-	"github.com/FoundationDB/fdb-kubernetes-operator/controllers"
-	corev1 "k8s.io/api/core/v1"
-
 	"github.com/spf13/cobra"
-	"k8s.io/apimachinery/pkg/runtime"
-	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
-	"sigs.k8s.io/controller-runtime/pkg/client"
-
-	ctx "context"
-
-	fdbtypes "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta1"
 )
 
 func newRemoveCmd(streams genericclioptions.IOStreams, rootCmd *cobra.Command) *cobra.Command {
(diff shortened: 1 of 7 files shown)


kubectl-fdb/cmd/remove.go
```diff
@@ -1,5 +1,5 @@
 /*
- * remove.go
+ * remove_instances.go
```
rbtcollins, line 2: I think this change is accidental?

kubectl-fdb/cmd/remove_instances.go
```diff
@@ -0,0 +1,205 @@
...
+		},
+		Example: `
+# Remove instances for a cluster in the current namespace
+kubectl fdb remove instances -c cluster instance-1 -i instance-2
+
+# Remove instances for a cluster in the namespace default
+kubectl fdb -n default remove instances -c cluster instance-1 instance-2
+
+# Remove instances for a cluster with the instance-id
+kubectl fdb -n default remove instances --use-instance-id -c cluster storage-1 storage-2
```
rbtcollins, line 108: These examples aren't super clear about the non-instance-id vs instance-id values. Perhaps thats just me?

kubectl-fdb/cmd/remove_instances.go
```diff
@@ -0,0 +1,205 @@
...
+`,
+	}
+
+	cmd.Flags().StringP("fdb-cluster", "c", "", "remove instance(s) from the provided cluster.")
+	cmd.Flags().BoolP("exclusion", "e", true, "define if the instances should be remove with exclusion.")
```
rbtcollins, line 113: grammar: "removed with exclusion."

kubectl-fdb/cmd/remove_instances.go
```diff
@@ -0,0 +1,205 @@
...
+	cmd.Flags().Bool("use-instance-id", false, "if set the operator will use the instance ID to remove it rather than the Pod name.")
```
rbtcollins, line 114: I suggest replacing "it" with "pod(s)", as this command may process multiple pods. 

Also Pod here should be lower cased, or Instances should be upper cased for consistency when talking about types, I think?

kubectl-fdb/cmd/remove_instances.go
```diff
@@ -0,0 +1,205 @@
...
+
+	for class, amount := range shrinkMap {
+		cluster.Spec.ProcessCounts.DecreaseCount(class, amount)
+	}
+
+	if !force {
+		confirmed := confirmAction(fmt.Sprintf("Remove %v from cluster %s/%s with exclude: %t and shrink: %t", instances, namespace, clusterName, withExclusion, withShrink))
+		if !confirmed {
+			print("Abort")
+			return nil
```
rbtcollins, line 194: I'm going to assume that returning nil here results in the cli exiting 0, which is probably undesirable. Much better IMO to put that Abort into a custom error and return that, then get a non-zero exit code to support anyone using automation.

kubectl-fdb/cmd/remove_instances.go
```diff
@@ -0,0 +1,205 @@
...
+		}
+	}
+
+	if withExclusion {
+		cluster.Spec.InstancesToRemove = append(cluster.Spec.InstancesToRemove, instances...)
+	} else {
+		cluster.Spec.InstancesToRemoveWithoutExclusion = append(cluster.Spec.InstancesToRemoveWithoutExclusion, instances...)
+	}
+
+	return kubeClient.Update(ctx.TODO(), &cluster)
```
rbtcollins, line 204: Does this do a CAS operation on the clusters previous state? I note that you're performing a list append on the clusters old state here and then updating - which means that, at least as I understand it, multiple kubectl users can collide; a semanticpatch would be better I think (or does kubeClient generate one under the hood?)

rbtcollins (changes requested):
Some suggested feedback done as single comments, sorry ;).

kubectl-fdb/cmd/remove.go
johscheuer: Thanks for catching this one!

kubectl-fdb/cmd/remove_instances.go
johscheuer: You are right, I'll think about how to make these examples more descriptive.

kubectl-fdb/cmd/remove_instances.go
johscheuer: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/500 :) We also had this discussion and I asked some people (without a clear answer) so we decided that we will move to the patch command.

johscheuer:
Thanks for your feedback!

kubectl-fdb/cmd/remove.go
```diff
@@ -21,155 +21,33 @@
...
+		Long:  "Subcommand to remove instances from a given cluster",
 
-			return removeInstances(kubeClient, cluster, instances, namespace, withExclusion, withShrink, force)
+		RunE: func(c *cobra.Command, args []string) error {
+			return c.Help()
 		},
 		Example: `
 # Remove instances for a cluster in the current namespace
-kubectl fdb remove -c cluster -i instance-1 -i instance-2
+kubectl fdb remove instances -c cluster instance-1 instance-2
```
brownleej, line 42: What are the trade-offs between implementing this as subcommands versus implementing this as different arguments to a single `remove` command?

brownleej: Or is the idea that there will be new `remove` subcommands in the future?

kubectl-fdb/cmd/remove.go
```diff
@@ -21,155 +21,33 @@
...
 
 # Remove instances for a cluster in the namespace default
-kubectl fdb -n default remove -c cluster -i instance-1 -i instance-2
+kubectl fdb -n default remove instances -c cluster instance-1 instance-2
```
brownleej, line 45: As I read the `instances` subcommand, this invocation would treat `instance-1` and `instance-2` as pod names. Is that correct? If so, should we update this example?

brownleej (changes requested):

kubectl-fdb/cmd/remove.go
johscheuer: The last one maybe we have one day `kubectl remove cluster` e.g. to delete multi-dc FDB clusters.

kubectl-fdb/cmd/remove.go
johscheuer: Good catch!

brownleej (approved):
👍

PR #509, opened 2021-02-22
Add test for kubectl rootcmd

johscheuer:
This PR is based on https://github.com/FoundationDB/fdb-kubernetes-operator/pull/505

I added a test case for the `root cmd` we can/will add more test cases for the specific command in a follow up. The simple test allows to test the cmd creation (e.g. all flags are set do we only use referenced flags etc. to prevent some mistakes).

### DIFF
--- a/kubectl-fdb/cmd/remove_instances.go
+++ b/kubectl-fdb/cmd/remove_instances.go
@@ -24,6 +24,8 @@ import (
 	"fmt"
 	"log"
 
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 
 	"github.com/FoundationDB/fdb-kubernetes-operator/controllers"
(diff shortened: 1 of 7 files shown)


kubectl-fdb/cmd/remove_instances.go
```diff
@@ -150,6 +155,19 @@ func getInstanceIDsFromPod(kubeClient client.Client, clusterName string, podName
 
 // removeInstances adds instances to the instancesToRemove field
 func removeInstances(kubeClient client.Client, clusterName string, instances []string, namespace string, withExclusion bool, withShrink bool, force bool) error {
+	var cluster fdbtypes.FoundationDBCluster
```
johscheuer, line 158: I moved this check upfront because otherwise the user won't notice that the cluster doesn't exit.

kubectl-fdb/cmd/root_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"testing"
+
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+)
+
+func TestRootCmd(t *testing.T) {
+	// We use these buffers to theoretically check the input/output
+	outBuffer := bytes.Buffer{}
+	errBuffer := bytes.Buffer{}
+	inBuffer := bytes.Buffer{}
```
johscheuer, line 35: e.g. with these buffers we can simulate input(s) and check the according output.

brownleej (approved):
👍

Issue #507, opened 2021-02-19
Idea: voluntary node evacuation


rbtcollins:
This idea came up during internal discussions about dealing with node maintenance in a baremetal cluster. In this (openshift) cluster the nodes are regularly drained, rebooted into a new kernel and then repopulated. We have LocalVolumeProvisioner volumes with storage and log volumes on the nodes, so during this period for any one node, FDB is running with reduced nodes, and has had to re-replicate data to get back up to the desired redundancy level.

Node draining as far as I know is a cordon (unschedulable=true + a NoSchedule taint) + a client side delete to the API to perform the drain with different logic applied depending on user flags.

Our specific configuration is to have 1 storage pod on every node, and then some nodes also have log pods. Running with a pod disruption budget is complicated as a result, because the nodes with 2 stateful pods get covered by the 1 PDB (because coordinators can be on any stateful pod), and then as a result the nodes with 1 log + 1 storage never drain.

It might be nice to have FDB observe the cordon and voluntarily give up the node - I realise this is not the exact semantics of cordon normally, so would suggest that it be an optional behaviour.

There is another thing we can do, and I'm filing a separate bug about it, since possibly these could be complementary.

That said I'm not sure there are other use cases for this behaviour - but this is really meant to spark a discussion. @simenl @larshagencognite

brownleej:
I think this could make sense as an extension of the automatic replacement feature. If a node is tainted or cordoned such that we know that the we couldn't reschedule the pod, the operator might as well replace it proactively. We'd want this behavior to be optional, since people may want to run in environments where the operator wouldn't have node access.

Issue #316, opened 2020-09-03
Create design to prevent user to remove to many Pods


johscheuer:
When thinking about https://github.com/FoundationDB/fdb-kubernetes-operator/issues/315 I came across the issue that the complete cluster was offline (obviously this was a test case) which could happen if a (human) operator has some fat fingers. We should optionally add `PodDIsruptionbudget` for the Pod groups to ensure that the cluster stays available even if someone would try to delete all Pods, see: https://kubernetes.io/docs/tasks/run-application/configure-pdb.

rbtcollins:
(Copied my detail from the dup I filed) - this is much larger than just fat fingers: it has implications on cluster upgrades, node autoscalers, horizontal pod autoscalers in future as well.

Foundation DB requires that a certain number of pods are running to retain quorum; Kubernetes supports not deliberately disrupting those pods if an API object called a PodDisruptionBudget is created in the same namespace as the pods, with a selector that matches the pods in question.

This gets a little tricky because of recruitment - unless it is possible to restrict the coordinator role to a subset of processes (is it?), we need to prevent disrupting (quorum-1) *processes* across all processes; and this needs to take into account the design for multi-storage-pods, which is on my todo to review (does that correctly provide anti-affinity for coordinators?). 

What happens to a cluster if all coordinators are killed atomically, even though the rest of the processes are live?

Anyhow, PDB's, great things, lets write one:

```
apiVersion: policy/v1beta1
kind: PodDisruptionBudget
metadata:
  name: {{fdb-clustername}}
spec:
  minAvailable: {{nodecount-2}}
  selector:
    matchLabels:
      fdbcluster: {{fdb-clustername}}
```

rbtcollins:
We now have some experience running with PDBs (we wrote a quick separate operator that just watches the status cluster health field) + cluster definition. When the cluster is healthy 1 pod disruption vs the planned cluster size is permitted; when it isn't, 0 disruptions are permitted.

Our experience so far:
 - naive PDBs stop nodepool upgrades in public cloud entirely / break that workflow
 - our second generation PDB permits public cloud upgrades but has issues with more constrained environments like private cloud (#507)
 - its probably worth having separate PDBs for stateless and stateful process types, since availability interacts differently to losses, and Kubernetes does not permit PDBs to overlap.

Our first generation PDB experiment just didn't permit disruption, requiring the operator to determine everything, which didn't work very well.
[...]

PR #512, opened 2021-02-22
Restrict coordinators to storage pods

rbtcollins:
This makes it possible to write PDBs that safely permit individual node
draining even when a single pod from each of multiple storage classes
are present on that node. See #507 and #316 for more context.

Possibly this should automatically fall back to using all stateful
storage classes in very small clusters? Or perhaps a configuration
switch?

Signed-off-by: Robert Collins <robert.collins@cognite.com>

### DIFF
--- a/controllers/change_coordinators.go
+++ b/controllers/change_coordinators.go
@@ -87,20 +87,28 @@ func (c ChangeCoordinators) Reconcile(r *FoundationDBClusterReconciler, context 
 		log.Info("Changing coordinators", "namespace", cluster.Namespace, "cluster", cluster.Name)
 		r.Recorder.Event(cluster, corev1.EventTypeNormal, "ChangingCoordinators", "Choosing new coordinators")
 
-		candidates := make([]localityInfo, 0, len(status.Cluster.Processes))
-		for _, process := range status.Cluster.Processes {
-			eligible := !process.Excluded && isStateful(process.ProcessClass) && !cluster.InstanceIsBeingRemoved(process.Locality[fdbtypes.FDBInstanceIDLabel])
-			if eligible {
-				candidates = append(candidates, localityInfoForProcess(process))
-			}
-		}
-
 		coordinatorCount := cluster.DesiredCoordinatorCount()
-		coordinators, err := chooseDistributedProcesses(candidates, coordinatorCount, processSelectionConstraint{
-			HardLimits: map[string]int{fdbtypes.FDBLocalityZoneIDKey: 1},
-		})
+		candidates := make([]localityInfo, 0, len(status.Cluster.Processes))
+		chooseCoordinators := func(candidates []localityInfo) ([]localityInfo, error) {
+			return chooseDistributedProcesses(candidates, coordinatorCount, processSelectionConstraint{
+				HardLimits: map[string]int{fdbtypes.FDBLocalityZoneIDKey: 1},
+			})
+		}
+		// Use all stateful pods if needed, but only storage if possible.
+		candidates = selectCandidates(cluster, status, candidates, fdbtypes.ProcessClassStorage)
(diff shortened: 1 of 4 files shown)


rbtcollins:
I'll fixup the tests tomorrow - but would love some feedback in the interim

johscheuer (commented):
I don't see. any issues with only using storage processes for the coordinators 👍

brownleej:
We need to consider how to handle multi-DC clusters where some DCs only run transaction logs, but we also want to have coordinators in those DCs. The repo has a multi_dc sample that works this way.

johscheuer:
Probably a "simple" way would be to check if storage servers exist in the Kubernetes cluster if yes prefer them otherwise also take TLog?

rbtcollins:
Can I get a more specific pointer to that multi_dc sample?

Yes, I think taking storage servers first and then log servers secondly is sufficient to meet our use case and the multi-dc use case.

johscheuer:
> Can I get a more specific pointer to that multi_dc sample?
> 
> Yes, I think taking storage servers first and then log servers secondly is sufficient to meet our use case and the multi-dc use case.

Here: https://github.com/FoundationDB/fdb-kubernetes-operator/tree/master/config/samples/multi_dc + https://github.com/FoundationDB/fdb-kubernetes-operator/tree/master/config/samples/multi_kc

controllers/change_coordinators.go
```diff
@@ -85,20 +85,37 @@ func (c ChangeCoordinators) Reconcile(r *FoundationDBClusterReconciler, context
 		log.Info("Changing coordinators", "namespace", cluster.Namespace, "cluster", cluster.Name)
 		r.Recorder.Event(cluster, "Normal", "ChangingCoordinators", "Choosing new coordinators")
 
+		coordinatorCount := cluster.DesiredCoordinatorCount()
 		candidates := make([]localityInfo, 0, len(status.Cluster.Processes))
-		for _, process := range status.Cluster.Processes {
-			eligible := !process.Excluded && isStateful(process.ProcessClass) && !cluster.InstanceIsBeingRemoved(process.Locality[FDBInstanceIDLabel])
-			if eligible {
-				candidates = append(candidates, localityInfoForProcess(process))
+		selectCandidates := func(candidates []localityInfo, class fdbtypes.ProcessClass) []localityInfo {
```
johscheuer, line 90: In general this looks good to me. Could you move the coordinator selection into its own method and add some unit tests?

docs/user_manual.md
```diff
@@ -492,6 +492,17 @@ The `dataCenter` field in the top level of the spec specifies what data center t
 
 Replicating across data centers will likely mean running your cluster across multiple Kubernetes clusters, even if you are using a single-Kubernetes replication strategy within each DC. This will mean taking on the operational challenges described in the "Multi-Kubernetes Replication" section above.
 
+# Managing disruption
```
johscheuer, line 495: Do you mind to add an example PDB to https://github.com/FoundationDB/fdb-kubernetes-operator/tree/master/config/samples that might be useful for people that want to use PDBs.

controllers/change_coordinators.go
rbtcollins: I'll look for a timeslice to do this. I've added the PDB example and pushed.

rbtcollins: @johscheuer this is now ready for review. sorry for the delay - timeslices are precious :)

rbtcollins: Actually sorry, I see the examples are tested now, so I'll fix that up too.

rbtcollins: @johscheuer actually actually ready now :)

johscheuer (approved):
LGTM 👍 Thanks!

Issue #508, opened 2021-02-21
update controller-runtime (and other dependencies)


johscheuer:
We should update the version of the [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime), our used version is pretty old  `0.4.0`. The newer releases has some bug fixes and some nice features we could probably leverage. Also we might want to update the [controller-tools](https://github.com/kubernetes-sigs/controller-tools) in a separate step.

Once we updated we can work on these issues:

- https://github.com/FoundationDB/fdb-kubernetes-operator/issues/300
- https://github.com/FoundationDB/fdb-kubernetes-operator/issues/334

PR #513, opened 2021-02-23
Update controller-runtime and dependencies

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/508 

This PR is updates the controller-runtime and the controller-tools (and all dependencies). I only made changes that are required to make everything work. Once we have this PR reviewed and merged I open some more issues to make use of some of the new features of the libs.


Makefile
```diff
@@ -5,7 +5,7 @@ IMG ?= fdb-kubernetes-operator:latest
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true,maxDescLen=0"
 
-CONTROLLER_GEN_VERSION ?= 0.2.4
+CONTROLLER_GEN_VERSION ?= 0.5.0
```
johscheuer, line 8: Seems like I came across this issue https://github.com/kubernetes-sigs/controller-tools/issues/529 🤔 I take a look if we can workaround this before upgrading the spec to apiextenstions v1

brownleej:
Looks like there's an issue with installing the CRD in the PR check: `no matches for kind "CustomResourceDefinition" in version "apiextensions.k8s.io/v1"`.

Makefile
brownleej: I wonder if there's a way to leave the core types that we're embedding in our CRD out of the CRD definition. Most of the content in our CRD is for types that we don't control and don't have an authoritative definition for, since they depend on the server's Kubernetes version.

johscheuer:
I need to check the 

> Looks like there's an issue with installing the CRD in the PR check: `no matches for kind "CustomResourceDefinition" in version "apiextensions.k8s.io/v1"`.

Yes, I needed to generate a v1beta1 CRD and ran into some other issues but now I got a working version with a workaround which is fine since we probably want to move to `v1` pretty soon anyway.

Makefile
johscheuer: As long as we want to have the possibility to override more or less all settings I don't see a way.

brownleej (approved):
👍

PR #514, opened 2021-02-23
Add a deny list to prevent specific instances of the operator from taking global operations.

### LINKED ISSUE #236: Blocking some instances of the operator from taking global operations
There may be some partition scenarios where an instance of the operator is able to talk to the cluster but cannot have its cluster spec updated. This could lead to problems where two instances of the operator have different views of the desired database configuration, and change the configuration back and forth. As an escape hatch for these kind of situations, I think we should have an option for a blocklist of instances of the operator that should not be able to take global operations. We should be able to add an entry on any instance of the operator and get it persisted in a way that will tell the blocked instance that it cannot take global operations.

We'll need to think carefully about how this gets set and unset. I think we want to use the cluster spec to make it easy for people to set through kubectl or whatever other tooling they use to deploy clusters. I think we'll need an explicit removal operation to limit the change of two instances of the operator trying to assert conflicting blocklists.

brownleej:
Resolves #236


controllers/lock_client.go
```diff
@@ -233,6 +248,79 @@ func (client *RealLockClient) ClearPendingUpgrades() error {
...
+			list[index] = string(value.Value)
+		}
+		return list, nil
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
+	return list.([]string), nil
```
johscheuer, line 276: We may want to check if the cast is possible to prevent a fatal error?

controllers/lock_client.go
```diff
@@ -233,6 +248,79 @@ func (client *RealLockClient) ClearPendingUpgrades() error {
...
+}
+
+// UpdateDenyList updates the deny list to match a list of entries.
+// This will return the complete deny list after these changes are made.
```
johscheuer, line 280: I think the second statement is not true, since only `err` is returned?

controllers/lock_client.go
```diff
@@ -233,6 +248,79 @@ func (client *RealLockClient) ClearPendingUpgrades() error {
...
+		}
+
+		values := tr.GetRange(keyRange, fdb.RangeOptions{}).GetSliceOrPanic()
+		denyListMap := make(map[string]bool, len(values))
+		for _, value := range values {
+			denyListMap[string(value.Value)] = true
+		}
+
+		for _, entry := range locks {
+			if entry.Allow && denyListMap[entry.ID] {
```
johscheuer, line 300: What should happen if `entry.Allow` is `false` and the ID is set in the `denyListMap`? I would assume that we want to add the ID to the deny list?

I also would expect that `locks` represents the current expected state of the system so do we really need to check the current state or should we just set the expected state from `locks`?

controllers/update_status.go
```diff
@@ -299,6 +299,23 @@ func (s UpdateStatus) Reconcile(r *FoundationDBClusterReconciler, context ctx.Co
...
+		denyList, err := lockClient.GetDenyList()
+		if err != nil {
+			return false, err
+		}
+		if len(denyList) == 0 {
+			denyList = nil
+		}
+		status.Locks.DenyList = denyList
+	} else {
+		status.Locks.DenyList = nil
```
johscheuer, line 316: Since we don't set this value anywhere above, shouldn't this be the default value?

johscheuer: Also the "new" `equality.Semantic.DeepEqual` method (should) take care of this:

```
An empty slice *is* equal to a nil slice for our purposes; same for maps.
```

johscheuer (changes requested):
A few comments from my side

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -2292,6 +2331,19 @@ type LockOptions struct {
 	// LockDurationMinutes determines the duration that locks should be valid
 	// for.
 	LockDurationMinutes *int `json:"lockDurationMinutes,omitempty"`
+
+	// DenyList manages configuration for whether an instance of the operator
+	// should be denied from taking locks.
+	DenyList []LockDenyListEntry `json:"denyList,omitempty"`
```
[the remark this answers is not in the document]
brownleej: I agree that it's potentially confusing. I'll give it some more thought.

controllers/lock_client.go
brownleej: I'd rather avoid checks for things that we believe to be impossible, to reduce code clutter. Given that we know that `err` is nil at this point, we should be confident that `list` is a valid slice.

controllers/lock_client.go
brownleej: The `locks` field can be set differently in different KCs, so we need to allow for keeping things in the deny list that a given instance of the operator doesn't have any deny list configuration for.

If `entry.Allow` is false and the ID is set in the `denyListMap`, then the deny list is correctly configured and we don't need to make changes here.

api/v1beta1/foundationdbcluster_types.go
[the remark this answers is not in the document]
brownleej: Just renaming it to AccessList would be confusing with the current semantics. We could invert the meaning of the `Allow` flag, but that may also give the impression that the access list should always be set, which is not true. The current formulation, though strange, may be the least confusing option.

johscheuer (approved):
LGTM 👍  I think the name `denyList` is not perfect but once we have a better name we can go ahead and rename it.

PR #515, opened 2021-02-23
Add locks before removing pods or including instances.

### LINKED ISSUE #238: Taking locks when including instances
When we are including instances, we should take a lock to prevent overlap between instances of the operator. Instances that are not able to get the lock should submit their inclusion list so that the lock owner can aggregate them and include everything in one go.

brownleej:
Resolves #238


johscheuer (approved):
LGTM 👍

PR #516, opened 2021-02-23
Use fault domains to determine which processes to shrink

### LINKED ISSUE #22: Use fault domains to determine which processes to shrink
When we do shrinks through the operator, we are choosing processes to remove based on the instance ID. I think it would be better to choose the processes such that we optimize the distribution across fault domains after the removal, with the instance ID as a tie-breaker.

brownleej:
Resolves #22

### DIFF
--- a/controllers/choose_removals.go
+++ b/controllers/choose_removals.go
@@ -42,7 +42,7 @@ func (c ChooseRemovals) Reconcile(r *FoundationDBClusterReconciler, context ctx.
 		}
 	}
 
-	currentCounts := fdbtypes.CreateProcessCountsFromProcessGroupStatus(cluster.Status.ProcessGroups).Map()
+	currentCounts := fdbtypes.CreateProcessCountsFromProcessGroupStatus(cluster.Status.ProcessGroups, true).Map()
 	desiredCountStruct, err := cluster.GetProcessCountsWithDefaults()
 	if err != nil {
 		return false, err
--- a/controllers/choose_removals.go
+++ b/controllers/choose_removals.go
@@ -49,26 +49,60 @@ func (c ChooseRemovals) Reconcile(r *FoundationDBClusterReconciler, context ctx.
 	}
 	desiredCounts := desiredCountStruct.Map()
 
+	adminClient, err := r.AdminClientProvider(cluster, r)
+	if err != nil {
+		return false, err
+	}
+	status, err := adminClient.GetStatus()
+	if err != nil {
+		return false, err
+	}
+	localityMap := make(map[string]localityInfo)
+	for _, process := range status.Cluster.Processes {
+		id := process.Locality[FDBLocalityInstanceIDKey]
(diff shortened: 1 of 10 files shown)


controllers/choose_removals.go
```diff
@@ -42,33 +42,67 @@ func (c ChooseRemovals) Reconcile(r *FoundationDBClusterReconciler, context ctx.
...
 		desiredCount := desiredCounts[processClass]
 
 		removedCount := currentCounts[processClass] - desiredCount
 
+		processClassLocality := make([]localityInfo, 0, currentCounts[processClass])
+
 		for _, processGroup := range cluster.Status.ProcessGroups {
-			if processGroup.ProcessClass == processClass && processGroup.Remove {
-				removedCount--
+			if processGroup.ProcessClass == processClass {
```
johscheuer, line 74: This can be switched to be a "guard" e.g.

```go
if processGroup.ProcessClass != processClass {
   continue
}
```

controllers/choose_removals.go
```diff
@@ -42,33 +42,67 @@ func (c ChooseRemovals) Reconcile(r *FoundationDBClusterReconciler, context ctx.
...
+				if processGroup.Remove {
+					if processGroup.ProcessClass == processClass && processGroup.Remove {
```
johscheuer, line 76:
```suggestion
					if processGroup.Remove {
```

the expression must be true otherwise we wouldn't be here.

johscheuer: I think we can drop the complete if statement in this line

controllers/choose_removals.go
```diff
@@ -42,33 +42,67 @@ func (c ChooseRemovals) Reconcile(r *FoundationDBClusterReconciler, context ctx.
...
+
+			log.Info("Chose remaining processes after shrink", "desiredCount", desiredCount, "options", processClassLocality, "selected", remainingProcesses)
+
+			remainingProcessMap := make(map[string]bool, len(remainingProcesses))
+			for _, localityInfo := range remainingProcesses {
+				remainingProcessMap[localityInfo.ID] = true
+			}
+
+			for _, processGroup := range cluster.Status.ProcessGroups {
+				if processGroup.ProcessClass == processClass && !remainingProcessMap[processGroup.ProcessGroupID] {
```
johscheuer, line 104: Do we actually have to check the process class here?

johscheuer: Wee pretty often use `cluster.Status.ProcessGroups` and then we filter to match `processGroup.ProcessClass == processClass` we may want to have an additional method for this to only return processGroups for a specific class.

johscheuer (changes requested):
A few comments.

controllers/choose_removals.go
brownleej: As currently implemented, we need to check the process class, because of how the remainingProcessMap is managed. I think we can extract it to a higher level, though.

johscheuer (approved):
LGTM 👍

PR #517, opened 2021-02-23
Add documentation on working with locks.

### LINKED ISSUE #239: Adding documentation on working with locks
Once we've filled out the global coordination features, we should add documentation to the user manual on working with locks and configuring the locking system.

brownleej:
Resolves #239


docs/user_manual.md
```diff
@@ -443,26 +445,7 @@ This tells the operator to use the value "zone2" as the fault domain for every p
...
-        storage: 5
-      automationOptions:
-        configureDatabase: false
-        killProcesses: false
-      faultDomain:
-        key: foundationdb.org/kubernetes-cluster
-        value: zone2
-        zoneIndex: 2
-        zoneCount: 5
+You should always specify an `instanceIDPrefix` when deploying an FDB cluster to multiple Kubernetes clusters. You should set it to a different value in each Kubernetes cluster. This will two multiple instances in different Kubernetes cluster from using the same instance ID.
```
johscheuer, line 448:
```suggestion
You must always specify an `instanceIDPrefix` when deploying an FDB cluster to multiple Kubernetes clusters. You must set it to a different value in each Kubernetes cluster. This will prevent instance ID duplicates in the different Kubernetes clusters.
```

docs/user_manual.md
```diff
@@ -588,3 +571,42 @@ The name of a cluster is immutable, and it is included in the names of all of th
...
+
+The locking system uses the `instanceIDPrefix` from the cluster spec to identify an instance of the operator. Make sure to set this to a unique value for each Kubernetes cluster, both to support the locking system and to prevent duplicate instance IDs.
+
+This locking system uses the FoundationDB cluster as its data source. This means that if the cluster is unavailable, no instance of the operator will be able to get a lock. If you hit a case where this becomes a blocker, you can disable the locking system by setting `lockOptions.disableLocks = true` in the cluster spec.
```
johscheuer, line 581:
```suggestion
This locking system uses the FoundationDB cluster as its data source. This means that if the cluster is unavailable, no instance of the operator will be able to get a lock. If you hit a case where this becomes an issue, you can disable the locking system by setting `lockOptions.disableLocks = true` in the cluster spec.
```

docs/user_manual.md
```diff
@@ -588,3 +571,42 @@ The name of a cluster is immutable, and it is included in the names of all of th
...
+
+In most cases, bounces will be done independently in each Kubernetes cluster, and the locking system will be used to ensure a minimum time between the different bounces and avoid multiple recoveries in a short span of time. During upgrades, however, all instances must be bounced at the same time. The operator will use the locking system to coordinate this. Each instance of the operator will store records indicating what processes it is managing and what version they will be running after the bounce. Each instance will then try to acquire a lock and confirm that every process reporting to the cluster is ready for the upgrade. If all processes are prepared, the operator will bounce all of them at once. If any instance of the operator is stuck and unable to prepare its processes for the upgrade, the bounce will not occur.
```
johscheuer, line 583: I would prefer to use `restarted` in this case since `bounce` is an ambiguous term.

docs/user_manual.md
```diff
@@ -588,3 +571,42 @@ The name of a cluster is immutable, and it is included in the names of all of th
...
+    kind: FoundationDBCluster
+    metadata:
+      name: sample-cluster
+    spec:
+      instanceIDPrefix: dc1
+      lockOptions:
+        denyList:
+          - id: dc2
+
+This will clear any locks held by `dc2`, and prevent it from taking further locks. In order to clear this deny list, you must change it to allow that instance again:
```
johscheuer, line 599: Is there a reason just removing the instance is not good enough? Since the name `denyList` list implies that any instance on this list is not allowed to take locks.

brownleej: Yes. We can talk about that more in the Deny List PR.

johscheuer (approved):
LGTM 👍

PR #518, opened 2021-02-23
Moves the management of more conditions into the process group status.

### LINKED ISSUE #412: Tracking incorrect pods and other failure conditions in the process group status
As part of the Process Group Status workflow, we should add tracking of process-level failure conditions to the process group status. The linked design has more details on what conditions this will add. This will replace multiple fields from the status: `incorrectProcesses`, `incorrectPods`, `failingPods`, and `missingProcesses`.

brownleej:
Resolves #412

### DIFF
--- a/config/samples/cluster_local.yaml
+++ b/config/samples/cluster_local.yaml
@@ -20,6 +20,7 @@ spec:
     general:
       customParameters:
         - "knob_disable_posix_kernel_aio=1"
+        - "locality_test=1"
       volumeClaimTemplate:
         spec:
           resources:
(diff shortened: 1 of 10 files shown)


config/samples/cluster_local.yaml
```diff
@@ -37,6 +38,9 @@ spec:
                 requests:
                   cpu: 250m
                   memory: 128Mi
+              env:
+                - name: JPB_TEST
```
johscheuer, line 42: We probably want to rename this env or remove it?

controllers/update_status.go
```diff
@@ -607,33 +575,29 @@ func validateInstance(r *FoundationDBClusterReconciler, context ctx.Context, clu
...
 			}
 		}
 	}
 
+	failing := false
 	for _, container := range instance.Pod.Status.ContainerStatuses {
-		if !container.Ready {
-			return true, false, needsSidecarConfInConfigMap, nil
-		}
+		failing = failing || !container.Ready
```
johscheuer, line 595: If `failing` is true we can return early.
```suggestion
		if container.Ready {
		     continue
		}
		 
		 failing = true
		 break
```

johscheuer (changes requested):

johscheuer (approved):
LGTM 👍

PR #521, opened 2021-02-24
Fix #520: Public NotReady pods to pod cluster IPs

### LINKED ISSUE #520: The ServiceIP approach can lead to grey failures
We have a partially analysed outage shortly after rolling out the service IP approach for coordinator stability. 

During a nodepool upgrade in AKS a log node took 43 minutes to shut down - why isn't clear, and its important to note that Kubernetes doesn't make any commitment about time to kill objects: webhooks and finalizers can arbitrarily extend the time it takes objects to be deleted.

In this cluster we have 6 log servers and triple replication.

During recovery, around the time the cluster reached initializing_transaction_servers, a log server pod was selected for termination. At that point the service IP started rejecting traffic (not atomically, but quite quickly).

However, the fdb server process was still alive and could reach out to the rest of the cluster, and continued to do so: we effectively had an asymmetric network break *on the session initiation level* as opposed to the individual packet level.

For some reason - we haven' t finished analysis yet - the process remained alive, and the pod in terminating state, for ~40 minutes.

Only after the process exited, did fdb cluster recovery resume (by starting over).

I think it was stuck in https://github.com/apple/foundationdb/blob/4c852687920c00e4c62c0111beb8ba83408390a2/fdbserver/masterserver.actor.cpp#L1512 somewhere, and I *think* that the half-working nature of the situation is what prevented FDB detecting this situation and recovering - but until we climb through the log server management code and look for heartbeats, timeouts etc and reason about that we can't really say. FDB 6.28 was in use.

rbtcollins:
The ClusterIP services are not doing true service discovery, instead
they are just fixed proxies for the pod, and so the readiness of the pod
is not relevant. Pod readiness automatically prevents traffic flow via
the ClusterIP if the pod isn't ready - either during startup, or during
the pod deletion lifecycle, but if the FDB process is running, outbound
connections may be made, and this sets up a grey-failed network scenario
which is poor for FDB.

Signed-off-by: Robert Collins <robert.collins@cognite.com>

### DIFF
--- a/controllers/pod_models.go
+++ b/controllers/pod_models.go
@@ -94,9 +94,10 @@ func GetService(cluster *fdbtypes.FoundationDBCluster, processClass string, idNu
 	return &corev1.Service{
 		ObjectMeta: metadata,
 		Spec: corev1.ServiceSpec{
-			Type:     corev1.ServiceTypeClusterIP,
-			Ports:    generateServicePorts(processesPerPod),
-			Selector: getMinimalSinglePodLabels(cluster, id),
+			Type:                     corev1.ServiceTypeClusterIP,
+			Ports:                    generateServicePorts(processesPerPod),
+			PublishNotReadyAddresses: true,
+			Selector:                 getMinimalSinglePodLabels(cluster, id),
 		},
 	}, nil
 }


johscheuer (approved):
Just checked the source code and it seems that this setting also includes terminating Pods (which is a little bit confusing from the name) 👍  LGTM

rbtcollins:
Yup, a terminating pod is implicitly NotReady

Issue #458, opened 2021-01-16
Prevent removing unknown labels


brownleej:
If someone labels a pod independently of the operator, the operator will attempt to remove the label. This can cause problems in situations where another controller is using labels to add additional information to pods, because the two controllers would fight back and forth changing the labels. We have more nuanced behavior with annotations where we leave unknown annotations in place. Should we apply the same behavior to labels?

PR #522, opened 2021-02-24
Prevent removing unkown labels in Pods

johscheuer:
fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/458


brownleej (approved):
👍

PR #527, opened 2021-02-26
Make ProcessClass a type

rbtcollins:
Sorry for this enormous out of the blue patch, and I'll understand if the answer is 'wat, nope.'

It came out of the other coordinator patch where a little more type safety would have been useful. And then the minimum set the tentacles reached to this size.

The basic idea is the usual benefits of type safety - no accidentally assigning random strings to ProcessClass places, etc. 

This version can probably be cleaned up a bit if the basic idea is not objectionable.

### DIFF
--- a/api/v1beta1/foundationdbcluster_types.go
+++ b/api/v1beta1/foundationdbcluster_types.go
@@ -78,7 +78,7 @@ type FoundationDBClusterSpec struct {
 	DatabaseConfiguration `json:"databaseConfiguration,omitempty"`
 
 	// Processes defines process-level settings.
-	Processes map[string]ProcessSettings `json:"processes,omitempty"`
+	Processes map[ProcessClass]ProcessSettings `json:"processes,omitempty"`
 
 	// ProcessCounts defines the number of processes to configure for each
 	// process class. You can generally omit this, to allow the operator to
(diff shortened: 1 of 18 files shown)


johscheuer (changes requested):
I think this makes definitely sense 👍 It maybe makes sense in a next PR to move the `ProcessClass` into it's own file and moving all related methods into this single place instead of having them separated across multiple files.

rbtcollins:
> 
> 
> I think this makes definitely sense 👍 It maybe makes sense in a next PR to move the `ProcessClass` into its own file and moving all related methods into this single place instead of having them separated across multiple files.
> 
> PS.: I think you need to rebase delete your `controller-gen` and the run `make clean all` the current makefile doesn't check if the `controller-gen` matches the expected version.

Done. All ready for merge I hope :).

rbtcollins:
@johscheuer ping

controllers/suite_test.go
```diff
@@ -288,3 +288,10 @@ func createTestClusterReconciler() *FoundationDBClusterReconciler {
 		LockClientProvider:  NewMockLockClient,
 	}
 }
+
+func testInstanceLabels(instanceName string, instanceClass fdbtypes.ProcessClass) map[string]string {
```
johscheuer, line 292: What is the reason that you added this function but only use it in one case?

johscheuer (commented):
One question otherwise it LGTM, will do some final testing today 👍

controllers/suite_test.go
rbtcollins: Ah, I was trying to use it in 3 places but then saw that the labels were actually subtly different. I can unwind this function, or look more aggressively for places to use it....

johscheuer: I think removing it makes the most sense in the current state, w can revisit this when we change something, or someone wants to refactor the tests :)

rbtcollins: Fixed, and fixed the semantic conflict that had been merged to master

rbtcollins: And fixed the conflict with #518 too - this is just large, not complex, would love to see it merged to stop the conflict chasing :)

johscheuer (approved):
LGTM 👍

PR #528, opened 2021-02-27
Remove all leading non JSON output

johscheuer:
We still have an issue when we use the `fdbcli status` e.g. for backups or when we want to call it in the kubectl plugin the new version of fdbcli should report warnings and errors to stderr (https://github.com/apple/foundationdb/pull/4332) so we shouldn't see this error anymore until all of the old version are deprecated we should have a workaround for these cases. This PR removes all leading characters of the JSON output that is not a `{`.


brownleej (approved):
👍

PR #529, opened 2021-02-27
Add e2e test to create CRDs against a newer version of the API server

johscheuer:
Initially I only wanted to fix this issue https://github.com/kubernetes-sigs/controller-tools/issues/529 (I didn't saw this issue because some older Kubernetes version ignore (e.g. 1.15 which is the default for kube-builder) and I wanted to add a small test to deploy the CRDs against our expected Kubernetes versions.

### DIFF
--- a/main.go
+++ b/main.go
@@ -48,6 +48,7 @@ func init() {
 func main() {
 	var metricsAddr string
 	var enableLeaderElection bool
+	var leaderElectionID string
 	var logFile string
 	var cliTimeout int
 	var deprecationOptions controllers.DeprecationOptions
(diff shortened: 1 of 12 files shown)


main.go
```diff
@@ -60,8 +61,8 @@ func main() {
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", true,
 		"Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
-	flag.StringVar(&logFile, "log-file", "", "The path to a file to write logs to.")
```
brownleej, line 63: Why are these two flags being removed?

brownleej (changes requested):

main.go
johscheuer: Good catch, these changes where from something different, seems like I mixed this up.

brownleej (approved):
👍

PR #535, opened 2021-03-01
Adding badges to readme

johscheuer:
Adding some "fancy" badges to show the current state of the repository (here is the rendered version: https://github.com/johscheuer/fdb-kubernetes-operator/tree/add-badges#overview).


brownleej (approved):
👍

PR #538, opened 2021-03-05
Add version to status output

johscheuer:
If you run multiple FDB clusters it would be nice to see in which version the specific clusters are.


brownleej (approved):
👍

Issue #482, opened 2021-02-05
The FDB plugin support FDB clusters that span across multiple Kubernetes clusters


johscheuer:
The FDB operator allows to manage FDB clusters that are span across multiple Kubernetes clusters. The kubectl FDB plugin should also supports actions on such clusters. One way would be to specify multiple context(s) or to have some kind of configuration file that defines in which clusters a FDB cluster is available. This issue is mostly here to keep track of this feature before implementing it we probably want to come up with a design doc to see how we can implement it.

brownleej:
This is a great idea! Maybe we could implement it as a kind of multi-KC apply operation, where we take a YAML for the cluster spec and apply it to all relevant KCs. We could infer some of that from the region configuration, and maybe add new fields to the spec that aren't used by the operator but can be used to specify where it needs to get deployed.

brownleej:
We also have test scripts for creating multi-KC clusters that could be rewritten to use this feature in the plugin.

johscheuer:
It probably makes sense to rework the current design of the `remove` command to be like `kubectl fdb create [type]` and `kubectl fdb delete [type]`  so it looks more like the normal kubectl commands. Adding an additional field would have the benefit that we don't need to create "yet another config file" and it would be more obvious where the FDB cluster is/should be running. I take a look at the multi dc/kc deployment and try to write down some implementable issues. One initial idea could be adding 2 new fields like `seed_kubernetes_cluster` and `kubernetes_clusters` (the latter one is just a list of additional Kubernetes contexts).

johscheuer:
Closing since the design is merged and we have more specific issues for the implementation.

PR #540, opened 2021-03-06
Initial design for multi dc/kc FDB clusters with the plugin

johscheuer:
Initial design to support multi dc/kc FDB clusters with the plugin see: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/482

### DIFF
--- a/docs/design/plugin_multi_fdb_support.md
+++ b/docs/design/plugin_multi_fdb_support.md
@@ -9,17 +9,20 @@
 ## Background
 
 The FDB Kubernetes operator supports to setup and operate FDB clusters span across
-multiple Kubernetes clusters. Currently the setup is more or less manual with the
-following steps: bootstrap the seed cluster and once the seed cluster is bootstrapped
-( actually reconciled) we can adjust the database configuration and bootstrap additional
+multiple Kubernetes clusters. The setup is more or less manual with the
+following steps:
+
+- Bootstrap the seed cluster and wait until it's reconciled
+- Adjust the database configuration and bootstrap additional
 FDB clusters managed by the Operator. Updating the cluster means adjusting all manifests
-across the Kubernetes clusters at the same time (or in a short time span). Additionally
-the tooling for managing FDB clusters (the kubectl FDB plugin) is currently not aware of
+across the Kubernetes clusters at the same time (or in a short time span).
+
+The tooling for managing FDB clusters (the kubectl FDB plugin) is currently not aware of
 FDB clusters spread across multiple Kubernetes clusters or even multiple manifests.
 
 ## General Design Goals
 
-* Propose a way how to define where a cluster is located.
+* Propose a way how to define the cluster location.
 * Bootstrap HA FDB clusters with `kubectl fdb`.
 * Make `kubectl fdb` aware of multiple FDB clusters forming an HA cluster.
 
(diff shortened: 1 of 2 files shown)


docs/design/plugin_multi_fdb_support.md
```diff
@@ -0,0 +1,134 @@
...
+## General Design Goals
+
+* Propose a way how to define where a cluster is located.
+* Bootstrap HA FDB clusters with `kubectl fdb`.
+* Make `kubectl fdb` aware of multiple FDB clusters forming an HA cluster.
+
+## Proposed Design
+
+The idea is to use two additional fields in the CRD that will be ignored by the operator but used by the `kubectl fdb plugin`.
+The fields are `seedCluster` and `targetClusters`, both contain the following `cluster objects`:
```
brownleej, line 29: We may want to elaborate more on how we plan to support multi-KC clusters where the KC is the primary zone in the replication configuration. I think we would want to cover the following things:

1. Setting the instanceIDPrefix to something separate from the DC ID. I think the best way to do this is to make the name mandatory, and set the instanceIDPrefix to the name, and then have a separate optional field for the name of the cluster object.
2. Setting the zoneIndex and zoneCount fields in the replication configuration. We can probably infer this from the list of targetClusters, but we may also want to allow setting it explicitly to cover the case of moving a part of a cluster from one zone to another.

docs/design/plugin_multi_fdb_support.md
```diff
@@ -0,0 +1,134 @@
...
+
+```yaml
+name: sample-cluster-dc3 # Optional: Will overwrite the cluster name in the resulting manifest otherwise uses the metadata
+dcID: dc3 # Must match with the data center ID specified under datacenters
+context: my-kube-ctx 3 # Optional: the kubconfig context to use otherwise the current context will be used
+namespace: dc3 # Optional: Otherwise the namespace from the metadata will be used
+processCounts: # Optional: otherwise the processCounts from he cluster Spec will be used (useful to change cluster specific fields e.g. satellites without SS)
```
brownleej, line 36: I think this is valid, but the operator should be able to infer the satellite process counts based on matching the DC ID with the region configuration.

docs/design/plugin_multi_fdb_support.md
brownleej: Under the suggestion 1 above, we would also want to validate that the name is unique within the targetClusters list. We could also simplify the seedCluster by having it take just the name instead of the full specification for the cluster.

docs/design/plugin_multi_fdb_support.md
```diff
@@ -0,0 +1,134 @@
...
+seedConnectionString: $connectionString # The seed connection string will be set once the initial cluster is bootstrapped
+```
+
+The Operator will be configured to ignore the fields `seedCluster` and `targetClusters` which means that these fields
+are only used by the `kubectl fdb plugin` and maybe the human operator. Also these fields are optional, if a FDB cluster
+is limited to a single namespace in a single Kubernetes cluster these fields can be omitted.
+
+The `kubectl fdb plugin` will take the following steps when creating a FDB cluster across multiple Kubernetes clusters (or namespaces):
+
+1. Create the seed cluster in the `seedcluster`.
```
brownleej, line 109: I think we'll need to change the configuration when creating the seed cluster in a couple ways:

1. If there is region config in the final DC configuration, the region config for the seed cluster should be set to a single data center, containing the DC for the seed cluster, since the other regions won't be available until a later step.
2. If we are using multi-KC replication for the zone, the seed cluster will need to be single-replicated until the other zones are available.

docs/design/plugin_multi_fdb_support.md
```diff
@@ -0,0 +1,134 @@
...
+instanceIDPrefix: $dcID
+dataCenter: $dcID
+seedConnectionString: $connectionString # The seed connection string will be set once the initial cluster is bootstrapped
+```
+
+The Operator will be configured to ignore the fields `seedCluster` and `targetClusters` which means that these fields
+are only used by the `kubectl fdb plugin` and maybe the human operator. Also these fields are optional, if a FDB cluster
+is limited to a single namespace in a single Kubernetes cluster these fields can be omitted.
+
+The `kubectl fdb plugin` will take the following steps when creating a FDB cluster across multiple Kubernetes clusters (or namespaces):
```
brownleej, line 107: Should we also describe how we are going to support updates to multi-DC/KC clusters? For instance, we will probably want a custom apply command that ensures that all targets have the same connection string.

docs/design/plugin_multi_fdb_support.md
brownleej: Given how finicky the region configuration structure in FDB is, there may be benefit in adding fields here that can be used to construct the region configuration. This would allow the user to leave the `region` section of the database configuration empty, and only put data center information here.

docs/design/plugin_multi_fdb_support.md
brownleej: We could also look at creation as a special case of the apply command, where it detects that the seed cluster is missing and creates it, and then moves on to the normal apply process.

brownleej (changes requested):

docs/design/plugin_multi_fdb_support.md
johscheuer: That's right, we can also remove this part in the first version.

docs/design/plugin_multi_fdb_support.md
johscheuer: That's a good point I will revisit this part for multi kc

brownleej (approved):
👍

Issue #539, opened 2021-03-06
Create a patch release for 0.29.0


johscheuer:
Currently 0.29.0 has a bug and won't start so we should release `0.29.1` that contains the fix: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/529

johscheuer:
solved: https://github.com/FoundationDB/fdb-kubernetes-operator/releases/tag/v0.30.0

PR #542, opened 2021-03-07
Correct v0.29.0 image tag in helm chart

funkypenguin:
Hey guys,

This PR simply corrects what seems to be an error in the latest helm chart - the desired container image should be `v0.29.0`, rather than `0.29.0` (which doesn't exist)

Cheers!
D


johscheuer (approved):
👍 Thanks! Just a heads-up `v0.29.0` has an issue: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/539 we will probably release tomorrow a patch release.

PR #543, opened 2021-03-08
Add helm chart publishing and tests

funkypenguin:
Hey guys!

This PR:

1. Moves the `helm` folder to `charts`, for better compatibility with the helm-provided chart-testing GitHub actions
2. Adds GitHub action for linting and testing charts on push, ensuring that charts actually run in a KinD cluster
3. Adds GitHub action to publish the helm chart when a push to master is made, which includes changes to the chart files. This would allow users to install the chart via a helm repo URL, rather than requiring them to clone the repo in order to deploy the chart.

I'm hoping this can make it in before 0.29.1, and can be used to validate the crashloop bugfix ;)

Cheers!
D


charts/.ci/values-kube-score.yaml
```diff
@@ -0,0 +1,12 @@
...
+operator:
+  imagePullPolicy: Always
+
+resources:
+  requests:
+    cpu: 1
+    memory: 100Mi
+  limits:
+    cpu: 1
+    memory: 100Mi
```
johscheuer, line 12: Missing newline

charts/.ci/lint-config.yaml
```diff
@@ -0,0 +1,6 @@
+rules:
+  # One blank line is OK
+  empty-lines: 
+    max-start: 1
+    max-end: 1
+    max: 1
```
johscheuer, line 6: Missing newline

charts/.ci/scripts/local-ct-lint.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+docker run --rm -it -w /repo -v `pwd`:/repo quay.io/helmpack/chart-testing ct lint --all
```
johscheuer, line 3: Missing newline

charts/.ci/scripts/local-ct-lint.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
johscheuer, line 1:
```suggestion
#!/usr/bin/env bash
```

charts/.ci/scripts/local-kube-score.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
johscheuer, line 1:
```suggestion
#!/usr/bin/env bash
```

charts/.ci/scripts/local-kube-score.sh
```diff
@@ -0,0 +1,10 @@
...
+
+helm template --values charts/.ci/values-kube-score.yaml charts/* -f charts/.ci/values-kube-score.yaml | kube-score score - \
```
johscheuer, line 3: This is inconsistent with the `chart-testing` script we should either expect that the user has these tools installed (and should tell the user in the Readme what is required) or use only docker images.

charts/.ci/scripts/local-ct-lint.sh
johscheuer: Also I would rather prefer to use a pinned version than using `latest`

.github/workflows/on-push-lint-charts.yml
```diff
@@ -0,0 +1,67 @@
+name: Lint and Test Charts
+
+on:
```
johscheuer, line 3: Any reason this is not included in the `pull_request.yml`?

.github/workflows/on-push-lint-charts.yml
```diff
@@ -0,0 +1,67 @@
...
+          python-version: 3.7
+
+      - name: Set up chart-testing
+        uses: helm/chart-testing-action@v2.0.1
+
+      - name: Run chart-testing (list-changed)
+        id: list-changed
+        run: |
+          changed=$(ct list-changed)
+          if [[ -n "$changed" ]]; then
```
johscheuer, line 54:
```suggestion
          if [[ -n "$(ct list-changed)" ]]; then
```

.github/workflows/on-push-lint-charts.yml
```diff
@@ -0,0 +1,67 @@
...
+          fetch-depth: 0
+
+      - name: Set up Helm
+        uses: azure/setup-helm@v1
+        with:
+          version: v${{ env.HELM_VERSION }}
+
+      - name: Set up kube-score
+        run: |
+          wget https://github.com/zegl/kube-score/releases/download/v${{ env.KUBE_SCORE_VERSION }}/kube-score_${{ env.KUBE_SCORE_VERSION }}_linux_amd64 -O kube-score
```
johscheuer, line 29: I'm not quite sure about the value here since we ignore most of the stuff anyway + we only check the rendered chart and not the other yaml files.

.github/workflows/on-push-master-publish-chart.yml
```diff
@@ -0,0 +1,40 @@
...
+on:
+  push:
+    branches:
+      - master
+      - main # assume that the branch name may change in future
+    paths:
+      - 'charts/**' 
+      - '.github/**'     
+env:
+  KUBE_SCORE_VERSION: 1.10.1
```
johscheuer, line 12:
```suggestion
```

.github/workflows/on-push-master-publish-chart.yml
```diff
@@ -0,0 +1,40 @@
...
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+
+      - name: Install Helm
+        uses: azure/setup-helm@v1
+        with:
+          version: v${{ env.HELM_VERSION }}
+
+      - name: Run chart-releaser
+        uses: helm/chart-releaser-action@v1.2.0
```
johscheuer, line 37: I don't think we meet all https://github.com/helm/chart-releaser-action#pre-requisites?

johscheuer (changes requested):

.github/workflows/on-push-master-publish-chart.yml
funkypenguin: We'd meet them if we created a `gh-pages` branch though :)

charts/.ci/lint-config.yaml
funkypenguin: Fixed :)

charts/.ci/scripts/local-ct-lint.sh
funkypenguin: Fixed!

.github/workflows/on-push-master-publish-chart.yml
funkypenguin: I'm not sure I understand the suggestion here, the suggest change seems to match the existing content?

.github/workflows/on-push-lint-charts.yml
funkypenguin: We're still subject to all the rest of the tests at https://github.com/zegl/kube-score/blob/master/README_CHECKS.md - the value (for me at least), is that we raise the bar on the chart for users (like me) who require kube-score passing in order to deploy into their cluster via CI. It'll also catch any snafus in future PRs, like accidentally turning on privileged mode in the deployment.

.github/workflows/on-push-lint-charts.yml
funkypenguin: I thought it would be simpler to keep the chart publishing as a self-contained workflow, acting only on the master branch. Also, I'm not familiar with `pull_request.yml`, and didn't want to cause confusion by interfering with it. I'm happy to refactor the chart publishing into `pull_request.yml` if you prefer.

.github/workflows/on-push-master-publish-chart.yml
johscheuer: We can remove the env variable since we never use it in this action.

.github/workflows/on-push-master-publish-chart.yml
johscheuer: That was one of my questions since it's never mentioned in the PR.

charts/.ci/scripts/local-ct-lint.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
```
johscheuer, line 1: I actually would prefer to have this in the `Makefile` or at least be called by the `Makefile` and otherwise I need to remember myself to call this script :)

charts/.ci/scripts/local-kube-score.sh
johscheuer: I would also prefer to have a `Makefile` target for that so we can simply run `make helm-lint` for example and call everything.

.github/workflows/on-push-lint-charts.yml
```diff
@@ -0,0 +1,67 @@
...
+          chmod 755 kube-score
+
+      - name: Kube-score generated manifests
+        run: helm template  --values charts/.ci/values-kube-score.yaml charts/* | ./kube-score score -
```
johscheuer, line 33: Can we call here `charts/.ci/scripts/local-kube-score.sh` instead? Otherwise we have to manage the exception list on multiple places.

johscheuer (changes requested):
Thanks for your patience just a few comments and I think we are good to go with that. Thanks for taking time to work on that and feel free to ping me once you changed everything.

johscheuer (commented):
@funkypenguin do you have any plans/time to move forward with this PR? I think except for some smaller things and probably a major rebase the changes are fine.

funkypenguin:
> @funkypenguin do you have any plans/time to move forward with this PR? I think except for some smaller things and probably a major rebase the changes are fine.

Yessir, I'll take care of it in the morning :)

funkypenguin:
Mmm. This rebase has gone... badly :) I'll create a fresh PR instead, shortly...

funkypenguin:
OK, moved to https://github.com/FoundationDB/fdb-kubernetes-operator/pull/899. The only input I've not effected is creating a dedicated Makefile (because I don't know how) :)

PR #544, opened 2021-03-08
Add validation for satellite

johscheuer:
I stumbled over this when I created a multi dc FDB cluster and used the `satellite` field as `priority`. This can be a little bit tricky to debug so I think it makes sense to add this as a validation check.


brownleej (approved):
👍

PR #545, opened 2021-03-11
Update the data-loader example to use the public image

johscheuer:
I change the default image to the public image that we push from this repo, so user doesn't have to build the image.

### DIFF
--- a/sample-apps/data-loader/Dockerfile
+++ b/sample-apps/data-loader/Dockerfile
@@ -1,7 +1,12 @@
 FROM python:3.8-slim
 
-COPY app.py /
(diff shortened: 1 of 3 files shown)


brownleej (approved):
👍

brownleej:
This change looks good. It looks like we have some PR checks failing in a way that's not connected with this change, but we should follow up on that separately.

johscheuer:
I already added a change to disable the cache for the binaries, somehow it's faulty and not really faster: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/546

PR #548, opened 2021-03-14
Make use of random bytes lib

johscheuer:
Make use of https://docs.python.org/3.9/library/random.html?highlight=random#functions-for-bytes in the sample app.


brownleej (approved):
👍

Issue #502, opened 2021-02-17
Logging all decisions to remove processes


brownleej:
We've seen some cases where processes were marked for removal and it wasn't clear why. We should add a log statement whenever we mark a process for removal identifying the reason, so we can search for this event in the logs when we see unexplained removals.

PR #553, opened 2021-03-17
Log reasons for replacement

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/502

### DIFF
--- a/controllers/replace_misconfigured_pods.go
+++ b/controllers/replace_misconfigured_pods.go
@@ -68,6 +68,7 @@ func (c ReplaceMisconfiguredPods) Reconcile(r *FoundationDBClusterReconciler, co
 		if processGroupStatus == nil {
 			return false, fmt.Errorf("unknown PVC %s in replace_misconfigured_pods", instanceID)
 		}
+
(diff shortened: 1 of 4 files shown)


controllers/replace_misconfigured_pods.go
```diff
@@ -92,9 +94,17 @@ func (c ReplaceMisconfiguredPods) Reconcile(r *FoundationDBClusterReconciler, co
 			if err != nil {
 				return false, err
 			}
+
 			if len(instances) > 0 {
 				processGroupStatus.Remove = true
 				hasNewRemovals = true
+
+				log.Info("replace instance",
```
brownleej, line 102:
```suggestion
				log.Info("Replace instance",
```

brownleej (approved):
Small request on making capitalization consistent. Otherwise LGTM.

controllers/replace_misconfigured_pods.go
johscheuer: Good catch, thanks!

PR #555, opened 2021-03-18
Block exclusions on missing processes

### LINKED ISSUE #363: Blocking exclusion until replacement pods are up
When we are doing a replacement, we create new pods and then move on to excluding the old pods. This can cause the operator to attempt an exclusion before the new pods are up. Generally, this will fail because of storage space limits, rather than doing real damage. If we are replacing processes in a satellite, an unsafe exclusion can take the database unavailable. I think that we should wait to start an exclusion until all processes are up, other than those pending removal.

Related issue on FDB: https://github.com/apple/foundationdb/issues/1292

JoshuaMcManus:
Fixes #363 

This is going to block on any process missing, not just the ones that we added as a replacement.

### DIFF
--- a/controllers/exclude_instances.go
+++ b/controllers/exclude_instances.go
@@ -70,6 +70,19 @@ func (e ExcludeInstances) Reconcile(r *FoundationDBClusterReconciler, context ct
 	}
 
 	if len(addresses) > 0 {
+		// Block excludes on missing processes not marked for removal
+		missingProcesses := make([]string, 0)
+		for _, processGroupStatus := range cluster.Status.ProcessGroups {
+			missingTime := processGroupStatus.GetConditionTime(fdbtypes.MissingProcesses)
+			if missingTime != nil && !processGroupStatus.Remove {
+				missingProcesses = append(missingProcesses, processGroupStatus.ProcessGroupID)
+			}
+		}
+		if len(missingProcesses) > 0 {
+			log.Info("Waiting for missing processes", "namespace", cluster.Namespace, "cluster", cluster.Name, "missingProcesses", missingProcesses)
+			return false, nil
+		}
+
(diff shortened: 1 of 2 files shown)


controllers/exclude_instances.go
```diff
@@ -70,6 +70,19 @@ func (e ExcludeInstances) Reconcile(r *FoundationDBClusterReconciler, context ct
 	}
 
 	if len(addresses) > 0 {
+		// Block excludes on missing processes not marked for removal
+		missingProcesses := make([]string, 0)
+		for _, processGroupStatus := range cluster.Status.ProcessGroups {
+			missingTime := processGroupStatus.GetConditionTime(fdbtypes.MissingProcesses)
+			if missingTime != nil && !processGroupStatus.Remove {
```
johscheuer, line 77: We probably also want to check for `MissingPod` https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/api/v1beta1/foundationdbcluster_types.go#L607 with the current design of the `ProcessGroups` it's a little bit harder but I think `MissingPod` and `MissingProcesses` should be fine. We might want to consider to create a new `ProcessGroup` with one of these conditions to be sure we don't run into a situation where we created the new `ProcessGroup` but the controller didn't set any condition on it.

johscheuer (changes requested):
LGTM, just one comment.

controllers/exclude_instances.go
brownleej: +1 to Johannes's suggestions. I think it makes sense to add the `MissingPod` condition when we create a new process group in AddProcessGroups, since at the time we add the process group that we don't have a corresponding pod yet.

johscheuer (approved):
LGTM 👍 mental note for myself we may want to revisit the code here since we iterate 3 times for all processGroups.

PR #556, opened 2021-03-18
Prevent reconciliation from getting stuck when there is a pod stuck In Terminating that has conditions on the process group.

brownleej:


mock-kubernetes-client/client/client_test.go
```diff
@@ -153,6 +153,37 @@ func TestDeletingObject(t *testing.T) {
 	g.Expect(k8serrors.IsNotFound(err)).To(BeTrue())
 }
 
+func TestDeletingPodInTerminating(t *testing.T) {
+	g := NewWithT(t)
+	client := &MockClient{}
+
+	pod1 := createDummyPod()
+	pod2 := createDummyPod()
```
johscheuer, line 161: What is the reason to have `pod2`?

mock-kubernetes-client/client/client.go
```diff
@@ -543,6 +551,30 @@ func (client *MockClient) DeleteAllOf(context ctx.Context, object ctrlClient.Obj
 	return fmt.Errorf("Not implemented")
 }
 
+// MockStuckTermination sets a flag determining whether an object should get stuck in terminating when it is deleted.
+func (client *MockClient) MockStuckTermination(object ctrlClient.Object, terminating bool) error {
```
johscheuer, line 555: With the current implementation it might be hard but I think it would be useful to set the `status.Phase` also to terminating (if we want to use this field some day).

johscheuer (changes requested):
Could you fix the lint error + one question/comment otherwise 👍

mock-kubernetes-client/client/client_test.go
brownleej: I wanted to add a test that when one pod is stuck in terminating, another pod can still terminate successfully.

mock-kubernetes-client/client/client.go
brownleej: I considered that, but figured it was more trouble than it's worth right now because the operator doesn't check it.

johscheuer (approved):
LGTM 👍 feel free to merge once tests passed

Issue #558, opened 2021-03-18
Making sure that one stuck pod doesn't block other pods from getting config map updates


brownleej:
When we are checking for updates to files from the config map, we stop as soon as we get an instance that has old configuration. The process for checking also calls APIs in the sidecar for copying files over, so if we have the ability to update an instance I think we're better off trying that as soon as we can, even if another instance is failing. I think we should try to make this more robust so we update every instance we can as soon as we can.

johscheuer:
That should also speed up the reconciliation.

Issue #557, opened 2021-03-18
Only log "Waiting for config update" in the operator when we do not have the right config


brownleej:
When we check for updates to files from the config map, we log a "Waiting for config update" message. This seems to get logged even when the config map is already updated in the pod. I think we only want to log a message when we have to retry reconciliation to see if the config is updated later.

PR #561, opened 2021-03-19
Update all ConfigMaps

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/558
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/557

### DIFF
--- a/controllers/pod_client.go
+++ b/controllers/pod_client.go
@@ -21,6 +21,7 @@
 package controllers
 
 import (
+	"context"
 	"crypto/sha256"
 	"crypto/tls"
 	"crypto/x509"
@@ -33,8 +34,10 @@ import (
 	"os"
 	"strings"
 	"sync"
+	"time"
 
 	fdbtypes "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta1"
+	"github.com/hashicorp/go-retryablehttp"
 	corev1 "k8s.io/api/core/v1"
 )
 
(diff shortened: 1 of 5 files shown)


controllers/pod_client.go
```diff
@@ -143,30 +146,39 @@ func (client *realFdbPodClient) getListenIP() string {
...
 	var err error
 
-	httpClient := &http.Client{}
+	protocol := "http"
+	retryClient := retryablehttp.NewClient()
+	retryClient.RetryMax = 10
+	retryClient.RetryWaitMax = 5 * time.Second
+	// Prevent logging
+	retryClient.Logger = nil
+	retryClient.CheckRetry = func(c context.Context, resp *http.Response, err error) (bool, error) {
```
johscheuer, line 158: If we change the way how we invoke the `CheckHash` we could leverage this to actually check multiple times with an exponential backoff.

controllers/update_config_map.go
```diff
@@ -81,20 +84,39 @@ func (u UpdateConfigMap) Reconcile(r *FoundationDBClusterReconciler, context ctx
 		return false, err
 	}
 
+	var errs []error
+	// We try to update all instances and if we observe an error we add it to the error list.
+	// TODO: should we try to update the instances concurrently?
```
johscheuer, line 89: We could also try to update all instances "at once" since we probably waiting for I/O the most time.

controllers/update_config_map.go
```diff
@@ -81,20 +84,39 @@ func (u UpdateConfigMap) Reconcile(r *FoundationDBClusterReconciler, context ctx
...
+		if err != nil {
+			log.Info("Update Pod metadata", "namespace", configMap.Namespace, "cluster", cluster.Name, "processGroupID", instance.GetInstanceID(), "error", err)
+			errs = append(errs, err)
+		}
+	}
+
+	if len(errs) > 0 {
+		// If we return an error we don't requeue
+		// So we just return that we can't continue but don't have an error
+		return false, nil
```
johscheuer, line 119: The way how our reconcile loop work is pretty weird (and we might want to change this). If we return an error we just ignore the requeue time 🤔

brownleej: Yeah, I think this is an oversight and we should add the requeueAfter time to the error handling path as well.

brownleej (approved):
👍

controllers/update_config_map.go
johscheuer: I'll open an issue for this one.

Issue #559, opened 2021-03-19
exclude_instances.go has multiple iterations


johscheuer:
Currently in exclude_instances.go we iterate multiple times over the processGroups IMHO we can simplify this code to only run once over the slice.

PR #562, opened 2021-03-19
Remove redundant loop

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/559

### DIFF
--- a/controllers/exclude_instances.go
+++ b/controllers/exclude_instances.go
@@ -41,6 +41,7 @@ func (e ExcludeInstances) Reconcile(r *FoundationDBClusterReconciler, context ct
 	defer adminClient.Close()
 
 	addresses := make([]string, 0)
+	missingProcesses := make([]string, 0)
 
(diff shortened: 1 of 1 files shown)


controllers/exclude_instances.go
```diff
@@ -40,62 +40,54 @@ func (e ExcludeInstances) Reconcile(r *FoundationDBClusterReconciler, context ct
...
 
-	removalCount := 0
-	for _, processGroup := range cluster.Status.ProcessGroups {
-		if processGroup.Remove {
-			removalCount++
-		}
+	addresses := make([]string, 0)
+	missingProcesses := make([]string, 0)
+
+	exclusions, err := adminClient.GetExclusions()
```
brownleej, line 46: Are we sure we want to check the exclusion list on every reconciliation, even in the common case where it won't be used?

brownleej (changes requested):

johscheuer:
I think for now the current state is fine.

PR #567, opened 2021-03-21
Update FDB version in examples

johscheuer:
We should build some tooling to ensure we use the same version in the examples and the operator/sidecar.


brownleej (approved):
👍

PR #568, opened 2021-03-22
Add support for PDBs.

### LINKED ISSUE #316: Create design to prevent user to remove to many Pods

brownleej:
Resolves #316

### DIFF
--- a/controllers/cluster_controller_test.go
+++ b/controllers/cluster_controller_test.go
@@ -37,6 +37,7 @@ import (
 	gomegatypes "github.com/onsi/gomega/types"
 
 	corev1 "k8s.io/api/core/v1"
+	policyv1beta1 "k8s.io/api/policy/v1beta1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -188,6 +189,17 @@ var _ = Describe(string(fdbtypes.ProcessClassClusterController), func() {
 				err := k8sClient.List(context.TODO(), services, getListOptions(cluster)...)
 				Expect(err).NotTo(HaveOccurred())
 				Expect(len(services.Items)).To(Equal(0))
+			})
+
+			It("should create pod disruption budgets", func() {
+				budgets := &policyv1beta1.PodDisruptionBudgetList{}
+				err := k8sClient.List(context.TODO(), budgets, getListOptions(cluster)...)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(len(budgets.Items)).To(Equal(4))
+				sort.Slice(budgets.Items, func(i, j int) bool {
+					return budgets.Items[i].Name < budgets.Items[j].Name
+				})
+				Expect(budgets.Items[0].Name).To(Equal("operator-test-1-cluster-controller"))
 			})
 
 			It("should fill in the required fields in the configuration", func() {
--- a/controllers/pod_models.go
+++ b/controllers/pod_models.go
@@ -30,6 +30,7 @@ import (
 	fdbtypes "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	policyv1beta1 "k8s.io/api/policy/v1beta1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
@@ -1155,6 +1156,11 @@ func NormalizeClusterSpec(spec *fdbtypes.FoundationDBClusterSpec, options Deprec
 		spec.CustomParameters = nil
 	}
 
+	if spec.EnablePodDisruptionBudget == nil {
+		enabled := options.UseFutureDefaults
+		spec.EnablePodDisruptionBudget = &enabled
+	}
+
 	// Validate customParameters
 	for processClass := range spec.Processes {
 		if setting, ok := spec.Processes[processClass]; ok {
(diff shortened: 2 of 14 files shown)


controllers/update_pdbs.go
```diff
@@ -0,0 +1,88 @@
...
+	policyv1beta1 "k8s.io/api/policy/v1beta1"
+	"k8s.io/apimachinery/pkg/api/equality"
+)
+
+// UpdatePDBs provides a reconciliation step for adding services to a cluster.
+type UpdatePDBs struct{}
+
+// Reconcile runs the reconciler's work.
+func (u UpdatePDBs) Reconcile(r *FoundationDBClusterReconciler, context ctx.Context, cluster *fdbtypes.FoundationDBCluster) (bool, error) {
+	newPDBs, err := GetPodDisruptionBudgets(cluster)
```
johscheuer, line 37: I think semantically this should be rather `desiredPDBs`?

controllers/update_pdbs.go
```diff
@@ -0,0 +1,88 @@
...
+	pdbs := &policyv1beta1.PodDisruptionBudgetList{}
+	err = r.Client.List(context, pdbs, getPodListOptions(cluster, "", "")...)
+	if err != nil {
+		return false, err
+	}
+
+	for _, pdb := range pdbs.Items {
+		newPDB, pdbDesired := newPDBMap[pdb.Name]
+		if pdbDesired {
+			if !equality.Semantic.DeepEqual(pdb.Spec, newPDB.Spec) {
```
johscheuer, line 56: I would invert this if statement and just `continue` if the PDB has the latest spec.

controllers/update_pdbs.go
```diff
@@ -0,0 +1,88 @@
...
+				log.Info("Updating PodDisruptionBudget", "cluster", cluster.Name, "namespace", cluster.Name, "pdb", pdb.Name)
+
+				err = r.Update(context, &pdb)
+				if err != nil {
+					return false, err
+				}
+			}
+			delete(newPDBMap, pdb.Name)
+		} else {
+			err = r.Delete(context, &pdb)
```
johscheuer, line 67: Do we want a log statement for deletion?

controllers/pod_models.go
```diff
@@ -1250,6 +1256,39 @@ func NormalizeClusterSpec(spec *fdbtypes.FoundationDBClusterSpec, options Deprec
 	return nil
 }
 
+// GetPodDisruptionBudgets generates the disruption budgets that we need to
+// configure for a cluster.
+func GetPodDisruptionBudgets(cluster *fdbtypes.FoundationDBCluster) ([]policyv1beta1.PodDisruptionBudget, error) {
```
johscheuer, line 1261: I would move this code part into `update_pdbs.go` I don't think that we should keep it in `pod_models`?

controllers/pod_models.go
```diff
@@ -1250,6 +1256,39 @@ func NormalizeClusterSpec(spec *fdbtypes.FoundationDBClusterSpec, options Deprec
...
+	if !*cluster.Spec.EnablePodDisruptionBudget {
```
johscheuer, line 1262: Should we check if the value is nil (in theory this shouldn't happen)?

controllers/pod_models.go
```diff
@@ -1250,6 +1256,39 @@ func NormalizeClusterSpec(spec *fdbtypes.FoundationDBClusterSpec, options Deprec
...
+		return nil, err
+	}
+	processCountsMap := processCounts.Map()
+	for _, processClass := range fdbtypes.ProcessClasses {
+		count := processCountsMap[processClass]
+		if count > 0 {
+			metadata := getObjectMetadata(cluster, nil, processClass, "")
+			metadata.OwnerReferences = buildOwnerReference(cluster.TypeMeta, cluster.ObjectMeta)
+			metadata.Name = fmt.Sprintf("%s-%s", cluster.Name, processClassSanitizationPattern.ReplaceAllString(string(processClass), "-"))
+			minAvailable := intstr.FromInt(count - 1)
```
johscheuer, line 1277: I think for now this is okay but in the end we could be smarter and use `DesiredFaultTolerance` for SS and maybe a better default for `stateless` processes.

controllers/update_pdbs.go
```diff
@@ -0,0 +1,88 @@
...
+		return false, err
+	}
+
+	newPDBMap := make(map[string]policyv1beta1.PodDisruptionBudget, len(newPDBs))
+	for _, pdb := range newPDBs {
+		newPDBMap[pdb.Name] = pdb
+	}
+
+	pdbs := &policyv1beta1.PodDisruptionBudgetList{}
+	err = r.Client.List(context, pdbs, getPodListOptions(cluster, "", "")...)
```
johscheuer, line 48: We might want to revisit the name `getPodListOptions` since we use this method in multiple places that are not Pod related. IMHO this is also a good candidate for an exposed method.

johscheuer (changes requested):

controllers/cluster_controller_test.go
```diff
@@ -190,6 +191,17 @@ var _ = Describe(string(fdbtypes.ProcessClassClusterController), func() {
 				Expect(len(services.Items)).To(Equal(0))
 			})
 
+			It("should create pod disruption budgets", func() {
```
rbtcollins, line 194: we don't check the value of the budgets here; is that ok?

controllers/pod_models.go
```diff
@@ -1250,6 +1256,39 @@ func NormalizeClusterSpec(spec *fdbtypes.FoundationDBClusterSpec, options Deprec
 	return nil
 }
 
+// GetPodDisruptionBudgets generates the disruption budgets that we need to
+// configure for a cluster.
+func GetPodDisruptionBudgets(cluster *fdbtypes.FoundationDBCluster) ([]policyv1beta1.PodDisruptionBudget, error) {
+	if !*cluster.Spec.EnablePodDisruptionBudget {
+		return nil, nil
+	}
+	budgets := make([]policyv1beta1.PodDisruptionBudget, 0, len(fdbtypes.ProcessClasses))
```
rbtcollins, line 1265: so, this makes one PDB per process class. So each process class can be interrupted independently ? But we'll also get a PDB for Proxy and Resolver and so on, even when none are in use. So this is saying that FDB can handle 
1 storage + 1 transaction + 1 stateless + 1 resolution + 1 unset + 1 log ... all down at once.

I don't think that that is correct, is it?

rbtcollins: What advice would we give on the forums for someone planning manual migrations outside of kubernetes. Would it be 'move one process at a time', or 'no more than one stateful and one stateless at a time, and never more than $coordinatorcount/2 -1 a a time' ? or something like that : thats what I think we should aim to match.

rbtcollins:
I think this is a good start - we wouldn't be able to drop our PDB daemon with this though, as it doesn't watch replicated data health : a fast moving disruption of LocalPersistentVolumes could wipe out the cluster with this PDB implementation

brownleej:
This has been on hold for a while due to concerns about whether it's a good fit for FDB's replication. In addition to the problem of not checking replication health, there's the problem of not being able to handle a case where there are multiple storage processes on a single node. In a case like that it's fine to take them both offline, because they're not replicas, but we don't have a good way of expressing that in PDBs right now. As an alternative we could consider a pod-level PDB that is rapidly updated by pod-level checks on replication health, but that would still require a fault-domain-aware system for limiting things across nodes. I'm debating whether to close this or try and continue in the direction in this PR.

johscheuer (commented):
@brownleej do you think it's worth moving forward with this PR or should we close it? If I remember correctly the last state was that PDB are not exactly what we want and we have to move to a custom WebHook logic (which is another beast) and if someone wants to use PDB they have to implement them outside of the operator (for now).

brownleej:
I think we should close it at this point.

PR #571, opened 2021-03-23
Update controller-runtime to prevent deprected message

johscheuer:
With the patch release we don't see the deprecation warnings anymore: https://github.com/elastic/cloud-on-k8s/issues/4220


brownleej (approved):
👍

Issue #581, opened 2021-03-24
Allow to configure hardcoded MinimumUptimeSecondsForBounce


johscheuer:
This will be useful for some test cases.

PR #582, opened 2021-03-24
Allow to specify the minimum uptime seconds

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/581

### DIFF
--- a/api/v1beta1/foundationdbcluster_types.go
+++ b/api/v1beta1/foundationdbcluster_types.go
@@ -273,7 +273,7 @@ type FoundationDBClusterSpec struct {
 	// MinimumUptimeSecondsForBounce defines the minimum time, in seconds, that the
 	// processes in the cluster must have been up for before the operator can
 	// execute a bounce.
-	// +kubebuilder:validation:Minimum=0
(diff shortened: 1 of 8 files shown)


api/v1beta1/foundationdbcluster_types.go
```diff
@@ -269,6 +269,13 @@ type FoundationDBClusterSpec struct {
 	// This means that you end up with ProcessCounts["storage"] * StorageServersPerPod
 	// storage processes
 	StorageServersPerPod int `json:"storageServersPerPod,omitempty"`
+
+	// MinimumUptimeSecondsForBounce defines the minimum time, in seconds, that the
+	// processes in the cluster must have been up for before the operator can
+	// execute a bounce.
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:default:=600
```
johscheuer, line 277: What happens if we have an old Spec with the missing value --> test

johscheuer: TLDR: If the object already exists the value will be 0 :(

brownleej (approved):
👍

PR #584, opened 2021-03-25
Make use of patch instead of update

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/500


kubectl-fdb/cmd/remove_instances.go
```diff
@@ -24,6 +24,8 @@ import (
 	"fmt"
 	"log"
 
+	"golang.org/x/net/context"
```
johscheuer, line 27: TODO: remove

brownleej (approved):
👍

Issue #269, opened 2020-06-19
Self conflict


rbtcollins:
I'm not entirely sure how this comes about, but the operator manages to self-conflict fairly regularly. <details>
<summary>Here's a log snippet</summary>

```
2020-06-19T03:23:29.742Z        INFO    controller      Retrying reconcilation  {"reason": "Waiting for pod fdb-test/fdb-test-cluster/fdb-test-cluster-stateless-9 to be ready"}
2020-06-19T03:24:08.122Z        INFO    controller      Running command {"namespace": "fdb-test", "cluster": "fdb-test-cluster", "path": "/usr/bin/fdb/6.2/fdbcli", "args": ["/usr/bin/fdb/6.2/fdbcli", "--exec", "status json", "-C", "/tmp/286657896", "--log", "--timeout", "30", "--log-dir", "/var/log/fdb"]}
2020-06-19T03:24:14.732Z        INFO    controller      Command completed       {"namespace": "fdb-test", "cluster": "fdb-test-cluster", "output": "{\n    \"client\" : {\n ..."}
2020-06-19T03:24:17.904Z        INFO    controller      Running command {"namespace": "fdb-test", "cluster": "fdb-test-cluster", "path": "/usr/bin/fdb/6.2/fdbcli", "args": ["/usr/bin/fdb/6.2/fdbcli", "--exec", "status json", "-C", "/tmp/262145242", "--log", "--timeout", "30", "--log-dir", "/var/log/fdb"]}
2020-06-19T03:24:22.624Z        INFO    controller      Command completed       {"namespace": "fdb-test", "cluster": "fdb-test-cluster", "output": "{\n    \"client\" : {\n ..."}
2020-06-19T03:24:22.639Z        INFO    controller      Running command {"namespace": "fdb-test", "cluster": "fdb-test-cluster", "path": "/usr/bin/fdb/6.2/fdbcli", "args": ["/usr/bin/fdb/6.2/fdbcli", "--exec", "status minimal", "-C", "/tmp/018675996", "--log", "--timeout", "30", "--log-dir", "/var/log/fdb"]}
2020-06-19T03:24:22.931Z        INFO    controller      Command completed       {"namespace": "fdb-test", "cluster": "fdb-test-cluster", "output": "The database is avai..."}
[...]

PR #585, opened 2021-03-25
Check resource version to prevent conflicts

johscheuer:
fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/269 I validated with this change I don't see any conflicts anymore.


brownleej (approved):
👍

PR #588, opened 2021-03-26
Add an option to put pods in a crash loop state for testing purposes

brownleej:


controllers/delete_pods_for_buggification.go
```diff
@@ -0,0 +1,104 @@
...
+
+	fdbtypes "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta1"
+)
+
+// DeletePodsForBuggification provides a reconciliation step for recreating
+// pods with new pod specs when buggifying the config.
+type DeletePodsForBuggification struct{}
+
+// Reconcile runs the reconciler's work.
+func (d DeletePodsForBuggification) Reconcile(r *FoundationDBClusterReconciler, context ctx.Context, cluster *fdbtypes.FoundationDBCluster) (bool, error) {
```
johscheuer, line 35: I wonder if we really need a new reconciler for this. If we change the args shouldn't the operator in the current state be able to detect that the args have changed (desired args != current args) and the Pod needs an update e.g. replace?

johscheuer (commented):
One comment otherwise the changes lgtm.

controllers/delete_pods_for_buggification.go
brownleej: I went back and forth on that. There's two things that motivated creating a new reconciler. The first reason is that we want to delete these pods at a much earlier stage in reconciliation, which allows us to delete them reliably even when the cluster is in a bad state. That's important for making sure that we can de-buggify the pods to get the cluster back into a healthy state. The other is that we want to delete these pods even when it is unsafe to do so. That means we have to skip a lot of the logic in the existing UpdatePods reconciler. We could make UpdatePods more complex to handle both cases, and run it in two places in the reconciliation loop, but that might lead to more problems or more surprising behavior.

johscheuer (approved):
LGTM 👍

PR #594, opened 2021-03-30
Allow to replace processGroup when resource change

johscheuer:


brownleej (approved):
👍

controllers/replace_misconfigured_pods.go
```diff
@@ -221,9 +221,61 @@ func instanceNeedsRemoval(cluster *fdbtypes.FoundationDBCluster, instance FdbIns
...
+		if err != nil {
+			return false, err
+		}
+
+		for idx, container := range desiredSpec.Containers {
+			if equality.Semantic.DeepEqual(container.Resources, instance.Pod.Spec.Containers[idx].Resources) {
+				continue
+			}
+
+			if resourcesNeedsReplacement(container.Resources, instance.Pod.Spec.Containers[idx].Resources) {
```
johscheuer, line 235: Now that I merged it I had another idea. It probably makes sense to just compare the sum of the requests (e.g. in the current implementation we probably also don't handle the case for additional containers).

PR #596, opened 2021-04-06
Calculate the sum of resource requirements

johscheuer:
This makes the replacement more robust and calculates the sum of all containers e.g. when a new sidecar is added.

### DIFF
--- a/controllers/replace_misconfigured_pods.go
+++ b/controllers/replace_misconfigured_pods.go
@@ -25,6 +25,8 @@ import (
 	"fmt"
 	"time"
 
+	"k8s.io/apimachinery/pkg/api/resource"
+
 	"k8s.io/apimachinery/pkg/api/equality"
 
 	fdbtypes "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta1"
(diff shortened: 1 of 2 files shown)


brownleej (approved):
👍

controllers/replace_misconfigured_pods.go
```diff
@@ -227,53 +229,55 @@ func instanceNeedsRemoval(cluster *fdbtypes.FoundationDBCluster, instance FdbIns
...
-
-			if resourcesNeedsReplacement(container.Resources, instance.Pod.Spec.InitContainers[idx].Resources) {
-				log.Info("Replace instance",
-					"namespace", cluster.Namespace,
-					"name", cluster.Name,
-					"processGroupID", instanceID,
-					"reason", "Resource requests have changed")
-				return true, nil
-			}
+		if resourcesNeedsReplacement(desiredSpec.InitContainers, instance.Pod.Spec.InitContainers) {
```
johscheuer, line 241: Thinking about this again we could be even smarter and check if one of init container wants more resources than all other containers and only then trigger a replacement (.e.g if the actual required resources increase) but I think the current solution should be fine and we can revisit this one if needed.

Issue #602, opened 2021-04-08
Bug: readOnlyRootFilesystem is not overwritten


johscheuer:
I currently have the following FDB Spec:

```yaml
  processes:
    general:
      podTemplate:
        spec:
          containers:
          - name: foundationdb
            securityContext:
              readOnlyRootFilesystem: false
    storage:
      allowTagOverride: false
      customParameters:
      - knob_report_dd_metrics=0
```

and I also tried:

```yaml
  processes:
    general:
      podTemplate:
        spec:
          containers:
          - name: foundationdb
            securityContext:
              readOnlyRootFilesystem: false
    storage:
      allowTagOverride: false
      customParameters:
      - knob_report_dd_metrics=0
      podTemplate:
        spec:
          containers:
          - name: foundationdb
            securityContext:
              readOnlyRootFilesystem: false
```

I would expect that `readOnlyRootFilesystem` is set to `false` but the created pods have the field set to `true`:

```yaml
  containers:
  - name: foundationdb
    securityContext:
      readOnlyRootFilesystem: true
```

I assume that this is a bug somewhere in the merge process.

johscheuer:
This is the place: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/controllers/pod_models.go#L236-L239

PR #603, opened 2021-04-08
Fix ReadOnlyRootFilesystem override

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/602


brownleej (approved):
👍

Issue #494, opened 2021-02-11
Use fdb plugin to analyze cluster


johscheuer:
I think it would be very helpful (especially in multi cluster environments) if we would have a `kubectl fdb analyze` command that checks for some common errors or misconfiguration. So a very simple step could be like `check for pending Pod of this cluster`. I'm not sure how far we can bring this but I'm optimistic that we can catch some simple issues here that help investigating during an incident (or maybe before triggering an update)

PR #605, opened 2021-04-09
Initial support for fdb analyze command

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/494

I got a few todo's in the code that I would cleanup afterwards.

### DIFF
--- a/controllers/controllers.go
+++ b/controllers/controllers.go
@@ -76,6 +76,9 @@ const (
 	// FDBLocalityDCIDKey represents the key in the locality map that holds
 	// the DC ID.
 	FDBLocalityDCIDKey = "dcid"
+
+	// FDBManagedByLabel represents the label that is used to represent that the resource is managed by the operator
+	FDBManagedByLabel = "fdb-managed-by-operator"
 )
 
 // metadataMatches determines if the current metadata on an object matches the
--- a/controllers/controllers.go
+++ b/controllers/controllers.go
@@ -76,9 +76,6 @@ const (
 	// FDBLocalityDCIDKey represents the key in the locality map that holds
 	// the DC ID.
 	FDBLocalityDCIDKey = "dcid"
-
-	// FDBManagedByLabel represents the label that is used to represent that the resource is managed by the operator
-	FDBManagedByLabel = "fdb-managed-by-operator"
 )
 
 // metadataMatches determines if the current metadata on an object matches the
--- a/kubectl-fdb/cmd/analyze.go
+++ b/kubectl-fdb/cmd/analyze.go
@@ -211,19 +211,12 @@ func analyzeCluster(cmd *cobra.Command, kubeClient client.Client, clusterName st
 	cmd.Printf("Checking cluster: %s/%s\n", namespace, clusterName)
 	patch := client.MergeFrom(cluster.DeepCopy())
 
-	// 1. Check if the cluster is available and healthy!
+	// 1. Check if the cluster is available
 	if cluster.Status.Health.Available {
 		printStatement(cmd, "Cluster is available", false)
 	} else {
 		foundIssues = true
 		printStatement(cmd, "Cluster is not available", true)
-	}
-
-	if cluster.Status.Health.Healthy {
-		printStatement(cmd, "Cluster is healthy", false)
-	} else {
-		foundIssues = true
-		printStatement(cmd, "Cluster is not healthy", true)
 	}
 
 	if cluster.Status.Health.FullReplication {
@@ -241,8 +234,7 @@ func analyzeCluster(cmd *cobra.Command, kubeClient client.Client, clusterName st
 		printStatement(cmd, "Cluster is not reconciled", true)
 	}
 
-	// We could add here more field from cluster.Status.Generations and check if they are present.
-
+	// We could add here more fields from cluster.Status.Generations and check if they are present.
 	var replaceInstances []string
 	// 3. Check for issues in processGroupID
 	processGroupIssue := false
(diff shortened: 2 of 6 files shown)


kubectl-fdb/cmd/analyze.go
```diff
@@ -0,0 +1,338 @@
...
+		printStatement(cmd, "Pods are all running and available", false)
+	} else {
+		foundIssues = true
+	}
+
+	// TODO: add more checks using the FDB status directly e.g. error message or something else or overloaded or matching versions
+	// We could add more auto fixes in the future for now this is fine.
+	// TODO: should we only replace/remove Pods if the cluster is fully replicated -> cluster.Status.Health.FullReplication?
+	// There shouldn't be an issue with replacing and removing a Pod.
+	if autoFix {
```
johscheuer, line 317: In the long term we should try to merge these conditions into one place and use the same set for the operator and the plugin.

johscheuer:
This is the sample output:

```bash
Checking cluster: default/sample-cluster
✔ Cluster is available
✔ Cluster is healthy
✔ Cluster is fully replicated
✔ Cluster is reconciled
✔ ProcessGroups are all in ready condition
✔ Pods are all running and available
```

and

```bash
Checking cluster:default/sample-cluster
✔ cluster is available
✔ cluster is healthy
✔ cluster is not reconciled
✖ processGroup: storage-1 has the following condition: MissingProcesses since 2021-04-09 18:02:13 +0000 GMT
✖ Pod default/storage-3 has unexpected Phase Pending with Reason:
✖ Pod default/storage-3 has an unready container: foundationdb
```

johscheuer:
I'm not sure where the lint issue comes from, locally I don't see it 🤔

kubectl-fdb/cmd/analyze.go
```diff
@@ -0,0 +1,346 @@
...
+
+func newAnalyzeCmd(streams genericclioptions.IOStreams) *cobra.Command {
+	o := NewFDBOptions(streams)
+
+	cmd := &cobra.Command{
+		Use:   "analyze",
+		Short: "Analyze if the given clusters have any issues",
+		Long:  "Analyze if the given clusters have any issues",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			autoFix, err := cmd.Flags().GetBool("auto-fix")
```
johscheuer, line 50: TODO: add a prompt that show the changes that it will take and ask for confirmation.

kubectl-fdb/cmd/analyze.go
```diff
@@ -0,0 +1,346 @@
...
+		color.Unset()
+		return
+	}
+
+	color.Set(color.FgGreen)
+	cmd.Printf("✔ %s\n", line)
+	color.Unset()
+}
+
+func replaceFailedInstance(cmd *cobra.Command, cluster *fdbtypes.FoundationDBCluster, instanceID string) {
```
johscheuer, line 166: We should also additionally check if we have enough resources.

johscheuer: I would add this in a second PR.

kubectl-fdb/cmd/analyze.go
```diff
@@ -0,0 +1,392 @@
...
+
+	// 1. Check if the cluster is available and healthy!
+	if cluster.Status.Health.Available {
+		printStatement(cmd, "Cluster is available", false)
+	} else {
+		foundIssues = true
+		printStatement(cmd, "Cluster is not available", true)
+	}
+
+	if cluster.Status.Health.Healthy {
```
brownleej, line 222: I'm not sure if this check is worth including. The `healthy` flag in FDB covers a lot of things that aren't obvious from the name, and this could lead to noise in the output.

kubectl-fdb/cmd/analyze.go
```diff
@@ -0,0 +1,392 @@
...
+			}
+		}
+
+		if deletePod && autoFix {
+			killPods = append(killPods, pod)
+		}
+
+		if pod.DeletionTimestamp != nil && pod.DeletionTimestamp.Add(5*time.Minute).Before(time.Now()) {
+			podIssue = true
+			statement := fmt.Sprintf("Pod %s/%s stuck in terminating should be deleted since %s", namespace, pod.Name, pod.DeletionTimestamp)
```
brownleej, line 320:
```suggestion
			statement := fmt.Sprintf("Pod %s/%s has been stuck in terminating since %s", namespace, pod.Name, pod.DeletionTimestamp)
```

controllers/controllers.go
```diff
@@ -76,6 +76,9 @@ const (
 	// FDBLocalityDCIDKey represents the key in the locality map that holds
 	// the DC ID.
 	FDBLocalityDCIDKey = "dcid"
+
+	// FDBManagedByLabel represents the label that is used to represent that the resource is managed by the operator
+	FDBManagedByLabel = "fdb-managed-by-operator"
```
brownleej, line 81: This seems to be used as a label when fetching pods in the analyze command, but it's not set by the operator.

brownleej (changes requested):

kubectl-fdb/cmd/analyze.go
johscheuer: That's right, that was also one of my concerns and is a good point. I will remove it.

controllers/controllers.go
johscheuer: I think I added it in the first place but removed it from the selector again. I will remove it completely.

brownleej (approved):
👍

Issue #168, opened 2020-03-14
Automatically setting the backup version based on the cluster version


brownleej:
The backup agents need to have their FDB version synced to the version of FDB that the cluster is running in order to make progress with the backup. We currently require that these two versions be configured independently. Instead, we should have the reconciliation for the backup use the cluster's version if the backup's version is unspecified, and the reconciliation for the cluster should automatically set the reconciliation for the backup during upgrades.

johscheuer:
Closing: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/607#issuecomment-822708694

PR #607, opened 2021-04-12
Use the fdb cluster version for the backup version

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/168

### DIFF
--- a/controllers/update_backup_agents.go
+++ b/controllers/update_backup_agents.go
@@ -54,7 +54,12 @@ func (u UpdateBackupAgents) Reconcile(r *FoundationDBBackupReconciler, context c
 		}
 	}
 
-	deployment, err := GetBackupDeployment(backup)
+	version, err := r.getBackupVersion(context, backup)
+	if err != nil {
+		return false, err
+	}
+
+	deployment, err := GetBackupDeployment(backup, version)
 	if err != nil {
 		r.Recorder.Event(backup, corev1.EventTypeWarning, "GetBackupDeployment", err.Error())
 		return false, err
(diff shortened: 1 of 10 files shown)


brownleej:
How will this behave when we upgrade the cluster? Will it automatically trigger a reconciliation of the backup?

johscheuer:
> How will this behave when we upgrade the cluster? Will it automatically trigger a reconciliation of the backup?

I have to validate this (I'm not sure if the Backup controller watches the clusters) but it will reconcile either with a watch event or the time trigger.

brownleej:
I've also been wondering if the idea of using the cluster spec to determine the configuration for the backup deployment is a little abstraction-breaking, since we now have this deployment based on a composite of two objects.

johscheuer:
I think the main question here is do we want to support Backups of non Kubernetes clusters (which probably don't have a cluster spec) otherwise I think this approach is reasonable and we currently use the cluster spec anyway.

controllers/update_backup_agents.go
```diff
@@ -54,7 +54,12 @@ func (u UpdateBackupAgents) Reconcile(r *FoundationDBBackupReconciler, context c
 		}
 	}
 
-	deployment, err := GetBackupDeployment(backup)
+	version, err := r.getBackupVersion(context, backup)
```
johscheuer, line 57: TODO: see if we can reconcile if the cluster status changes.

johscheuer:
Closing since we decided not to add this dependency

PR #608, opened 2021-04-12
Ignore removed processes for bounce

johscheuer:
We can ignore processes that are marked for removal during the bounce.


brownleej (approved):
👍

Issue #493, opened 2021-02-11
Plugin should be able to kill/restart all fdbservers


johscheuer:
It would be helpful if the kubectl plugin offers the possibility to restart (actually `pkill fdbserver` of all Pods of a cluster. This is especially useful with the multi cluster support.

PR #609, opened 2021-04-12
Initial support for the restart command for the plugin

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/493


kubectl-fdb/cmd/restart.go
```diff
@@ -0,0 +1,191 @@
...
+		}
+
+		processes = append(processes, pod.Name)
+	}
+
+	return processes, nil
+}
+
+func restartProcesses(cmd *cobra.Command, restConfig *rest.Config, kubeClient *kubernetes.Clientset, processes []string, namespace string) error {
+	for _, process := range processes {
```
johscheuer, line 182: TODO: add prompt like for the remove instances.

kubectl-fdb/cmd/restart.go
```diff
@@ -0,0 +1,203 @@
...
+			}
+			clusterName, err := cmd.Flags().GetString("fdb-cluster")
+			if err != nil {
+				return err
+			}
+			allProcesses, err := cmd.Flags().GetBool("all-processes")
+			if err != nil {
+				return err
+			}
+			allProcessesIncorrectCMD, err := cmd.Flags().GetBool("all-processes-with-incorrect-cmd")
```
brownleej, line 61: I wonder if there's a way we can make this a little more generic. Maybe we could allow filtering by any condition?

brownleej: I'm also worried we could find ourselves in a place where we have multiple different commands implementing their own process / pod filtering logic in a way that leads to duplication of effort.

brownleej (changes requested):

kubectl-fdb/cmd/restart.go
johscheuer: Making this more generic sounds good 👍 I already planned as a next step to review the current mechanism and to merge (at least `remove instances` and this one) with a similar logic.

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -640,6 +640,30 @@ const (
 	MissingProcesses ProcessGroupConditionType = "MissingProcesses"
 )
 
+// GetProcessGroupConditionType returns the ProcessGroupConditionType for the matching string or an error
+func GetProcessGroupConditionType(processGroupConditionType string) (ProcessGroupConditionType, error) {
```
brownleej, line 644: Could we do this through reflection instead of a switch? That might make it easier to add more conditions in the future.

brownleej (approved):
👍

Small suggestion but I think it's good to go as it is.

api/v1beta1/foundationdbcluster_types.go
johscheuer: Good point, I take look 👍

johscheuer: I would leave it as it is and we could refactor this in a later step. We probably have to revisit this anyway once we move to the FoundationDBProcessGroup.

PR #613, opened 2021-04-14
Remove Pods before running the bounce command

johscheuer:
Otherwise the waiting for a bounce can block the deletion of Pods.


brownleej:
Couldn't this instead put us in a position where the deletion of pods blocks bouncing processes? I think that would be more likely based on the safety checks and throttling we have for deleting pods.

johscheuer:
That's right. With https://github.com/FoundationDB/fdb-kubernetes-operator/pull/608 we shouldn't see my specific issue anymore (where a Pod is replaced but stuck in Pending and blocks the bounce step because the process doesn't have an address).

PR #617, opened 2021-04-16
When configuring sidecar, use the version number in cluster Status.Ru…

sears:
…nningVersion instead of trying to parse it from the foundationdb image name

I confirmed that (at least in the scenario I tested), the cluster status version is populated before this is called.

### DIFF
--- a/controllers/update_status.go
+++ b/controllers/update_status.go
@@ -25,7 +25,6 @@ import (
 	"fmt"
 	"reflect"
 	"sort"
-	"strings"
 	"time"
 
 	fdbtypes "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta1"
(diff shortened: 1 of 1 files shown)


controllers/update_status.go
```diff
@@ -599,8 +598,7 @@ func validateInstance(r *FoundationDBClusterReconciler, context ctx.Context, clu
 	var needsSidecarConfInConfigMap bool
 	for _, container := range instance.Pod.Spec.Containers {
 		if container.Name == "foundationdb" {
-			image := strings.Split(container.Image, ":")
-			version, err := fdbtypes.ParseFdbVersion(image[len(image)-1])
+			version, err := fdbtypes.ParseFdbVersion(cluster.Status.RunningVersion)
```
johscheuer, line 601: I think this is fine in the short-term but wen probably want to use the version of the process group (not yet implemented).

johscheuer (approved):

PR #618, opened 2021-04-16
Use the same mechanism to generate the sidecar image

johscheuer:
Before we used different methods to generate the sidecar image which lead to a different behaviour when someone sets a custom sidecar.


sears (approved):
I confirmed this works with with custom images in my test environment.

PR #619, opened 2021-04-16
Add validation to the FDB version

johscheuer:
That should help to catch a misconfiguration in the FDB version before the operator has to complain.

### DIFF
--- a/api/v1beta1/foundationdb_version.go
+++ b/api/v1beta1/foundationdb_version.go
@@ -1,0 +1,151 @@
+/*
+ * foundationdb_version.go
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2021 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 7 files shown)


api/v1beta1/foundationdb_version.go
```diff
@@ -0,0 +1,151 @@
+/*
```
johscheuer, line 1: I move this code into a separate file only for readability.

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -70,6 +70,7 @@ func init() {
 // FoundationDBClusterSpec defines the desired state of a cluster.
 type FoundationDBClusterSpec struct {
 	// Version defines the version of FoundationDB the cluster should run.
+	// +kubebuilder:validation:Pattern:=^(\d+)\.(\d+)\.(\d+)$
```
brownleej, line 73: This is fine for now, but we may need to change it in the future to support pre-release versions with tags. When that comes we'll need to update the pattern both in this file and in the foundationdb_version.go file.

brownleej (approved):
👍

api/v1beta1/foundationdbcluster_types.go
johscheuer: That's fair 👍 I just think it's useful to provide as much validation as possible (or at least have the same validation in the code and the CRD) during the creation time otherwise the creation will fail and you have to dig through the logs or the events.

PR #620, opened 2021-04-16
Remove unused scheme

johscheuer:
This is an old artefact from the upgrade of the controller-runtime and is not required.


brownleej (approved):
👍

PR #621, opened 2021-04-16
Initial add of an e2e test

johscheuer:
Todos:

- [x] Rewrite all tests to use Ginkgo (so we can use the selector)
- [x] Add a make target to run `e2e`
- [ ] ~Define what `e2e` tests we want to run in CI (main limitation will be the runner cpu/memory)~
- [ ] ~Print state if a test fails~
- [X] Create documentation and follow up tasks for the e2e test suite

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -139,7 +139,7 @@ jobs:
         kubectl rollout status deploy fdb-kubernetes-operator-controller-manager --timeout=120s
         # Check if all resources can be applied
         kubectl apply --dry-run=server -f ./config/samples
-        make test
+        RUN_E2E=true make test
         # Change the image back to the default otherwise the next check will complain
         cd config/manager && kustomize edit set image controller=fdb-kubernetes-operator:latest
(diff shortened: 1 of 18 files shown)


.github/workflows/pull_request.yml
```diff
@@ -134,15 +134,13 @@ jobs:
         kubectl apply -f https://raw.githubusercontent.com/FoundationDB/fdb-kubernetes-operator/master/config/crd/bases/apps.foundationdb.org_foundationdbrestores.yaml
         # Generate secrets used by the operator
         ./config/test-certs/generate_secrets.bash
-        IMG=localhost:5000/fdb-kubernetes-operator:latest make clean all docker-build docker-push deploy
+        SKIP_TEST=1 IMG=localhost:5000/fdb-kubernetes-operator:latest make clean all docker-build docker-push deploy
```
johscheuer, line 137: We skip the tests here because we want to deploy the operator first and then pull all tests including the e2e test(s)

brownleej (approved):
👍

Issue #622, opened 2021-04-17
Current 0.32 macOS kubectl fdb plugin is not a valid binary


mologie:
Hello, I noticed that the recently released 0.32 kubectl-fdb binary cannot be run on macOS. It starts with a bunch of zeroes, so it's not a valid binary (no mach-o header).

Additionally the checksum file does not match what's downloaded. I have tried to download again with no change. Interestingly though the tar.gz file extracted just fine.

```
# hiccup @ reginn in ~/Downloads/kubectl-fdb-v0.32.0-macos [16:02:31] 
$ ./kubectl-fdb 
zsh: exec format error: ./kubectl-fdb

# hiccup @ reginn in ~/Downloads/kubectl-fdb-v0.32.0-macos [16:02:34] C:130
$ xxd kubectl-fdb | head -n10
00000000: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000010: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000020: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000030: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000040: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000050: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000060: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000070: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000080: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000090: 0000 0000 0000 0000 0000 0000 0000 0000  ................

# hiccup @ reginn in ~/Downloads [16:05:40] 
$ cat kubectl-fdb-v0.32.0-macos.tar.gz.sha256 
31246b18fddcf92394328fa8a86234497c06d5974c0a19b584210015bad8ff2f

# hiccup @ reginn in ~/Downloads [16:06:16] 
$ sha256sum kubectl-fdb-v0.32.0-macos.tar.gz          
58e300521b84a01fb0d9afd8ebaa446e0d7bcdfbaabe4e778cd1e4a8ce0ec682  kubectl-fdb-v0.32.0-macos.tar.gz
```

My gut says a build machine of yours might have broken RAM (garbled binary during tar'ing, garbled sha256 input during checksumming).

johscheuer:
Thanks for reporting! There might be a problem with the GitHub actions.

[...]

PR #623, opened 2021-04-17
Make use of sha256sum instead of hashFiles and check if the final binary works

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/622

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -20,6 +20,8 @@ jobs:
       - name: Create Release
         id: create_release
         uses: actions/create-release@v1
+        # Otherwise we can't rerun this to build new binaries
+        continue-on-error: true
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -20,6 +20,8 @@ jobs:
       - name: Create Release
         id: create_release
         uses: actions/create-release@v1
+        # Otherwise we can't rerun this to build new binaries
+        continue-on-error: true
```
johscheuer, line 24: This can be useful if the workflow succeeds but we want to rebuild the binaries currently this will fail.

.github/workflows/release.yaml
```diff
@@ -77,8 +87,6 @@ jobs:
           asset_path: ./bin/kubectl-fdb.tar.gz
           asset_name: ${{ steps.set_package_name.outputs.package_name }}
           asset_content_type: application/gzip
-      - name: Create sha256 for assert
-        run: echo "${{ hashFiles('./bin/kubectl-fdb.tar.gz') }}" > ./bin/${{ steps.set_package_name.outputs.package_name }}.sha256
```
johscheuer, line 81: the `hashFiles` function does not what we expected:

```
This function calculates an individual SHA-256 hash for each matched file, and then uses those hashes to calculate a final SHA-256 hash for the set of files.
```

brownleej (approved):
👍

PR #625, opened 2021-04-21
Fix the reconcile order for updating the storage server per Pod

johscheuer:
I guess we introduced a bug when we refactored the process group creation. In the current behaviour the config map is never updated and the Pods can't start. I simply split the `UpdateConfigMap` into two different methods.

### DIFF
--- a/controllers/update_pod_config.go
+++ b/controllers/update_pod_config.go
@@ -1,0 +1,93 @@
+/*
+ * update_config_map.go
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2019 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


e2e/cluster_test.go
```diff
@@ -154,51 +154,62 @@ var _ = Describe("[e2e] cluster tests", func() {
...
-			Expect(true).To(BeTrue())
+			clusterReconciled(runtimeClient, testCluster)
 		})
 
-		AfterEach(func() {
-			err := runtimeClient.Delete(context.Background(), testCluster)
-			Expect(err).NotTo(HaveOccurred())
+		When("changing the storageServerPerPod", func() {
+			BeforeEach(func() {
+				clusterReconciled(runtimeClient, testCluster)
```
johscheuer, line 162: I assume we either hit out resource limit or we take longer than expected. I add some debug statements for the failure.

johscheuer:
For now I will skip the test and I will create a follow up PR to allow running all e2e tests e.g. for local tests.

controllers/update_pod_config.go
```diff
@@ -0,0 +1,93 @@
+/*
+ * update_config_map.go
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2019 Apple Inc. and the FoundationDB project authors
```
brownleej, line 6:
```suggestion
 * Copyright 2021 Apple Inc. and the FoundationDB project authors
```

brownleej (approved):
👍

Issue #552, opened 2021-03-16
Make it easier for operator clients to avoid the FDB C++ headers.


sears:
We recently started using string constants from the package `"github.com/FoundationDB/fdb-kubernetes-operator/controllers"` which created a compile time dependency on the C++ FDB headers.  We'd rather not complicate our build environment by taking a direct dependency on the headers.  After discussing offline, we came up with two ideas:

- Client-facing constants (and related methods) could be consolidated in an API package
- Implementation code that requires the C++ client could be put in its own package to avoid polluting the rest of the logic with build-time dependencies.

Either approach seems reasonable to me.

johscheuer:
I think both approaches make sense the first one has the benefit that all constants are in a common place and the latter one has the benefit that we can share public methods from the controller 👍

johscheuer:
With the latest changed in the code only the usage of the `fdbclient` package should require the c++ headers. I'll close the issue. If there are still concerns/issue real free to reopen it.

Issue #624, opened 2021-04-17
Remove the requirement for the fdb-client


johscheuer:
Related to https://github.com/FoundationDB/fdb-kubernetes-operator/issues/552 the plugin currently requires the fdb client packages installed which limits some of the usability (e.g. running it in a container). We should remove this dependency from the plugin as long as we don't use any features of the fdb client.

```bash
$ docker run -ti ubuntu:20.04
# Install plugin
./kubectl-fdb version --client-only
./kubectl-fdb: error while loading shared libraries: libfdb_c.so: cannot open shared object file: No such file or directory
$  dpkg -i fdb.deb
...
$ ./kubectl-fdb version --client-only
kubectl-fdb: v0.32.0
```

johscheuer:
Fixed in: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/663

PR #631, opened 2021-04-24
Move labels into api package

johscheuer:
That solves partially https://github.com/FoundationDB/fdb-kubernetes-operator/issues/552 since most people want to use the labels.


johscheuer (commented):
This is also the first step for https://github.com/FoundationDB/fdb-kubernetes-operator/issues/624 (there are a few methods that are still used in the plugin that should be moved to the api package or a common package).

brownleej (approved):
👍

PR #635, opened 2021-04-27
Pull the FDB binaries for the operator from init containers to decouple FDB versions and operator versions.

### LINKED ISSUE #610: Using the sidecar to provide binaries to the operator instead of packaging all of them in the image
We currently provide the FDB client binaries in the operator docker image. This means we have to update the operator image to support new versions of FoundationDB. I think this is a coupling that we should avoid. As an alternative, we should consider deploying the operator with the FDB sidecar image as an init container, so it can copy the binaries into a shared volume that the main image can use. We have an example of this pattern in our client examples.

brownleej:
Resolves #610

### DIFF
--- a/helm/fdb-operator/templates/manager/deployment.yaml
+++ b/helm/fdb-operator/templates/manager/deployment.yaml
@@ -29,12 +29,68 @@ spec:
       - name: tmp
         emptyDir: {}
       - name: logs
-        emptyDir: {}        
+        emptyDir: {}
+      - name: fdb-binaries
+        emptyDir: {}
+      initContainers:
+        - name: foundationdb-kubernetes-init-6-1
+          image: foundationdb/foundationdb-kubernetes-sidecar:6.1.13-1
+          args:
+            - "--copy-library"
+            - "6.1"
+            - "--copy-binary"
+            - "fdbcli"
+            - "--copy-binary"
+            - "fdbbackup"
+            - "--copy-binary"
+            - "fdbrestore"
+            - "--output-dir"
+            - "/var/output-files/6.1.12"
+            - "--init-mode"
+          volumeMounts:
+            - name: fdb-binaries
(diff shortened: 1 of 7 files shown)


sears (approved):

helm/fdb-operator/templates/manager/deployment.yaml
```diff
@@ -29,12 +29,68 @@ spec:
...
+            - "--copy-library"
+            - "6.1"
+            - "--copy-binary"
+            - "fdbcli"
+            - "--copy-binary"
+            - "fdbbackup"
+            - "--copy-binary"
+            - "fdbrestore"
+            - "--output-dir"
+            - "/var/output-files/6.1.12"
```
johscheuer, line 48: If we copied the files to `/var/output-files/6.1` we wouldn't ned to copy them around in the manager?

main.go
```diff
@@ -130,6 +132,11 @@ func main() {
 		DeprecationOptions:  deprecationOptions,
 	}
 
+	if err = MoveFDBBinaries(); err != nil {
```
johscheuer, line 135: I think instead of moving the files in the manager code we should just copy the files to the correct place e.g. `/var/output-files/6.1`, `/var/output-files/6.2`, `/var/output-files/6.3`, etc.

johscheuer (changes requested):

helm/fdb-operator/templates/manager/deployment.yaml
brownleej: We still would, because the directory format is different. I plan to file a follow-up issue on changing the behavior of the sidecar to make the copying in the manager unnecessary, but I don't want to block on that because changes to the sidecar have a longer release cycle.

johscheuer (approved):

johscheuer:
LGTM 👍

PR #638, opened 2021-04-27
Remove action cache

johscheuer:
Since we added the cache we saw some weird failures and the benefit is not that huge.


johscheuer:
I think the cache is fine the issue was rather this one: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/643

Issue #641, opened 2021-04-28
The operator should set the trace format for fdbcli if FDB_NETWORK_OPTION_TRACE_FORMAT is set


johscheuer:
We use the env variable `FDB_NETWORK_OPTION_TRACE_FORMAT` to set the trace log format, that works fine except for all `fdbcli` calls so we should set `--trace-format` to the same value like `FDB_NETWORK_OPTION_TRACE_FORMAT` if set (otherwise we just use `xml` which is the default).

PR #642, opened 2021-04-28
Set the trace format in fdbcli

johscheuer:
fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/641


brownleej (approved):
👍

Issue #648, opened 2021-04-28
Test for root command failing


brownleej:
I was just running the test suite on one of my branches through a normal `make` command, and I got an error in the tests for running the root command: 

```
    Unexpected error:
        <*errors.errorString | 0xc0002ea2e0>: {
            s: "unknown shorthand flag: 'g' in -ginkgo.skip=^\\[e2e\\]",
        }
        unknown shorthand flag: 'g' in -ginkgo.skip=^\[e2e\]
    occurred
```

I think the issue is that when we leave the `RUN_E2E` flag unset, it sets a parameter on the start command for `go test`, which gets picked up when executing the root command inside the test.

brownleej:
@johscheuer Since you were looking at these flags recently, do you think you could take a look at this as well?

johscheuer:
That should be fixed with https://github.com/FoundationDB/fdb-kubernetes-operator/pull/633/files#diff-be9d711a60d5b19e7e761b57312c68c6d5f33e2954be751ecb968caf294b6bb6 since one test was not ported to Ginkgo. I verify that later and close the issue.

PR #649, opened 2021-04-29
Reset the args to prevent test failure

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/648

Interestingly the args passed to the framework will be passed to the `cmd` if we don't overwrite it 🤔


brownleej (approved):
👍

Issue #650, opened 2021-04-29
Metric for remove instances (without exclusion)


johscheuer:
That metrics could be interesting to see how many removals a cluster has over time.

PR #652, opened 2021-04-29
Add metrics for instances to remove

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/650


brownleej (approved):
👍 Looks good other than replacing `Exclusoon` with `Exclusion` in a couple places.

Issue #640, opened 2021-04-27
Automatic replacement deadlock


johscheuer:
We have a possible deadlock when automatic replacements are enable and a Pod get's created but is stuck in `Pending` (never get's an IP address assigned). After 30 min (or later) the operator will see that the `processGroup` has the `MissingProcesses` for more then 30 min and will attempt to replace it by setting it to the `instancesToRemove` list but that will never complete since the `processGroup` has no addresses. I think we can be smarter here and just remove the `processGroup` without exclusion.

brownleej:
I'm hesitant to go down the path of automatically removing processes without exclusion, given the risks of making mistakes in that area. It may also not be super helpful to do automatic replacements when a newly creating pod gets stuck in pending, since whatever replacement pod we create would likely hit the same blocker.

johscheuer:
I see I missed my second option here when I created the issue 🙈 I'm on your side that we should be very carful to remove processes without exclusion. I thin it should be safer to just ignore processes that have no addresses in the automatic replacement (since we would add them to the list but the replacement/exclusion could never finish). We already seen some cases where the automatic replacement would have helped: If a node is long enough healthy to be considered as schedulable but then crashes before the Pod get's started (and an IP address assigned) the automatic replace would help.

brownleej:
We might still want to guard against the case where no pod is able to schedule, to make sure that doesn't lead to the operator creating pods over and over again.

brownleej:
It's possible that in almost all cases where replacing a pod with no IP would be helpful, deleting the pod would accomplish the same thing. What do you think?

johscheuer:
I think that sounds reasonable and we probably can build a method to delete a Pod for the following conditions:

1. Pod has no address for given time period
2. Pod has a failing container for a given time period

We already saw the 2. case where for some kind of reasons the sidecar died and wasn't restarted (or ran into a backoff) the fdbserver process itself was fine but every other changes where blocked because the sidecar was not reachable.

johscheuer:
I'll create a new issue for the Pod deletion. In general I think that's a good idea on the other hand we might want to track how often we deleted a Pod (e.g. to prevent deleting the whole time a Pod that is stuck where the deletion doesn't help).

PR #653, opened 2021-04-29
Ignore process groups without an address

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/640 for now we just ignore process groups without an address.


brownleej (approved):
👍

Issue #377, opened 2020-11-12
Calculate last-applied-config-map process specific and not for all


johscheuer:
Currently the `last-applied-config-map` annotation is calculated over the whole `ConfigMap` rather than the interesting part for the process. This has some side effects  like if we change a specific `ConfigMap` for e.g. `storage` processes all processes will be "incorrect" which will create some unnecessary noise in the cluster. A better solution would be to take only the interesting part of the `ConfigMap` for the process into account.

PR #654, opened 2021-04-30
Calculate ConfigMap hash based on process class and servers per pod

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/377

With this change it's not that confusing why all processes must be "updated" when we change either the number of storage server per Pod or a class specific setting.

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -404,7 +404,7 @@ func setMonitorConfForFilename(cluster *fdbtypes.FoundationDBCluster, data map[s
 	return nil
 }
 
-func getConfigMapEntry(pClass fdbtypes.ProcessClass, serversPerPod int) string {
+func getConfigMapMonitorConfEntry(pClass fdbtypes.ProcessClass, serversPerPod int) string {
 	if serversPerPod > 1 {
 		return fmt.Sprintf("fdbmonitor-conf-%s-density-%d", pClass, serversPerPod)
 	}
@@ -417,7 +417,7 @@ func GetConfigMap(cluster *fdbtypes.FoundationDBCluster) (*corev1.ConfigMap, err
 	data := make(map[string]string)
 
 	connectionString := cluster.Status.ConnectionString
-	data[clusterFile] = connectionString
+	data[clusterFileKey] = connectionString
 	data["running-version"] = cluster.Status.RunningVersion
 
 	var caFile strings.Builder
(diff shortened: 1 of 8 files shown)


controllers/cluster_controller.go
```diff
@@ -707,15 +698,22 @@ func GetJSONHash(object interface{}) (string, error) {
 	return hex.EncodeToString(specHash), nil
 }
 
-// GetDynamicConfHash gets a hash of the data from the config map holding the
+// getDynamicConfHash gets a hash of the data from the config map holding the
```
johscheuer, line 701: I don't think that we want to expose this method same for the method below.

johscheuer (commented):
I tested an update of the operator and as expected after one reconcile loop the new labels are updated the cluster won't be changed over that time.

controllers/cluster_controller.go
```diff
@@ -707,15 +698,22 @@ func GetJSONHash(object interface{}) (string, error) {
...
 // cluster's dynamic conf.
 //
 // This will omit keys that we do not expect the pods to reference.
-func GetDynamicConfHash(configMap *corev1.ConfigMap) (string, error) {
-	var data = make(map[string]string, len(configMap.Data))
-	for key, value := range configMap.Data {
-		data[key] = value
+func getDynamicConfHash(configMap *corev1.ConfigMap, pClass fdbtypes.ProcessClass, serversPerPod int) (string, error) {
```
brownleej, line 705: Are we sure that there aren't other entries in the config map that are relevant to the processes? For instance, I think  we include the CA file in the config map.

brownleej: Though that may raise further questions about how we're rolling out changes like that and what the purpose if including them in the hash is.

controllers/controllers.go
```diff
@@ -32,6 +32,10 @@ var log = logf.Log.WithName("controller")
 // DefaultCLITimeout is the default timeout for CLI commands.
 var DefaultCLITimeout = 10
 
+const (
+	clusterFile = "cluster-file"
```
brownleej, line 36: This feels like a name that might conflict with other variable names. Should we make it something more like `clusterFileKey`, kind of like what we have down with the constants for labels on pods?

brownleej (changes requested):

controllers/cluster_controller.go
johscheuer: Yes, the `ca-file` is part of the configMap. I can add that and I guess it makes sense from a rollout perspective to include it.

brownleej (approved):
👍

PR #657, opened 2021-05-03
Move some shared code into an internal package

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/624

The PR moves some code into an internal package that is only used in this repo (and not public available) to reduce the number of methods we expose and to remove the need to have the client-library installed when you want to use the `kubectl-fdb` plugin. In the future we might move some additional methods into the [Internal package](https://dave.cheney.net/2019/10/06/use-internal-packages-to-reduce-your-public-api-surface).


brownleej:
Can you add some documentation on what we want to go into the `internal` package? If the goal is to limit the scope of the FDB client dependency, then can we narrowly target that dependency as the thing to move out?

johscheuer:
I think it makes sense to first tackle: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/681

PR #663, opened 2021-05-05
Move the FDB client dependencies into their own sub-package.

brownleej:
This should allow importing anything from the `controllers` sub-package without taking on a dependency on the FDB client library.


fdbclient/admin_client.go
```diff
@@ -0,0 +1,588 @@
...
+	log.Info("Command completed", "namespace", client.Cluster.Namespace, "cluster", client.Cluster.Name, "output", debugOutput)
+	return outputString, nil
+}
+
+// GetStatus gets the database's status
+func (client *cliAdminClient) GetStatus() (*fdbtypes.FoundationDBStatus, error) {
+	adminClientMutex.Lock()
+	defer adminClientMutex.Unlock()
+	// This will call directly the database and fetch the status information
+	// from the system key.
```
johscheuer, line 204:
```suggestion
	// from the system key space.
```

johscheuer (changes requested):
Looks good in general just a question if we want to make these methods/struct public available or if we want to move them in an internal package.

config/test-certs/cert.pem
```diff
@@ -1,28 +1,28 @@
 -----BEGIN CERTIFICATE-----
```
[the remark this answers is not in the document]
brownleej: Yes, I think we should commit it. This is a self-signed that we use for testing, so I don't think there's a security risk in committing it.

fdbclient/admin_client.go
```diff
@@ -0,0 +1,588 @@
+/*
```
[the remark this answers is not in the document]
brownleej: Most of the stuff in fdbclient is private and unexpected in this PR, but we want an exported type for the DatabaseClientProvider.

controllers/admin_client.go
```diff
@@ -22,40 +22,16 @@ package controllers
...
-	if flag == "" {
-		return 20
-	}
-	result, err := strconv.Atoi(flag)
-	if err != nil {
-		panic(err)
-	}
-	return result
-}
-
```
johscheuer, line 58: I think we should also move the `AdminClient interface` implementation into the new package and only keep the mock inside of controllers.

controllers/fdb_client.go
```diff
@@ -1,110 +1,78 @@
...
+	return p.AdminClientProvider(cluster, kubernetesClient)
+}
 
-		statusBytes := transaction.Get(fdb.Key(statusKey)).MustGet()
-		if len(statusBytes) == 0 {
-			return nil, err
-		}
+// CleanUpCache removes the cache entry for a cluster.
+func (p legacyDatabaseClientProvider) CleanUpCache(namespace string, name string) {
+}
```
johscheuer, line 60: It seems like the implementation is missing here?

fdbclient/admin_client.go
```diff
@@ -0,0 +1,592 @@
...
+	// Cluster is the reference to the cluster model.
+	Cluster *fdbtypes.FoundationDBCluster
+
+	// clusterFilePath is the path to the temp file containing the cluster file
+	// for this session.
+	clusterFilePath string
+}
+
+// NewCliAdminClient generates an Admin client for a cluster
+func NewCliAdminClient(cluster *fdbtypes.FoundationDBCluster, _ client.Client) (controllers.AdminClient, error) {
```
johscheuer, line 72: I guess `_ client.Client` is only there for backwards compatibility?

fdbclient/admin_client.go
```diff
@@ -0,0 +1,592 @@
...
+func (command cliCommand) getClusterFileFlag() string {
+	if command.binary == "fdbrestore" {
+		return "--dest_cluster_file"
+	}
+	return "-C"
+}
+
+// getBinaryPath generates the path to an FDB binary.
+func getBinaryPath(binaryName string, version string) string {
+
```
johscheuer, line 129: Could you remove the newline here?

fdbclient/admin_client.go
```diff
@@ -0,0 +1,592 @@
...
+	hardTimeout := DefaultCLITimeout
+	args := make([]string, 0, 9)
+	args = append(args, command.args...)
+	if len(args) == 0 {
+		args = append(args, "--exec", command.command)
+	}
+
+	args = append(args, command.getClusterFileFlag(), client.clusterFilePath, "--log")
+
+	if binaryName == "fdbcli" {
```
johscheuer, line 159: Nothing that you have to solve in this PR but I think it makes sense to move that into a method.

fdbclient/admin_client.go
```diff
@@ -0,0 +1,592 @@
...
+		} else {
+			results[address] = "Unknown"
+		}
+	}
+	return results
+}
+
+// KillInstances restarts processes
+func (client *cliAdminClient) KillInstances(addresses []string) error {
+
```
johscheuer, line 367: Could you remove the newline here?

johscheuer (changes requested):

controllers/admin_client.go
brownleej: That would require the controllers package to take a dependency on the fdbclient package, which means it would take a dependency on the FDB client library.

controllers/fdb_client.go
brownleej: I don't think we have a way to expose the cache clearing behavior, so we have a dummy implementation instead.

fdbclient/admin_client.go
brownleej: Yes

controllers/admin_client.go
johscheuer: Right, I missed that.

johscheuer (approved):
LGTM 👍

Issue #661, opened 2021-05-05
Sidecar should have a livenessProbe per default


johscheuer:
Currently the sidecar only has a `readinessProbe` to check if the Python process already has started up we should additionally add a `livenessProbe` so that the container get's restarted when the Python process hangs. We've seen a few Python processes in the Sidecar that have stopped working and killing the process resolves the issue. With a `livenessProbe` we don't have to manually kill the process. For the `livenessProbe` I would propose a default `periodSeconds` of `30s` and a `failureThreshold` of 2 which means after 2 minutes the container will be killed if not responding.

PR #666, opened 2021-05-06
Allow to set a livenessProbe for the sidecar

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/661

### DIFF
--- a/controllers/pod_models.go
+++ b/controllers/pod_models.go
@@ -518,6 +518,21 @@ func configureSidecarContainer(container *corev1.Container, initMode bool, insta
 		}
 
 		sidecarEnv = append(sidecarEnv, corev1.EnvVar{Name: "FDB_INSTANCE_ID", Value: instanceID})
+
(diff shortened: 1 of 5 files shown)


controllers/pod_models.go
```diff
@@ -518,6 +518,21 @@ func configureSidecarContainer(container *corev1.Container, initMode bool, insta
 		}
 
 		sidecarEnv = append(sidecarEnv, corev1.EnvVar{Name: "FDB_INSTANCE_ID", Value: instanceID})
+
+		if !initMode && cluster.Spec.SidecarContainer.EnableLivenessProbe && container.LivenessProbe == nil {
+			// We can't use a HTTP handler here since the server
+			// requires a client certificate
+			container.LivenessProbe = &corev1.Probe{
```
brownleej, line 525: Should we capture a follow-up task to remove the readiness probe, since it's not doing a lot of good for us right now?

brownleej (approved):
👍

PR #671, opened 2021-05-07
Enable using custom image tags

Daniel-B-Smith:
This is related to: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/610

The previous version regex would only match image tags that were exclusively a version tag (e.g. "7.0.0" but not "foo-7.0.0"). When building custom images in development, we want to be able to add some additional context to avoid conflicts / confusing with official releases. This PR allows for the version to be extracted from image tags like "foo-7.0.0".


johscheuer (changes requested):
You also have to adjust the regex here: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/api/v1beta1/foundationdbcluster_types.go#L73 (and run `make clean all`) otherwise the CRD will prevent setting the version.

Daniel-B-Smith:
I _think_ enough of the make command succeeded to get the changes needed.

config/samples/deployment.yaml
```diff
@@ -1,219 +0,0 @@
-apiVersion: v1
```
johscheuer, line 1: It seems that one step has failed since the deployment is removed. Let's have a look at this on Monday.

config/crd/bases/apps.foundationdb.org_foundationdbclusters.yaml
```diff
@@ -7604,7 +7604,7 @@ spec:
                 updatePodsByReplacement:
                   type: boolean
                 version:
-                  pattern: ^(\d+)\.(\d+)\.(\d+)$
+                  pattern: (\d+)\.(\d+)\.(\d+)
```
johscheuer, line 7607: We may want to run/add some tests to ensure versions like `prerelease-5.6.2` and `test-1.2.3-test` etc. are actually working like expected.

johscheuer: That's fine to only test the `ParseFdbVersion` the Kubernetes validation should behave the same way.

johscheuer (approved):

johscheuer:
Thanks!

PR #673, opened 2021-05-08
Make use of combined output

johscheuer:
In `fdbcli` support to output errors and warnings on stderr instead of stdout (https://github.com/apple/foundationdb/pull/4332) was added. To read all messages (stdout and stderr) we have to use `CombinedOutput` instead of `Output` only.


brownleej (approved):
👍

Issue #647, opened 2021-04-28
Merge ConfirmExclusionCompletion and RemovePods


johscheuer:
Currently we have two reconcile loops for removing Pods (actually we have a few more that are involved): `ConfirmExclusionCompletion` checks if all processes are excluded and `RemovePods` will remove the Pods and then include them. If a process is excluded there is actually no need to wait until all other processes are excluded. We should combine both methods and allow the reconciler to remove Pods that are excluded (or where we skip the exclusion) without waiting for all processes.

PR #674, opened 2021-05-08
Merge exclude check and process group removal

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/647

Merges the logic of both reconcilers. We allow to remove all process groups to be removed that are excluded (we don't have to wait for all to finish, since it's safe to remove an excluded process group).


brownleej (approved):
👍

PR #675, opened 2021-05-10
Ignore process group on error instead of setting condition

johscheuer:
Before we always set `IncorrectCommandLine` to true if we couldn't get a Pod client. In this case we ignore the failure and skip the Pod (we could also return the error).


brownleej (approved):
👍

PR #676, opened 2021-05-10
Add explicit schema for podTemplateSpec.metadata

jkylling:
Unknown metadata fields are pruned https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#field-pruning
If you change from v1beta CRDs to v1 CRDs the metadata field is pruned and the operator believes that every podSpec is incorrect.
Related to https://github.com/kubernetes-sigs/controller-tools/issues/448

### DIFF
--- a/config/crd/bases/apps.foundationdb.org_foundationdbbackups.yaml
+++ b/config/crd/bases/apps.foundationdb.org_foundationdbbackups.yaml
@@ -65,6 +65,23 @@ spec:
                 podTemplateSpec:
                   properties:
                     metadata:
+                      properties:
+                        annotations:
+                          additionalProperties:
+                            type: string
+                          type: object
+                        finalizers:
+                          items:
+                            type: string
+                          type: array
+                        labels:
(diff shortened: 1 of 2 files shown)


jkylling:
Thanks! Would it be possible to add unit tests to catch issues like this?

Another tiny annoyance when upgrading the CRDs is that some extra default values are populated (e.g., `podSpec.containers[].ports[].protocol` was empty, and is now TCP). The operator detects that the last applied podSpec is different from what is now defined in the foundationdbcluster CR and decides to bounce the pod, although the resulting pod is identical (except of `metadata.annotations["foundationdb.org/last-applied-spec"]`). Would the operator be able to avoid most no-ops like this.

johscheuer (commented):
The changes look in general good, I'm just not quite sure if we want to use a beta version directly. Let me evaluate that and come back to you.

johscheuer (approved):
LGTM 👍  Thanks!

Issue #174, opened 2020-03-21
Documenting how to configure backups in the user manual


brownleej:
Once we have the core features of backup management in place, we should add an example of how to configure backups in the user manual.

PR #679, opened 2021-05-11
Add instructions on testing a restore in our backup sample.

brownleej:
Remove a parameter from the backup and restore commands that is only available on fdbcli.
Extend the snapshot period in the local sample to prevent lots of small snapshots from slowing things down.

### DIFF
--- a/config/samples/cluster_with_backup.yaml
+++ b/config/samples/cluster_with_backup.yaml
@@ -39,6 +39,10 @@
 # 10. Uncomment the FoundationDBRestore section below and apply the YAML again.
 # 11. Wait for all resources to be reconciled.
 # 12. Open a CLI and check the test key.
+#
+# Once that is done, you can clean up the backup by running:
+#     kubectl exec deployment/sample-cluster-backup-agents -- fdbbackup delete -d "blobstore://minio@minio-service:9000/sample-cluster?bucket=fdb-backups"
+#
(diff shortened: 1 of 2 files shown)


config/samples/cluster_with_backup.yaml
```diff
@@ -21,6 +21,28 @@
 # If you are testing this in Docker Desktop, you can browse the local MinIO
 # instance at https://localhost:9000. Note: This will use a certificate that
 # is not in your browser's trust store, so you will get a security warning.
+#
+# If you want to test a restore, you can take the following steps:
```
johscheuer, line 25: I wonder if the user manual (or user manual backup) would be a better place to document these steps?

config/samples/cluster_with_backup.yaml
```diff
@@ -78,7 +100,8 @@ spec:
   version: 6.2.20
   clusterName: sample-cluster
   accountName: minio@minio-service:9000
-  snapshotPeriodSeconds: 60
```
johscheuer, line 81: Could you update the version above to `6.2.30`?

johscheuer (approved):

config/samples/cluster_with_backup.yaml
brownleej: Eventually it will, and there's a separate issue on that: #174

PR #680, opened 2021-05-12
Upload binaries instead of tar package

johscheuer:


johscheuer:
Here is a test release: https://github.com/johscheuer/fdb-kubernetes-operator/releases/tag/v0.0.0.1233

JoshuaMcManus (commented):
Do we also need to update the documentation on installing the binary since we'll no longer need to unzip the file?

johscheuer:
> Do we also need to update the documentation on installing the binary since we'll no longer need to unzip the file?

Good, point I adjust the docs 👍

brownleej (approved):
👍

Issue #685, opened 2021-05-14
External state leak into tests


rbtcollins:
Just ran go test ./... coming back to my PR after a while and it failed like so:

```
Failure [0.005 seconds]
[AfterSuite] AfterSuite 
/home/robertc/src/cognitedata/fdb-kubernetes-operator/e2e/cluster_test.go:52

  Unexpected error:
      <*errors.errorString | 0xc0002fa210>: {
          s: "no Auth Provider found for name \"azure\"",
      }
      no Auth Provider found for name "azure"
  occurred

  /home/robertc/src/cognitedata/fdb-kubernetes-operator/e2e/cluster_test.go:60
  ```

possibly this is just a bug, but if not - I certainly can understand wanting to be able to interact with external clusters from test code from time to time, but I think the default should be to be hermetic, with external interactions strictly opt-in....

johscheuer:
You should use `make test` instead of the go command directly that will set everything to ignore the e2e test. We have to document that we don't support `go ./...`.

rbtcollins:
Educating every go developer who ever encounters the project to use make test *before* they try `go test ./...` seems like a likely-to-fail scenario. I don't think we have to support `go test ./...`, but being safe if someone does use it is a different matter.

johscheuer:
I agree on the latter and it should be a small fix using build tags.

PR #686, opened 2021-05-14
Make use of build tags to prevent running e2e tests with go test

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/685 CC: @rbtcollins


brownleej (approved):
👍

Issue #683, opened 2021-05-13
Add threading to the sidecar


johscheuer:
Currently the sidecar only uses one thread which means a requests will block all other requests. I think we should add threading to the HTTP server in the sidecar. That allows that neither the operator or the readiness/livenessProbes will block and also we should be more robust to requests that are stuck. The only thing that we need to keep in mind is that we might want to change the way how we copy the files (e.g. use a random name for the temp file).

Issue #563, opened 2021-03-19
Remove unused endpoints in sidecar


johscheuer:
The sidecar has currently some unused endpoints that we probably want to remove:

- `copy_binaries`
- `copy_libraries`
- `refresh_certs`
- `restart`

brownleej:
I'm not sure if it makes sense to put this under the 1.0 milestone, since the sidecar is versioned independently of the operator.

johscheuer:
I'll close this issue, with the efforts for the new sidecar image we are not investing that much time into the current sidecar.

PR #688, opened 2021-05-16
Refactor sidecar

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/563
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/683

I added threading to the sidecar and removed the client auth for the `ready` endpoint, that will allow us to actually use a HTTP probe instead of an TCP probe.

### DIFF
--- a/foundationdb-kubernetes-sidecar/sidecar.py
+++ b/foundationdb-kubernetes-sidecar/sidecar.py
@@ -32,7 +32,12 @@ import stat
 import time
 import traceback
 import sys
+import tempfile
 from pathlib import Path
+
+from http.server import HTTPServer, BaseHTTPRequestHandler
+from socketserver import ThreadingMixIn
+import threading
 
 from watchdog.observers import Observer
 from watchdog.events import FileSystemEventHandler
@@ -58,7 +63,7 @@ class Config(object):
         )
         parser.add_argument(
             "--tls",
-            help=("This flag enables TLS for incoming " "connections"),
+            help=("This flag enables TLS for incoming connections"),
             action="store_true",
         )
         parser.add_argument(
@@ -103,7 +108,7 @@ class Config(object):
         )
         parser.add_argument(
             "--input-dir",
-            help=("The directory containing the input files " "the config map."),
+            help=("The directory containing the input files the config map."),
             default="/var/input-files",
         )
         parser.add_argument(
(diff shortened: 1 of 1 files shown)


foundationdb-kubernetes-sidecar/sidecar.py
```diff
@@ -359,7 +360,12 @@ def send_text(self, text, code=200, content_type="text/plain", add_newline=True)
         self.end_headers()
         self.wfile.write(response)
 
-    def check_request_cert(self):
+    def check_request_cert(self, path):
+        # The ready endpoint will not require a client certificate
+        # to allow the kubelet to do HTTP probes.
+        if path == "/ready":
```
johscheuer, line 366: Since we require a certificate on the TLS layer we can drop that. In general it would be good to have a way to use the HTTP probe.

foundationdb-kubernetes-sidecar/sidecar.py
```diff
@@ -483,21 +489,12 @@ def do_POST(self):
...
         try:
-            if not self.check_request_cert():
+            if not self.check_request_cert(self.path):
                 return
             if self.path == "/copy_files":
                 self.send_text(copy_files())
-            elif self.path == "/copy_binaries":
-                self.send_text(copy_binaries())
-            elif self.path == "/copy_libraries":
-                self.send_text(copy_libraries())
```
brownleej, line 493: I'm not clear on why we're getting rid of these endpoints.

brownleej (changes requested):
It looks like we're removing endpoints that the sidecar currently supports. Is that something we need to do in this PR, or can we have a separate discussion on that?

foundationdb-kubernetes-sidecar/sidecar.py
johscheuer: That's something separate. The operator doesn't use these endpoints but we can also leave them as they are.

brownleej: I'd rather have a separate discussion. Even though the operator doesn't use these endpoints, there's a possibility that someone else would, and we might want them to be available in some debugging scenarios.

brownleej (approved):
👍

Issue #667, opened 2021-05-07
Add a skip setting to not reconcile clusters


johscheuer:
It can be useful to have a `Skip` setting that can be enable to prevent the operator from doing any further actions on the cluster (and the related objects). The idea is to add this setting in the cluster spec and if it is enable we should skip the reconciliation loop and return a success (to prevent requeueing).

PR #697, opened 2021-05-20
Add an option to skip cluster reconciliation

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/667

### DIFF
--- a/controllers/cluster_controller_test.go
+++ b/controllers/cluster_controller_test.go
@@ -575,7 +575,6 @@ var _ = Describe(string(fdbtypes.ProcessClassClusterController), func() {
 						Expect(err).NotTo(HaveOccurred())
 					})
 
-
 					It("should not replace the pod", func() {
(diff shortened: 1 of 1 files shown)


brownleej (approved):
👍

thorn126:
I tested in my env, it sounds when I set `skip: true`, somehow, it got reset to `false`, could you kindly share some insights? More details could be found https://forums.foundationdb.org/t/shutdown-scale-down-fdb-on-ocp-kubernetes-and-keep-the-pvc/3236/6 .
Some background: we need to implement shutdown feature in our controller which depends on the apple operator's skip function. thank you.

johscheuer:
I responded to your forum post. Please check if you run any additional operator that manages the spec and resets you changes. The operator itself will never change the cluster spec.

thorn126:
thank you @johscheuer , it works without our operator. I'll check about our own controller.

PR #700, opened 2021-05-20
Add emptyMonitorConf as an option to disable all fdbserver processes …

Daniel-B-Smith:
…in a cluster

### DIFF
--- a/controllers/cluster_controller_test.go
+++ b/controllers/cluster_controller_test.go
@@ -279,6 +279,53 @@ var _ = Describe(string(fdbtypes.ProcessClassClusterController), func() {
 			})
 		})
 
+		Context("when buggifying an empty fdbmonitor conf", func() {
+			BeforeEach(func() {
+				cluster.Spec.Buggify.EmptyMonitorConf = true
+				err = k8sClient.Update(context.TODO(), cluster)
+				Expect(err).NotTo(HaveOccurred())
+
+				// The buggify config causes the reconciliation loop to never finish with the mock admin client.
+				shouldCompleteReconciliation = false
+			})
+
+			It("should update conf and bounce processes", func() {
+				pods := &corev1.PodList{}
+				err = k8sClient.List(context.TODO(), pods, getListOptions(cluster)...)
+				Expect(len(pods.Items)).To(Equal(len(originalPods.Items)))
+				sortPodsByID(pods)
+
+				cm, _ := GetConfigMap(cluster)
+
+				for _, file := range cm.Data {
+					Expect(file).To(Not(ContainSubstring("fdbserver")))
+				}
+
+				adminClient, err := newMockAdminClientUncast(cluster, k8sClient)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(adminClient).NotTo(BeNil())
+
+				// Converted the killed servers into a set since they may have been killed more than once during
+				// reconciliation.
+				killedSet := map[string]struct{}{}
+				for _, addr := range adminClient.KilledAddresses {
+					killedSet[addr] = struct{}{}
+				}
+
+				// All of the processes in the cluster should be killed.
+				processes := map[string]struct{}{}
+				for _, processGroup := range cluster.Status.ProcessGroups {
+					for i, addr := range processGroup.Addresses {
+						// +1 since the process list uses 1-based indexing.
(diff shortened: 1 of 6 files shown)


brownleej:
This may also require changes to ensure that we kill the existing fdbserver processes when we set this flag.

Daniel-B-Smith:
Do you mean to verify that the server processes are killed? If the fdbmonitor.conf file is updated, I'm pretty sure it will kill the existing processes (will test, obviously).

Daniel-B-Smith:
I eventually figured out what you meant by ensuring the processes are killed. The unit tests are passing, and the operator is behaving as expected on our EKS cluster.

controllers/update_status.go
```diff
@@ -448,7 +449,8 @@ func CheckAndSetProcessStatus(r *FoundationDBClusterReconciler, cluster *fdbtype
 			versionMatch = process.Version == cluster.Spec.Version || process.Version == fmt.Sprintf("%s-PRERELEASE", cluster.Spec.Version)
 		}
 
-		correct = correct && versionMatch
+		// If the `EmptyMonitorConf` is set, the commandline is by definition wrong since there should be no running processes.
+		correct = correct && versionMatch && !cluster.Spec.Buggify.EmptyMonitorConf
```
johscheuer, line 453: Shouldn't that be rather an `||` to set the `correct` to true if we set `EmptyMonitorConf` otherwise all processes will have the `IncorrectCommandLine` but since we explicitly don't want yo have any running process that shouldn't set the condition.

controllers/cluster_controller_test.go
```diff
@@ -279,6 +279,53 @@ var _ = Describe(string(fdbtypes.ProcessClassClusterController), func() {
...
+					Expect(file).To(Not(ContainSubstring("fdbserver")))
+				}
+
+				adminClient, err := newMockAdminClientUncast(cluster, k8sClient)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(adminClient).NotTo(BeNil())
+
+				// Converted the killed servers into a set since they may have been killed more than once during
+				// reconciliation.
+				killedSet := map[string]struct{}{}
```
johscheuer, line 310: I like that you use the empty struct here :) normally I try to prevent that and choose to use a `bool`.

johscheuer: I like the empty struct implementation I only figured out that it's a little bit confusing for people that are not used to go.

controllers/update_status.go
johscheuer: I have to try that in my dev environment. My expectation would be that once we killed the processes (and they are not running anymore) we don't want to have the `IncorrectCommandLine` condition anymore since we reached the desired state. If we still set the `IncorrectCommandLine` even if all processes are shutdown that might be confusing for users without additional knowledge.

Daniel-B-Smith: I'm certainly open to suggestions. I don't think there is a well resolved state for the operator when every fdbserver process is shut down. I also don't expect this to be used in normal operation.

johscheuer (commented):
Looking at it the cluster will never reconcile:

```
NAME                                                       GENERATION   RECONCILED   HEALTHY   VERSION   AGE
foundationdbcluster.apps.foundationdb.org/sample-cluster   2            1                      6.2.30    14m
```

because the operator an't change the coordinators (which makes sense):

```
"controllers.ChangeCoordinators","error":"unable to fetch connection string: The database is unavailable; type `status' for more information.\n"
```

I would prefer to make changes to reach a reconciled state everything else will be confusing for the user (are the processes are already stopped or not?) also all process groups are obviously in a "bad" state:

```yaml
  - addresses:
    - 10.1.12.25
    processClass: storage
    processGroupConditions:
    - timestamp: 1622186960
      type: IncorrectCommandLine
    - timestamp: 1622187302
      type: MissingProcesses
    processGroupID: storage-4
```

I'll chat later with @brownleej what we expect and how we can move forward with this.

controllers/update_status.go
brownleej: I think it's fine for buggify conditions to leave the cluster in an unreconcilable state, as long as we can change the options back after that and get the cluster back to a good state.

johscheuer (approved):
LGTM 👍

Issue #479, opened 2021-02-04
Move NormalizeClusterSpec into cluster specific file


johscheuer:
Noted in: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/474/files#r568097751

PR #705, opened 2021-05-23
Refactor code to move shared code into internal package

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/479


brownleej (approved):
👍

Issue #702, opened 2021-05-21
Only track the current IP address


johscheuer:
In a recent change we added the tracking for IP addresses. Instead of tracking all IP addresses (and the potential issue of having conflicts) we should only track the current IP address. If the process is under removal we should track all IP addresses since we need to exclude and include all of them.

PR #707, opened 2021-05-23
Only track latest address

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/702

The first part is removing the `MockPodIP` and the second part is the actual bug fix.


brownleej (approved):
👍

Issue #713, opened 2021-05-25
Coordinator selection bug


johscheuer:
Currently there is a bug in the coordinator selection introduces with the change to prefer storage processes https://github.com/FoundationDB/fdb-kubernetes-operator/pull/512 from my tests only clusters that use the dcid (e.g. spread across multiple DCs in the same Kubernetes clusters are affected) and only HA clusters are affected. The issue is that only storage Pods will be selected (and no Pods in the satellite). I'm working on a fix for that and hopefully I can provide a PR for that tomorrow. This issue is here for providing visibility of the bug. Once we solved the issue we create a patch release. I will add more testing for HA clusters to be sure we actually fix the issue.

Fun fact we didn't saw that issue with out local testing since we had less storage processes (and only storage processes) running then the number of desired coordinators.

PR #714, opened 2021-05-25
Fix coordinator selection

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/713


controllers/change_coordinators_test.go
```diff
@@ -40,79 +42,285 @@ var _ = Describe("Reconcile.SelectCandidates", func() {
...
-		Context("with candidates", func() {
-			var candidates []localityInfo
-			JustBeforeEach(func() {
-				candidates = make([]localityInfo, 0, len(status.Cluster.Processes))
+			candidates, err = selectCoordinators(cluster, status)
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		When("all processes are healthy", func() {
+			FIt("should only select storage processes in primary and remote and Tlog in satellite", func() {
```
brownleej, line 171:
```suggestion
			It("should only select storage processes in primary and remote and Tlog in satellite", func() {
```

controllers/change_coordinators_test.go
```diff
@@ -40,79 +42,285 @@ var _ = Describe("Reconcile.SelectCandidates", func() {
...
+
+			// generate status for 2 dcs and 1 sate
+			status.Cluster.Processes = generateProcessInfo(2, 2)
+
+			candidates, err = selectCoordinators(cluster, status)
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		When("all processes are healthy", func() {
+			FIt("should only select storage processes in primary and remote and Tlog in satellites", func() {
```
brownleej, line 229:
```suggestion
			It("should only select storage processes in primary and remote and Tlog in satellites", func() {
```

controllers/change_coordinators_test.go
johscheuer: Yeah, I still have to add some more tests to ensure the selection works also in some other weird cases or return the correct error. I actually thought I marked thee PR as draft 🤔

brownleej (approved):
👍

PR #715, opened 2021-05-25
Make use of the new db client

johscheuer:
Make usr of `DatabaseClientProvider` to prevent issues with cleaning up the DB cache.


brownleej (approved):
👍

PR #719, opened 2021-05-27
Allow specifying portions of the connection string

Daniel-B-Smith:
For the YCSB fast download project, we want to be able to specify the database name and coordinator generation ID without specifying the IP addresses. The former must match the values from the files being downloaded, but we don't know the values of the latter until the pods have already been created.

I am less than confident in this API. Arguably, `PartialConnectionString` should replace `SeedConnectionString` as the latter is simply a special case of the former. Currently, `PartialConnectionString` can set the database name and/or generation ID but cannot set the list of coordinator IPs.


api/v1beta1/foundationdbcluster_types.go
```diff
@@ -94,6 +94,11 @@ type FoundationDBClusterSpec struct {
 	// After the initial reconciliation, this will not be used.
 	SeedConnectionString string `json:"seedConnectionString,omitempty"`
 
+	// PartialConnectionString provides a way to specify part of the connection string
+	// (e.g. the database name and coordinator generation) without specifying the entire
+	// string.
+	PartialConnectionString ConnectionString `json:"partialConnectionString,omitempty"`
```
brownleej, line 100: Am I right in thinking that this cannot be used to specify the coordinators? If so, it may be good to make that explicit in the documentation.

Daniel-B-Smith: Done.

johscheuer: I wonder if we should merge that with `SeedConnectionString` otherwise we have multiple different settings to change the behaviour e.g. adding a bool `onlyForInitialize` and then. deprecate `SeedConnectionString`?

Daniel-B-Smith: We wouldn’t need the bool as the semantics of both are only for initialization. The reason to keep both is that forcing users to decompose a connection string they already have into its constituent parts is rather annoying. I see a case for both ways and am happy to go with whichever option the two of you would prefer.

brownleej: I think they're going to solve different purposes, and there's a good chance their usage will diverge in the long-terms. I see `seedConnectionString` as a hack for bootstrapping multi-KC clusters, and we want to develop a more robust solution that also allows instances of the operator to discover new connection strings from their peers if they get partitioned.

brownleej (approved):
👍

api/v1beta1/foundationdbcluster_types.go
johscheuer: That sounds reasonable. My only concern is/was that we add to many different settings that solve a similar problem. In the end it should be easy for an user to know which settings are required for a specific task.

johscheuer (approved):
LGTM 👍

api/v1beta1/foundationdbcluster_types.go
johscheuer: One NIT: wee may want to document if both `seedConnectionString` and `partialConnectionString` are set `seedConnectionString` will be used and `partialConnectionString` will be ignored (?) and the TLDR would be don't use both :)

Daniel-B-Smith:
@johscheuer good point. Added!

PR #734, opened 2021-06-01
Apply an annotation when we see an out-of-date config map.

### LINKED ISSUE #731: Using annotations to speed up config map updates
The Kubernetes docs suggest using annotations to trigger faster updates to config maps (https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically). If we made use of this in our operator, it could produce a meaningful speed-up to reconciliation, as it is common to see reconciliation spinning for a few minutes on config map syncing. We could set an annotation with the last time stamp when we saw an incorrect config map, and clear it when we see the correct contents.

brownleej:
Resolves #731

I did a test run of this locally, and it reduced the initial cluster creation time from 115 seconds to 40 seconds.

### DIFF
--- a/api/v1beta1/foundationdb_labels.go
+++ b/api/v1beta1/foundationdb_labels.go
@@ -29,6 +29,10 @@ const (
 	// config map.
 	LastConfigMapKey = "foundationdb.org/last-applied-config-map"
 
+	// OutdatedConfigMapKey provides the annotation name we use to store the
+	// timestamp when we saw an outdated config map.
+	OutdatedConfigMapKey = "foundationdb.org/outdated-config-map-seen"
+
 	// BackupDeploymentLabel provides the label we use to connect backup
 	// deployments to a cluster.
 	BackupDeploymentLabel = "foundationdb.org/backup-for"
--- a/controllers/update_pod_config.go
+++ b/controllers/update_pod_config.go
@@ -71,7 +71,7 @@ func (u UpdatePodConfig) Reconcile(r *FoundationDBClusterReconciler, context ctx
 			err = r.PodLifecycleManager.UpdateMetadata(r, context, cluster, instance)
 			if err != nil {
 				allSynced = false
-				log.Info("Update Pod metadata", "namespace", configMap.Namespace, "cluster", cluster.Name, "processGroupID", instance.GetInstanceID(), "error", err)
+				log.Info("Update Pod ConfigMap annotation", "namespace", cluster.Namespace, "cluster", cluster.Name, "processGroupID", instance.GetInstanceID(), "error", err)
 			}
 			continue
 		}
--- a/controllers/update_pod_config.go
+++ b/controllers/update_pod_config.go
@@ -66,10 +66,18 @@ func (u UpdatePodConfig) Reconcile(r *FoundationDBClusterReconciler, context ctx
 		if !synced {
 			allSynced = false
 			log.Info("Update dynamic Pod config", "namespace", cluster.Namespace, "cluster", cluster.Name, "processGroupID", instance.GetInstanceID(), "synced", synced, "error", err)
+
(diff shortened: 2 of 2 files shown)


api/v1beta1/foundationdb_labels.go
```diff
@@ -29,6 +29,10 @@ const (
 	// config map.
 	LastConfigMapKey = "foundationdb.org/last-applied-config-map"
 
+	// OutdatedConfigMapKey provides the annotation name we use to store the
+	// timestamp when we saw an outdated config map.
+	OutdatedConfigMapKey = "foundationdb.org/outdated-config-map-seen"
```
johscheuer, line 34: I wonder if we should rather put that as `new-config-map` or. `config-map-update` and putting the timestamp there when the ConfigMap changes (then we could directly set the annotation for all Pods.

controllers/update_pod_config.go
```diff
@@ -66,10 +66,18 @@ func (u UpdatePodConfig) Reconcile(r *FoundationDBClusterReconciler, context ctx
...
+			err = r.PodLifecycleManager.UpdateMetadata(r, context, cluster, instance)
+			if err != nil {
+				allSynced = false
+				log.Info("Update Pod metadata", "namespace", configMap.Namespace, "cluster", cluster.Name, "processGroupID", instance.GetInstanceID(), "error", err)
+			}
 			continue
 		}
 
 		instance.Metadata.Annotations[fdbtypes.LastConfigMapKey] = configMapHash
+		delete(instance.Metadata.Annotations, fdbtypes.OutdatedConfigMapKey)
```
johscheuer, line 80: Should we check if the Annotations contain that key and only then trigger an update?

controllers/update_pod_config.go
```diff
@@ -66,10 +66,18 @@ func (u UpdatePodConfig) Reconcile(r *FoundationDBClusterReconciler, context ctx
 		if !synced {
 			allSynced = false
 			log.Info("Update dynamic Pod config", "namespace", cluster.Namespace, "cluster", cluster.Name, "processGroupID", instance.GetInstanceID(), "synced", synced, "error", err)
+
+			instance.Metadata.Annotations[fdbtypes.OutdatedConfigMapKey] = time.Now().Format(time.RFC3339)
+			err = r.PodLifecycleManager.UpdateMetadata(r, context, cluster, instance)
+			if err != nil {
+				allSynced = false
+				log.Info("Update Pod metadata", "namespace", configMap.Namespace, "cluster", cluster.Name, "processGroupID", instance.GetInstanceID(), "error", err)
```
johscheuer, line 74:
```suggestion
				log.Info("Update Pod ConfigMap annotation", "namespace", cluster.Namespace, "cluster", cluster.Name, "processGroupID", instance.GetInstanceID(), "error", err)
```

That should make it a little bit more clear where the error happened.

johscheuer (changes requested):
The code itself looks good, just a question how/when we want to set this annotation.

api/v1beta1/foundationdb_labels.go
johscheuer: Thinking more about it I guess we want both.

brownleej: I think it will be helpful to be able to poke the kubelet with pod updates more than once per change, in case there's a blip and things get dropped. Having a separate annotation that gets triggered when we update the desired config map may also be useful, but with the implementation for `OutdatedConfigMapKey` it only shows up for a brief duration, which means we're avoiding clutter.

controllers/update_pod_config.go
brownleej: We're always going to be doing the update by the time we reach this branch, because we know the value for `LastConfigMapKey`, so I don't think checking if it already contains the `OutdatedConfigMapKey` is necessary.

johscheuer: Right 👍

johscheuer (approved):
LGTM 👍

Issue #678, opened 2021-05-11
Newly created process groups should have a condition


johscheuer:
We should add the `MissingPod`, `MissingPVC`, and `MissingService` condition to all newly created process groups.

johscheuer:
Pr was merged.

PR #739, opened 2021-06-02
Set the missing conditions when creating new process groups

johscheuer:
I thought we got an issue for that.


api/v1beta1/foundationdbcluster_types.go
```diff
@@ -484,12 +484,18 @@ func (processGroupStatus *ProcessGroupStatus) IsExcluded(remainingMap map[string
...
-		Remove:                 false,
-		Excluded:               false,
-		ProcessGroupConditions: make([]*ProcessGroupCondition, 0),
+		ProcessGroupID: processGroupID,
+		ProcessClass:   processClass,
+		Addresses:      addresses,
+		Remove:         false,
+		Excluded:       false,
+		ProcessGroupConditions: []*ProcessGroupCondition{
+			NewProcessGroupCondition(MissingProcesses),
```
johscheuer, line 493: I also thought about adding a new `Initialized` condition that will be reset the first time it appears in the update status reconciler.

brownleej:
This looks like it's addressing #678. Is that correct?

brownleej (approved):
👍

johscheuer:
> This looks like it's addressing #678. Is that correct?

Yes, I didn't find it when I searched 🤦

PR #742, opened 2021-06-04
Show available instead of healthy in the kubectl get command

johscheuer:
I think the healthy output is too flaky to be useful.


brownleej (approved):
👍

Issue #718, opened 2021-05-26
selectCoordinators should be deterministic


johscheuer:
Currently `selectCoordinators` in the https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/controllers/change_coordinators.go is not deterministic and returns different coordinators for multiple runs (even when the state is unchanged). That is not a problem currently since the operator only selects new coordinators when the old ones are not valid anymore. We should still change the code to always return a deterministic set of coordinators.

PR #747, opened 2021-06-07
Ensure that the coordinator selection is deterministic

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/718

The sorting makes the coordinator selection more expensive but I think that's worth the deterministic output and also we only select new coordinators when a coordinator is not available.

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -1097,6 +1097,39 @@ type localityInfo struct {
 
 	// The locality map.
 	LocalityData map[string]string
+
+	Class fdbtypes.ProcessClass
+}
+
+// These indexes are used for sorting and since we sort ascending
+func getClassIndex(cls fdbtypes.ProcessClass) int {
+	switch cls {
+	case fdbtypes.ProcessClassStorage:
+		return 0
+	case fdbtypes.ProcessClassLog:
+		return 1
+	case fdbtypes.ProcessClassTransaction:
+		return 2
+	}
+
+	return math.MaxInt32
+}
+
+// This will sort the processes according to the following rules:
+// First all storage processes next log processes and then tlog processes.
+// Inside each process class the processes will be sorted by the ID (lower IDs come first).
+// We have to do this to ensure we get a deterministic result for selecting the candidates
+// otherwise we get a (nearly) random result since processes are stored in a map which is by definition
+// not sorted and doesn't return values in a stable way.
+func sortLocalities(processes []localityInfo) {
+	// Sort the processes for ID to ensure we have a stable input
+	sort.SliceStable(processes, func(i, j int) bool {
+		if processes[i].Class == processes[j].Class {
+			return processes[i].ID < processes[j].ID
+		}
+
+		return getClassIndex(processes[i].Class) < getClassIndex(processes[j].Class)
+	})
 }
 
 // localityInfoForProcess converts the process information from the JSON status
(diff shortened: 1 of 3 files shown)


controllers/cluster_controller.go
```diff
@@ -1097,6 +1097,39 @@ type localityInfo struct {
...
+// First all storage processes next log processes and then tlog processes.
+// Inside each process class the processes will be sorted by the ID (lower IDs come first).
+// We have to do this to ensure we get a deterministic result for selecting the candidates
+// otherwise we get a (nearly) random result since processes are stored in a map which is by definition
+// not sorted and doesn't return values in a stable way.
+func sortLocalities(processes []localityInfo) {
+	// Sort the processes for ID to ensure we have a stable input
+	sort.SliceStable(processes, func(i, j int) bool {
+		if processes[i].Class == processes[j].Class {
+			return processes[i].ID < processes[j].ID
```
johscheuer, line 1128: This will sort alphanumeric, we could also sort only based on the instance ID number but I don't think that the additional effort adds anything.

controllers/cluster_controller.go
```diff
@@ -1097,6 +1097,39 @@ type localityInfo struct {
 
 	// The locality map.
 	LocalityData map[string]string
+
+	Class fdbtypes.ProcessClass
+}
+
+// These indexes are used for sorting and since we sort ascending
+func getClassIndex(cls fdbtypes.ProcessClass) int {
+	switch cls {
```
brownleej, line 1106: I'm nervous about the possibility for the non-exhaustiveness to introduce non-determinism down the line. When we're using this, maybe we could fall back to a direct string comparison for unknown process classes?

controllers/cluster_controller.go
brownleej: I agree.

controllers/cluster_controller.go
```diff
@@ -1188,35 +1222,42 @@ func chooseDistributedProcesses(processes []localityInfo, count int, constraint
...
 	}
 
+	// Sort the processes to ensure a deterministic result
+	sortLocalities(processes)
+
 	for len(chosen) < count {
 		choseAny := false
 
 		for _, process := range processes {
-			if !chosenIDs[process.ID] {
```
brownleej, line 1195: Is anything changing in the behavior of this section, or is this just rearranging the code to reduce nesting?

brownleej (changes requested):

controllers/cluster_controller.go
johscheuer: So there are two things we need to keep in mind:

1.) Currently the coordinator selection will try storage -> log -> transaction processes all other processes are currently ignore so in theory the default case should never happen.
2.) If the default case will happen this only means that all processes that are not storage/log/transaction will have the least preference (have the highest index) and all processes in that "bucket" will be sorted by their ID so the behaviour is deterministic.

controllers/cluster_controller.go
johscheuer: I only reduced the nesting to make it more readable.

controllers/cluster_controller.go
brownleej: You're right, since we're sorting by IP we don't have a risk of non-determinism.

brownleej (approved):
👍

PR #748, opened 2021-06-07
Build future versions

sears:


brownleej:
We intend to remove all of the copying from the docker images in the future, in favor of the new behavior of using init containers to bring in the binaries, so I'm not sure if this is necessary.

sears:
Good to know.   Daniel was blocked on not having it, and it's what I'm currently using to build 7.0 images.  I'm OK with keeping it around as a doomed feature branch until we can switch to the init containers approach.

sears:
When will you have the new container design working?   This is blocking creation of CI pipelines.

johscheuer:
> When will you have the new container design working? This is blocking creation of CI pipelines.

It's already working, we just have to change the way how the operator is created in the CI tests. I didn't know we had a todo for that. I'll look into that.

PR #749, opened 2021-06-08
Ensure we never remove a coordinator

johscheuer:
That's an additional safety guard to never remove a coordinator to ensure we don't remove a coordinator (since the exclusion doesn't have any effect on the coordinator state). We should wait until a new coordinator is chosen and it's safe to remove the Pod.


api/v1beta1/foundationdbcluster_types.go
```diff
@@ -2609,3 +2609,29 @@ func (clusterStatus *FoundationDBClusterStatus) AddStorageServerPerDisk(serversP
...
+
+// None represents an empty struct used for creating set where we don't care about the value.
+type None struct{}
+
+// GetCoordinatorSet returns the current coordinator addresses as a set based on the information
+// in the connection string of the FoundationDBClusterStatus
+func (cluster FoundationDBCluster) GetCoordinatorSet() map[string]None {
+	coordinators := make(map[string]None)
+	// Split the connection string to get only the addresses after the @
+	conSplit := strings.Split(cluster.Status.ConnectionString, "@")
```
brownleej, line 2621: Rather than relying on parsing the connection string, it might be cleaner to get it from the database status.

brownleej (changes requested):

api/v1beta1/foundationdbcluster_types.go
brownleej: We could also look at the process roles so we can check this by instance ID instead of using the address as an indirection. That may be more future-proof.

johscheuer: That sounds reasonable, my idea was to prevent to have to many status calls against the FDB cluster but I guess that's a valid case.

internal/foundationdb_status.go
```diff
@@ -0,0 +1,41 @@
...
+func GetCoordinatorsFromStatus(status *fdbtypes.FoundationDBStatus) map[string]None {
+	coordinators := make(map[string]None)
+
+	for _, pInfo := range status.Cluster.Processes {
+		for _, roleInfo := range pInfo.Roles {
+			if roleInfo.Role != string(fdbtypes.ProcessRoleCoordinator) {
+				continue
+			}
+
+			coordinators[pInfo.Address] = None{}
```
brownleej, line 35: Since we've got the full process information here, I wonder if it would be identify the coordinators by process group ID rather than the address.

brownleej (changes requested):

internal/foundationdb_status.go
johscheuer: That makes sense and simplifies that part.

brownleej (approved):
👍

Issue #695, opened 2021-05-18
Unifying logging between replacing misconfigured pods and replacing failed pods


brownleej:
When we mark a process for removal in the replace_misconfigured_pods action, we log a message `Replace instance`, with a reason field providing more information. When we mark a process for removal in the replace_failed_pods action, we log a message `Replacing failed process group` with its own special fields. I think we should log the same message in these actions to make it easier to find replacements regardless of the cause.

Issue #604, opened 2021-04-09
Allow to define how may Pods should automatically replaced


johscheuer:
Currently we only replace one failed Pod when we enable the feature. I think it makes sense to allow the user to specify how many Pods should/can be replaced.

johscheuer:
Think about that a little bit we would implement that as some kind of a `max inflight replacements` which would be the upper limit of replacements for the automatic replacements. If the user would specify `5` that means the controller will evaluate how many process groups are currently marked for removal and if the limit is not reached the controller can replace as many process groups until the limit is reached. A concrete example a user sets the limit to `5` and the operator has `3` ongoing replacements, that means the operator will replace at most `2` process groups.

PR #754, opened 2021-06-10
Allow to define the number of concurrent automatic replacements

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/604 
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/695


controllers/replace_failed_pods.go
```diff
@@ -52,29 +53,50 @@ func chooseNewRemovals(cluster *fdbtypes.FoundationDBCluster) bool {
...
 			// replacing more failed pods.
-			return false
+			removalCnt++
 		}
 	}
+	maxReplacements := cluster.GetMaxConcurrentReplacements() - removalCnt
 
+	hasReplacement := false
 	for _, processGroupStatus := range cluster.Status.ProcessGroups {
+		if maxReplacements == 0 {
```
johscheuer, line 71:
```suggestion
		if maxReplacements <= 0 {
```

There are cases where we have a negative number --> Add a test case for that.

controllers/replace_failed_pods.go
```diff
@@ -52,29 +53,50 @@ func chooseNewRemovals(cluster *fdbtypes.FoundationDBCluster) bool {
 		return false
 	}
 
+	// The maximum number of removals will be the defined number in the cluster spec
+	// minus all currently ongoing removals e.g. process groups marked fro removal but
+	// not fully excluded.
+	removalCnt := 0
```
brownleej, line 59: Can we change this to `removalCount`? My brain consistently parses the word `cnt` as expanding to something else, so I would rather we spell it out 😊

brownleej (changes requested):

controllers/replace_failed_pods.go
johscheuer: Sure :)

controllers/replace_failed_pods_test.go
```diff
@@ -113,6 +113,36 @@ var _ = Describe("replace_failed_pods", func() {
...
+					It("should return true", func() {
+						Expect(result).To(BeTrue())
+					})
+
+					It("should mark the process group for removal", func() {
+						Expect(getRemovedProcessGroupIDs(cluster)).To(Equal([]string{"storage-2", "storage-3"}))
+					})
+				})
+
+				When("max concurrent replacements is set to one", func() {
```
johscheuer, line 132:
```suggestion
				When("max concurrent replacements is set to zero", func() {
```

09harsh (approved):
LGTM 👍

Issue #711, opened 2021-05-24
kubectl analyze should not print error message with auto-fix


johscheuer:
When we run `kubectl fdb analyze --auto-fix ..` we shouldn't print the error message about the cluster has issue since that's unexpected and we should only print errors that happens during the `auto-fix` apply. Additionally we should print a message when the user confirms the action to show that the `auto-fix` action is triggered.

Issue #708, opened 2021-05-24
kubectl analyze should also report process groups marked for removal


johscheuer:
Currently the analyze command skips process groups that are marked fro removal. Instead of skipping we should print those out as a warning otherwise that's confusing for the user why the cluster is not reconciled but the analyze command doesn't print out any further information.

PR #755, opened 2021-06-11
Print message for removals in analyze

johscheuer:
fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/708
fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/711


brownleej (approved):
👍

Issue #764, opened 2021-06-17
PVC stuck in terminating is blocking reconciliation


brownleej:
When we have a pod that is stuck in terminating, but the pod is fully excluded, we want to allow reconciliation to go forward. We have special logic in the RemovePods action that does this. However, if we have a PVC stuck in terminating, reconciliation gets blocked. Generally speaking if a pod is stuck in terminating, its associated PVC will be stuck in terminating too, so we need to account for this.

johscheuer:
Same for Services. Good catch 👍

PR #765, opened 2021-06-18
Don't block if PVC or service is in terminating state

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/764

### DIFF
--- a/controllers/cluster_controller_test.go
+++ b/controllers/cluster_controller_test.go
@@ -607,10 +607,8 @@ var _ = Describe(string(fdbtypes.ProcessClassClusterController), func() {
 				})
 
 				Context("with a PVC stuck in terminating", func() {
-					var originalPod *corev1.Pod
-
 					BeforeEach(func() {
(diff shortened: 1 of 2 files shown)


controllers/cluster_controller_test.go
```diff
@@ -600,11 +600,70 @@ var _ = Describe(string(fdbtypes.ProcessClassClusterController), func() {
...
+						Expect(err).NotTo(HaveOccurred())
+						Expect(adminClient).NotTo(BeNil())
+						Expect(adminClient.ReincludedAddresses).To(HaveLen(0))
 						Expect(adminClient.ExcludedAddresses).To(Equal([]string{
 							originalPods.Items[firstStorageIndex].Status.PodIP,
 						}))
 					})
+
+					It("should remove the Pod", func() {
+						pod := &corev1.Pod{}
```
johscheuer, line 658: Add a comment that this would not be possible in a real Kubernetes env.

ltsampros (approved):

PR #766, opened 2021-06-18
helm chart improvements

EladDolev:
Hello there !

Using your operator for the last year and very happy with it

Offering here a few improvements to the Helm chart
Some of these are opinionated, and I am more than welcome to have those changed

Motivation:
- Fix a few bugs and indentations
- Increase conformance to other community charts
- Have the chart more configurable for common use-cases
- BREAKING CHANGE: adapt deployment's labels to Helm's [best practices](https://helm.sh/docs/chart_best_practices/labels/)

### DIFF
--- a/helm/fdb-operator/values.yaml
+++ b/helm/fdb-operator/values.yaml
@@ -1,12 +1,71 @@
-operator:
-  name: fdb-kubernetes-operator-controller-manager
-  image: foundationdb/fdb-kubernetes-operator
+---
+image:
+  repository: foundationdb/fdb-kubernetes-operator
   tag: v0.36.0
-  role: fdb-kubernetes-operator-manager-role
-  rolebinding: fdb-kubernetes-operator-manager-rolebinding
-  replicas: 1
-  globalMode:
-    enabled: false
+  pullPolicy: IfNotPresent
+
+initContainers:
+  resources:
+    limits:
+      cpu: 10m
+      memory: 50Mi
+    requests:
+      cpu: 10m
+      memory: 50Mi
+
+  init_6_1:
+    image:
+      repository: foundationdb/foundationdb-kubernetes-sidecar
+      tag: 6.1.13-1
+      pullPolicy: IfNotPresent
+  init_6_2:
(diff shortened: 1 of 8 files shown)


helm/fdb-operator/values.yaml
```diff
@@ -1,12 +1,71 @@
...
+      pullPolicy: IfNotPresent
+  init_6_2:
+    image:
+      repository: foundationdb/foundationdb-kubernetes-sidecar
+      tag: 6.2.30-1
+      pullPolicy: IfNotPresent
+  init_6_3:
+    image:
+      repository: foundationdb/foundationdb-kubernetes-sidecar
+      tag: 6.3.10-1
```
EladDolev, line 29: I understand changing these (at least this one that I've tries) are not supported at the moment, but I do hope they will be in the future

johscheuer: It should be supported or what error did you see?

helm/fdb-operator/values.yaml
```diff
@@ -1,12 +1,71 @@
...
+initContainers:
+  resources:
+    limits:
+      cpu: 10m
+      memory: 50Mi
+    requests:
+      cpu: 10m
+      memory: 50Mi
+
+  init_6_1:
```
johscheuer, line 16: Can we move this into a list we iterate over? Otherwise we have to change the helm chart again for `7.0.0` and all other new not minor releases.

johscheuer (changes requested):
Thanks for your PR! Just a few comments from my side.

helm/fdb-operator/values.yaml
EladDolev: I failed to realize how the operator is working
I changed it to `tag: 6.3.13-1` and it broke 6.3 versions
I tried with 6.3.13 and 6.3.10 and got a message saying something like "version 6.3.x is not supported"

after figuring it out I've replaced with iteration

johscheuer (approved):
LGTM 👍 Thanks!

Issue #746, opened 2021-06-07
Cleanup old fdbcli logs


johscheuer:
We have seen that the operator creates a lot of smaller files with the `fdbcli` logs. This is a limitation of the non existing management API and therefore the usage of `fdbcli`. Every time we run a command with `fdbcli` if will create a new trace file that will never be cleaned up. We should ensure that the operator removes files that are older than a given threshold e.g. default ~2h. Otherwise the operator will create a lot of files which are never removed.

```bash
$ kubectl controller-manager-555fc8849f-7tllg -c log-forwarder -- bash
$ ls -lah /var/log/operator-logs | wc -l
1033
$ grep -lRiI 'CLICommandLog'  /var/log/operator-logs | wc -l
1021
```

PR #779, opened 2021-06-22
Remove old logs

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/746

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -121,6 +121,11 @@ func (r *FoundationDBClusterReconciler) Reconcile(ctx context.Context, request c
 	}
 	defer adminClient.Close()
 
+	err = adminClient.CleanupOldLogs()
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
 	supportedVersion, err := adminClient.VersionSupported(cluster.Spec.Version)
 	if err != nil {
 		return ctrl.Result{}, err
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -121,11 +121,6 @@ func (r *FoundationDBClusterReconciler) Reconcile(ctx context.Context, request c
 	}
 	defer adminClient.Close()
 
-	err = adminClient.CleanupOldLogs()
-	if err != nil {
-		return ctrl.Result{}, err
-	}
-
 	supportedVersion, err := adminClient.VersionSupported(cluster.Spec.Version)
 	if err != nil {
 		return ctrl.Result{}, err
--- a/fdbclient/admin_client.go
+++ b/fdbclient/admin_client.go
@@ -27,6 +27,7 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
@@ -199,6 +200,48 @@ func (client *cliAdminClient) runCommand(command cliCommand) (string, error) {
 	}
 	log.Info("Command completed", "namespace", client.Cluster.Namespace, "cluster", client.Cluster.Name, "output", debugOutput)
 	return outputString, nil
+}
+
+// CleanupOldLogs removes old fdbcli log files.
+func (client *cliAdminClient) CleanupOldLogs() error {
+	logDir := os.Getenv("FDB_NETWORK_OPTION_TRACE_ENABLE")
+	if logDir == "" {
+		return nil
+	}
+
+	err := filepath.Walk(logDir, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if info.IsDir() {
+			return nil
+		}
+
+		if !strings.HasPrefix(info.Name(), "trace") {
+			return nil
+		}
+
+		// If the file is newer than 1 hour skip it
+		if info.ModTime().Add(1 * time.Hour).Before(time.Now()) {
+			return nil
+		}
+
+		isCliFile, err := regexp.Compile(`\.1\.\d+`)
+		if err != nil {
+			return err
+		}
+
+		// These files are the one from the fdb lib and will be automatically rotated
+		if isCliFile.MatchString(info.Name()) {
+			return nil
+		}
+
+		log.Info("Delete old log file", "namespace", client.Cluster.Namespace, "cluster", client.Cluster.Name, "file", info.Name())
+		return os.Remove(path)
+	})
+
+	return err
 }
 
 // GetStatus gets the database's status
--- a/fdbclient/admin_client.go
+++ b/fdbclient/admin_client.go
@@ -227,6 +227,9 @@ func (client *cliAdminClient) CleanupOldLogs() error {
 			return nil
 		}
 
+		// Files from the lib will have the format:
+		// trace.$IP.1.&randnumber...json
+		// with this regexp we check for the middle part.
 		isCliFile, err := regexp.Compile(`\.1\.\d+`)
 		if err != nil {
 			return err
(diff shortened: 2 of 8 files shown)


fdbclient/admin_client.go
```diff
@@ -201,6 +202,51 @@ func (client *cliAdminClient) runCommand(command cliCommand) (string, error) {
...
+
+		// If the file is newer than 1 hour skip it
+		if info.ModTime().Add(1 * time.Hour).Before(time.Now()) {
+			return nil
+		}
+
+		// Files from the lib will have the format:
+		// trace.$IP.1.&randnumber...json
+		// with this regexp we check for the middle part.
+		isCliFile, err := regexp.Compile(`\.1\.\d+`)
```
johscheuer, line 233: Another more expensive (but more reliable?) approach would be to check the content of the file e.g. if `CLICommandLog` is contained.

fdbclient/admin_client.go
```diff
@@ -201,6 +202,51 @@ func (client *cliAdminClient) runCommand(command cliCommand) (string, error) {
...
+		if info.IsDir() {
+			return nil
+		}
+
+		if !strings.HasPrefix(info.Name(), "trace") {
+			return nil
+		}
+
+		// If the file is newer than 1 hour skip it
+		if info.ModTime().Add(1 * time.Hour).Before(time.Now()) {
```
johscheuer, line 226: Thinking more about this we could also add a flag for this to make it configurable.

fdbclient/admin_client.go
```diff
@@ -201,6 +202,66 @@ func (client *cliAdminClient) runCommand(command cliCommand) (string, error) {
...
+	// Files from the lib will have the format:
+	// trace.$IP.1.&timestamp...json (or xml)
+	// with this regexp we check for the middle part.
+	isLibFile, err := regexp.Compile(`\.1\.\d{10,}`)
+	if err != nil {
+		return false, err
+	}
+
+	// These files are the one from the fdb lib and will be automatically rotated
+	return isLibFile.MatchString(info.Name()), nil
```
johscheuer, line 229: That should be the other way e.g. the negative one since we want to remove the cli files not the lib files.

fdbclient/admin_client.go
```diff
@@ -201,6 +202,66 @@ func (client *cliAdminClient) runCommand(command cliCommand) (string, error) {
...
+
+	// If the file is newer than 30 minutes skip it.
+	// We assume that the log shipper as pushed the logs away in under 30 minutes.
+	if info.ModTime().Add(30 * time.Minute).Before(now) {
+		return false, nil
+	}
+
+	// Files from the lib will have the format:
+	// trace.$IP.1.&timestamp...json (or xml)
+	// with this regexp we check for the middle part.
```
johscheuer, line 222: TODO: add an example for which files we want to skip and which to keep.

fdbclient/admin_client.go
```diff
@@ -201,6 +202,66 @@ func (client *cliAdminClient) runCommand(command cliCommand) (string, error) {
...
+		remove, err := shouldRemoveLogFile(info, time.Now())
+		if err != nil {
+			return err
+		}
+
+		if !remove {
+			return nil
+		}
+
+		log.Info("Delete old log file", "file", info.Name())
```
johscheuer, line 253: TODO: move to debug and only return a counter of deleted files.

fdbclient/admin_client_test.go
```diff
@@ -99,50 +132,121 @@ var _ = Describe("admin_client_test", func() {
...
+		)
+	})
+
+	When("", func() {
+		type testCase struct {
+			fileInfo os.FileInfo
+			time     time.Time
+			expected bool
+		}
+		DescribeTable("Should return wther the file should be removed",
```
ltsampros, line 190: typo.

fdbclient/admin_client_test.go
```diff
@@ -99,50 +132,121 @@ var _ = Describe("admin_client_test", func() {
...
+					fileInfo: dummyFileInfo{
+						name:    "trace.10.1.14.36.1.1625057172.rmuWOn.0.1.xml",
+						isDir:   true,
+						modTime: time.Now().Add(-45 * time.Minute),
+					},
+					time:     time.Now(),
+					expected: false,
+				},
+			),
+			Entry("When the file is a library file",
```
ltsampros, line 239: fdbcli ?

fdbclient/admin_client_test.go
```diff
@@ -99,50 +132,121 @@ var _ = Describe("admin_client_test", func() {
...
+				testCase{
+					input:       "}",
+					expected:    "",
+					expectedErr: fmt.Errorf("the JSON string doesn't contain a starting '{'"),
+				},
+			),
+		)
+	})
+
+	When("", func() {
```
ltsampros, line 184: missing description

fdbclient/admin_client.go
ltsampros: I agree this should be configurable.

controllers/cluster_controller.go
```diff
@@ -121,6 +121,11 @@ func (r *FoundationDBClusterReconciler) Reconcile(ctx context.Context, request c
 	}
 	defer adminClient.Close()
 
+	err = adminClient.CleanupOldLogs()
```
ltsampros, line 124: I think we should be adding a context.WithTimeout() for this with a configurable timeout, just in case the underlying fs/disk is slow for any reason. That way we can at least ensure we won't block the loop for more than `timeout`.

ltsampros: In the future I also think we should have a flag to control whether we keep or discard older files for the rare case that someone needs to debug using older fdbcli trace files.

ltsampros (commented):
LGTM with a couple of minor comments.

controllers/cluster_controller.go
johscheuer: Added a flag to enable/disable this feature (default enabled).

setup/setup.go
```diff
@@ -164,6 +169,17 @@ func StartManager(
 		}
 	}
 
+	if operatorOpts.CleanUpOldLogFile {
+		setupLog.V(1).Info("setup log file cleaner", "LogFileMinAge", operatorOpts.LogFileMinAge.String())
```
johscheuer, line 173: That's the weird format for debug logs :)

setup/setup.go
```diff
@@ -164,6 +169,17 @@ func StartManager(
...
+		ticker := time.NewTicker(operatorOpts.LogFileMinAge)
+		go func() {
+			for {
+				<-ticker.C
+				internal.CleanupOldCliLogs(operatorOpts.LogFileMinAge)
```
johscheuer, line 178: After our chat I think moving this into a go routine with a ticker should be the best option.

setup/setup.go
```diff
@@ -67,6 +70,8 @@ func (o *Options) BindFlags(fs *flag.FlagSet) {
 	fs.StringVar(&o.LogFile, "log-file", "", "The path to a file to write logs to.")
 	fs.IntVar(&o.CliTimeout, "cli-timeout", 10, "The timeout to use for CLI commands.")
 	fs.IntVar(&o.MaxConcurrentReconciles, "max-concurrent-reconciles", 1, "Defines the maximum number of concurrent reconciles for all controllers.")
+	fs.BoolVar(&o.CleanUpOldLogFile, "cleanup-old-cli-logs", true, "Defines if the operator should delete old fdbcli log files.")
+	fs.DurationVar(&o.LogFileMinAge, "log-file-min-age", 5*time.Minute, "Defines the minimum age od fdbcli log files before removing when \"--cleanup-old-cli-logs\" is set.")
```
ltsampros, line 74: typo: `od`

setup/setup.go
ltsampros: Agreed!

setup/setup.go
johscheuer: Ups :) Thanks 👍

ltsampros (approved):
LGTM!

PR #780, opened 2021-06-22
Fallback to Pod IP and not to the machine ID

johscheuer:
I ran into the following error where `FDB_MACHINE_ID` pointed to the node name and obviously on the node itself no FDB process was running. Instead of using the `getaddrinfo` we should just fallback to the `FDB_POD_IP` variable or the IP in the container (should be the same).

```bash
Traceback (most recent call last):
  File "//sidecar.py", line 662, in <module>
    copy_files()
  File "//sidecar.py", line 556, in copy_files
    config = Config.shared()
  File "//sidecar.py", line 297, in shared
    cls.shared_config = Config()
  File "//sidecar.py", line 228, in __init__
    address_info = socket.getaddrinfo(
  File "/usr/local/lib/python3.9/socket.py", line 953, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
socket.gaierror: [Errno -3] Temporary failure in name resolution
```

### DIFF
--- a/foundationdb-kubernetes-sidecar/sidecar.py
+++ b/foundationdb-kubernetes-sidecar/sidecar.py
@@ -227,7 +227,7 @@ class Config(object):
         if self.substitutions["FDB_PUBLIC_IP"] == "":
             # As long as the public IP is not set fallback to the
             # Pod IP address.
-            pod_ip = os.getenv('FDB_POD_IP')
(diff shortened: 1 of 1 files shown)


09harsh (approved):
LGTM 👍

Issue #758, opened 2021-06-16
Controlling what process classes can serve as coordinators


brownleej:
We have some complex logic to preferentially select storage nodes as coordinators. This was put in place to make it easier to define pod disruption budgets that can specifically target coordinators. Rather than having this implicit preference, I think it would be better to allow people to specify what processes are eligible to be coordinators. This would allow people to restrict coordinators to storage nodes if their architecture sees advantages in it, with stronger guarantees because they can specify that they don't want to fall back to other process classes. It could also allow people to use a special coordinator process class which could allow them to use different storage options for them or otherwise customize the coordinator configuration.

johscheuer:
Should we put a small design doc together for this one? I'm happy to help writing it down.

johscheuer:
Thinking about this, we would also get the opportunity to limit the services created by the operator only to classes that are selected for coordinators.

PR #781, opened 2021-06-25
Allow to define which processes should take part at the coordinator selection

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/758

I'll add some additional test cases and I will open a follow up issue to support different coordinator settings in different DC's (currently everything would be global). The default will be using all stateful processes.

Todos:

- [x] Documentation
- [x] Additional tests changing the setting


api/v1beta1/coordinator_selection.go
```diff
@@ -0,0 +1,56 @@
...
+ * limitations under the License.
+ */
+
+package v1beta1
+
+import "math"
+
+// CoordinatorSelectionSetting defines the process class and the priority of it.
+// A higher priority means that the process class is preferred over another.
+type CoordinatorSelectionSetting struct {
```
brownleej, line 27: Putting this type in its own file means that it doesn't get automatically included in the cluster_spec doc. I think we're better off keeping all of the types for a given CRD in the same file.

api/v1beta1/coordinator_selection.go
```diff
@@ -0,0 +1,56 @@
...
+	ProcessClass ProcessClass `json:"processClass,omitempty"`
+	Priority     int          `json:"priority,omitempty"`
```
brownleej, line 29: Is the priority field necessary? I think the only reason we had a prioritization system before was to infer the correct behavior in the satellites, and now that can be expressed explicitly.

api/v1beta1/coordinator_selection.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+// IsEligibleAsCandidate checks if the given process has the right process class to be considered a valid coordinator
+func (cluster *FoundationDBCluster) IsEligibleAsCandidate(pClass ProcessClass) bool {
+	if len(cluster.Spec.CoordinatorSelection) == 0 {
+		return pClass.IsStateful()
```
brownleej, line 35: The way this check is structured suggests that if someone expressly adds a stateless class to the CoordinatorSelection list, it would allow recruiting that as a coordinator. I think that would open the door to dangerous behavior, since we wouldn't be giving that process a persistent volume.

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -296,6 +296,14 @@ type FoundationDBClusterSpec struct {
...
 	// +kubebuilder:default:=false
 	Skip bool `json:"skip,omitempty"`
+
+	// CoordinatorSelection defines which process classes are eligible for coordinator selection.
+	// If empty all stateful processes classes are equally eligible.
+	// A higher priority means that a process class is preferred over another process class.
+	// If the FoundationDB cluster is sans across multiple Kubernetes clusters or DCs the
+	// CoordinatorSelection must match in all FoundationDB cluster resources otherwise
+	// the coordinator selection process could conflict.
+	CoordinatorSelection []CoordinatorSelectionSetting `json:"coordinatorSelection,omitempty"`
```
brownleej, line 306: The word `sans` doesn't sound right here, but I'm not sure what word you were going for.

api/v1beta1/coordinator_selection.go
```diff
@@ -0,0 +1,56 @@
...
+		if pClass == setting.ProcessClass {
+			return true
+		}
+	}
+
+	return false
+}
+
+// GetClassPriority returns the priority for a class. This will be used to sort the processes for coordinator selection
+func (cluster *FoundationDBCluster) GetClassPriority(pClass ProcessClass) int {
```
brownleej, line 48: This name sounds a bit too generic for something that is only used in coordinator selection.

controllers/cluster_controller.go
```diff
@@ -1331,6 +1320,7 @@ func checkCoordinatorValidity(cluster *fdbtypes.FoundationDBCluster, status *fdb
 	}
 
 	allAddressesValid := true
+	allEligble := true
```
brownleej, line 1323:
```suggestion
	allEligible := true
```

controllers/cluster_controller.go
```diff
@@ -1368,6 +1358,11 @@ func checkCoordinatorValidity(cluster *fdbtypes.FoundationDBCluster, status *fdb
 		if isCoordinator {
 			coordinatorZones[process.Locality[fdbtypes.FDBLocalityZoneIDKey]]++
 			coordinatorDCs[process.Locality[fdbtypes.FDBLocalityDCIDKey]]++
+
+			if !cluster.IsEligibleAsCandidate(process.ProcessClass) {
+				log.Info("Process class of process is not eligible as coordinator", "namespace", cluster.Namespace, "name", cluster.Name, "process", process.Locality[fdbtypes.FDBLocalityInstanceIDKey], "class", process.ProcessClass, "address", address)
+				allEligble = false
```
brownleej, line 1364:
```suggestion
				allEligible = false
```

docs/manual/index.md
```diff
@@ -14,4 +14,5 @@ This document assumes that you are generally familiar with Kubernetes. For more
 1. [Controlling Fault Domains](fault_domains.md)
 1. [Running with TLS](tls.md)
 1. [Backup](backup.md)
+1. [Coordinator Selection](coordinator_selection.md)
```
brownleej, line 17: If we're going to introduce this as a new document, it would be good to have a section at the bottom linking to the next document, and to update the `Backup` document to link to `Coordinator Selection`. On the other hand, it's short enough that it might be fine to fit it into the `Controlling Fault Domains` doc, which already talks about disruption budgets and talks about how coordinators are selected. We'll need to update that doc one way or the other.

brownleej (changes requested):

api/v1beta1/coordinator_selection.go
johscheuer: I thought it make sense to have the ability to say prefer process class X. I think that's useful for multiple processes and you want to be able to prefer a class over another.

api/v1beta1/coordinator_selection.go
johscheuer: That was also a concern but I was not sure if we want to limit the coordinators to only stateful services but I guess that makes sense.

api/v1beta1/foundationdbcluster_types.go
johscheuer: 🤦 that should be `spans`

docs/manual/index.md
johscheuer: I think merging it into `Controlling Fault Domains` make sense.

docs/manual/index.md
```diff
@@ -16,6 +16,4 @@ This document assumes that you are generally familiar with Kubernetes. For more
 1. [Controlling Fault Domains](fault_domains.md)
 1. [Running with TLS](tls.md)
 1. [Backup](backup.md)
-1. [Technical Design](technical_design.md)
-1. [Debugging](debugging.md)
```
brownleej, line 20: Was it intentional to delete these? I think we still want them there.

brownleej (approved):
Small thing about the index in the user manual. Otherwise looks good to me.

docs/manual/index.md
johscheuer: Nope 🤔 Adding this back.

PR #786, opened 2021-06-25
Remove the cache for the database

johscheuer:
So far we've seen more issues with the cache then it solved and I'm not sure if actually helps that much.


brownleej:
I think it would be good to run a local test with this and perform a handful number of operations, and see if it produces a bunch of different connected clients in the database status.

johscheuer:
I'll run some tests tomorrow 👍

johscheuer:
So far my test only showed 3 connected clients (I assume because of the multi-version client since we use 3 versions)

brownleej (approved):
👍

Issue #738, opened 2021-06-02
Reference the kubectl plugin in our docs


johscheuer:
Currently we never reference the kubectl plugin. I think it's useful to provide this information and make it easier to find for everyone.

johscheuer:
Mentioned in https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/docs/manual/debugging.md

PR #787, opened 2021-06-26
Add new sections in the user manual

### LINKED ISSUE #736: Provide a PR template
We should provide a PR template to make it easier to contribute to the project and hopefully we can also use that template to auto generate our changelog. I'll try to put a template together this week.

brownleej:
# Description

This expands the user manual to cover core technical design and debugging. 

Resolves #736

## Type of change

- Documentation Update

# Discussion

No.

# Testing

Manual testing in dev environment?*

N/A

*Is there we need to perform additional testing once this is merged, or perform in a larger testing environment?*

No.

# Documentation

Yes

# Follow-up

No

No.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,6 +1,6 @@
 # Description
 
-*Please include a summary of the change and which radar is addressed. If this change resolves an issue, please include the issue number in the description.*
+*Please include a summary of the change and which issue is addressed. If this change resolves an issue, please include the issue number in the description.*
 
 ## Type of change
 
@@ -9,7 +9,7 @@
 - Bug fix (non-breaking change which fixes an issue)
 - New feature (non-breaking change which adds functionality)
 - Breaking change (fix or feature that would cause existing functionality to not work as expected)
-- Documentation Update
+- Documentation
 - Other
 
 # Discussion
@@ -19,9 +19,9 @@
 # Testing
 
 *Please describe the tests that you ran to verify your changes. Unit tests?
-Manual testing in dev environment?*
+Manual testing?*
 
-*Is there we need to perform additional testing once this is merged, or perform in a larger testing environment?*
+*Do we need to perform additional testing once this is merged, or perform in a larger testing environment?*
 
 # Documentation
 
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,40 @@
+# Description
+
+*Please include a summary of the change and which radar is addressed. If this change resolves an issue, please include the issue number in the description.*
+
+## Type of change
+
+*Please select one of the options below.*
+
+- Bug fix (non-breaking change which fixes an issue)
+- New feature (non-breaking change which adds functionality)
+- Breaking change (fix or feature that would cause existing functionality to not work as expected)
+- Documentation Update
+- Other
+
+# Discussion
+
+*Are there any design details that you would like to discuss further?*
+
+# Testing
+
+*Please describe the tests that you ran to verify your changes. Unit tests?
+Manual testing in dev environment?*
+
+*Is there we need to perform additional testing once this is merged, or perform in a larger testing environment?*
+
+# Documentation
+
+*Did you update relevant documentation within this repository?*
+
+*If this change is adding new functionality, do we need to describe it in our user manual?*
+
+*If this change is adding or removing subreconcilers, have we updated the core technical design doc to reflect that?*
+
+*If this change is adding new safety checks or new potential failure modes, have we documented and how to debug potential issues?*
+
+# Follow-up
+
+*Are there any follow-up issues that we should pursue in the future?*
+
+*Does this introduce new defaults that we should re-evaluate in the future?*
--- a/docs/manual/debugging.md
+++ b/docs/manual/debugging.md
@@ -91,13 +91,13 @@ Example:
 ```yaml
 metadata:
   annotations:
-    touch: touch1
+    foundationdb.org/reconcile: now
 ```
 
 or simply run:
 
 ```bash
-kubectl annotate fdb cluster force-reconcile="$(date)" --overwrite
+kubectl annotate fdb cluster foundationdb.org/reconcile="$(date)" --overwrite
 ```
 
 ## Reconciliation Not Completing
--- a/docs/manual/debugging.md
+++ b/docs/manual/debugging.md
@@ -25,7 +25,7 @@ Error:
 found issues for cluster example-cluster. Please check them
 ```
 
-The plugin can also resolve these issues automatically:
+The plugin can also resolve most of these issues automatically:
 
 ```
 $ kubectl fdb analyze example-cluster --auto-fix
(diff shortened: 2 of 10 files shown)


brownleej:
This may also address #738

docs/manual/debugging.md
```diff
@@ -0,0 +1,133 @@
...
+✖ ProcessGroup: transaction-5 has the following condition: MissingProcesses since 2021-04-27 02:57:24 +0100 BST
+✖ Pod default/example-cluster-storage-2 has unexpected Phase Pending with Reason:
+✖ Pod default/example-cluster-storage-2 has an unready container: foundationdb
+✖ Pod default/example-cluster-storage-2 has an unready container: foundationdb-kubernetes-sidecar
+✖ Pod default/example-cluster-storage-2 has an unready container: trace-log-forwarder
+Error:
+found issues for cluster example-cluster. Please check them
+```
+
+The plugin can also resolve these issues automatically:
```
johscheuer, line 28:
```suggestion
The plugin can also resolve most of these issues automatically:
```

There are still some corner cases that the plugin doesn't catch.

docs/manual/debugging.md
```diff
@@ -0,0 +1,133 @@
...
+✖ ProcessGroup: storage-2 has the following condition: PodFailing since 2021-03-25 18:23:22 +0000 GMT
+✖ ProcessGroup: storage-2 has the following condition: IncorrectConfigMap since 2021-03-26 19:20:56 +0000 GMT
+✖ ProcessGroup: transaction-5 has the following condition: MissingProcesses since 2021-04-27 02:57:24 +0100 BST
+✖ Pod default/example-cluster-storage-2 has unexpected Phase Pending with Reason:
+✖ Pod default/example-cluster-storage-2 has an unready container: foundationdb
+✖ Pod default/example-cluster-storage-2 has an unready container: foundationdb-kubernetes-sidecar
+✖ Pod default/example-cluster-storage-2 has an unready container: trace-log-forwarder
+Replace instances [stateless-5 storage-2 transaction-5] in cluster default/example-cluster [y/n]: y
+Error:
+found issues for cluster example-cluster. Please check them
```
johscheuer, line 47: In the current version the error message shouldn't be printed when we use `auto-fix` to prevent confusion.

docs/manual/debugging.md
```diff
@@ -0,0 +1,133 @@
...
+
+When using this feature, read carefully what the plugin wants to do and only confirm the dialog when you are sure that you want to do these actions.
+
+## Pods stuck in Pending
+
+If you have Pods that are failing to launch, because they are stuck in either a pending or terminating state, you can address that by replacing the failing instance. You can do that using a [plugin command](#replacing-pods-with-the-kubectl-plugin).
+
+## Replacing Pods with the kubectl plugin
+
+Let's assume we are working with the cluster `example-cluster`, and the pod `example-cluster-storage-1` is failing to launch.
```
johscheuer, line 58: The assumption here is that the Pod was at least once up and running and got an IP address otherwise the `failing to launch` will result in a process group without an IP address and the replacement with exclusion will fail.

docs/manual/debugging.md
```diff
@@ -0,0 +1,133 @@
...
+
+```yaml
+$ kubectl fdb remove instances --exclusion=false -c example-cluster example-cluster-storage-1
+Remove [storage-1] from cluster default/example-cluster with exclude: false and shrink: false [y/n]:
+# Confirm with 'y' if the excluded Pod is the correct one
+```
+
+**NOTE**: This is a very dangerous operation.
+This will delete the pod and the PVC without checking that the data has been re-replicated.
+You should only due this after checking that the database is available, that the database has not had any data loss, and that the pod is not currently running. You can confirm the first and second check by looking in the cluster status.
```
johscheuer, line 83:
```suggestion
You should only due this after checking that the database is available, has not had any data loss, and that the pod is currently not running. You can confirm the first and second check by looking at the cluster status.
```

docs/manual/debugging.md
```diff
@@ -0,0 +1,133 @@
...
+## Reconciliation Not Running
+
+If reconciliation is not complete, and there are no recent messages in the operator logs for the cluster, it may be that the reconciliation is backing off due to repeated failures. It should eventually retry the reconciliation. If you want to force it to run reconciliation again immediately, you can edit the cluster metadata. The operator will receive an event about the change and start reconciling. The best no-op change to make is a new annotation.
+
+Example:
+
+```yaml
+metadata:
+  annotations:
+    touch: touch1
```
johscheuer, line 94:
```suggestion
    foundationdb.org/recocile: now
```

docs/manual/debugging.md
```diff
@@ -0,0 +1,133 @@
...
+```
+
+or simply run:
+
+```bash
+kubectl annotate fdb cluster force-reconcile="$(date)" --overwrite
```
johscheuer, line 100:
```suggestion
kubectl annotate fdb cluster foundationdb.org/recocile="$(date)" --overwrite
```

johscheuer: Just as an idea for some standard naming.

docs/manual/technical_design.md
```diff
@@ -0,0 +1,334 @@
...
+              allow: true
+```
+
+Once that is done and the change is reconciled, you can clear the deny list in the spec.
+
+See the [LockOptions](../cluster_spec.md#LockOptions) documentation for more options for customizing the locking system.
+
+## Cluster Reconciliation
+
+The cluster reconciler runs the following subreconcilers:
```
johscheuer, line 85: Should we rather point to the code here?

docs/manual/technical_design.md
```diff
@@ -0,0 +1,334 @@
...
+
+This action requires a lock.
+
+### RemoveServices
+
+The `RemoveServices` subreconciler deletes any services that are no longer required for the cluster.
+
+### RemovePods
+
+The `RemovePods` subreconciler deletes any pods that are marked for removal and have been fully excluded, meaning that they are not serving any roles or holding any data.
```
johscheuer, line 260: I'll open an issue to rename this to `RemoveProcessGroup` (https://github.com/FoundationDB/fdb-kubernetes-operator/issues/790).

johscheuer (changes requested):
Thanks for taking the time writing everything down!

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Type of change
+
+*Please select one of the options below.*
+
+- Bug fix (non-breaking change which fixes an issue)
+- New feature (non-breaking change which adds functionality)
+- Breaking change (fix or feature that would cause existing functionality to not work as expected)
+- Documentation Update
+- Other
```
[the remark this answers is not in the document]
brownleej: Making it a checklist creates noise in the GitHub summary because it interprets the checklist as subtasks that need to be marked off.

docs/manual/technical_design.md
brownleej: I think it's good to have the list of subreconcilers embedded in this doc, to give a sense of the structure for the subsections below.

.github/PULL_REQUEST_TEMPLATE.md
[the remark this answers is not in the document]
johscheuer: That's a valid concern 👍

PR #788, opened 2021-06-26
Enable automatic replacement by default.

### LINKED ISSUE #501: Enabling automatic replacements by default
We recently introduced an option to automatically replace failed processes, and left it disabled by default. Once we've had some experience with this feature in real environments, we should enable it by default.

brownleej:
Resolves #501 

This is marked to only take effect in the future. This prepares us to make us the default as part of the 1.0 release.

### DIFF
--- a/api/v1beta1/foundationdb_version.go
+++ b/api/v1beta1/foundationdb_version.go
@@ -155,3 +155,29 @@ func (version FdbVersion) HasSidecarCrashOnEmpty() bool {
 func (version FdbVersion) HasNonBlockingExcludes() bool {
 	return version.IsAtLeast(FdbVersion{Major: 6, Minor: 3, Patch: 5})
 }
+
+// Versions provides a shorthand for known versions.
+// This is only to be used in testing.
+var Versions = struct {
+	NextMajorVersion, NextPatchVersion,
+	WithSidecarInstanceIDSubstitution, WithoutSidecarInstanceIDSubstitution,
+	WithCommandLineVariablesForSidecar, WithEnvironmentVariablesForSidecar,
+	WithBinariesFromMainContainer, WithoutBinariesFromMainContainer,
+	WithRatekeeperRole, WithoutRatekeeperRole,
+	WithSidecarCrashOnEmpty, WithoutSidecarCrashOnEmpty,
+	Default FdbVersion
+}{
+	Default:                              FdbVersion{Major: 6, Minor: 2, Patch: 20},
+	NextPatchVersion:                     FdbVersion{Major: 6, Minor: 2, Patch: 21},
+	NextMajorVersion:                     FdbVersion{Major: 7, Minor: 0, Patch: 0},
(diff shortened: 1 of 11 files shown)


api/v1beta1/foundationdb_version.go
```diff
@@ -155,3 +155,29 @@ func (version FdbVersion) HasSidecarCrashOnEmpty() bool {
 func (version FdbVersion) HasNonBlockingExcludes() bool {
 	return version.IsAtLeast(FdbVersion{Major: 6, Minor: 3, Patch: 5})
 }
+
+// Versions provides a shorthand for known versions.
+// This is only to be used in testing.
+var Versions = struct {
```
johscheuer, line 161: If this is only intended for testing purpose we should move it into the `internal` package and not expose it to the outside.

internal/deprecations_test.go
```diff
@@ -52,4 +57,777 @@ var _ = Describe("[internal] deprecations", func() {
...
+					err = NormalizeClusterSpec(spec, DeprecationOptions{UseFutureDefaults: true, OnlyShowChanges: true})
+					Expect(err).NotTo(HaveOccurred())
+				})
+
+				It("should be equal to the version with the old explicit defaults", func() {
+					Expect(reflect.DeepEqual(originalSpec, spec)).To(BeTrue())
+				})
+			})
+		})
+	})
```
johscheuer, line 832: I assume these tests are just moved?

internal/deprecations_test.go
```diff
@@ -52,4 +57,777 @@ var _ = Describe("[internal] deprecations", func() {
...
+					originalSpec = spec.DeepCopy()
+				})
+
+				JustBeforeEach(func() {
+					err = NormalizeClusterSpec(spec, DeprecationOptions{UseFutureDefaults: true, OnlyShowChanges: true})
+					Expect(err).NotTo(HaveOccurred())
+				})
+
+				It("should be equal to the version with the old explicit defaults", func() {
+					Expect(reflect.DeepEqual(originalSpec, spec)).To(BeTrue())
```
johscheuer, line 828: I believe a `semantic.DeepEqual` is the right choice here (like in https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/controllers/update_status.go#L312)

kubectl-fdb/cmd/deprecation_test.go
```diff
@@ -43,6 +43,8 @@ import (
 )
 
 var _ = Describe("[plugin] deprecation command", func() {
+	var boolValue = true
```
johscheuer, line 46: This is always true? Can we rename the variable to `trueVal`?

johscheuer (changes requested):

api/v1beta1/foundationdb_version.go
brownleej: I tried that but couldn't get the imports to work, because it created a circular import.

internal/deprecations_test.go
brownleej: Yes.

PR #792, opened 2021-06-28
Rename sub reconciler RemovePods

### LINKED ISSUE #790: Rename RemovePods to RemoveProcessGroups
We have a sub reconciler called `RemovePods` when we moved towards the process groups we also added more logic to actually take care of the complete process group so we should rename the reconciler to `RemoveProcessGroups`.

DeviousLab:
Aims to resolve #790, there might have been instances I could have missed, let me know if there are! Decided not to rename the remove_pods.go file to the requested name incase it might have broken something else.


controllers/remove_pods.go
```diff
@@ -31,12 +31,12 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-// RemovePods provides a reconciliation step for removing pods as part of a
+// RemoveProcessGroups provides a reconciliation step for removing pods as part of a
```
johscheuer, line 34:
```suggestion
// RemoveProcessGroups provides a reconciliation step for removing process groups as part of a
```

johscheuer (changes requested):
Could you please rename the file.

DeviousLab:
I renamed the file now!

johscheuer (changes requested):
There are still some things that should be renamed e.g. the `removePod` method and here: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/792/files#diff-500f2eb6fa10c52fbd7405c9346261cc11738d485e9b04b0d50a6523f18f3859L345-L347 also please ensure that all comments/descriptions are updated.

There are still some issues in the code please run at least `make fmt lint test` to ensure the all lint issues are solved and the tests are passing.

DeviousLab:
I don't think there should be any more errors now, let me know if there are.

johscheuer:
Could you resolve the conflict? Then this should be good to go.

DeviousLab:
@johscheuer Conflicts should be sorted out now!

johscheuer:
Seems like your rebase got wrong all current changes are reverted?

johscheuer (changes requested):
Your PR still shows 94 files changed + all recent changes from other PRs are reverted could you please fix that?

DeviousLab:
Should look good now? Sorry for the prolonged issues that keep arising.

scripts/setup_kind_local_registry.sh
```diff
@@ -29,3 +29,16 @@ EOF
 # connect the registry to the cluster network
 # (the network may already be connected)
 docker network connect "kind" "${reg_name}" || true
+
```
johscheuer, line 32: This change can be dropped

johscheuer (changes requested):
Only one comment, the rest looks good.

scripts/setup_kind_local_registry.sh
DeviousLab: Dropped the change to the setup script

mock-kubernetes-client/client/client_test.go
```diff
@@ -99,12 +99,12 @@ var _ = Describe("[mock client]", func() {
 			Expect(len(pod.Spec.Containers)).To(Equal(1))
 			Expect(pod.Spec.Containers[0].Name).To(Equal("test-container"))
 			Expect(pod.ObjectMeta.Generation).To(Equal(int64(1)))
-			Expect(pod.Status.PodIP).To(Equal("1.1.1.1"))
+			Expect(pod.Status.PodIP).To(Equal("1.1.0.1"))
```
johscheuer, line 102: The tests are failing because of this change.

johscheuer (changes requested):

mock-kubernetes-client/client/client_test.go
DeviousLab: Changed the IP

mock-kubernetes-client/client/client.go
```diff
@@ -745,8 +745,8 @@ func (client *MockClient) SetPodIntoFailed(context ctx.Context, object ctrlClien
 	return client.Update(context, pod)
 }
 
-// RemovePodIP sets the IP address of the Pod to an empty string
-func (client *MockClient) RemovePodIP(pod *corev1.Pod) error {
+// RemoveProcessGroupIP sets the IP address of the Pod to an empty string
+func (client *MockClient) RemoveProcessGroupIP(pod *corev1.Pod) error {
```
johscheuer, line 749: This should still be `RemovePodIP` since we are acting on the Pod level here. The `process group` is only a logical construct to group multiple resources like Pod, PVC, Service to map to the same underlying FoundationDB process.

johscheuer (changes requested):
I missed in my last review that you changed the method in the mock client. we should leave the method there since the method is targeted for Pods and not the logical `process group`.

mock-kubernetes-client/client/client.go
DeviousLab: Reverted the change

johscheuer (approved):
Thanks 👍

PR #793, opened 2021-06-28
Make the health check in the database more narrowly-targeted.

### LINKED ISSUE #244: Allowing database configuration changes when a cluster is unhealthy because of connection string issues
If a cluster has a client that cannot write its cluster file, it reports itself as unhealthy. This blocks configuration changes through the operator, even though there's no relationship between the configuration change and the problem with the cluster health. I think we should look for a more nuanced way of checking whether a cluster is healthy enough to accept configuration changes.

brownleej:
# Description

This uses the data health instead of the overall cluster health to gate database configuration changes.

Resolves #244 

## Type of change

# Discussion

I made this new check continue reconciliation instead of requeuing it. This matches the behavior of the check on availability, and I think this is a pattern we want to follow more in the future.

# Testing

Manual testing?*

I ran manual tests of changing database configuration, including creating an HA cluster.

No.

# Documentation

I've added a description of this check to the user manual.

# Follow-up

I think we should look into other uses of the database health flag and see if we want to replace them with more narrowly-defined checks.


johscheuer (approved):
LGTM 👍

Issue #668, opened 2021-05-07
Remove the readinessProbe for the sidecar


johscheuer:
In https://github.com/FoundationDB/fdb-kubernetes-operator/pull/666 we added support for a livenessProbe in the sidecar. before cutting the 1.0.0 release we want to remove the readinessProbe for the sidecar since we don't get any benefits of it (we always call directly the sidecar).

brownleej:
I think the smoothest way to handle this transition is:
1. Introduce a new setting to enable the readiness probe, defaulting to true
2. In the next major release, default that to false and deprecate it
3. In the major release after that, remove the setting

This means we would need to go through two major releases to fully remove the code, but it means that there is no risk of a change to people's pod specs coming in a minor release.

brownleej:
We've already introduced the new setting, which is all we can do in the 1.0 release.

brownleej:
I think we can leave this as-is, until we have a broader sweep to remove support for the older sidecar model.

Issue #85, opened 2020-01-11
Running clusters with IPv6


brownleej:
Kubernetes has dual-stack IPv4/IPv6 support in alpha now. When running a cluster on that configuration, users will want to control whether their cluster is running on IPv4 or IPv6. We should have an option for this, and a mechanism to transition between IPv4 and IPv6. The transition mechanism can follow the pattern we use for enabling and disabling TLS.

brownleej:
I think we should explore using migration as a transition mechanism instead, to simplify the initial work.

johscheuer:
I'm not sure if migration is there really helpful since all Pods will have an IPv4 and IPv6 address in Dual-Stack we only have to change the [Service config](https://kubernetes.io/docs/concepts/services-networking/dual-stack/#dual-stack-service-configuration-scenarios) if we use Services. Otherwise it's just `IPv4 -> IPv4/IPv6 -> IPv6` if we want to move from IPv4 to IPv6 otherwise we have to execute these step in the reverse order. We also might want to support Dual-Stack clusters?

brownleej:
I'm going to close this with the linked PR, and follow additional issues for more nuanced aspects of IPv6 support.

Issue #352, opened 2020-10-20
Avoid modeling addresses as strings


brownleej:
We have hit multiple bugs due to different parts of the system using different format for addresses. For instance, some parts  use `$ip:$port:$tls`, where others use `$ip:$port`, and others simply use `$ip`. This may suggest a deeper problem of using strings to represent something that deserves a first-class type, with consistent comparison operations and automatic serialization into parts of the system that need a specific form.

johscheuer:
Once https://github.com/FoundationDB/fdb-kubernetes-operator/pull/380 is merged we can revisit the representation. I also did some work in the IPv6 PR https://github.com/FoundationDB/fdb-kubernetes-operator/pull/341 but I think it makes sense to solve this issue first and then add the IPv6 support (should simplify the IPv6 PR and later the Dual-Stack support).

PR #341, opened 2020-10-10
Initial support for IPv6 clusters

johscheuer:
This PR adds the capability to manage IPv6 FDB clusters running in a Kubernetes IPv6 only cluster

With the changes I'm able to bootstrap a FDB cluster with IPv6:

```bash
$ kubectl get nodes -o wide
NAME       STATUS   ROLES    AGE   VERSION   INTERNAL-IP                          EXTERNAL-IP   OS-IMAGE             KERNEL-VERSION   CONTAINER-RUNTIME
master     Ready    master   18h   v1.19.2   fd4a:fc40:8cfb:0:5054:ff:fea2:8107   <none>        Ubuntu 20.04.1 LTS   5.4.0-1024-kvm   containerd://1.3.3-0ubuntu2
worker-0   Ready    <none>   18h   v1.19.2   fd4a:fc40:8cfb:0:5054:ff:fe4e:fac6   <none>        Ubuntu 20.04.1 LTS   5.4.0-1024-kvm   containerd://1.3.3-0ubuntu2
worker-1   Ready    <none>   18h   v1.19.2   fd4a:fc40:8cfb:0:5054:ff:feb6:bec4   <none>        Ubuntu 20.04.1 LTS   5.4.0-1024-kvm   containerd://1.3.3-0ubuntu2
worker-2   Ready    <none>   18h   v1.19.2   fd4a:fc40:8cfb:0:5054:ff:fe4a:ebea   <none>        Ubuntu 20.04.1 LTS   5.4.0-1024-kvm   containerd://1.3.3-0ubuntu2

```

and the coordinators are listening on the IPv6 addresses:

```bash
fdb> coordinators
Cluster description: sample_cluster
Cluster coordinators (3): [fd44:fe56:7891:2f3a:901a:62f2:12d3:88c2]:4500,[fd44:fe56:7891:2f3a:9905:f003:c7bd:fb0d]:4500,[fd44:fe56:7891:2f3a:fd43:e24:a84c:2b37]:4500
Type `help coordinators' to learn how to change this information.
```

the cluster is healthy and can be used:

```bash
 kubectl get fdb
NAME             GENERATION   RECONCILED   HEALTHY
sample-cluster   1                         true
```

There are some minor issues with bouncing the processes that will be fixed before we can merge these changes.

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/85 -> I would create an additional issue fur dual-stack and the transition from an IPv4 cluster to an IPv6 (or dual-stack) cluster.

[...]

PR #794, opened 2021-06-28
Pod IP selection rules for dual-stack environments

brownleej:
# Description

This issue provides an option for using specific addresses from the `podIPs` list in dual-stack environments. This should unblock testing with IPv6 in those environments, but there is more work to be done to fill out the IPv6 solution.

This relates to #85, but we should talk more about how we want to track the feature work for IPv6.

This also includes an option to disable the readiness probe for the sidecar, as discussed in #668. I added this here because the current probes can only go to the main pod IP, as discussed in https://github.com/kubernetes/kubernetes/issues/101324.

## Type of change

# Discussion

I chose to use a pattern string instead of encoding knowledge about the address family. This gives more flexibility but will make it less obvious how to configure IPv6 through this path.

When changing the IP pattern, this does a rolling bounce. That bounce will change coordinator IPs, meaning that the operator has to repeatedly select new coordinators. I considered alternatives like migrating to new pods or configuring multiple listeners, but I think it's not worth the complexity for our first pass at this.

# Testing

Manual testing?*

I created a cluster in a testing environment and switched between patterns to select the IPv4 address and the IPv6 address.

# Documentation

No. I think we want to do more testing before documenting IPv6, but we should track it as part of the bigger project.

# Follow-up

We have more aspects of the IPv6 story that we should fill out:

* Cleaner migration paths
* IPv6 support for services
* IPv6 support for probes


brownleej:
I hit an issue with this where the `black` checker was giving contradictory requirements to `pycodestyle`. I think we're better off restricting ourselves to only one of these checkers.

.github/workflows/pull_request.yml
```diff
@@ -63,7 +63,6 @@ jobs:
       - name: Run Python lint
         run: |
           pycodestyle --max-line-length=120 ./**/*.py
-          black --check --diff ./
```
johscheuer, line 66: I would like to still have both, they target for different things and we just have to configure them properly. `black` ensures a common formatting (e.g. like `go fmt` a nd `pycodestyle` also checks some other code issues.

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -1563,23 +1567,28 @@ type ProcessAddress struct {
 // representation.
 func ParseProcessAddress(address string) (ProcessAddress, error) {
 	result := ProcessAddress{}
-	components := strings.Split(address, ":")
 
-	if len(components) < 2 {
+	ipEnd := strings.Index(address, "]:") + 1
```
johscheuer, line 1571: Instead of building this custom logic I would prefer to use `net.SplitHostPort` that should make everything more robust.

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -1808,10 +1817,13 @@ type DataCenter struct {
 // ContainerOverrides provides options for customizing a container created by
 // the operator.
 type ContainerOverrides struct {
-	// EnableLivenessProbe defines if the sidecar should have a livenessProbe in addition
-	// to the readinessProbe. This setting will be enabled per default in the 1.0.0 release.
+	// EnableLivenessProbe defines if the sidecar should have a livenessProbe.
+	// This setting will be ignored on the main container.
```
johscheuer, line 1821: You removed the `This setting will be enabled per default in the 1.0.0 release.` was that intended?

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -2390,6 +2403,27 @@ type ServiceConfig struct {
...
+	// cluster.
+	HeadlessService *bool `json:"headlessService,omitempty"`
+
+	// PublicIPSource specifies what source a process should use to get its
+	// public IPs.
+	//
+	// This supports the values `pod` and `service`.
+	PublicIPSource *PublicIPSource `json:"publicIPSource,omitempty"`
+
+	// PodIPIndex tells the pod to use a specific entry from the podIPs list
```
johscheuer, line 2419: Where is the `PodIPIndex`?

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -2390,6 +2403,27 @@ type ServiceConfig struct {
 	PublicIPSource *PublicIPSource `json:"publicIPSource,omitempty"`
 }
 
+// RoutingConfig allows configuring routing to our pods, and services that sit
+// in front of them.
+type RoutingConfig struct {
+	// Headless determines whether we want to run a headless service for the
+	// cluster.
+	HeadlessService *bool `json:"headlessService,omitempty"`
```
johscheuer, line 2411: Strictly speaking we never use the headless service for routing.

controllers/cluster_controller.go
```diff
@@ -918,13 +918,51 @@ func (instance FdbInstance) GetPublicIPs() []string {
...
 	if source == "" || source == string(fdbtypes.PublicIPSourcePod) {
-		// TODO for dual-stack support return PodIPs
-		return []string{instance.Pod.Status.PodIP}
+		return getPublicIpsForPod(instance.Pod)
 	}
 
 	return []string{instance.Pod.ObjectMeta.Annotations[fdbtypes.PublicIPAnnotation]}
 }
 
+func getPublicIpsForPod(pod *corev1.Pod) []string {
```
johscheuer, line 927:
```suggestion
func getPublicIPsForPod(pod *corev1.Pod) []string {
```

controllers/cluster_controller.go
```diff
@@ -918,13 +918,51 @@ func (instance FdbInstance) GetPublicIPs() []string {
...
+	var podIPPattern *regexp.Regexp
+
+	if pod == nil {
+		return []string{}
+	}
+
+	for _, container := range pod.Spec.Containers {
+		if container.Name == "foundationdb-kubernetes-sidecar" {
```
johscheuer, line 935: Can we invert this if statement?

controllers/cluster_controller.go
```diff
@@ -918,13 +918,51 @@ func (instance FdbInstance) GetPublicIPs() []string {
...
+			for indexOfArgument, argument := range container.Args {
+				if argument == "--public-ip-pattern" && indexOfArgument < len(container.Args)-1 {
```
johscheuer, line 937: Is there a case where `public-ip-pattern` is set without an argument?

controllers/cluster_controller.go
```diff
@@ -918,13 +918,51 @@ func (instance FdbInstance) GetPublicIPs() []string {
...
+
+	if podIPPattern != nil {
+		podIPs := pod.Status.PodIPs
+		matchingIPs := make([]string, 0, len(podIPs))
+
+		for _, ip := range podIPs {
+			if podIPPattern.Match([]byte(ip.IP)) {
+				matchingIPs = append(matchingIPs, ip.IP)
+			}
+		}
```
johscheuer, line 959: Wouldn't be easier to just use IPv4 or IPv6 instead of that IP pattern matching?

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -2390,6 +2403,27 @@ type ServiceConfig struct {
...
+	//
+	// This supports the values `pod` and `service`.
+	PublicIPSource *PublicIPSource `json:"publicIPSource,omitempty"`
+
+	// PodIPIndex tells the pod to use a specific entry from the podIPs list
+	// instead of the default podIP. The entry be the first one that matches the
+	// pattern that is provided in this field.
+	// This feature is only supported in FDB 7.0 or later, and requires
+	// dual-stack support in your Kubernetes environment.
+	PodIPPattern *string `json:"podIPPattern,omitempty"`
```
johscheuer, line 2424: Any reason to use any IP pattern instead of IPv4 or IPv6?

controllers/cluster_controller_test.go
```diff
@@ -3876,6 +3876,82 @@ var _ = Describe(string(fdbtypes.ProcessClassClusterController), func() {
...
+		Context("with a pod IP pattern configured", func() {
+			BeforeEach(func() {
+				pattern := ":"
+				cluster.Spec.Routing.PodIPPattern = &pattern
+				pod, err := GetPod(cluster, "storage", 1)
+				Expect(err).NotTo(HaveOccurred())
+				pod.Status.PodIP = "1.1.1.1"
+				pod.Status.PodIPs = []corev1.PodIP{
+					{IP: "1.1.1.2"},
+					{IP: "2001:db8::ff00:42:8329"},
```
johscheuer, line 3917: What happens here if we have an IPv4 address in an IPv6 address? Or any other IPv6 formats that might be confused for the wrong address family here.

foundationdb-kubernetes-sidecar/sidecar.py
```diff
@@ -315,9 +354,15 @@ def start(cls):
...
         """
         config = Config.shared()
-        (address, port) = config.bind_address.split(":")
+        port_index = config.bind_address.rindex(":")
+        address = config.bind_address[:port_index]
+        port = config.bind_address[port_index+1:]
         log.info(f"Listening on {address}:{port}")
-        server = ThreadingHTTPServer((address, int(port)), cls)
+
+        if ":" in address:
```
johscheuer, line 362: Please use `ipaddress.ip_address(address).version == 6` instead of checking for `:` the latter is error prone.

foundationdb-kubernetes-sidecar/sidecar.py
```diff
@@ -305,6 +326,24 @@ def is_at_least(self, target_version):
...
+            return string
+
+        ips = string.split(",")
+        matching_ips = [ip for ip in ips if pattern.search(ip)]
+        if len(matching_ips) == 0:
+            raise Exception("Failed to find IP matching %s in %s" % (pattern, ips))
+        ip = matching_ips[0]
+        if ":" in ip:
+            ip = "[%s]" % ip
+        return ip
```
johscheuer, line 341: The returned IP address for IPv6 will be invalid since `[::]` is not an valid IPv6 address

foundationdb-kubernetes-sidecar/sidecar.py
```diff
@@ -305,6 +326,24 @@ def is_at_least(self, target_version):
...
 
+    @classmethod
+    def extract_desired_ip(cls, pattern, string):
+        if string == "":
+            return string
+
+        ips = string.split(",")
+        matching_ips = [ip for ip in ips if pattern.search(ip)]
+        if len(matching_ips) == 0:
+            raise Exception("Failed to find IP matching %s in %s" % (pattern, ips))
```
johscheuer, line 337:
```suggestion
            raise Exception(f"Failed to find IP matching {pattern} in {ips}")
```

foundationdb-kubernetes-sidecar/sidecar.py
```diff
@@ -305,6 +326,24 @@ def is_at_least(self, target_version):
...
+        ip = matching_ips[0]
+        if ":" in ip:
```
johscheuer, line 339: Please use `ipaddress.ip_address(ip).version == 6` here

johscheuer (changes requested):
I'm a little bit concerned about the custom string parsing logic, that will be error prone (at least from my past experience) and we should try to make use of the according libraries in go/Python. I'm also not sure if `PodIPPattern` is a good idea to allow any arbitrary regex? I would rather prefer to just have an `IPv4` or `IPv6` option.

api/v1beta1/foundationdbcluster_types.go
brownleej: Yeah, but I felt that "Routing" was closer than "Service" for this group of settings, since in the long term the plan is to use the headless service for discovery. Do you have any other suggestions for what we could call this?

api/v1beta1/foundationdbcluster_types.go
brownleej: How would that work with the custom flags in our addresses like the `:tls` suffix?

api/v1beta1/foundationdbcluster_types.go
brownleej: We don't generally document future defaults, and I think it would be better to document them in a systematic way.

api/v1beta1/foundationdbcluster_types.go
brownleej: This is an artifact of an earlier form of the solution.

api/v1beta1/foundationdbcluster_types.go
brownleej: I felt like `Routing` captured the intent of these fields better than `Service`, but I'm open to other suggestions about how to make the meaning clearer.

foundationdb-kubernetes-sidecar/sidecar.py
brownleej: We're enclosing the addresses in brackets because everywhere we are using it in the monitor conf, it gets paired with a port, so it needs the brackets. We could modify the monitor conf to contain the brackets, but I think that would lead to more challenges with the rollouts.

controllers/cluster_controller_test.go
brownleej: I've updated it to use the built-in IP parsing.  Does that address this concern?

api/v1beta1/foundationdbcluster_types.go
johscheuer: What about `Discovery`? I mean that's also an overloaded term. Otherwise I don't have any better idea :(  Like always:

"There are 2 hard problems in computer science: cache invalidation, naming things, and off-by-1 errors." :)

api/v1beta1/foundationdbcluster_types.go
johscheuer: I got a solution for that in mind (more or less the same FDB currently does). I would work on that in https://github.com/FoundationDB/fdb-kubernetes-operator/issues/352 that should simplify thee logic here. For the PR itself that's fine and we can improve things step by step.

controllers/cluster_controller.go
johscheuer: This is still unchanged? Or is GitHub just trolling me?

controllers/cluster_controller_test.go
johscheuer: Yes, thanks!

foundationdb-kubernetes-sidecar/sidecar.py
```diff
@@ -148,6 +148,13 @@ def __init__(self):
             "--main-container-version",
             help=("The version of the main foundationdb container in the pod"),
         )
+        parser.add_argument(
+            "--public-ip-family",
+            help=(
+                "Tells the sidecar to treat the public IP as a comma-separated "
+                "list, and use the first entry in the specified IP family"
```
johscheuer, line 155: Could you add an example what the sidecar actually expects?

foundationdb-kubernetes-sidecar/sidecar.py
```diff
@@ -305,6 +328,24 @@ def is_at_least(self, target_version):
...
+        if string == "":
+            return string
+
+        ips = string.split(",")
+        matching_ips = [ip for ip in ips if ipaddress.ip_address(ip).version == version]
+        if len(matching_ips) == 0:
+            raise Exception(f"Failed to find IPv{version} entry in {ips}")
+        ip = matching_ips[0]
+        if version == 6:
+            ip = "[%s]" % ip
```
johscheuer, line 342:
```suggestion
            ip = f"[{ip}]"
```

foundationdb-kubernetes-sidecar/sidecar.py
johscheuer: I think for the current state but we might to revisit this. Sadly Python doesn't have https://golang.org/pkg/net/#JoinHostPort (at least to my knowledge) and I remember doing the same fun stuff: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/341/files#diff-361ca47879a27b8387987496c32b845dbea647c753333511aebc6d93a19be6edR505-R521 I think I just moved it to a later state to prevent parsing the address for the web server again.

kubectl-fdb/cmd/deprecation.go
```diff
@@ -193,3 +193,29 @@ func checkDeprecation(cmd *cobra.Command, kubeClient client.Client, inputCluster
 	cmd.Printf("%d cluster(s) without deprecations\n", clusterCounter)
 	return nil
 }
+
+func getMinimalYAML(object interface{}) ([]byte, error) {
```
johscheuer, line 197: That could also be a `FoundationDBClusterSpec`? I assume you used an `interface{}` to use that method later for different structs?

johscheuer (approved):
Only a few comments other than that LGTM for the first steps towards IPv6.

kubectl-fdb/cmd/deprecation.go
brownleej: Yes, that was my thinking.

PR #795, opened 2021-06-28
Refactor reconciliation results to use a custom structure

### LINKED ISSUE #370: Adding events when reconciliation ends early
There are some cases where the operator gets stuck in reconciliation but doesn't record any events explaining why. For instance, this can happen if we are in the AddPods step and are stuck because of pods that are left in terminating. To give us more consistent debugging methods, I think we should record an event whenever reconciliation ends early. It should include either the error message or an additional context message provided by the failing step.

### LINKED ISSUE #385: Make the SubReconciler use ctrl.Result as return argument
This change allows to use specific requeue times e.g. for https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/controllers/bounce_processes.go#L38 instead of trying yo bounce the processes the whole time we just can requeue the request for the amount of time that we need to wait anytime.

brownleej:
# Description

This introduces a new Requeue type to allow subreconcilers to return requeue results with custom messages and requeue times. It also introduces new events that get stored on clusters when reconciliation ends early.

Resolves #370 
Resolves #385

## Type of change

# Discussion

I chose to use a custom structure instead of `ctrl.Result`. This allows us to add custom messages that are separate from errors, and to propagate errors through the same result type. I think this makes the code for returning these results more readable.

# Testing

Manual testing?*

I updated the unit tests to capture scenarios where the requeue result is changing here, and ran operations in my local environment to confirm that I could see the new events.

No.

# Documentation

I've updated the debugging guide to reference the new event and log messages.

# Follow-up

No.


controllers/cluster_controller_test.go
```diff
@@ -1101,10 +1101,9 @@ var _ = Describe(string(fdbtypes.ProcessClassClusterController), func() {
 					generations, err := reloadClusterGenerations(cluster)
 					Expect(err).NotTo(HaveOccurred())
 					Expect(generations).To(Equal(fdbtypes.ClusterGenerationStatus{
-						Reconciled:             originalVersion,
-						NeedsBounce:            originalVersion + 1,
-						NeedsMonitorConfUpdate: originalVersion + 1,
-						HasUnhealthyProcess:    originalVersion + 1,
+						Reconciled:          originalVersion,
+						NeedsBounce:         originalVersion + 1,
+						HasUnhealthyProcess: originalVersion + 1,
```
johscheuer, line 1106: Was the previous state incorrect?

controllers/controllers.go
```diff
@@ -21,14 +21,27 @@
...
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 var log = logf.Log.WithName("controller")
 
 const (
 	clusterFileKey = "cluster-file"
+
+	// podSchedulingDelayDuration determines how long we should delay a requeue
+	// of reconciliation when a pod is not fully scheduled.
```
johscheuer, line 43:
```suggestion
	// of reconciliation when a pod is not ready.
```

Or if ready is not the correct phrase we should drop the fully a Pod is either scheduled or not.

controllers/controllers.go
```diff
@@ -66,3 +79,39 @@ func mergeMap(target map[string]string, desired map[string]string) bool {
...
+	err := requeue.Error
+	if err != nil && k8serrors.IsConflict(err) {
+		err = nil
+		if requeue.Delay == time.Duration(0) {
+			requeue.Delay = time.Minute
+		}
+	}
+
+	logger.Info("Reconciliation terminated early", "subReconciler", fmt.Sprintf("%T", subReconciler), "message", requeue.Message, "requeueAfter", requeue.Delay)
+	recorder.Event(object, corev1.EventTypeNormal, "ReconciliationEndedEarly", requeue.Message)
```
johscheuer, line 110:
```suggestion
	recorder.Event(object, corev1.EventTypeNormal, "ReconciliationTerminatedEarly", requeue.Message)
```

controllers/controllers.go
```diff
@@ -66,3 +79,39 @@ func mergeMap(target map[string]string, desired map[string]string) bool {
...
+
+	if err != nil {
+		logger.Error(err, "Error in reconciliation", "subReconciler", fmt.Sprintf("%T", subReconciler))
```
johscheuer, line 113: With that change we will print the error twice once in the Info statement above and a second time here. Assuming that most requeue structs won't set the message when an error occurs.

controllers/controllers.go
```diff
@@ -66,3 +79,39 @@ func mergeMap(target map[string]string, desired map[string]string) bool {
...
+
+	// Error provides an error that we encountered that forced a requeue.
+	Error error
+
+	// Message provides a log message that explains the reason for the requeue.
+	Message string
+}
+
+// processRequeue interprets a requeue resulit from a subreconciler.
+func processRequeue(requeue *Requeue, subReconciler interface{}, object runtime.Object, recorder record.EventRecorder, logger logr.Logger) (ctrl.Result, error) {
```
johscheuer, line 96: Before we aborted early when the generation was changed win the current implementation we ignore that. I guess that's intended?

johscheuer (changes requested):

controllers/cluster_controller.go
```diff
@@ -744,22 +710,22 @@ func getDynamicConfHash(configMap *corev1.ConfigMap, pClass fdbtypes.ProcessClas
 	return GetJSONHash(data)
 }
 
-func (r *FoundationDBClusterReconciler) getPodClient(cluster *fdbtypes.FoundationDBCluster, instance FdbInstance) (FdbPodClient, error) {
+func (r *FoundationDBClusterReconciler) getPodClient(cluster *fdbtypes.FoundationDBCluster, instance FdbInstance) (FdbPodClient, string) {
```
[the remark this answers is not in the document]
brownleej: This was something I deliberated, and I should have brought this up in the description. I think that the common cases where the we can't get the pod client, it's because of a temporary condition where we need to wait, and I think we want to move away from using errors to model that. However, it also feels a little awkward to propagate strings from a method like this.

controllers/controllers.go
brownleej: I think checking the generation was a legacy of when the operator would post updates to the spec of the cluster itself, and that's something we want to prevent.

controllers/cluster_controller.go
```diff
@@ -579,10 +549,6 @@ func GetMonitorConf(cluster *fdbtypes.FoundationDBCluster, processClass fdbtypes
 
 // GetStartCommand builds the expected start command for an instance.
 func GetStartCommand(cluster *fdbtypes.FoundationDBCluster, instance FdbInstance, podClient FdbPodClient, processNumber int, processCount int) (string, error) {
-	if instance.Pod == nil {
```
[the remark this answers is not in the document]
brownleej: We don't use it anywhere else in this method, so I don't think we need to check for it.

controllers/cluster_controller_test.go
brownleej: I think the previous state was reflecting reconciliation not making it as far as it ideally could in this configuration.

controllers/cluster_controller.go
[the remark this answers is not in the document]
johscheuer: Assuming we don't care about error's that happen here we could change this in a way that `getPodClient` logs the issue (e.g. the same string that we currently put ass message) and then just always return nil or the client. The only problem here would be when a "real" error occurs. Or we return 3 values `(FdbPodClient, string, error)` that would be a compromise and we still would catch "real" errors. If that makes sense.

controllers/cluster_controller_test.go
johscheuer: Oh right 👍

controllers/cluster_controller.go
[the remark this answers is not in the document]
johscheuer: That makes sense, saved a few CPU cycles 👍

johscheuer (approved):
LGTM 👍

PR #796, opened 2021-06-28
Support defining custom match labels for identifying resources owned by the operator

### LINKED ISSUE #290: Using match labels to identify which pods are owned by the operator

brownleej:
# Description

This adds customization options for the match labels the operator uses, as well as options for defining additional labels to add to the resources it creates.

Resolves #290 

## Type of change

# Discussion

I considered adding support for adding new match labels without first defining them on the resources, but I hit a wall with that, due to the complexity of juggling multiple sets of selectors. This is particularly challenging when we're defining services, since the services can only support a single selector. I think trying to make it more magical would introduce new corner cases, so I think we're better off establishing a multi-step process for updating the labels, which matches built-in Kubernetes resources like Deployment.

# Testing

Manual testing?*

I added unit tests, created a cluster with custom labels, and updated it with new labels.

No.

# Documentation

I added documentation on configuring the labels.

# Follow-up

No.


09harsh (approved):
LGTM

PR #798, opened 2021-06-28
Wait to set the HasPendingRemoval flag in the status until we’ve started terminating the resources.

brownleej:
# Description

This adds a new process group condition to track when we've terminated resources, and uses that to track when we want to set the special `HasPendingRemoval` field in the generation status. This means that if we exclude the processes but aren't able to start the termination, we won't set this field, and won't mark the cluster as reconciled.

## Type of change

# Discussion

This may not be necessary, since the space for failures between finishing exclusions and terminating pods is pretty small, but I think termination is the right condition to act off of.

# Testing

Manual testing?*

I did a replacement in my local environment.

No.

# Documentation

I don't think this requires any documentation updates.

# Follow-up

No.

### DIFF
--- a/api/v1beta1/foundationdbcluster_types.go
+++ b/api/v1beta1/foundationdbcluster_types.go
@@ -719,7 +719,7 @@ const (
 	MissingService ProcessGroupConditionType = "MissingService"
 	// MissingProcesses represents a process group that misses a process.
 	MissingProcesses ProcessGroupConditionType = "MissingProcesses"
-	// Terminating represents a process group whose resources are being
+	// ResourcesTerminating represents a process group whose resources are being
 	// terminated.
 	ResourcesTerminating ProcessGroupConditionType = "ResourcesTerminating"
 	// ReadyCondition is currently only used in the metrics.
--- a/api/v1beta1/foundationdbcluster_types.go
+++ b/api/v1beta1/foundationdbcluster_types.go
@@ -719,6 +719,9 @@ const (
 	MissingService ProcessGroupConditionType = "MissingService"
 	// MissingProcesses represents a process group that misses a process.
 	MissingProcesses ProcessGroupConditionType = "MissingProcesses"
+	// Terminating represents a process group whose resources are being
+	// terminated.
+	ResourcesTerminating ProcessGroupConditionType = "ResourcesTerminating"
 	// ReadyCondition is currently only used in the metrics.
 	ReadyCondition ProcessGroupConditionType = "Ready"
 )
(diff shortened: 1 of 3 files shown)


api/v1beta1/foundationdbcluster_types.go
```diff
@@ -1361,11 +1364,17 @@ func (cluster *FoundationDBCluster) CheckReconciliation() (bool, error) {
 	cluster.Status.Generations = ClusterGenerationStatus{Reconciled: cluster.Status.Generations.Reconciled}
 
 	for _, processGroup := range cluster.Status.ProcessGroups {
-		if processGroup.Remove && !processGroup.Excluded {
-			cluster.Status.Generations.NeedsShrink = cluster.ObjectMeta.Generation
-			reconciled = false
-		} else if processGroup.Remove {
-			cluster.Status.Generations.HasPendingRemoval = cluster.ObjectMeta.Generation
+		if processGroup.Remove {
```
johscheuer, line 1367: NIT: could we invert that check and just continue if the process group is not marked for deletion?

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -1361,11 +1364,17 @@ func (cluster *FoundationDBCluster) CheckReconciliation() (bool, error) {
...
+			terminating := false
+			for _, condition := range processGroup.ProcessGroupConditions {
+				terminating = terminating || condition.ProcessGroupConditionType == ResourcesTerminating
+			}
```
johscheuer, line 1371:
```suggestion
			if processGroup.GetConditionTime(ResourcesTerminating) != nil {
				terminating = true
			}
```

johscheuer: With that change we could also merge this if statement with the statement below.

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -1361,11 +1364,17 @@ func (cluster *FoundationDBCluster) CheckReconciliation() (bool, error) {
...
+			if processGroup.Excluded && terminating {
```
johscheuer, line 1372:
```suggestion
			if processGroup.Excluded || terminating {
```

That should rather be an or. We could have an instance that is in the terminating state but not excluded (e.g. removal without exclusion). I'm also not sure if we really care about the excluded state, I would argue that we only care about the terminating state because if the process group is in a terminating state that means that the process group was either successfully excluded or we don't want to exclude it.

johscheuer: And in the current implementation we never set `processGroup.Excluded` to true and update the field. I created an issue for that.

controllers/update_status.go
```diff
@@ -539,6 +546,13 @@ func validateInstances(r *FoundationDBClusterReconciler, context ctx.Context, cl
 		}
 	}
 
+	for _, processGroupStatus := range processGroups {
```
johscheuer, line 549: This case happens when the Pod of a process group is removed but not all other resources like the PVC or service right? Could we add a short comment above that loop?

johscheuer (changes requested):

api/v1beta1/foundationdbcluster_types.go
brownleej: If a process group is excluded but not in a terminating state, we don't want to set `HasPendingRemoval` yet.

brownleej: As it stands today, we could probably just rely on `terminating` here. I'm not sure what behavior we want if a process group is marked for removal, but isn't excluded yet, and the pod gets deleted for some other reason.

As an alternative here, we could bring in the `ExclusionSkipped` flag.

johscheuer: I think for now just using `terminating` should be good. We could open a follow up issue and once we populate the exclude flag again (https://github.com/FoundationDB/fdb-kubernetes-operator/pull/800) we can change that to:

`(processGroup.Excluded || processGroup. ExclusionSkipped) && terminating `

johscheuer (approved):
LGTM 👍 With the follow up issue we are good to go depending on our timeline we might be able to get this in before our next relese.

Issue #799, opened 2021-06-29
Set exclude field for process groups


johscheuer:
With the merge of our exclusion check and the removal of the process groups we never update the `Excluded` field of process groups we only set it for internal use: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/controllers/remove_pods.go#L303 That can be confusing for resources that have a pending removal e.g. because a resource is stuck in terminating. We should ensure that we update the field.

PR #800, opened 2021-06-29
Don't skip update status after process group removal

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/799 this change will allow to set the excluded field for process groups. There is no need to block the next reconciler if not all process groups are removed.

## Type of change

# Discussion

# Testing

Locally. I'll add later a test case with a Pod stuck in terminating.

# Documentation

Nothing.

# Follow-up

No

### DIFF
--- a/controllers/remove_pods.go
+++ b/controllers/remove_pods.go
@@ -50,15 +50,12 @@ func (u RemovePods) Reconcile(r *FoundationDBClusterReconciler, context ctx.Cont
 		return nil
 	}
 
-	allRemoved, removedProcessGroups := r.removeProcessGroups(context, cluster, processGroupsToRemove)
+	removedProcessGroups := r.removeProcessGroups(context, cluster, processGroupsToRemove)
 	err = includeInstance(r, context, cluster, removedProcessGroups)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


johscheuer:
Result is like expected:

```yaml
    generations:
      hasPendingRemoval: 2
      reconciled: 2
  # ...
    - addresses:
      - 10.1.14.98
      excluded: true
      processClass: storage
      processGroupID: storage-3
      remove: true
    - addresses:
      - 10.1.14.106
      processClass: storage
      processGroupID: storage-4
```

I just added for testing to the `storage-3` Pod the following finalizer:

```yaml
 finalizers:
   - johannes.org/test
```

controllers/remove_pods.go
```diff
@@ -48,13 +48,13 @@ func (u RemovePods) Reconcile(r *FoundationDBClusterReconciler, context ctx.Cont
 		return allExcluded, nil
```
johscheuer, line 48: TODO: we should probably also return true here. Just need to be requeued again.

brownleej (approved):
👍

PR #808, opened 2021-07-02
Add initContainer securityContext to helm chart

funkypenguin:
# Description

Hey guys! This PR simply uses the existing logic in the helm chart to insert a securityContext for each of the initContainers. We stumbled across this when trying to update to the latest chart version - our CI uses kube-score to lint our manifests, and wouldn't permit creating a pod without securityContext for its initContainers :)

## Type of change

# Discussion

# Testing

Successfully validate charts against kube-score, and deployed into multiple clusters without incident.

No

# Documentation

From what I can tell, the helm chart is not documented to the degree of detail which would require any changes resulting from this PR. If I'm wrong, I'm happy to update the docs if you point me to them.

# Follow-up

*Are there any follow-up issues that we should pursue in the future?* No

*Does this introduce new defaults that we should re-evaluate in the future?* No


helm/fdb-operator/values.yaml
```diff
@@ -74,3 +74,11 @@ initContainersResources:
...
     cpu: 10m
     memory: 50Mi
+
+initContainerSecurityContext:
+  allowPrivilegeEscalation: false
+  privileged: false
+  capabilities:
+    drop:
+    - all
+  readOnlyRootFilesystem: true
```
johscheuer, line 84: Missing newline

johscheuer (changes requested):
Only one small thing otherwise LGTM 👍

johscheuer (commented):
Friendly ping :)

funkypenguin:
Oops! Friendly pong :)

johscheuer (approved):
LGTM 👍

PR #809, opened 2021-07-02
Add namespace to clusterrolebinding in global mode (helm chart)

funkypenguin:
# Description

When deploying the helm chart in global mode, a clusterrolebinding is created. This PR fixes the fact that the clusterrolebinding's subject needs to include a namespace, and sets the namespace to the release namespace. (A namespace is not required when deploying a simple roleBinding, when global mode is disabled)

## Type of change

# Discussion

*Are there any design details that you would like to discuss further?* No

# Testing

Tested in multiple clusters, and observed to work (previously the helm install would fail, having generated an invalid clusterrolebinding)

No

# Documentation

*Did you update relevant documentation within this repository?* No, since this PR just fixes something which should already work as documented :)

# Follow-up

*Are there any follow-up issues that we should pursue in the future?* No

*Does this introduce new defaults that we should re-evaluate in the future?* No

### DIFF
--- a/helm/fdb-operator/templates/rbac/rbac_role_binding.yaml
+++ b/helm/fdb-operator/templates/rbac/rbac_role_binding.yaml
@@ -20,3 +20,6 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: {{ include "fdb-operator.serviceAccountName" . }}
+  {{- if .Values.globalMode.enabled }}
+  namespace: {{ .Release.Namespace }}
+  {{- end }}


helm/fdb-operator/templates/rbac/rbac_role_binding.yaml
```diff
@@ -20,3 +20,6 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: {{ include "fdb-operator.serviceAccountName" . }}
+  {{- if .Values.globalMode.enabled }}
+  namespace: {{ .Release.Namespace }}
+  {{- end }}
```
johscheuer, line 25: Missing newline

johscheuer (changes requested):
One small thing otherwise LGTM 👍

johscheuer (commented):
Friendly ping :)

funkypenguin:
Oops. Pushed :)

johscheuer (approved):
LGTM 👍

PR #813, opened 2021-07-03
Make use of net.IP for address

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/352

## Type of change

# Discussion

Once we have support for DNS we have to revisit this (but that was the case anyway :))

# Testing

Locally and a test cluster locally and no other testing is needed.

# Documentation

Nothing to update only internals.

# Follow-up

That should make the handling of the dual-stack support simpler in the operator.


09harsh (approved):
LGTM

Issue #348, opened 2020-10-20
Support three_data_hall redundancy


simenl:
Many cloud providers have Availability Zones within their regions. These zones are independent from each other, such that failures affecting multiple zones are less likely.

We'd like to run FoundationDB with awareness of these zones, so that FoundationDB is available and suffers no data-loss during a full outage in a single availability zone. 
The [three_data_hall](https://apple.github.io/foundationdb/configuration.html#choosing-a-redundancy-mode) redundancy seems like a great fit, using the availability zone as `locality_data_hall` in each process.

simenl:
**Idea**:
Add an option to provide a list of three availability zones (or 'data halls')  in the cluster configuration. And also a corresponding label key. The availability zone is often provided as a label on the kubernetes node, as `topology.kubernetes.io/zone`.
```
availability_zone_key: 'topology.kubernetes.io/zone'
availability_zones: ['europe-west1-b', 'europe-west1-c', 'europe-west1-d']
```

The kubernetes operator would use a nodeSelector to decide which availability zone new pods would be placed on. 
E.g:
```
spec:
  nodeSelector:
    topology.kubernetes.io/zone: europe-west1-b
```

The availability zone would be passed down to the pod, so that the fdb process can set the correct `locality_data_hall`.

The kubernetes operator would need to make sure that the number of pods in each availability zone is balanced for each process type.

brownleej:
Are you thinking of a topology where there is a single Kubernetes cluster that spans an entire region? Would you be using the hostname as the lower-order fault domain within each availability zone?

simenl:
> Are you thinking of a topology where there is a single Kubernetes cluster that spans an entire region? Would you be using the hostname as the lower-order fault domain within each availability zone?

Yes, a single multi-zonal kubernetes cluster spanning the whole region (3 availability zones). 
The hostname, unique per node, would be used as a lower-order fault domain to ensure that logs are not stored on the same node within an availability zone.

rbtcollins:
Hi; I'm a colleague of @simenl 's - Ideally we would have the largest non-AZ failure domain as the lower-order fault domain within the AZ, but few cloud providers expose that (e.g. we believe it to be rack in AKS) - we're working with our providers on getting precise failure domain information though.

brownleej:
[...]

Issue #499, opened 2021-02-17
Automatic replacements if fault-domain is violated


johscheuer:
In the automatic [replacement design doc](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/docs/design/automatic_replacements.md) we initially focused on the replacement of Pod that are not running. I think we should extend this by also allowing the operator to replace Pods that violate our fault-domain. settings e.g. with the `affinityTerms` that we set and in the future the [topologySpreadConstraints](https://github.com/FoundationDB/fdb-kubernetes-operator/issues/361).

brownleej:
Agreed, I think it makes sense to extend automatic replacements to automatic rebalancing across fault domains. I'm not sure if it will be necessary with topology spread constraints, since we may have stricter rules in that model, but we'll have to get deeper into that design to be sure.

johscheuer:
I'm going to close this one. I believe in the current state this doesn't make too much sense, but once we implement the logical fault domain design https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/docs/design/bin_pack_fault_domains.md we have to implement it for this setup.

PR #818, opened 2021-07-05
Add design for three_data_hall and three_datacenter

johscheuer:
# Description

Design for https://github.com/FoundationDB/fdb-kubernetes-operator/issues/348 to support `three_data_hall` and `three_datacenter`

## Type of change

# Discussion

I think I covered all requirements for these redundancy modes.

# Testing

None.

# Documentation

None.

# Follow-up

Once the design is approach wI will create the according issues to work on.

### DIFF
--- a/docs/design/thee_datahall.md
+++ b/docs/design/thee_datahall.md
@@ -1,0 +1,197 @@
+# Three datahall and three datacenter redundancy mode
+
+## Metadata
+
+* Authors: @johscheuer
+* Created: 2020-07-05
+* Updated: 2020-07-05
+
+## Background
+
+Many Cloud Provder offer regional Kubernetes clusters that are span across multiple avalability zones (AZ).
+An avalability zone is an isolated zone in a region and offers users the ability ro create systems that run atop of these failure domains.
+For on-premise clusters this could be a rack, a datahall or a data center.
+The current implementation of the operator only supports either single FoundationDB clusters with the redundancy modes `single`, `dobule`, `triple` or HA clusters span across multiple Kubernetes clusters.
+FoundationDB offers the redundnacy mode [three_data_hall](https://apple.github.io/foundationdb/configuration.html#single-datacenter-modes) or [three_datacenter](https://apple.github.io/foundationdb/configuration.html#datacenter-aware-mode) that fits well in these enviornments without the overhead of the HA solution.
+The current requirement for `three_data_hall` and `three_datacenter` is to have 3 AZs and at least two different failure zones per `data_hall`/`datacenter`.
+
+## General Design Goals
+
+The goal of this design is to support the redundancy modes `three_data_hall` and `three_datacenter` in different deployment models:
+
+* Single Kubernetes cluster across multiple AZs.
+* One Kubernetes cluster per AZ.
+
+The implemenation should be flexible enough to give the user a choice of the used AZ.
+
+## Current Implementation
+
+The current version of the operator doesn't support this redundancy mode.
+The locality is currently a mix of some constants, a configurable zone and two settings at the cluster level [locality configuration](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/controllers/cluster_controller.go#L631-L642).
+
+## Proposed Design
+
+The design is split into the two main parts of the deployment: configure the locality and how to deploy the Pods across the cluster.
+
+### Configure locality
+
+The `three_data_hall` deployment requires the `locality_data_hall` locality to be set and the `three_datacenter` requires the `locality_dcid` to be set.
+There must be at least and at most 3 different values for these localities.
+Additionaly to these localties at least two different `locality_zone` per `data_hall`/`datacenter` is required to fullfill the requirements.
+The `locality_zone` can be configured with [FoundationDBClusterFaultDomain](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/docs/cluster_spec.md#foundationdbclusterfaultdomain).
+In addition to that the user can configure additional variables that the sidecar will use for substition in the [FoundationDBClusterSpec](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/docs/cluster_spec.md#foundationdbclusterspec) with the `sidecarVariables` list.
+The current implementation would allow to set `dataHall: $AZ` and in `sidecarVariables` we would list `AZ` to define the `locality_data_hall` based on an environment variable that will be replaced by the sidecar with the actual value.
+Instead of having these different mechanisims spread across different settings I propose to have a new `localities` setting in the `FoundationDBClusterSpec`:
+
+```yaml
+localities:
+- key: "locality_data_hall"
+  value: ""
+  valueFromEnv: ""
+  valueFromNode: "topology.kubernetes.io/zone"
+  topologyKey: ""
+- key: "locality_zone"
+  value: ""
+  valueFromEnv: "MyFancyZone"
+  valueFromNode: ""
+  topologyKey: "kubernetes.io/hostname"
+```
+
+The `topologyKey` will be used for `topologySpreadConstraints` and for `PodAntiAffinity` and only has to be set if `valueFromNode` is empty, otherwise it will default to that value.
+The `FoundationDBClusterFaultDomain` would then be reduced to only have `key`, `zoneCount` and `zoneIndex` since the locality will be configured in `localities`.
+For all `localities` that define a `valueFromEnv` we would add the key to the `--substitute-variable` flag.
+For `valueFromNode` we have to modify the sidecar to allow it to read labels from Kubernetes nodes (see [Related Links](#related-links)) and pass that information to the according new flag.
+This change should provide the most flexibility to the user to define the required/wanted localities.
+We would set `locality_instance_id` and `locality_machineid` to the current defaults but also allow the user to define custom localities.
+After that change we should deprecate `cluster.Spec.DataCenter` and `cluster.Spec.DataHall`.
+
+### Deployment model
+
+#### Single regional Kubernetes cluster
+
+The assumption here is that the user has a regional Kubernetes cluster that spreads across at least 3 different AZs.
+To ensure that all Pods are spread evenly across the AZs the operator should use [Pod Topology Spread Constraints](https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints).
+
+```yaml
+  topologySpreadConstraints:
+  - maxSkew: 1
+    # the topologyKey will be read from the locality
+    topologyKey: topology.kubernetes.io/zone 
+    whenUnsatisfiable: DoNotSchedule
+    labelSelector:
+      matchLabels:
+        foundationdb.org/fdb-cluster-name: test-cluster
+```
+
+The `topologySpreadConstraints` above ensures that all Pods are evenly spread across the different `topology.kubernetes.io/zone` and must be set by the operator.
+If a regional cluster contains more than 3 AZ and a user only want to use 3 specific AZs the user has to define a `NodeAffinity`:
+
+```yaml
+  affinity:
+    nodeAffinity:
+      requiredDuringSchedulingIgnoredDuringExecution:
+        nodeSelectorTerms:
+        - matchExpressions:
+          - key:  topology.kubernetes.io/zone 
+            operator: In
+            values:
+            - az1
+            - az2
+            - az3
+```
+
+The `NodeAffinity` must be set by the user and the operator doesn't take any action to automatically set the value.
+The user would only require to create one `FoundationDBCluster` and set the `redundancy_mode` to `three_data_hall` or `three_datacenter`.
+
+#### Multiple Kubernetes clusters
+
+In this deployment scenario a user would have 3 different Kubernetes clusters where each cluster spans across a different AZ.
+One requirement is that all Pods in the different AZs are able to communicate.
+The deployment pattern is similar to the [multi Kubernetes deployment](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/docs/manual/fault_domains.md#option-2-multi-kubernetes-replication) but we can't use the special key here since that would mean that the `locality_zone` field would be set to the Kubernetes clusters zone ID and for `three_data_hall` or `three_datacenter` we need at least 2 zones per `data_hall`/`data_center`.
+The initial deployment would be splitt into two phases:
+
+1. Deploy the seed cluster into one of the Kubernetes clusters and wait until it's fully reconciled.
+1. Copy the connection string and use it as `seedConnectionString` for the other two clusters and set `redundancy_mode: "three_data_hall"`.
+
+The initial cluster spec could look like this:
+
+```yaml
+apiVersion: apps.foundationdb.org/v1beta1
+kind: FoundationDBCluster
+metadata:
+  labels:
+    cluster-group: test-cluster
+  name: test-cluster
+spec:
+  version: 6.2.30
+  instanceIDPrefix: az1
+  databaseConfiguration:
+    redundancy_mode: "triple"
+  localities:
+  - key: "locality_data_hall"
+    valueFromNode: "topology.kubernetes.io/zone"
+```
+
+With this configuration we will use the default fault domain the Kubernetes nodes as `locality_zone`.
+When the clutser is reconciled we can create the `FoundationDBCluster` spec in the other two Kubernetes clusters and set the `redundancy_mode: "three_data_hall"`:
+
+```yaml
+apiVersion: apps.foundationdb.org/v1beta1
+kind: FoundationDBCluster
+metadata:
+  labels:
+    cluster-group: test-cluster
+  name: test-cluster
+spec:
+  version: 6.2.30
+  instanceIDPrefix: $az
+  seedConnectionString: $connectionString
+  databaseConfiguration:
+    redundancy_mode: "three_data_hall"
+  localities:
+  - key: "locality_data_hall"
+    valueFromNode: "topology.kubernetes.io/zone"
+```
+
+The processes in the other Kubernetes clusters will join the current `FoundationDBCluster`.
+Once enough processes joined the cluster one of the operator will select 5 Coordinators span across 3 `data_halls`.
+
+### Coordinator selection
+
+The coordinator selecion must be adjusted to select at most 2 coordinators in the same `data_hall` or `datacenter` depending on the redundancy mode.
+Currently the selection happens in the [cluster_controller](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/controllers/cluster_controller.go#L1283-L1292) with a small extension we can support the coordinator selection:
+
+```go
+func getHardLimits(cluster *fdbtypes.FoundationDBCluster) map[string]int {
+    // Multi region cluster (HA)
+    if cluster.Spec.UsableRegions > 1 {
+        return map[string]int{
+            fdbtypes.FDBLocalityZoneIDKey: 1, 
+            fdbtypes.FDBLocalityDCIDKey: int(math.Floor(float64(cluster.DesiredCoordinatorCount()) / 2.0))
+        }
+    }
+
+    // Single region cluster (non-HA)
+    req := map[string]int{fdbtypes.FDBLocalityZoneIDKey: 1}
+
+    if cluster.Spec.RedundancyMode == fdbtypes.RedundancyModeThreeDataHall {
+        req[fdbtypes.FDBLocalityDataHallKey] = 2
+    }
+
+    if cluster.Spec.RedundancyMode == fdbtypes.RedundancyModeThreeDatacenter {
+        req[fdbtypes.FDBLocalityDatacenterKey] = 2
+    }
+
+    return req
+}
+```
+
+## Related Links
+
+Links to other pages that inform or relate to this design.
+
+* [Allow the sidecar to read labels from nodes](https://github.com/FoundationDB/fdb-kubernetes-operator/issues/817)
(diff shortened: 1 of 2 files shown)


docs/design/thee_datahall.md
```diff
@@ -0,0 +1,197 @@
...
+- key: "locality_data_hall"
+  value: ""
+  valueFromEnv: ""
+  valueFromNode: "topology.kubernetes.io/zone"
+  topologyKey: ""
+- key: "locality_zone"
+  value: ""
+  valueFromEnv: "MyFancyZone"
+  valueFromNode: ""
+  topologyKey: "kubernetes.io/hostname"
```
johscheuer, line 57: We could also add a flag here if a `topologySpreadConstraints` should be created for that locality.

docs/design/thee_datahall.md
```diff
@@ -0,0 +1,197 @@
...
+  seedConnectionString: $connectionString
+  databaseConfiguration:
+    redundancy_mode: "three_data_hall"
+  localities:
+  - key: "locality_data_hall"
+    valueFromNode: "topology.kubernetes.io/zone"
+```
+
+The processes in the other Kubernetes clusters will join the current `FoundationDBCluster`.
+Once enough processes joined the cluster one of the operator will select 5 Coordinators span across 3 `data_halls`.
```
johscheuer, line 157: For simplicity we would still create the `topologySpreadConstraints` with the assumption that the `topologySpreadConstraints` works when only one zone is available. Needs to be tested.

docs/design/thee_datahall.md
```diff
@@ -0,0 +1,197 @@
...
+    // Multi region cluster (HA)
+    if cluster.Spec.UsableRegions > 1 {
+        return map[string]int{
+            fdbtypes.FDBLocalityZoneIDKey: 1, 
+            fdbtypes.FDBLocalityDCIDKey: int(math.Floor(float64(cluster.DesiredCoordinatorCount()) / 2.0))
+        }
+    }
+
+    // Single region cluster (non-HA)
+    req := map[string]int{fdbtypes.FDBLocalityZoneIDKey: 1}
```
johscheuer, line 175:
```suggestion
    req := map[string]int{fdbtypes.FDBLocalityZoneIDKey: 1}
    if cluster.Spec.UsableRegions > 1 {
        req[fdbtypes.FDBLocalityDCIDKey] = int(math.Floor(float64(cluster.DesiredCoordinatorCount()) / 2.0))
    }
```

This change allows to use `three_data_hall` with multi region clusters. We only need an additional check if we are using multi region and throw an error if `three_datacenter` is set.

docs/design/thee_datahall.md
```diff
@@ -0,0 +1,197 @@
...
+  - key: "locality_data_hall"
+    valueFromNode: "topology.kubernetes.io/zone"
+```
+
+The processes in the other Kubernetes clusters will join the current `FoundationDBCluster`.
+Once enough processes joined the cluster one of the operator will select 5 Coordinators span across 3 `data_halls`.
+
+### Coordinator selection
+
+The coordinator selecion must be adjusted to select at most 2 coordinators in the same `data_hall` or `datacenter` depending on the redundancy mode.
```
johscheuer, line 161:
```suggestion
The coordinator selection must be adjusted to select at most 2 coordinators in the same `data_hall` or `datacenter` depending on the redundancy mode.
```

johscheuer:
CC @rbtcollins and @simenl feedback is appreciated

docs/design/thee_datahall.md
```diff
@@ -0,0 +1,197 @@
...
+
+```yaml
+  topologySpreadConstraints:
+  - maxSkew: 1
+    # the topologyKey will be read from the locality
+    topologyKey: topology.kubernetes.io/zone 
+    whenUnsatisfiable: DoNotSchedule
+    labelSelector:
+      matchLabels:
+        foundationdb.org/fdb-cluster-name: test-cluster
```
simenl, line 83: I assume that the process classes needs to be evenly spread across zones:

```
      matchLabels:
        foundationdb.org/fdb-cluster-name: test-cluster
        fdb-process-class: <process-class>
```

docs/design/thee_datahall.md
```diff
@@ -0,0 +1,197 @@
...
+This change should provide the most flexibility to the user to define the required/wanted localities.
+We would set `locality_instance_id` and `locality_machineid` to the current defaults but also allow the user to define custom localities.
+After that change we should deprecate `cluster.Spec.DataCenter` and `cluster.Spec.DataHall`.
+
+### Deployment model
+
+#### Single regional Kubernetes cluster
+
+The assumption here is that the user has a regional Kubernetes cluster that spreads across at least 3 different AZs.
+To ensure that all Pods are spread evenly across the AZs the operator should use [Pod Topology Spread Constraints](https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints).
```
simenl, line 73: Does the operator need to ensure that pods remain balanced after removing/excluding pods? 
E.g. after scaling down the cluster.

docs/design/thee_datahall.md
johscheuer: Yes that makes sense, seems like I forgot to add the process class label.

docs/design/thee_datahall.md
johscheuer: I think that would be a useful feature. During a replacement everything should be fine we could only run in an issue when we do a shrink but the [chooseDistributedProcesses](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/controllers/cluster_controller.go#L1198) should take care of that. I will add some additional information in the design and we should add enough test cases to get confidence.

We also have an open issue for automatic balancing: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/499.

brownleej: Do we want to include topology spread constraints here, or use anti-affinities in this context and make topology spread constraints a follow up? Do we have concerns about older versions of Kubernetes where topology spread constraints may not be available?

docs/design/thee_datahall.md
```diff
@@ -0,0 +1,197 @@
...
+  valueFromNode: "topology.kubernetes.io/zone"
+  topologyKey: ""
+- key: "locality_zone"
+  value: ""
+  valueFromEnv: "MyFancyZone"
+  valueFromNode: ""
+  topologyKey: "kubernetes.io/hostname"
+```
+
+The `topologyKey` will be used for `topologySpreadConstraints` and for `PodAntiAffinity` and only has to be set if `valueFromNode` is empty, otherwise it will default to that value.
```
brownleej, line 60: Should we explore using a pre-defined list of values for the data hall instead of spread constraints and picking up data halls dynamically?

docs/design/thee_datahall.md
```diff
@@ -0,0 +1,197 @@
...
+There must be at least and at most 3 different values for these localities.
+Additionaly to these localties at least two different `locality_zone` per `data_hall`/`datacenter` is required to fullfill the requirements.
+The `locality_zone` can be configured with [FoundationDBClusterFaultDomain](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/docs/cluster_spec.md#foundationdbclusterfaultdomain).
+In addition to that the user can configure additional variables that the sidecar will use for substition in the [FoundationDBClusterSpec](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/docs/cluster_spec.md#foundationdbclusterspec) with the `sidecarVariables` list.
+The current implementation would allow to set `dataHall: $AZ` and in `sidecarVariables` we would list `AZ` to define the `locality_data_hall` based on an environment variable that will be replaced by the sidecar with the actual value.
+Instead of having these different mechanisims spread across different settings I propose to have a new `localities` setting in the `FoundationDBClusterSpec`:
+
+```yaml
+localities:
+- key: "locality_data_hall"
```
brownleej, line 48: If all of the keys are going to begin with `locality_`, could we omit it for more readability?

docs/design/thee_datahall.md
```diff
@@ -0,0 +1,197 @@
...
+  value: ""
+  valueFromEnv: ""
+  valueFromNode: "topology.kubernetes.io/zone"
```
brownleej, line 51: What's the behavior if someone sets multiple of these `value` fields on a single locality?

brownleej (changes requested):

docs/design/thee_datahall.md
```diff
@@ -0,0 +1,197 @@
...
+  topologyKey: ""
+- key: "locality_zone"
```
brownleej, line 53: How would we support the behavior of the `foundationdb.org/none` fault domain that we use for testing?

docs/design/thee_datahall.md
```diff
@@ -0,0 +1,197 @@
...
+  value: ""
+  valueFromEnv: "MyFancyZone"
+  valueFromNode: ""
+  topologyKey: "kubernetes.io/hostname"
+```
+
+The `topologyKey` will be used for `topologySpreadConstraints` and for `PodAntiAffinity` and only has to be set if `valueFromNode` is empty, otherwise it will default to that value.
+The `FoundationDBClusterFaultDomain` would then be reduced to only have `key`, `zoneCount` and `zoneIndex` since the locality will be configured in `localities`.
```
brownleej, line 61: How would we be using the `FoundationDBClusterFaultDomain` type in this new model? Is it possible we could deprecate `zoneCount` and `zoneIndex` in favor of the new structures from the design on multi-DC support in the plugin?

docs/design/thee_datahall.md
brownleej: If we're going to move to topology spread constraints, how will we control the rollout of that change?

docs/design/thee_datahall.md
brownleej: Given that the spec already has the AZ hardcoded, could we use `value: $az` instead of setting `valueFromNode`?

docs/design/thee_datahall.md
brownleej: I think we want to use 9 coordinators in three data hall mode, which would mean we would be fine with up to 4 coordinators in a single data hall.

brownleej (changes requested):

docs/design/thee_datahall.md
johscheuer: I think the topology spread makes more sense in this specific use case. Topology spread is supported (stable) since `1.19` and beta in `1.18` (auto enabled) so I think using topology spread and increasing the required version of Kubernetes should be fine? Currently `1.19` is the oldest "supported" version of Kubernetes: https://kubernetes.io/releases/

docs/design/thee_datahall.md
```diff
@@ -0,0 +1,197 @@
...
+Many Cloud Provder offer regional Kubernetes clusters that are span across multiple avalability zones (AZ).
+An avalability zone is an isolated zone in a region and offers users the ability ro create systems that run atop of these failure domains.
+For on-premise clusters this could be a rack, a datahall or a data center.
+The current implementation of the operator only supports either single FoundationDB clusters with the redundancy modes `single`, `dobule`, `triple` or HA clusters span across multiple Kubernetes clusters.
+FoundationDB offers the redundnacy mode [three_data_hall](https://apple.github.io/foundationdb/configuration.html#single-datacenter-modes) or [three_datacenter](https://apple.github.io/foundationdb/configuration.html#datacenter-aware-mode) that fits well in these enviornments without the overhead of the HA solution.
+The current requirement for `three_data_hall` and `three_datacenter` is to have 3 AZs and at least two different failure zones per `data_hall`/`datacenter`.
+
+## General Design Goals
+
+The goal of this design is to support the redundancy modes `three_data_hall` and `three_datacenter` in different deployment models:
```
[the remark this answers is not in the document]
johscheuer: I'm happy to remove it I just thought it makes sense to support it (I'm also not sure about the state) since the requirements are similar to `three_data_hall`. The benefit would be to have 6 replicas for the storage system instead of 3.

docs/design/thee_datahall.md
johscheuer: That sounds like a good first step 👍 I'll change that part once we discover the need for more flexibility we can change that.

docs/design/thee_datahall.md
johscheuer: Sounds good!

docs/design/thee_datahall.md
johscheuer: In a naive way I would have defined a "priority". If multiple values are set we will that the following order:

1. value
2. valueFromEnv
3. valueFromNode

docs/design/thee_datahall.md
johscheuer: I'll take a look at that and comeback 👍

docs/design/thee_datahall.md
johscheuer: The idea would be to deprecated `FoundationDBClusterFaultDomain` I just have to look into the special keys for `foundationdb.org/none` and the multi KC case. The later actually sounds good 👍

docs/design/thee_datahall.md
johscheuer: That's fine for I just oriented at the FoundationDB docs saying 5 coordinators are recommended.

docs/design/thee_datahall.md
johscheuer: Sure, I think both are valid use cases or at least they should produce the same result :)

docs/design/thee_datahall.md
```diff
@@ -0,0 +1,197 @@
...
+    cluster-group: test-cluster
+  name: test-cluster
+spec:
+  version: 6.2.30
+  instanceIDPrefix: az1
+  databaseConfiguration:
+    redundancy_mode: "triple"
+  localities:
+  - key: "locality_data_hall"
+    valueFromNode: "topology.kubernetes.io/zone"
```
johscheuer, line 132: See: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/818/files#r665583201

docs/design/thee_datahall.md
johscheuer: The rollout would require a replacement for stateful Pods (storage, log) and for stateless Pods we could do a recreate. For simplicity we might just want to replace all Pods.

docs/design/three_datahall.md
```diff
@@ -0,0 +1,204 @@
...
+  topologyKey: "kubernetes.io/hostname"
+```
+
+If multiple fields of the `value`, `valueFromEnv` or `valueFromNode` are set the following order will be used:
+
+1. `value`
+1. `valueFromEnv`
+1. `valueFromNode`
+
+The `topologyKey` will be used for `topologySpreadConstraints` and for `PodAntiAffinity` and only has to be set if `valueFromNode` is empty, otherwise it will default to that value.
```
brownleej, line 67: Could we combine the `topologyKey` and `valueFromNode` fields, or do we think there is a use case for them to be different? I see the zone example where we have `valueFromNode` blank, but since `valueFromNode` is the lowest priority option in the different `value...` fields, I think it would have the same behavior whether it was set or not in that case.

docs/design/three_datahall.md
```diff
@@ -0,0 +1,204 @@
...
+The `FoundationDBClusterFaultDomain` will be deprecated, the `zoneCount` and `zoneIndex` will be read from the new [multi-cluster field](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/docs/design/plugin_multi_fdb_support.md#proposed-design).
```
brownleej, line 68: Does this imply that we will have to either implement the plugin's multi-DC support first, or incorporate that work into this project?

docs/design/three_datahall.md
```diff
@@ -0,0 +1,204 @@
...
+For all `localities` that define a `valueFromEnv` we would add the key to the `--substitute-variable` flag.
+For `valueFromNode` we have to modify the sidecar to allow it to read labels from Kubernetes nodes (see [Related Links](#related-links)) and pass that information to the according new flag.
+This change should provide the most flexibility to the user to define the required/wanted localities.
+We would set `locality_instance_id` and `locality_machineid` to the current defaults but also allow the user to define custom localities.
+After that change we should deprecate `cluster.Spec.DataCenter` and `cluster.Spec.DataHall`, those values never had an affect.
```
brownleej, line 73: Maybe this is just picking at wording, but they have an effect in the sense that they are used to populate locality fields today. I think that just means that we will need to translate them into the new locality structure as part of the deprecation.

docs/design/three_datahall.md
```diff
@@ -0,0 +1,204 @@
...
+        - matchExpressions:
+          - key:  topology.kubernetes.io/zone 
+            operator: In
+            values:
+            - az1
+            - az2
+            - az3
+```
+
+The `NodeAffinity` must be set by the user and the operator doesn't take any action to automatically set the value.
```
brownleej, line 111: Would the operator still set the topology spread constraints in this case? Would that conflict with this node affinity because the spread constraint is taking into consideration AZs other than the ones the user wants to use?

Maybe it would be worthwhile to have the locality configuration support using either a topology spread constraint or a combination of node affinity and anti-affinity rules. This may also overlap with the design on the new fault domain management.

docs/design/three_datahall.md
```diff
@@ -0,0 +1,204 @@
...
+The user would only require to create one `FoundationDBCluster` and set the `redundancy_mode` to `three_data_hall` or `three_datacenter`.
```
brownleej, line 112: Is the user required to define locality rules as part of this configuration, or are there reasonable defaults we can apply?

docs/design/three_datahall.md
```diff
@@ -0,0 +1,204 @@
...
+The `locality_zone` can be configured with [FoundationDBClusterFaultDomain](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/docs/cluster_spec.md#foundationdbclusterfaultdomain).
+In addition to that the user can configure additional variables that the sidecar will use for substitution in the [FoundationDBClusterSpec](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/docs/cluster_spec.md#foundationdbclusterspec) with the `sidecarVariables` list.
+The current implementation would allow to set `dataHall: $AZ` and in `sidecarVariables` we would list `AZ` to define the `locality_data_hall` based on an environment variable that will be replaced by the sidecar with the actual value.
+Instead of having these different mechanisms spread across different settings I propose to have a new `localities` setting in the `FoundationDBClusterSpec`.
+The key will always be prefixed with `locality_`:
+
+```yaml
+localities:
+- key: "data_hall"
+  value: ""
```
brownleej, line 50: If this is fully replacing the FaultDomain structure, do we need a special mechanism to mimic the behavior of the `foundationdb.org/none` fault domain strategy, or is that something we can reproduce with `valueFromEnv`?

docs/design/three_datahall.md
```diff
@@ -0,0 +1,204 @@
...
+```yaml
+apiVersion: apps.foundationdb.org/v1beta1
+kind: FoundationDBCluster
+metadata:
+  labels:
+    cluster-group: test-cluster
+  name: test-cluster
+spec:
+  version: 6.2.30
+  instanceIDPrefix: $az
```
brownleej, line 156: I wonder if there's a way we can infer a reasonable value for `instanceIDPrefix` to minimize duplicate configuration and reduce the risk of someone being surprised by problems with conflicting instance IDs.

brownleej (changes requested):

docs/design/three_datahall.md
brownleej: We also have logic to set environment variables for fault domain information based on fields on the pod, which is how we inject the node name today. Is that kind of behavior going to be handled through `valueFromNode` instead?

docs/design/three_datahall.md
```diff
@@ -0,0 +1,204 @@
...
+
+### Deployment model
+
+#### Single regional Kubernetes cluster
+
+The assumption here is that the user has a regional Kubernetes cluster that spreads across at least 3 different AZs.
+To ensure that all Pods are spread evenly across the AZs the operator should use [Pod Topology Spread Constraints](https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints).
+
+```yaml
+  topologySpreadConstraints:
```
johscheuer, line 83: We probably want to remove that and look at this in another design/issue. There are still some questions about how that will behave.

docs/design/three_datahall.md
brownleej: Conclusion from discussion: implement the multi-DC support first.

docs/design/three_datahall.md
brownleej: Conclusion from discussion: combine them.

docs/design/three_datahall.md
brownleej: Conclusion from discussion: emit an event if they don't have valid locality fields for their redundancy mode.

docs/design/three_datahall.md
brownleej: Conclusion from discussion: This makes more sense in the context of multi-DC support for the plugin.

johscheuer: TODO: `instanceIDPrefix` -> `processGroupIDPrefix`

docs/design/three_datahall.md
johscheuer: This would be something we could mimic with `valueFromEnv` and then use use the Pod name. If those fields are "on the Pod" I think that would be a case for `valueFromEnv` the idea of `valueFromNode` is to read labels from the node that "hosts" the specific Pod.

johscheuer:
CC @simenl if you want to have a final look.

brownleej (approved):
👍

Issue #547, opened 2021-03-12
Review exposed methods


johscheuer:
We should revisit all exposed methods of the operator and check if we actually want to expose them.

PR #821, opened 2021-07-06
Refactor cluster controller

johscheuer:
# Description

Refactoring to have most of the methods in a better describing file. Fixes https://github.com/FoundationDB/fdb-kubernetes-operator/issues/547 there are only a few methods that I want to revisit once this PR is merged but this one is huge enough already to merge that stepwise.

## Type of change

- Bug fix (non-breaking change which fixes an issue) / Refactoring

# Discussion

I moved most of the methods into the `internal` package to reduce the methods we expose.

# Testing

Unit tests and locally

# Documentation

None.

# Follow-up

Once this PR is merged I take a second look to ensure we only expose what we want to expose.

### DIFF
--- a/controllers/add_pods.go
+++ b/controllers/add_pods.go
@@ -24,6 +24,8 @@ import (
 	ctx "context"
 	"fmt"
 
+	"github.com/FoundationDB/fdb-kubernetes-operator/internal"
+
 	fdbtypes "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
@@ -35,7 +37,7 @@ type AddPods struct{}
 
 // Reconcile runs the reconciler's work.
 func (a AddPods) Reconcile(r *FoundationDBClusterReconciler, context ctx.Context, cluster *fdbtypes.FoundationDBCluster) *Requeue {
-	configMap, err := GetConfigMap(cluster)
+	configMap, err := internal.GetConfigMap(cluster)
 	if err != nil {
 		return &Requeue{Error: err}
 	}
@@ -51,7 +53,7 @@ func (a AddPods) Reconcile(r *FoundationDBClusterReconciler, context ctx.Context
 		return &Requeue{Error: err}
 	}
 
-	instances, err := r.PodLifecycleManager.GetInstances(r, cluster, context, getPodListOptions(cluster, "", "")...)
+	instances, err := r.PodLifecycleManager.GetInstances(r, cluster, context, internal.GetPodListOptions(cluster, "", "")...)
 	if err != nil {
 		return &Requeue{Error: err}
 	}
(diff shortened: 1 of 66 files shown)


brownleej:
Do we want to look at the exposed methods in the other packages as part of #547?

controllers/add_pods.go
```diff
@@ -35,7 +37,7 @@ type AddPods struct{}
 
 // Reconcile runs the reconciler's work.
 func (a AddPods) Reconcile(r *FoundationDBClusterReconciler, context ctx.Context, cluster *fdbtypes.FoundationDBCluster) *Requeue {
```
brownleej, line 39: Should we consider moving the subreconcilers into the internal package? Having them in the public package means that we wouldn't be able to fully remove one without a major release.

brownleej: If we moved them into the internal package we could also look into creating subpackages for different reconcilers to keep the code more organized.

controllers/cluster_controller.go
```diff
@@ -45,24 +39,19 @@ import (
...
-
 // FoundationDBClusterReconciler reconciles a FoundationDBCluster object
 type FoundationDBClusterReconciler struct {
 	client.Client
 	Recorder            record.EventRecorder
 	Log                 logr.Logger
 	InSimulation        bool
 	PodLifecycleManager PodLifecycleManager
-	PodClientProvider   func(*fdbtypes.FoundationDBCluster, *corev1.Pod) (FdbPodClient, error)
+	PodClientProvider   func(*fdbtypes.FoundationDBCluster, *corev1.Pod) (internal.FdbPodClient, error)
```
brownleej, line 54: If this field refers to an internal type, does it mean that something that imports the `controllers` package would be unable to create a `FoundationDBClusterReconciler`?

internal/process_config.go
```diff
@@ -0,0 +1 @@
+package internal
```
brownleej, line 1: What is this file for?

brownleej (changes requested):

controllers/add_pods.go
johscheuer: Sounds good, I would prefer to open a follow up issue for that and move the subreconcilers in another PR. This PR is already touching too many files :) https://github.com/FoundationDB/fdb-kubernetes-operator/issues/827

internal/process_config.go
johscheuer: Oh, I guess I wanted to move something here and then didn't use it. I'll remove it!

johscheuer (commented):
I just noted that I have to add the Apache header at some places.

controllers/cluster_controller.go
johscheuer: Oh right, I'll check that. Good point 👍

johscheuer:
> Do we want to look at the exposed methods in the other packages as part of #547?

I would look at the other packages in another PR/step to have multiple refactoring PRs instead of one huge.

controllers/cluster_controller.go
johscheuer: I'll created a method for that, which only exposes fields that should be changed. We could also think about exposing a smaller surface before 1.0 but I would solve that in another PR.

09harsh (approved):

PR #823, opened 2021-07-06
Add additional logs for reconcile state and don't reset the reconciled value

johscheuer:
# Description

Add additional logs for the reconcile state.

## Type of change

- 
# Discussion

None

# Testing

Unit tests and locally.

# Documentation

None

# Follow-up


api/v1beta1/foundationdbcluster_types.go
```diff
@@ -1482,8 +1498,6 @@ func (cluster *FoundationDBCluster) CheckReconciliation() (bool, error) {
 
 	if reconciled {
 		cluster.Status.Generations.Reconciled = cluster.ObjectMeta.Generation
-	} else if cluster.Status.Generations.Reconciled == cluster.ObjectMeta.Generation {
-		cluster.Status.Generations.Reconciled = 0
```
johscheuer, line 1486: We might want to discuss that change offline.

johscheuer: I'll add this back and we can discuss this in another issue :)

brownleej (approved):
👍

PR #829, opened 2021-07-08
Correct cluster key for logging

johscheuer:
# Description

Fix some log messages.
## Type of change

# Discussion

No.

# Testing

Unit testing.

# Documentation

No.

# Follow-up

No.

### DIFF
--- a/controllers/pod_client.go
+++ b/controllers/pod_client.go
@@ -312,11 +312,14 @@ func UpdateDynamicFiles(client FdbPodClient, filename string, contents string, u
 		if err != nil {
 			return false, err
 		}
-
 		// We check this more or less instantly, maybe we should add some delay?
 		match, err = client.CheckHash(filename, contents)
 		if !match {
-			log.Info("Waiting for config update", "namespace", client.GetPod().Namespace, "pod", client.GetPod().Name, "file", filename)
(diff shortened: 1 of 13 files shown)


controllers/pod_client.go
```diff
@@ -312,11 +312,14 @@ func UpdateDynamicFiles(client FdbPodClient, filename string, contents string, u
...
 			return false, err
 		}
-
 		// We check this more or less instantly, maybe we should add some delay?
 		match, err = client.CheckHash(filename, contents)
 		if !match {
-			log.Info("Waiting for config update", "namespace", client.GetPod().Namespace, "pod", client.GetPod().Name, "file", filename)
+			log.Info("Waiting for config update",
+				"namespace", client.GetPod().Namespace,
+				"cluster", client.GetPod().Labels[fdbtypes.FDBClusterLabel],
```
johscheuer, line 320: I think that should be fine, the only case where we would hit an issue would be when the Pod doesn't have any label (nil pointer exception). I can also make the cluster optional since for backup and restore Pods we might don't have that label?

johscheuer: Should make use of `GetCluster` instead of `GetPod`.

ltsampros (approved):
LGTM

PR #844, opened 2021-07-11
Remove sort from initial file generation

johscheuer:
# Description

Remove a useless sort of the instances. The problem is that we always sort even if we don't have enough processes for our desired coordinator count and in [chooseDistributedProcesses](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/1dade55f1efa1ee6135cec5846538d7490785c5d/controllers/cluster_controller.go#L1226) we also sort the processes so there is no need to pre-sort them.

## Type of change

# Discussion

Nothing.

# Testing

locally.

# Documentation

None.

# Follow-up

None.

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -777,27 +777,6 @@ func sortPodsByID(pods *corev1.PodList) {
 	})
 }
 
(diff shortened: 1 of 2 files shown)


ltsampros (approved):
LGTM

Issue #716, opened 2021-05-26
Adding a Sidecar unreachable condition


johscheuer:
We should add a condition that indicates that the operator is not able to reach the sidecar of a process group. Once we moved to a system that doesn't need the direct sync we can drop that condition.

PR #851, opened 2021-07-14
Add condition for sidecar unreachable

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/716

## Type of change

# Discussion

# Testing

Locally and e2e.

# Documentation

None

# Follow-up

None.

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -154,6 +154,7 @@ func (r *FoundationDBClusterReconciler) Reconcile(ctx context.Context, request c
 
 	originalGeneration := cluster.ObjectMeta.Generation
 	normalizedSpec := cluster.Spec.DeepCopy()
+	delayedRequeue := false
 
 	for _, subReconciler := range subReconcilers {
 		// We have to set the normalized spec here again otherwise any call to Update() for the status of the cluster
@@ -166,10 +167,18 @@ func (r *FoundationDBClusterReconciler) Reconcile(ctx context.Context, request c
 			continue
 		}
 
+		if requeue.DelayedRequeue {
+			clusterLog.Info("Delaying requeue for sub-reconciler",
+				"subReconciler", fmt.Sprintf("%T", subReconciler),
+				"message", requeue.Message)
+			delayedRequeue = true
+			continue
+		}
+
 		return processRequeue(requeue, subReconciler, cluster, r.Recorder, clusterLog)
 	}
 
-	if cluster.Status.Generations.Reconciled < originalGeneration {
+	if cluster.Status.Generations.Reconciled < originalGeneration || delayedRequeue {
 		clusterLog.Info("Cluster was not fully reconciled by reconciliation process", "status", cluster.Status.Generations)
 
 		return ctrl.Result{Requeue: true}, nil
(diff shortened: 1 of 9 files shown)


controllers/cluster_controller.go
```diff
@@ -166,10 +167,18 @@ func (r *FoundationDBClusterReconciler) Reconcile(ctx context.Context, request c
 			continue
 		}
 
+		if requeue.DelayedRequeue {
```
johscheuer, line 170: I don't make use of that yet, but the plan is to use this for delayed requeue e.g. when we want to iterate further but in the end queue the cluster again.

controllers/controllers.go
```diff
@@ -109,9 +113,10 @@ func processRequeue(requeue *Requeue, subReconciler interface{}, object runtime.
 	recorder.Event(object, corev1.EventTypeNormal, "ReconciliationTerminatedEarly", requeue.Message)
 
 	if err != nil {
-		logger.Error(err, "Error in reconciliation", "subReconciler", fmt.Sprintf("%T", subReconciler), "requeueAfter", requeue.Delay)
+		curLog.Error(err, "Error in reconciliation")
 		return ctrl.Result{}, err
 	}
-	logger.Info("Reconciliation terminated early", "subReconciler", fmt.Sprintf("%T", subReconciler), "message", requeue.Message, "requeueAfter", requeue.Delay)
+	logger.Info("Reconciliation terminated early", "message", requeue.Message)
```
johscheuer, line 119:
```suggestion
	curLog.Info("Reconciliation terminated early", "message", requeue.Message)
```

controllers/update_pod_config.go
```diff
@@ -65,13 +86,22 @@ func (u UpdatePodConfig) Reconcile(r *FoundationDBClusterReconciler, context ctx
 		synced, err := r.updatePodDynamicConf(cluster, instance)
 		if !synced {
 			allSynced = false
-			log.Info("Update dynamic Pod config", "namespace", cluster.Namespace, "cluster", cluster.Name, "processGroupID", instance.GetInstanceID(), "synced", synced, "error", err)
+			hasUpdate = true
+			curLogger.Info("Update dynamic Pod config", "synced", synced, "error", err)
+
+			if err != nil {
+				if _, ok := err.(net.Error); ok {
```
johscheuer, line 93: I think we should iterate over the error and unwrap until we hit the end.

controllers/update_status_test.go
```diff
@@ -380,5 +380,53 @@ var _ = Describe("update_status", func() {
 				Expect(removalCount).To(BeNumerically("==", 1))
 			})
 		})
+
+		When("one instance is not reachable", func() {
+			var unreachableProcessGroup string
+
+			BeforeEach(func() {
+				unreachableProcessGroup = instances[0].GetInstanceID()
+				instances[0].Pod.Annotations[mockUnreachableAnnotation] = "banana"
```
09harsh, line 389: Shouldn't we need to clear this after test completes

johscheuer: We create a complete new cluster object for each test so we don't have to remove anything.

Issue #733, opened 2021-06-01
Keeping pending pods from blocking reconciliation


brownleej:
It can be frustrating when reconciliation is blocked indefinitely on a single pod stuck in pending. While we want to ensure that the cluster has the desired capacity, we also want to limit how much external events can cause reconciliation to be considered incomplete. Maybe we should special case pending pods in the reconciliation, similar to how we special case terminating pods, and set the reconciled generation while also setting another generation to indicate the pending pods problem.

If we're going to generalize that, we might also want to improve the UX and documentation on it. We could introduce a new "weaklyReconciled" generation to capture these kind of states, and add information in our user manual on how users should treat weak reconciliation and full reconciliation.

PR #853, opened 2021-07-14
Keep pending pod from blocking

johscheuer:
# Description

This PR is based on: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/851
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/733

## Type of change

# Discussion

We added a  while back a check to ensure we only exclude when all Pods are up and running: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/controllers/exclude_instances.go#L76-L83 we have to ways to handle this either remove this check completely or define a configurable threshold.

# Testing

Locally, will do some additional tests.

# Documentation

Need to be updated.

# Follow-up

None.


api/v1beta1/foundationdbcluster_types.go
```diff
@@ -743,6 +743,11 @@ const (
 	// ResourcesTerminating represents a process group whose resources are being
 	// terminated.
 	ResourcesTerminating ProcessGroupConditionType = "ResourcesTerminating"
+	// SidecarUnreachable represents a process group where the sidecar is not reachable
+	// because of networking or TLS issues.
+	SidecarUnreachable ProcessGroupConditionType = "SidecarUnreachable"
+	// PodPending represents a process group where the pod is in a pending state.
+	PodPending ProcessGroupConditionType = "PendingPod"
```
johscheuer, line 750:
```suggestion
	PodPending ProcessGroupConditionType = "PodPending"
```

controllers/update_status_test.go
```diff
@@ -428,5 +428,32 @@ var _ = Describe("update_status", func() {
...
+
+		When("a Pod is stuck in Pending", func() {
+			var pendingProcessGroup string
+
+			BeforeEach(func() {
+				pendingProcessGroup = instances[0].GetInstanceID()
+				instances[0].Pod.Status.Phase = corev1.PodPending
+			})
+
+			It("should be mark the process group as Pod pending", func() {
```
johscheuer, line 440: remove the `be`

09harsh (approved):
one add on, rest looks good to me

controllers/update_pods.go
```diff
@@ -35,36 +35,41 @@ type UpdatePods struct{}
...
-	removals := make(map[string]bool)
+	instanceProcessGroupMap := make(map[string]FdbInstance, len(instances))
+	for _, instance := range instances {
+		instanceProcessGroupMap[instance.GetInstanceID()] = instance
+	}
+
 	for _, processGroup := range cluster.Status.ProcessGroups {
 		if processGroup.Remove {
-			removals[processGroup.ProcessGroupID] = true
+			logger.V(1).Info("Ignore removed Pod")
```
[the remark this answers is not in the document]
johscheuer: You mean adding the process group ID?

controllers/cluster_controller.go
```diff
@@ -1307,6 +1307,17 @@ func checkCoordinatorValidity(cluster *fdbtypes.FoundationDBCluster, status *fdb
 			continue
 		}
 
+		processGroupStatus := processGroups[process.Locality["instance_id"]]
+		pendingRemoval := processGroupStatus != nil && processGroupStatus.Remove
+		if processGroupStatus != nil && processGroupStatus.GetConditionTime(fdbtypes.PodPending) != nil {
```
brownleej, line 1312: I wonder if we should factor this into a more general method for detecting if a process group is failing in a way that means we should skip it in checks like this. Maybe this would use the same kind of logic as we use for automatic replacements?

docs/development.md
```diff
@@ -0,0 +1,10 @@
+# Development
```
johscheuer, line 1: That file should be removed and was added by accident.

09harsh (approved):

johscheuer:
I validated that our e2e pipeline passes for these changes.

PR #862, opened 2021-07-21
Allow to only set the IP address without any ports and add more tests

johscheuer:
# Description

This allows to also define only an IP address as process address (we use this for Pod actions like exclusion and inclusion) and adding some additional test cases.

## Type of change

- 
# Discussion

None.

# Testing

Unit tests, locally and e2e.

# Documentation

Nothing.

# Follow-up

Nothing.


ltsampros (approved):
LGTM

PR #863, opened 2021-07-21
Initial design on backup and restore in the operator.

brownleej:
# Description

This adds a design for doing backup and restore in the operator. Much of this work has already been done in the operator, but this proposes significant new features and alterations to the existing structures.

### DIFF
--- a/docs/design/backup_and_restore.md
+++ b/docs/design/backup_and_restore.md
@@ -30,7 +30,7 @@ We will create a FoundationDBBackup resource with the following spec:
     * Backup path (optional, defaults to backup name)
     * Timestamp Format (optional, formatted as a Go time format description, defaults to "-2006-01-02T15:04:05Z-07:00")
 * Backup tag (optional)
-* Key range to restore (optional, defaults to the full keyspace)
+* Key range to back up (optional, defaults to the full keyspace)
 * Snapshot duration (optional, defaults to 7 days)
 * Backup agent template (optional)
 * Parameters for backup URL (optional)
@@ -91,7 +91,6 @@ The backup reconciler will run the following stages:
 * Configure backup agents
 * Toggle backup running
 * Modify backup
-* Toggle backup paused
 * Check backup status
 
 We will use the `fdbbackup` tool to modify the backup state. We will use the backup status that is found within the database status to determine the current state of the backup.
@@ -122,6 +121,8 @@ The version of FoundationDB that the backup agents run will be taken from the cl
 
 In the `Toggle Backup Running` stage, we will start or stop the backup, if necessary. If the desired state is `stopped`, and there is a backup running on the configured tag, this will run the `fdbbackup abort` command. If the desired state is `running` or `paused`, and there is no backup running on the configured tag, this will run the `fdbbackup start` command.
 
+This will also update the pause state in the backup. If the desired state is `paused` and the backup is not paused, this will run the `fdbbackup pause` command. If the desired state is `running` and the backup is paused, this will run the `fdbbackup resume` command.
+
 In order to support restarting backups when configuration changes, the operator will automatically include a timestamp when it starts a backup. The format of the timestamp will be configurable, and configuring it to an empty string will omit the timestamp. When a change is detected that requires restarting a backup, we will stop the backup and start a new one. We will continue to track both backups in the status of the resource until the old backup is fully deleted.
 
 Doing the initial snapshot for a backup will require more bandwidth than subsequent snapshots, because the initial snapshot tries to complete as fast as it can while subsequent snapshots complete more slowly based on the snapshot interval. If we update all backups at once, it would cause every backup to stop and then start a new backup. This creates a gap in restorability until the new backup completes its initial snapshot. If all of the backups are doing this at once, it could lead to rate limiting on the destination that causes all of the backups to slow down, creating a longer aggregate gap in restorability. To prevent this, we will have a cross-backup throttle to limit how many backups can be in the initial snapshot phase at a time. The default limit will be 1. If this limit has already been reached, then the operator will not start or restart any backups, but it will be able to stop a backup if the backup has a desired state of `stopped`. This limit will only apply within a single KC.
@@ -140,10 +141,6 @@ Changing the following fields will require running the modify command:
 
 * Backup URL parameters
 * Snapshot time
-
-#### Toggle Backup Paused
-
-In the `Toggle Backup Paused` stage, we will pause or resume the backup, if necessary. If the desired state is `paused` and the backup is not paused, this will run the `fdbbackup pause` command. If the desired state is `running` and the backup is paused, this will run the `fdbbackup resume` command.
 
 ### Reconciling Restores
 
--- a/docs/design/backup_and_restore.md
+++ b/docs/design/backup_and_restore.md
@@ -81,6 +81,8 @@ We will create a FoundationDBRestore resource with the following spec:
 
 All fields in a restore are immutable.
 
+If the restore version and restore timestamp are both specified, the version will take precedence.
+
 ### Reconciling Backups
 
 The backup reconciler will run the following stages:
@@ -174,6 +176,8 @@ We will have a special case of backup expiration for deleting an entire backup. 
 The backup expiration job will have a maximum execution time of 1 hour. If the operator detects a pod that has been running for longer than this limit, it will delete the pod and start another one for the next backup in the queue.
 
 If a backup resource has multiple low-level backups that have data, we will run the expiry on the last entry in the list. If that entry is restorable and has the minimum restorable data, any older backups will be fully deleted, and removed from the resource status. Otherwise, they will be left intact.
+
+The retry interval when we hit the concurrency limit for expiration pods will be customizable in the start command for the operator.
 
 ### Coordinating Operations
 
--- a/docs/design/backup_and_restore.md
+++ b/docs/design/backup_and_restore.md
@@ -1,0 +1,198 @@
+# Backup and Restore in the Operator
+
+## Metadata
+
+* Authors: @brownleej
+* Created: 2021-07-20
+* Updated: 2021-07-20
+
+## Background
+
+FoundationDB has built-in support for backup and restore through the `fdbbackup` and `fdbrestore` tools. We want to support exposing this functionality through the operator, with custom resources representing continuous backups and one-time restores.
+
+## General Design Goals
+
+* Allow managing backup lifecycle independently of cluster lifecycle
+* Share backup tooling with the open-source community
+* Provide an abstraction that can be used to manage different backup methodologies
+
+## Proprosed Design
+
+### Backup Resource
+
+We will create a FoundationDBBackup resource with the following spec:
+
+* Source cluster name (optional, defaults to backup name, immutable)
+* Desired state (running, stopped, paused) (optional, defaults to running)
+* Backup destination:
+    * Backup account (optional, defaults to first account in the credentials file)
+    * Backup endpoint (optional, defaults to first account in the credentials file)
+    * Backup path (optional, defaults to backup name)
+    * Timestamp Format (optional, formatted as a Go time format description, defaults to "-2006-01-02T15:04:05Z-07:00")
+* Backup tag (optional)
+* Key range to restore (optional, defaults to the full keyspace)
+* Snapshot duration (optional, defaults to 7 days)
+* Backup agent template (optional)
+* Parameters for backup URL (optional)
+* Expiry:
+    * Maximum retention time (optional, defaults to no expiry)
+    * Minimum restorable time (optional, defaults to no restorability protection in expiry)
+    * Enable Deletion Finalizer (optional, default is true)
+
+The resource will have the following status:
+
+* Observed generation
+* Conditions (list)
+    * Condition Type
+    * Timestamp
+* Backups (list)
+    * Tag
+    * Destination
+    * Oldest Data
+        * Version
+        * Timestamp
+    * Latest Restorable Point
+        * Version
+        * Timestamp
+
+In the output for `kubectl get`, we will include the name of the backup, the generation, whether it is fully reconciled, the oldest data timestamp, and the latest restorable timestamp. For this purpose a backup will be considered fully reconciled if it has no conditions in its status.
+
+The possible conditions on the backup will be:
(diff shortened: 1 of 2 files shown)


docs/design/backup_and_restore.md
```diff
@@ -0,0 +1,198 @@
...
.... This will be done through the `fdbbackup expire` command. Because this command blocks, and requires non-trivial resources to run, this will be run in a separate pod from the main controller. The operator will maintain an in-memory queue of backup expiration jobs, with one job for each backup. On start-up, it will populate the queue, and on reconciliation it will ensure that the backup has an entry in the expiration queue.  The operator will have a limit for the concurrency of backup expiration pods, with a default of 1. The operator will have a dedicated goroutine for monitoring this queue and creating pods. If the number of active backup expiration pods is less than the concurrency limit, the queueing goroutine will pop the first job off the queue, create a pod for running that backup expiration, and put a new entry on the end of the queue. If the number of active backup expiration pods is greater than or equal to the concurrency limit, this goroutine will sleep for 10 minutes and try again. It will also check to make sure that there is only one backup expiration running for a given backup at once. Pods that have started termination will not count against this concurrency limit.
```
JoshuaMcManus, line 170: Do we want to make the sleep time configurable for starting new goroutines?

docs/design/backup_and_restore.md
```diff
@@ -0,0 +1,198 @@
...
+* Share backup tooling with the open-source community
+* Provide an abstraction that can be used to manage different backup methodologies
+
+## Proprosed Design
+
+### Backup Resource
+
+We will create a FoundationDBBackup resource with the following spec:
+
+* Source cluster name (optional, defaults to backup name, immutable)
```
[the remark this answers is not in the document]
brownleej: The "backup name" here is the name from the metadata.

docs/design/backup_and_restore.md
```diff
@@ -0,0 +1,198 @@
...
+### Restore Resource
+
+We will create a FoundationDBRestore resource with the following spec:
+
+* Backup URL
+* Destination cluster name
+* Desired state (running, stopped) (optional, defaults to running)
+* Key range to restore (optional, defaults to the full keyspace)
+* Version to restore (optional, defaults to latest version)
+* Timestamp to restore (optional, defaults to latest version)
```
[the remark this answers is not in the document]
brownleej: I don't know of a way to do that in the CRD, so maybe the best we can do is establish which one takes precedence.

docs/design/backup_and_restore.md
brownleej: Yes, it should be configurable.

docs/design/backup_and_restore.md
```diff
@@ -0,0 +1,202 @@
...
.... This will be done through the `fdbbackup expire` command. Because this command blocks, and requires non-trivial resources to run, this will be run in a separate pod from the main controller. The operator will maintain an in-memory queue of backup expiration jobs, with one job for each backup. On start-up, it will populate the queue, and on reconciliation it will ensure that the backup has an entry in the expiration queue.  The operator will have a limit for the concurrency of backup expiration pods, with a default of 1. The operator will have a dedicated goroutine for monitoring this queue and creating pods. If the number of active backup expiration pods is less than the concurrency limit, the queueing goroutine will pop the first job off the queue, create a pod for running that backup expiration, and put a new entry on the end of the queue. If the number of active backup expiration pods is greater than or equal to the concurrency limit, this goroutine will sleep for 10 minutes and try again. It will also check to make sure that there is only one backup expiration running for a given backup at once. Pods that have started termination will not count against this concurrency limit.
```
sears, line 172: Nit:  "to expire backup data older than a certain age" isn't quite right.  Should this read "to expire backup data that is no longer needed to recover to any version that is newer than the expiry date"?

It could make a difference if there are retention rules that force us to delete data that was deleted prior to a certain date, but when doing so would leave us with an incomplete backup.

docs/design/backup_and_restore.md
```diff
@@ -0,0 +1,202 @@
...
+
+We will have a special case of backup expiration for deleting an entire backup. The operator will add finalizers to the backup resource to indicate that when the backup resource is deleted, we must run an `fdbbackup delete` command. There will be an option for turning off this finalizer in the backup spec. When the backup expiration job identifies that a backup needs to be deleted, it will run an `fdbbackup delete` command rather than an `fdbbackup expire` command, and when that completes it will clear the deletion finalizer. At this point the backup resource will get cleaned up.
+
+The backup expiration job will have a maximum execution time of 1 hour. If the operator detects a pod that has been running for longer than this limit, it will delete the pod and start another one for the next backup in the queue.
```
sears, line 176: Does this imply that the job that took over an hour won't ever run (until it's rescheduled the next day, or whatever the expiration frequency is)?  If the job is slow because the backup is too big to quickly delete, then won't the first timeout start a process of digging an ever-deepening hole?

docs/design/backup_and_restore.md
```diff
@@ -0,0 +1,202 @@
...
+
+If any of these checks fail, the backup will be marked as unreconciled, but reconciliation will continue through the later stages whether it is reconciled or not.
+
+#### Configure Backup Agents
+
+In the `Configure Backup Agents` stage, we will create a backup deployment that runs FDB's `backup_agent` process. The agents will be configured through a deployment object in the cluster spec. The deployment will have the following defaults:
+
+* The deployment will have the same name as the backup resource
+* The deployment will have 3 replicas
+* The deployment will have an init container running the `foundationdb/foundationdb-kubernetes-sidecar` image, configured to copy the cluster file from the same config map that the operator creates for the cluster.
```
johscheuer, line 116: The limitation of this is that we can't use this mechanism to backup/restore clusters in a different namespace/cluster since the `ConfigMap` can only be consumed as a volume in the same namespace. I think for the initial steps that's fine but I could think of a setup where the backup agents are running in a different namespace to ensure the backups are not deleted with the clusters if someone deletes the namespace (intended or not).

*edit* I see you mention that below. Are we planning to support that in the future or should we just assume this as a limitation of the design?

docs/design/backup_and_restore.md
```diff
@@ -0,0 +1,202 @@
...
+* The deployment will have a main container running the `foundationdb/foundationdb` image, running a `backup_agent` process.
+
+The version of FoundationDB that the backup agents run will be taken from the cluster resource. There must be a cluster resource in the same namespace with the name given in the cluster spec in order to run a backup. This version will also be used for the `fdbbackup` commands. In order to keep these versions in sync, the cluster reconciler will have a stage to check for backups that need to be updated. This stage will find any backup resources that are connected to the cluster and set an annotation, `foundationdb.org/backup-cluster-version` with the version in the cluster spec. During upgrades, the cluster reconciler will set this annotation, which will trigger a reconciliation of the backup that will update the backup agents.
```
johscheuer, line 119: The last part implies that the cluster controller will change the deployment that is actually managed by the backup reconciler. Do you know if it's possible to watch for other resource events? I would prefer (if that's possible in a simple way) that the backup reconciler watched the cluster resources without the cluster reconciler having to interfere with the backup resources.

I think I spent some time in the past on that topic but couldn't find anything useful but maybe the controller runtime has some new features for that.

docs/design/backup_and_restore.md
```diff
@@ -0,0 +1,202 @@
...
+
+Doing the initial snapshot for a backup will require more bandwidth than subsequent snapshots, because the initial snapshot tries to complete as fast as it can while subsequent snapshots complete more slowly based on the snapshot interval. If we update all backups at once, it would cause every backup to stop and then start a new backup. This creates a gap in restorability until the new backup completes its initial snapshot. If all of the backups are doing this at once, it could lead to rate limiting on the destination that causes all of the backups to slow down, creating a longer aggregate gap in restorability. To prevent this, we will have a cross-backup throttle to limit how many backups can be in the initial snapshot phase at a time. The default limit will be 1. If this limit has already been reached, then the operator will not start or restart any backups, but it will be able to stop a backup if the backup has a desired state of `stopped`. This limit will only apply within a single KC.
+
+Changing the following fields will require restarting a backup:
+
+* Backup destination
+* Tag
+* Key Range
```
johscheuer, line 133: All settings in `Backup destination` will probably also cause a restart?

docs/design/backup_and_restore.md
```diff
@@ -0,0 +1,202 @@
...
+In the `Modify Backup` stage, we will update parameters for the backup. If the backup parameters do not match the configured set, this will run the `fdbbackup modify` command.
+
+Changing the following fields will require running the modify command:
+
+* Backup URL parameters
+* Snapshot time
+
+#### Toggle Backup Paused
+
+In the `Toggle Backup Paused` stage, we will pause or resume the backup, if necessary. If the desired state is `paused` and the backup is not paused, this will run the `fdbbackup pause` command. If the desired state is `running` and the backup is paused, this will run the `fdbbackup resume` command.
```
johscheuer, line 146: We need this step to ensure we did the initial backup once? Otherwise I'm not sure if we really need this step of if we could merge it with `Toggle Backup Running`.

docs/design/backup_and_restore.md
```diff
@@ -0,0 +1,202 @@
...
+
+We will have a special case of backup expiration for deleting an entire backup. The operator will add finalizers to the backup resource to indicate that when the backup resource is deleted, we must run an `fdbbackup delete` command. There will be an option for turning off this finalizer in the backup spec. When the backup expiration job identifies that a backup needs to be deleted, it will run an `fdbbackup delete` command rather than an `fdbbackup expire` command, and when that completes it will clear the deletion finalizer. At this point the backup resource will get cleaned up.
```
johscheuer, line 174: The Pods for deleting backups will also count to the backup expiration pods limit?

docs/design/backup_and_restore.md
johscheuer: The concurrency limit is per Backup or global? I would interpret it like per Backup but I might be wrong.

johscheuer: Since we are managing bare Pods (that's an assumption since we are talking about a `backup expiration pod limit`, please correct me) what happens when the expire is canceled (or actually the Pod is deleted) during that command? Would a `ReplicaSet` or a `Job` make more sense?

brownleej: If the pod is deleted, the operator will start a new expiry job for the next cluster in the queue.

docs/design/backup_and_restore.md
brownleej: My understanding of the backup expire command is that it is somewhere between those two ideas: the semantics of the command are to delete data older than a certain age, but it has safety checks to prevent deleting data that is needed to restore to a certain point. The point in time for that safety check is provided in a different parameter.

docs/design/backup_and_restore.md
brownleej: The assumption here is that a job that takes an hour before getting cancelled will still be able to make some progress. The job doesn't have a regular frequency. After a job is cancelled, it will move onto an expiration job for the next backup in the queue, and once it has cycled through every other backup, it should come back to the backup that had the job cancelled previously.

docs/design/backup_and_restore.md
```diff
@@ -0,0 +1,202 @@
...
+
+We will create a FoundationDBBackup resource with the following spec:
+
+* Source cluster name (optional, defaults to backup name, immutable)
+* Desired state (running, stopped, paused) (optional, defaults to running)
+* Backup destination:
+    * Backup account (optional, defaults to first account in the credentials file)
+    * Backup endpoint (optional, defaults to first account in the credentials file)
+    * Backup path (optional, defaults to backup name)
+    * Timestamp Format (optional, formatted as a Go time format description, defaults to "-2006-01-02T15:04:05Z-07:00")
```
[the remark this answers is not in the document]
brownleej: Yes.

docs/design/backup_and_restore.md
brownleej: I'm happy to accept it as a limitation of the design. If we want to keep backup data as a safety mechanism after a namespace is deleted, we can look into expressing that as a wait time in the finalizer logic.

docs/design/backup_and_restore.md
brownleej: The annotation I'm describing here would be set on the `FoundationDBBackup` resource, rather than the deployment. I don't know if there's a way for the backup reconciler to respond to events on the cluster object.

docs/design/backup_and_restore.md
brownleej: Yes, that's what I was alluding to on line 131.

docs/design/backup_and_restore.md
brownleej: You're right. These two actions could probably be merged into a single action.

docs/design/backup_and_restore.md
brownleej: Yes.

docs/design/backup_and_restore.md
brownleej: There are two different limits at play here:

* Global concurrency limit, which is configurable. This is the number of overall backup expiration jobs we can run at one time. Each job will target a single backup.
* Per-backup concurrency limit, which is hard-coded to 1 for safety reasons.

brownleej: We may need to think more about whether the "global" concurrency limit is scoped to the Kubernetes cluster or to the namespace.

brownleej: That's an interesting idea. Would we have one `BackupExpiry` object for each backup, or a single `BackupExpiryQueue` object? A single queue object sounds like it could be a problem since it would mean a resource of unbounded size, but if we are reconciling each expiry independently then we would have additional challenges around fairness. We could track the lastExpiryStartTime in the status on the BackupExpiry, and then on every reconciliation pull up the full list sorted by that time to determine if it is valid to start a new job for the given BackupExpiry. It could be an odd fit for the reconciliation model, though, because I don't think the BackupExpiry resource would ever reach a reconciled state. Maybe we could consider it "reconciled" if the last run completed successfully and the last expiry start time was in the last X hours, and rely on the controller periodically checking every resource to make sure jobs keep going.

johscheuer (approved):
LGTM 👍

PR #865, opened 2021-07-23
Revert the change for the CRD to use coordinators as strings

johscheuer:
# Description

Reverting the change that coordinators in the connection string (CRD) are handled as `ProcessAddress`.

## Type of change

# Discussion

# Testing

Local, e2e.

# Documentation

None.

# Follow-up

None.


brownleej (approved):
👍

Issue #369, opened 2020-11-05
Replacing instances when changing volume name


brownleej:
When we change the name in the PVC metadata in the cluster spec, it will cause the operator to expect a new volume name for the PVCs. We should handle this by replacing instances that use a non-matching volume name.

johscheuer:
PR was already merged.

PR #867, opened 2021-07-26
Replace pods if PVC name in cluster spec changes

AnnigeriShambu:
# Description

This is to solve [issue 369](https://github.com/FoundationDB/fdb-kubernetes-operator/issues/369), Replacing pods when changing volume name

## Type of change

# Discussion

NA
# Testing
Local testing
Manual testing?*

# Documentation

# Follow-up

### DIFF
--- a/controllers/replace_misconfigured_pods.go
+++ b/controllers/replace_misconfigured_pods.go
@@ -89,6 +89,25 @@ func (c ReplaceMisconfiguredPods) Reconcile(r *FoundationDBClusterReconciler, co
 			return &Requeue{Error: err}
 		}
 
+		if pvc.Name != desiredPVC.Name {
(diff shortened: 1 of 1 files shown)


controllers/replace_misconfigured_pods.go
```diff
@@ -89,6 +89,25 @@ func (c ReplaceMisconfiguredPods) Reconcile(r *FoundationDBClusterReconciler, co
 			return &Requeue{Error: err}
 		}
 
+		if pvc.Name != desiredPVC.Name {
```
johscheuer, line 92: Can you please write test cases for that? I also think it makes sense to move everything in the PVC for loop into a private method (that should make the testing easier, similar to `instanceNeedsRemoval`.

johscheuer (changes requested):

AnnigeriShambu:
Please refer https://github.com/FoundationDB/fdb-kubernetes-operator/pull/868

PR #870, opened 2021-07-29
Prevent double-encoding IPv6 addresses when translating between pod IPs and process addresses.

brownleej:
# Description

This fixes a bug where IP addresses would be enclosed in two sets of brackets when we generated the connection string for IPv6 clusters.

## Type of change

# Discussion

This does fairly primitive parsing, but I don't know if there's a cleaner way of expressing what we're trying to do.

# Testing

Manual testing?*

I added new unit tests for IPv6 clusters, and created a real IPv6 cluster in a test environment.

No.

# Documentation

No.

No.

No.

No.

# Follow-up

No.

No.

### DIFF
--- a/api/v1beta1/foundationdbcluster_types.go
+++ b/api/v1beta1/foundationdbcluster_types.go
@@ -1886,6 +1886,12 @@ func GetProcessPort(processNumber int, tls bool) int {
 // the primary address.
 func (cluster *FoundationDBCluster) GetFullAddressList(address string, primaryOnly bool, processNumber int) []ProcessAddress {
 	addrs := make([]ProcessAddress, 0, 2)
+
+	// If the address is already enclosed in brackets, remove them since they
+	// will be re-added automatically in the ProcessAddress logic.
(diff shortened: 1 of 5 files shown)


api/v1beta1/foundationdbcluster_types.go
```diff
@@ -1886,6 +1886,12 @@ func GetProcessPort(processNumber int, tls bool) int {
 // the primary address.
 func (cluster *FoundationDBCluster) GetFullAddressList(address string, primaryOnly bool, processNumber int) []ProcessAddress {
 	addrs := make([]ProcessAddress, 0, 2)
+
+	// If the address is already enclosed in brackets, remove them since they
+	// will be re-added automatically in the ProcessAddress logic.
+	if strings.HasPrefix(address, "[") && strings.HasSuffix(address, "]") {
```
johscheuer, line 1892: We could directly use `TrimPrefix` and `TrimSuffix` the complexity will be the same but we touch the string only once.

johscheuer (approved):
LGTM 👍 Only one note from my side feel free to ignore it or change it :)

PR #874, opened 2021-08-06
Add additional checks that we have an empty exclusion list after doing replacements or migrations.

brownleej:
# Description

This adds some more checks to our tests for replacements to make sure we do not have stray entries in the exclusion list.

## Type of change

- Other

# Discussion

No

# Testing

Manual testing?*

I ran the unit tests.

No

# Documentation

No

# Follow-up

No

No


johscheuer (approved):
LGTM 👍

Issue #732, opened 2021-06-01
Keeping pending pods from blocking monitor conf updates


brownleej:
When a pod is stuck in pending, and we want to change the start commands for the fdbserver process, all of the changes to start commands get blocked on one bad pod. I think we can make this process more robust to pending pods. We could delete any pending pod with an incorrect config map, with the understanding that the recreated pod would come up with the latest config. We might also be able to just update the annotation to have the latest hash, but that could introduce race conditions.

We might also need to make sure that the IncorrectCommandLine condition gets cleared when a process is not running so that we don't try to bounce a process that's not there anymore. This wouldn't happen commonly, but we could see it if we start an update to the start commands and then have a process go into pending while that change is being rolled out.

PR #876, opened 2021-08-11
Keeping pending pods from blocking monitor conf updates

09harsh:
# Description
This PR handles the issue #732 by skipping all the Pending pods while doing a bounce. 
## Type of change

[x]New feature (non-breaking change which adds functionality)
- Documentation
- Other

# Discussion
NO

# Testing
Wrote a test case to check validity

# Documentation
Afaik no doc update is required

# Follow-up
NO


controllers/bounce_processes.go
```diff
@@ -60,7 +60,7 @@ func (b BounceProcesses) Reconcile(r *FoundationDBClusterReconciler, context ctx
 		}
 	}
 
-	processesToBounce := fdbtypes.FilterByCondition(cluster.Status.ProcessGroups, fdbtypes.IncorrectCommandLine, true)
+	processesToBounce := fdbtypes.FilterByCondition(cluster.Status.ProcessGroups, fdbtypes.IncorrectCommandLine, true, true)
```
johscheuer, line 63: Is there a good reason to change the method signature here and not use `SkipProcessGroup` in the for loop below? I would prefer that instead of changing the method since `SkipProcessGroup` also checks if a Pod is in the Pending state for a given time.

johscheuer (changes requested):

controllers/bounce_processes.go
09harsh: Completely agree on this, let me make the changes.

controllers/bounce_processes_test.go
```diff
@@ -98,6 +98,30 @@ var _ = Describe("BounceProcesses", func() {
...
+			Expect(requeue).To(BeNil())
+		})
+
+		It("should not kill the pending processes", func() {
+			addresses := make([]string, 0, 1)
+			processGroupAddresses := fdbtypes.FindProcessGroupByID(cluster.Status.ProcessGroups, "storage-1").Addresses
+			for _, address := range processGroupAddresses {
+				addresses = append(addresses, fmt.Sprintf("%s:4501", address))
+			}
+			sort.Strings(adminClient.KilledAddresses)
```
johscheuer, line 120: Why do we sort the addresses here?

johscheuer (approved):
LGTM 👍

PR #878, opened 2021-08-11
More flexible rules for customizing docker images

### LINKED ISSUE #387: Supporting customization of tags in the built-in images

brownleej:
# Description

This adds more options for customizing docker images, in order to cover more use cases like testing upgrades and having permanent image variants.

Resolves #387 

## Type of change

# Discussion

I considered deprecating the ability to specify custom images in the pod templates, since the new image configs can cover almost all the use cases for that, and I think it would lead to less confusion about how we form the images. However, that change wasn't covered in the original issue, and the pod templates allow customizing images on a per-process-class basis, so I did not want to deprecate support for that without further discussion.

This uses an implicit default tag suffix of `-1` for the sidecar image, and though it allows changing the suffix, it does not allow specifying a blank suffix. That matches the current behavior, but I don't like having defaults that people cannot replace with empty values. However, users can still specify a full tag as a workaround for this. I considered making the strings for these fields string pointers to allow us to distinguish between nil and an empty string, but that added significant complexity and didn't seem worth it to me.

# Testing

Manual testing?*

I added unit tests for how we form images with different variations on the image config, and created local clusters with different kinds of image names.

No.

# Documentation

Yes.

Yes, I've added a section on this.

# Follow-up

We may want an issue to discuss deprecating support for images in the pod spec, as I mentioned above. I also noticed in the course of this work that the deprecated fields in `ContainerOverrides` are not handled by the spec normalization, and are not caught in the deprecation checker, so we need to fix that before the 1.0 release.

No.


api/v1beta1/foundationdbcluster_types.go
```diff
@@ -2084,6 +2071,52 @@ type ContainerOverrides struct {
 	SecurityContext *corev1.SecurityContext `json:"securityContext,omitempty"`
 }
 
+// ImageConfig provides a policy for customizing an image.
+type ImageConfig struct {
```
johscheuer, line 2075: We should add some documentation here that the values are merged if multiple entries are matching.

internal/pod_models.go
```diff
@@ -195,7 +184,7 @@ func GetPodSpec(cluster *fdbtypes.FoundationDBCluster, processClass fdbtypes.Pro
 		versionString = cluster.Spec.Version
 	}
 
-	image, err := GetImage(cluster.Spec.MainContainer.ImageName, mainContainer.Image, "foundationdb/foundationdb", versionString, processSettings.GetAllowTagOverride())
+	image, err := GetImage(mainContainer.Image, cluster.Spec.MainContainer.ImageConfigs, versionString, processSettings.GetAllowTagOverride())
```
johscheuer, line 187: Should we deprecate `processSettings.GetAllowTagOverride()` with this change?

johscheuer (approved):
One NIT otherwise LGTM 👍

internal/pod_models.go
brownleej: I think we can talk about that in the context of deprecating custom images in the pod specs in general.

Issue #855, opened 2021-07-15
Add a log field reconciler


johscheuer:
We should add a log field `reconciler` for all of our log messages. That makes it easier for a human operator to group log messages based on the reconciler. The idea would be to the the reconciler name in each log statement  e.g. all log messages emitted by `BounceProcesses` reconciler should have the field `"reconciler": "BounceProcesses"`. Without this we don't have a way to see all log messages for a specific reconciler.

PR #879, opened 2021-08-12
Adding a log field reconciler

09harsh:
# Description

Adding a log field reconciler to all the reconcilers. This PR handles the issue #855 

## Type of change

# Discussion
No

# Testing

Manual testing by running operator locally.

# Documentation
Nothing to update

# Follow-up
Nothing to follow up

### DIFF
--- a/config/crd/bases/apps.foundationdb.org_foundationdbbackups.yaml
+++ b/config/crd/bases/apps.foundationdb.org_foundationdbbackups.yaml
@@ -2,7 +2,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.6.0-beta.0
+    controller-gen.kubebuilder.io/version: v0.6.1
   creationTimestamp: null
   name: foundationdbbackups.apps.foundationdb.org
 spec:
--- a/config/crd/bases/apps.foundationdb.org_foundationdbbackups.yaml
+++ b/config/crd/bases/apps.foundationdb.org_foundationdbbackups.yaml
@@ -2,7 +2,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.6.1
+    controller-gen.kubebuilder.io/version: v0.6.0-beta.0
   creationTimestamp: null
   name: foundationdbbackups.apps.foundationdb.org
 spec:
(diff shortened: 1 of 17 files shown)


config/crd/bases/apps.foundationdb.org_foundationdbbackups.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.6.1
+    controller-gen.kubebuilder.io/version: v0.6.0-beta.0
```
johscheuer, line 5: Can you fix that please? You have to delete `controller-gen` and then run it again.

controllers/add_pods.go
```diff
@@ -38,13 +38,14 @@ type AddPods struct{}
 // Reconcile runs the reconciler's work.
 func (a AddPods) Reconcile(r *FoundationDBClusterReconciler, context ctx.Context, cluster *fdbtypes.FoundationDBCluster) *Requeue {
 	configMap, err := internal.GetConfigMap(cluster)
+	logger := log.WithValues("cluster", cluster.Name, "reconciler", "AddPods")
```
johscheuer, line 41: Can we add `"namespace", cluster.Namespace,` that will always be the same like the cluster.

johscheuer (changes requested):
Can you please fix the generated files.

config/crd/bases/apps.foundationdb.org_foundationdbbackups.yaml
09harsh: done

controllers/add_pods.go
09harsh: done

controllers/add_pods.go
```diff
@@ -97,7 +98,7 @@ func (a AddPods) Reconcile(r *FoundationDBClusterReconciler, context ctx.Context
 				}
 				ip := service.Spec.ClusterIP
 				if ip == "" {
-					log.Info("Service does not have an IP address", "namespace", cluster.Namespace, "cluster", cluster.Name, "podName", pod.Name)
+					logger.Info("Service does not have an IP address", "podName", pod.Name)
```
johscheuer, line 101: Instead of the pod name we should log here the process group ID to be consistent with the rest of the logging system.

controllers/add_pods.go
```diff
@@ -38,13 +38,14 @@ type AddPods struct{}
 // Reconcile runs the reconciler's work.
 func (a AddPods) Reconcile(r *FoundationDBClusterReconciler, context ctx.Context, cluster *fdbtypes.FoundationDBCluster) *Requeue {
 	configMap, err := internal.GetConfigMap(cluster)
+	logger := log.WithValues("cluster", cluster.Name, "namespace", cluster.Namespace, "reconciler", "AddPods")
```
johscheuer, line 41:
```suggestion
	logger := log.WithValues("namespace", cluster.Namespace, "cluster", cluster.Name, "reconciler", "AddPods")
```
In most places we have namespace as the first pair (it shouldn't matter in the end).

johscheuer (changes requested):
Just a few last things to make our logging more consistent after that 👍

johscheuer (approved):
LGTM 👍

PR #880, opened 2021-08-12
A few minor refactors.

RenxuanW:
# Description
1. Refactor DesiredFaultTolerance() so that it can be leveraged by chaos testing framework;
2. Refactor MinimumFaultDomains() too;
3. Make RedundancyMode an enum type.

## Type of change

- Minor refactor

# Discussion

# Testing

Manual testing?*

# Documentation

# Follow-up

### DIFF
--- a/api/v1beta1/foundationdbcluster_types.go
+++ b/api/v1beta1/foundationdbcluster_types.go
@@ -1358,9 +1358,9 @@ func (cluster *FoundationDBCluster) GetProcessCountsWithDefaults() (ProcessCount
 }
 
 // DesiredFaultTolerance returns the number of replicas we should be able to
-// lose when the cluster is at full replication health.
-func (cluster *FoundationDBCluster) DesiredFaultTolerance() int {
-	switch cluster.Spec.RedundancyMode {
+// lose given a redundancy mode.
+func DesiredFaultTolerance(redundancyMode RedundancyMode) int {
+	switch redundancyMode {
 	case RedundancyModeSingle:
 		return 0
 	case RedundancyModeDouble, RedundancyModeUnset:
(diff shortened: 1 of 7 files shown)


api/v1beta1/foundationdbcluster_types.go
```diff
@@ -1205,7 +1205,7 @@ func (cluster *FoundationDBCluster) GetProcessSettings(processClass ProcessClass
 func (cluster *FoundationDBCluster) GetRoleCountsWithDefaults() RoleCounts {
 	counts := cluster.Spec.RoleCounts.DeepCopy()
 	if counts.Storage == 0 {
-		counts.Storage = 2*cluster.DesiredFaultTolerance() + 1
```
brownleej, line 1208: I think we'd be better off keeping `cluster.DesiredFaultTolerance()` as a short-hand for `DesiredFaultTolerance(cluster.Spec.RedundancyMode)`, both because it's shorter and because it avoids removing a publicly exposed method.

brownleej (approved):
Small suggestion, but other than that this looks good to me.

api/v1beta1/foundationdbcluster_types.go
RenxuanW: It doesn't make much sense to me, as DesiredFaultTolerance(cluster.Spec.RedundancyMode) needs to be public anyways. Are there any possible client calling cluster.DesiredFaultTolerance()? I mean other repos that depends of this repo.

johscheuer: "Are there any possible client calling cluster.DesiredFaultTolerance()"

Since this repo is public to the world we don't know who or where this function is used and we should try our best to not break anything (even we could with the current `0.x.x` version).

johscheuer (approved):
LGTM 👍

api/v1beta1/foundationdbcluster_types.go
RenxuanW: Then isn't changing RedundancyMode from string to enum gonna break things anyway?

johscheuer: None of these methods are taking the redundancy mode as an argument (at least as they were). It will only break things if someone was setting the redundancy mode in the database configuration programatically. I think that latter is better than changing the method signature and it should be not that much code to have e.g.

```go
func (cluster *FoundationDBCluster) DesiredFaultTolerance() int {
  return DesiredFaultTolerance(cluster.Spec.RedundancyMode)
}
```

PR #881, opened 2021-08-14
Printout Pod names instead of Pod struct

johscheuer:
# Description

Print Pod names instead of Pod structs when running the analyze command.

## Type of change

# Discussion

# Testing

Unit tests + locally

# Documentation

-
# Follow-up

-


brownleej (approved):
👍

PR #885, opened 2021-08-18
Replace pods with the PodFailing condition.

### LINKED ISSUE #882: NeedsReplacement should check PodFailing
I think it makes sense to check in ` NeedsReplacement` if a Pod is failing for a log time and then replace it, there could be different issues like the Sidecar crashed and doesn't get restarted from the Kubelet (for a various reasons).

https://github.com/FoundationDB/fdb-kubernetes-operator/blob/705b85a53c9057a4ca30a56c20eba93a43e365b1/api/v1beta1/foundationdbcluster_types.go#L450-L458

brownleej:
# Description

This replaces pods when they have been marked as failing for 30 minutes.

Resolves #882

## Type of change

# Discussion

No

# Testing

Manual testing?*

I added unit tests for the NeedsReplacement method. I also tested in my local environment by putting the sidecar container for a pod in a crash loop state and waiting for the automatic replacement to kick in.

No.

# Documentation

Yes.

# Follow-up

No.

No.

### DIFF
--- a/api/v1beta1/foundationdbcluster_types.go
+++ b/api/v1beta1/foundationdbcluster_types.go
@@ -64,6 +64,8 @@ type FoundationDBClusterList struct {
 	Items           []FoundationDBCluster `json:"items"`
 }
 
+var conditionsThatNeedReplacement = []ProcessGroupConditionType{MissingProcesses, PodFailing}
+
 func init() {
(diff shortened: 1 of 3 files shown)


ltsampros (approved):
LGTM

PR #886, opened 2021-08-18
Allow exclusions when another process is missing.

### LINKED ISSUE #806: Allowing the operator to start exclusions when there are multiple failing processes
We have a safety check to prevent starting exclusions when there are pods that are not reporting to the cluster and are not flagged for removal. This could be a problem if someone has multiple processes that are failing in that way. If they only replace one of them, the replacement would get stuck until they started a replacement for the second. Ideally we want people to be able to do replacements gradually if it is a better fit for their environment.

It's also a problem with automatic replacements, since we have a limit on the number of processes that can be replaced at once. If someone has this set to 1, and sees 2 pods fail, the operator won't be able to automatically replace them.

I think we should try to find a more nuanced check that will allow replacements to move forward in some of these failure scenarios.

brownleej:
# Description

This adds a more permissive safety check on missing processes before we start an exclusion.

Resolves #806 

## Type of change

# Discussion

I put in hardcoded limits here, instead of adding new settings. I don't know if this is something people really need to tweak, and I'm not sure if it's worth the complexity of adding new fields to the spec.

# Testing

Manual testing?*

I added unit tests for the core logic, and I ran manual tests in my dev environment.

No.

# Documentation

Yes.

# Follow-up

No.

No.

### DIFF
--- a/controllers/cluster_controller_test.go
+++ b/controllers/cluster_controller_test.go
@@ -952,7 +952,7 @@ var _ = Describe(string(fdbtypes.ProcessClassClusterController), func() {
 			})
 		})
 
-		Context("with a missing process and pending exclusion", func() {
+		Context("with missing processes and pending exclusion", func() {
 			var adminClient *MockAdminClient
 
 			BeforeEach(func() {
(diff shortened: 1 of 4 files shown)


AnnigeriShambu (approved):
LGTM

PR #889, opened 2021-08-19
Adds additional checks for deprecated fields in container overrides.

### LINKED ISSUE #888: Handle the deprecated fields on ContainerOverrides in NormalizeClusterSpec
We have several fields in the `ContainerOverrides` struct, such as `Env`, that are flagged as being deprecated in the documentation, but are not evaluated in the `NormalizeClusterSpec` method. That means that when values are set for these fields, they are not detected by our deprecation checking tool. We should add logic in `NormalizeClusterSpec` to move values from these fields into the new, supported locations.

brownleej:
# Description

This moves the logic for handling 3 deprecated fields in ContainerOverrides into the NormalizeClusterSpec method.

Resolves #888 

## Type of change

# Discussion

No

# Testing

Manual testing?*

Manual tests and unit tests

No

# Documentation

I don't think there are any documentation updates necessary.

# Follow-up

No.

No.

### DIFF
--- a/controllers/cluster_controller_test.go
+++ b/controllers/cluster_controller_test.go
@@ -1717,67 +1717,6 @@ var _ = Describe(string(fdbtypes.ProcessClassClusterController), func() {
 			})
 		})
 
-		Context("with a change to environment variables with the deprecated field", func() {
-			BeforeEach(func() {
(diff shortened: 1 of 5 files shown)


johscheuer (approved):
LGTM 👍

PR #892, opened 2021-08-24
Add a command to fix coordinator IPs after a majority of coordinators get new IPs.

brownleej:
# Description

This provides tooling and documentation to help with a potential failure scenario where all coordinators get deleted and come up with new IPs.

## Type of change

# Discussion

The new command doesn't work for multi-KC clusters, because any given invocation will only have access to the process group information (and thus the new addresses) for the processes in that KC. I think we should address that in the future, as part of the broader work on improving multi-KC support in the plugin.

I changed the logic around tracking process group IPs to keep old IPs around when the database is unavailable. This was necessary to be able to use the process group address list to compare old and new IPs. I verified that once the database recovers, we clear out the old IPs, so we're still fulfilling the requirement to prevent the address list from growing indefinitely.

# Testing

Manual testing?*

I ran manual tests of deleting all pods and then recovering the cluster.

No.

# Documentation

Yes.

# Follow-up

I think we'll want a follow-up issue on multi-KC support for the fix-coordinator-ips command.

No.

### DIFF
--- a/docs/manual/debugging.md
+++ b/docs/manual/debugging.md
@@ -121,7 +121,18 @@ The `ExcludeInstances` subreconciler can get stuck if it needs to exclude proces
 
 Any step that requires a lock can get stuck indefinitely if the locking is blocked. See the section on [Coordinating Global Operations](fault_domains.md#coordinating-global-operations) for more background on the locking system. You can see if the operator is trying to take a lock by looking in the logs for the message `Taking lock on cluster`. This will identify why the operator needs a lock. If another instance of the operator has a lock, you will see a log message `Failed to get lock`, which will have an `owner` field that tells you what instance has the lock, as well as an `endTime` field that tells you when the lock will expire. You can then look in the logs for the instance of the operator that has the lock and see if that operator is stuck in reconciliation, and try to get it unstuck. Once the operator completes reconciliation and the lock expires, your original instance of the operator should able to get the lock for itself.
 
-### Running CLI Commands
+## Coordinators Getting New IPs
+
+The FDB cluster file contains a list of coordinator IPs, and if the coordinator processes are not listening on those IPs, the database will be unavailable. If you have your processes listening on their pod IPs, and a majority of the coordinator pods are deleted in a short window, the operator will not be able to automatically recover the cluster. You can fix this through a manual recovery process:
+
+1. Identify the coordinator processes based on the IPs in the connection string, and the addresses in the process group status
+2. Replace the coordinator IPs in the connection string with the new IPs for those processes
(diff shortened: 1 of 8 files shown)


docs/manual/debugging.md
```diff
@@ -121,7 +121,18 @@ The `ExcludeInstances` subreconciler can get stuck if it needs to exclude proces
...
+
+1. Identify the coordinator processes based on the IPs in the connection string, and the addresses in the process group status
+2. Replace the coordinator IPs in the connection string with the new IPs for those processes
+3. Edit the file `/var/fdb/data/fdb.cluster` in the `foundationdb` container in each pod to contain the new connection string, and kill the fdbserver processes.
+4. Edit the `connectionString` in the cluster status, or the `seedConnectionString` in the cluster spec, to contain the new connection string.
+
+To simplify this process, the kubectl-fdb plugin has a command that encapsulates these steps. You can run `kubectl fdb fix-coordinator-ips -c example-cluster`, and that should update everything with the modified connection string, bring the cluster back up, and allow the operator to continue with any further reconciliation work.
```
johscheuer, line 133: Should we state here that this process won't change anything when all coordinators have the correct IP?

kubectl-fdb/cmd/fix_coordinator_ips.go
```diff
@@ -0,0 +1,246 @@
...
+
+func runFixCoordinatorIPs(kubeClient client.Client, cluster *fdbtypes.FoundationDBCluster, context string, namespace string, dryRun bool) error {
+	patch := client.MergeFrom(cluster.DeepCopy())
+	err := updateIPsInConnectionString(cluster)
+	if err != nil {
+		return err
+	}
+
+	if dryRun {
+		log.Printf("New connection string: %s", cluster.Status.ConnectionString)
```
johscheuer, line 220: Any reason to print this message only during a dry run?

kubectl-fdb/cmd/fix_coordinator_ips_test.go
```diff
@@ -0,0 +1,233 @@
...
+					for index, command := range commands {
+						Expect(command.Args).To(Equal(input.ExpectedCommands[index]))
+					}
+				}
+			},
+			Entry("instance with valid pod",
+				testCase{
+					ExpectedCommands: [][]string{
+						{
+							"/usr/local/bin/kubectl",
```
johscheuer, line 104: My only concern here would be that the test cases assume that `kubectl` is installed under `/usr/local/bin/kubectl` if not the tests will be failing with a different value. We could somehow inject the path for testing only, that would allow these tests to be run without having `kubectl` in that location (or installed at all).

docs/manual/debugging.md
brownleej: In that event it will still run the same steps, including killing the fdbserver processes, so I think it's better to leave the wording as it is.

PR #897, opened 2021-09-01
Return err if file not found instead of nil

AnnigeriShambu:
# Description

Currently, If the `fdbcli` binary is not present `nil` is returned instead of an error string [here](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/3fac0fbfed79abb4785c77b03dcc69acf11d9981/fdbclient/admin_client.go#L422). This makes debugging difficult.  

## Type of change

# Discussion

NA
# Testing

Manual testing?*
Local testing

NA
# Documentation

NA
NA
NA
NA
# Follow-up

NA
NA


johscheuer (approved):
LGTM 👍

brownleej (approved):
👍

PR #898, opened 2021-09-01
Disable non-blocking excludes due to concerns about the handling of missing processes.

brownleej:
# Description

This disables non-blocking excludes because of concerns that missing processes can be permanently deleted before their data is re-replicated.

## Type of change

# Discussion

There are ways that we could modify the design to be safe with the current exclude behavior, but I think that will take more time and we're better off doing a patch release to protect durability.

# Testing

Manual testing?*

I ran a test in my local environment of excluding a process under load.

No.

# Documentation

The documentation around this method was incorrect before, but with the change in this PR it becomes correct again.

# Follow-up

We should track the work to enable non-blocking excludes again.

No.

### DIFF
--- a/api/v1beta1/foundationdb_version.go
+++ b/api/v1beta1/foundationdb_version.go
@@ -153,7 +153,7 @@ func (version FdbVersion) HasSidecarCrashOnEmpty() bool {
 // This is currently set to false across the board, pending investigation into
 // potential bugs with non-blocking excludes.
 func (version FdbVersion) HasNonBlockingExcludes() bool {
-	return version.IsAtLeast(FdbVersion{Major: 6, Minor: 3, Patch: 5})
+	return false
 }
 
 // NextMajorVersion returns the next major version of FoundationDB.


api/v1beta1/foundationdb_version.go
```diff
@@ -153,7 +153,7 @@ func (version FdbVersion) HasSidecarCrashOnEmpty() bool {
 // This is currently set to false across the board, pending investigation into
 // potential bugs with non-blocking excludes.
 func (version FdbVersion) HasNonBlockingExcludes() bool {
-	return version.IsAtLeast(FdbVersion{Major: 6, Minor: 3, Patch: 5})
+	return false
```
johscheuer, line 156: Like discussed offline I think we should a setting to https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/docs/cluster_spec.md#automaticreplacementoptions to allow the users to decide what they want to use + allow us to test the different exclusion modes easier. The logic should be in the end:

```go
return version.IsAtLeast(FdbVersion{Major: 6, Minor: 3, Patch: 5}) && useNonBlockingExclude
```

johscheuer (changes requested):

johscheuer (approved):
LGTM 👍 Could you create an issue for adding the settings so we can take that later.

PR #899, opened 2021-09-02
Publish helm chart via GH pages, add tests

funkypenguin:
This is a recreation of https://github.com/FoundationDB/fdb-kubernetes-operator/pull/543, which required a level of git-fu/rebasing beyond my abilities ;)

This PR:

1. Moves the helm folder to charts, for better compatibility with the helm-provided chart-testing GitHub actions
2. Adds GitHub action for linting and testing charts on push, ensuring that charts actually run in a KinD cluster
3. Adds GitHub action to publish the helm chart when a push to master is made, which includes changes to the chart files. This would allow users to install the chart via a helm repo URL, rather than requiring them to clone the repo in order to deploy the chart.

In order to effect the chart publishing, we'd need a branch in the repo called `gh-pages`

Cheers!
D


johscheuer (commented):
The changes LGTM, could you. just add a short documentation here: https://github.com/FoundationDB/fdb-kubernetes-operator#running-the-operator how to install the operator using helm (a subsection under that point makes IMHO the most sense). Just to make it easier for people to consume the new helm chart.

Without being an expert in hosting helm charts on GH pages :) we will always host (only) one version and that will be the latest version based on pushes on master/main, correct?

funkypenguin:
I've made a brief addition to the README :) 

> Without being an expert in hosting helm charts on GH pages :) we will always host (only) one version and that will be the latest version based on pushes on master/main, correct?

Actually no, you'll end up bumping the chart version with every change (_either to the chart templates or to the application_), and your "repo" will contain the latest, as well as all the preceding versions. This means that a user is free to deploy the latest version (_the default behaviour_), but also any other historical version. It's helpful for teams (_like mine_) who like to version everything which goes into our cluster, and bump manually :)

For an example of how this works, see https://github.com/portainer/k8s/tree/gh-pages, which produces a complete list of hosted versions, at https://portainer.github.io/k8s/index.yaml

Cheers!
D

johscheuer:
@brownleej if you have some time, could you take a look?

README.md
```diff
@@ -19,6 +19,14 @@ kubectl apply -f https://raw.githubusercontent.com/FoundationDB/fdb-kubernetes-o
 kubectl apply -f https://raw.githubusercontent.com/foundationdb/fdb-kubernetes-operator/master/config/samples/deployment.yaml
 ```
 
+Or by installing the helm chart (*see values.yaml [here](charts/fdb-operator/values.yaml)*):
+
+```bash
+helm repo add fdb-kubernetes-operator https://foundationdb.github.io/fdb-kubernetes-operator/
```
brownleej, line 25: This won't work at the time this is merged, because we don't have the GitHub pages configured. I'd rather we track that as a follow-up issue, and defer any links to those pages until they are live.

.github/workflows/on-push-master-publish-chart.yml
```diff
@@ -0,0 +1,40 @@
...
+        run: |
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+
+      - name: Install Helm
+        uses: azure/setup-helm@v1
+        with:
+          version: v${{ env.HELM_VERSION }}
+
+      - name: Run chart-releaser
```
brownleej, line 36: Will this publish to the root of repo in the gh-pages branch? We've talked recently about using GitHub pages to host other content, like our user manual and other docs, so if we're going to publish the Helm charts there I would rather have those charts in a subdirectory of the gh-pages.

johscheuer:
Changes are merged in the other PR. Thanks @funkypenguin for your help on this!

Issue #901, opened 2021-09-02
Add additional fault tolerance check before deleting a ProcessGroup


johscheuer:
We should define a setting that allows to user to define the threshold for fault tolerance when removing ProcessGroups e.g. as a user I should be able to define that the operator should only remove ProcessGroups when the fault tolerance is at the desired level or allow to remove ProcessGroups when the desired fault tolerance is `1`. Since FDB `6.3.13` we have detailed information for `storage`, `log` and `coordinator` fault tolerance, we should use these to allow the user to specify the appropriate surge they are comfortable with. In a follow up we could extend this setting to also allow to configure these settings based on the region for multi region clusters. I think https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/docs/cluster_spec.md#foundationdbclusterautomationoptions is the right place to put this configuration in. 

We could use a configuration like:

```yaml
excludeOptions:
    generalSurge: 1
    storageSurge: 1
    logSurge: 1
    coordinatorSurge: 1
```

This configuration would allow a different of 1 for each fault tolerance e.g. meaning we have one missing process where the data is not fully replicated. To reflect the current behaviour the default values would be `0`. The operator should always use the most specific value e.g. if `storageSurge` is set we will use that value if `storageSurge` is not set we will use `generalSurge` and if both are unset we will use `0` as default. As a note these surges will only be checked for stateful processes.

I'm open to a better naming/structure since this is only a proposal from my side.

johscheuer:
@brownleej do you think we need a design doc got that?

brownleej:
I think if we're going to make this highly customizable, and break it out by specific kinds of fault tolerance, it's worth putting it into a design doc.

brownleej:
On the other hand, we could make this simpler by adding a hard rule that we don't delete any pods when fault tolerance is degraded in any way, and give users the fallback of manually deleting pods if this causes things to get stuck.

johscheuer:
> ~On the other hand, we could make this simpler by adding a hard rule that we don't delete any pods when fault tolerance is degraded in any way, and give users the fallback of manually deleting pods if this causes things to get stuck.~
> 
> Crossing this out now that I see there's a PR for this exact thing :)

[...]

PR #903, opened 2021-09-05
Add setting to enforce full replication of the cluster before deleting Pods

johscheuer:
# Description

This is the "dumb" version of https://github.com/FoundationDB/fdb-kubernetes-operator/issues/901 without any additional checks (e.g. are we targeting a stateless process). I decided to use `false` as the default since that reflects the current behaviour and the user can/has decide if that's a valid risk or not.

## Type of change

# Discussion

I decided to query the FDB cluster for the latest status, one could also argue that the "cached" status from the FDB cluster object should be good enough.

# Testing

Unit tests.

# Documentation

I added a few lines about that feature but I think it's self describing.

# Follow-up

https://github.com/FoundationDB/fdb-kubernetes-operator/issues/901


controllers/remove_process_groups_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+	fdbtypes "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+
+	"k8s.io/utils/pointer"
+)
+
+var _ = Describe("remove_process_groups", func() {
+	var cluster *fdbtypes.FoundationDBCluster
+	//var fakeConnectionString string
```
johscheuer, line 39: TODO remove these lines

controllers/remove_process_groups_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+	BeforeEach(func() {
+		cluster = internal.CreateDefaultCluster()
+		//fakeConnectionString = "operator-test:asdfasf@127.0.0.1:4501"
+	})
+
+	Describe("Reconciliation", func() {
```
johscheuer, line 46: Remove the describe block

controllers/remove_process_groups.go
```diff
@@ -51,6 +52,25 @@ func (u RemoveProcessGroups) Reconcile(r *FoundationDBClusterReconciler, context
 		return nil
 	}
 
+	// We don't use the "cached" of the cluster status from the CRD to minimize the window between data loss (e.g. a node
+	// or a set of Pods is not reachable anymore). We still end up with the risk to actually query the FDB cluster and after that
+	// query the cluster gets into a degraded state.
+	// We could be smarter here and only block removals that target stateful processes by e.g. filtering hose out of the
```
johscheuer, line 58:
```suggestion
	// We could be smarter here and only block removals that target stateful processes by e.g. filtering those out of the
```

ltsampros (approved):
I like the PR but happy to discuss the following points:
1) `EnforceFullReplicationForDeletion` should be true by default
2) Do we need the knob in the first place, if we make this behaviour the default one?

johscheuer:
> I like the PR but happy to discuss the following points:
> 
> 1. `EnforceFullReplicationForDeletion` should be true by default

I decided to default to `false` to not change the current behaviour. Changing this setting is more or less a noop from the cluster point of view. Since we are still in `0.*` we could argue that this is fine. Let's chat about that offline and decide what's the best way to go forward.

> 2. Do we need the knob in the first place, if we make this behaviour the default one?

I think it can be useful for a user to decide if they require these stricter checks or not. Also it's useful to be able to disable this check if something is not working as expected or the cluster is in a weird state and disabling this check would actually help (I don't have a specific case here in mind).

brownleej (approved):
👍

johscheuer:
Given the offline discussions and the feedback in the related issue I make the default `true` but still leave the setting here for user to enable/disable the setting.

PR #906, opened 2021-09-07
Rename destinationClusterName to clusterName to match the other resources

johscheuer:
# Description

In the other resources we always use `clusterName`, for consistency we should use the same in the restore CRD. I think the context already describes that this will be the 'destination" for the restore.

## Type of change

# Discussion

None.

# Testing

Local.

# Documentation

Adjusted.

# Follow-up

None.

### DIFF
--- a/api/v1beta1/foundationdbrestore_types.go
+++ b/api/v1beta1/foundationdbrestore_types.go
@@ -51,6 +51,7 @@ type FoundationDBRestoreSpec struct {
 	// DestinationClusterName provides the name of the cluster that the data is
 	// being restored into.
 	// Deprecated: Use ClusterName instead.
+	// +kubebuilder:validation:Optional
 	DestinationClusterName string `json:"destinationClusterName"`
 
 	// clusterName provides the name of the cluster that the data is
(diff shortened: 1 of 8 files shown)


api/v1beta1/foundationdbrestore_types.go
```diff
@@ -50,8 +50,15 @@ type FoundationDBRestoreList struct {
 type FoundationDBRestoreSpec struct {
 	// DestinationClusterName provides the name of the cluster that the data is
 	// being restored into.
+	// Deprecated: Use ClusterName instead.
```
brownleej, line 53: `DestinationClusterName` matches the naming in the `fdbrestore` command. Even though we don't support it yes in the operator, there are some cases where you want to pass a source cluster to a restore, so that you can pull timekeeper data from the original cluster. This would make `ClusterName` ambiguous.

brownleej (changes requested):

api/v1beta1/foundationdbrestore_types.go
johscheuer: That makes sense 👍

PR #908, opened 2021-09-07
Check for fault tolerance and not cluster health when deleting Pods

johscheuer:
# Description

Ironically the FDB cluster health is a bad indicator of the actual cluster health. Instead we should look if the cluster is fully replicated and if the cluster is available. Otherwise different things that are unrelated to the actual cluster health can block deletions e.g. a test process that is not running any more or clients with a wrong cluster file.

## Type of change

# Discussion

None.

# Testing

Locally.

# Documentation

None.

# Follow-up

None.


brownleej (approved):
👍

PR #910, opened 2021-09-07
Remove unused status fields

johscheuer:
# Description

Remove unused/deprecated fields from cluster status. Most of the fields are not set anymore so we can go ahead and remove them (they are useless anyway). 

## Type of change

# Discussion

None

# Testing

Locally.

# Documentation

None.

# Follow-up

None.


johscheuer:
I guess I just was motivated to get rid of some code :)

PR #917, opened 2021-09-13
Check if the cluster has the desired fault tolerance before removing Process Groups

johscheuer:
# Description

This ensures we only replace failed Pods when we meet the desired fault tolerance.

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/920

## Type of change

# Discussion

I decided not to add a setting to enable/disable this setting.

# Testing

Local, unit testing

# Documentation

I'll add some additional docs.

# Follow-up

-


controllers/cluster_controller.go
```diff
@@ -699,6 +699,32 @@ func (r *FoundationDBClusterReconciler) hasFullReplication(cluster *fdbtypes.Fou
 	return status.Cluster.FullReplication, nil
 }
 
+func (r *FoundationDBClusterReconciler) hasDesiredFaultTolerance(cluster *fdbtypes.FoundationDBCluster) (bool, error) {
+	adminClient, err := r.DatabaseClientProvider.GetAdminClient(cluster, r)
```
johscheuer, line 703: If the version of FDB is lower than 6.2 just approve it since we are missing the zone fault tolerance information and 6.1 will be deprecated in 1.0.

controllers/cluster_controller.go
```diff
@@ -699,6 +699,44 @@ func (r *FoundationDBClusterReconciler) hasFullReplication(cluster *fdbtypes.Fou
...
+
+	expectedFaultTolerance := cluster.DesiredFaultTolerance()
+	log.V(0).Info("Has desired fault tolerance",
+		"namespace", cluster.Namespace,
+		"cluster", cluster.Name,
+		"expectedFaultTolerance", expectedFaultTolerance,
+		"maxZoneFailuresWithoutLosingData", status.Cluster.FaultTolerance.MaxZoneFailuresWithoutLosingData,
+		"maxZoneFailuresWithoutLosingAvailability", status.Cluster.FaultTolerance.MaxZoneFailuresWithoutLosingAvailability)
+
+	return internal.HasDesiredFaultTolerance(
```
brownleej, line 734: What's the advantage of looking at two different fields here, rather than relying on `cluster.full_replication`?

controllers/replace_failed_pods_test.go
```diff
@@ -49,63 +52,90 @@ var _ = Describe("replace_failed_pods", func() {
...
+				ProcessGroupConditionType: fdbtypes.MissingProcesses,
+				Timestamp:                 time.Now().Add(-1 * time.Hour).Unix(),
+			})
 		})
 
-		Context("with no missing processes", func() {
+		Context("with no other removals", func() {
 			It("should return false", func() {
-				Expect(result).To(BeFalse())
+				Expect(result).To(BeTrue())
```
brownleej, line 80: This condition seems like it's no longer matching the `It` statement above it. Does that need a change as well?

internal/fault_tolerance.go
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package internal
+
+// HasDesiredFaultTolerance checks if the cluster has the desired fault tolerance.
+func HasDesiredFaultTolerance(expectedFaultTolerance int, maxZoneFailuresWithoutLosingData int, maxZoneFailuresWithoutLosingAvailability int) bool {
+	return !(maxZoneFailuresWithoutLosingData < expectedFaultTolerance || maxZoneFailuresWithoutLosingAvailability < expectedFaultTolerance)
```
brownleej, line 25: This is pretty nit-picky, but given the confusion over this check recently, I wonder if it would be clearer if we distributed the negation to the two different conditions rather than negating the larger statement.

brownleej (changes requested):

brownleej:
I think we should talk more about whether this check is desirable. In some cases, replacements may be necessary to recover fault tolerance. I'm also not sure that we need this check, because starting a replacement does not in itself put fault tolerance at risk.

controllers/cluster_controller.go
johscheuer: I believe (I'll dig into the code to confirm this) is that the `cluster.full_replication` is only looking at the data but is not including the availability part (I'm not too sure about that). I'll investigate and find out what's the best solution to figure out the clusters fault tolerance status. As a side product I'm hopefully able to clarify that on the FDB doc (they are not totally clear about the different "metrics").

controllers/replace_failed_pods_test.go
johscheuer: Good catch I'll check that test case. I assume fat fingers when I moved the test cases outside the second describe block.

internal/fault_tolerance.go
johscheuer: I'll take a look and see if I can make that statement more expressive + adding some additional comments what we are doing (and why).

controllers/replace_failed_pods.go
```diff
@@ -48,12 +48,23 @@ func (c ReplaceFailedPods) Reconcile(r *FoundationDBClusterReconciler, context c
...
+	// If we are not able to replace any failed instance because the fault tolerance dropped
+	// we don't want to block any subsequent steps.
+	hasDesiredFaultTolerance, err := r.hasDesiredFaultTolerance(cluster)
+	if err != nil {
+		return false
+	}
+
+	if !hasDesiredFaultTolerance {
+		return false
+	}
```
johscheuer, line 66: Thinking more about this I think you're right John. With the assumption that the exclude command will block until the data is replicated + the check to only delete/remove Pods when we meet our fault tolerance (or a fully replicated) we shouldn't need this check here for the following reasons:

- The automatic replacements will be limited by the number of concurrent replacements (default `1`)
- We will only remove the process group and all associated resources (Pod, PVC, ..) when the cluster meets the fault tolerance req. (or is fully replicated)

controllers/remove_process_groups.go
```diff
@@ -58,14 +58,20 @@ func (u RemoveProcessGroups) Reconcile(r *FoundationDBClusterReconciler, context
...
 	if cluster.GetEnforceFullReplicationForDeletion() {
-		fullyReplicated, err := r.hasFullReplication(cluster)
+		adminClient, err := r.DatabaseClientProvider.GetAdminClient(cluster, r)
 		if err != nil {
 			return &Requeue{Error: err}
 		}
+		defer adminClient.Close()
 
-		if !fullyReplicated {
+		fullReplication, err := hasDesiredFaultTolerance(adminClient, cluster)
```
johscheuer, line 67: That should be `hasDesiredFaultTolerance`.

docs/manual/replacements_and_deletions.md
```diff
@@ -27,6 +27,10 @@ The following conditions are currently eligible for replacement:
 * `MissingProcesses`: This indicates that a process is not reporting to the database.
 * `PodFailing`: This indicates that one of the containers is not ready.
 
+The operator will only replace failed Pods when the cluster meets the desired fault tolerance.
```
johscheuer, line 30: Remove this block.

controllers/cluster_controller.go
```diff
@@ -684,19 +684,44 @@ func GetPodSpec(cluster *fdbtypes.FoundationDBCluster, processClass fdbtypes.Pro
...
+	version, err := fdbtypes.ParseFdbVersion(cluster.Spec.Version)
 	if err != nil {
 		return false, err
 	}
-	defer adminClient.Close()
+
+	// TODO (johscheuer): https://github.com/FoundationDB/fdb-kubernetes-operator/issues/743
+	// For FDB versions lower than 6.2.0 we just approve the replacement.
+	// In 6.1 we only have machine information.
+	if !version.IsAtLeast(fdbtypes.FdbVersion{Major: 6, Minor: 2, Patch: 0}) {
```
brownleej, line 696: Rather than hardcoding a version here, can we add a method on the `FDBVersion` type? That's the general pattern we've followed, so it should make it easier to document the behavior of older versions in a centralized place, and remove it when we come to that issue.

controllers/cluster_controller.go
```diff
@@ -684,19 +684,44 @@ func GetPodSpec(cluster *fdbtypes.FoundationDBCluster, processClass fdbtypes.Pro
...
+	if !status.Client.DatabaseStatus.Available {
+		log.V(0).Info("Cluster is not available",
+			"namespace", cluster.Namespace,
+			"cluster", cluster.Name)
+
+		return false, nil
+	}
+
+	expectedFaultTolerance := cluster.DesiredFaultTolerance()
+	log.V(0).Info("Has desired fault tolerance",
```
brownleej, line 714: Instead of having the log message make an assertion like "Has desired fault tolerance", could we have it indicate that we're checking the fault tolerance, and include a value in the message that says whether the fault tolerance check is passing? I worry that the current form could give a misleading impression that we have the desired fault tolerance when we really don't.

controllers/remove_process_groups.go
```diff
@@ -58,14 +58,20 @@ func (u RemoveProcessGroups) Reconcile(r *FoundationDBClusterReconciler, context
...
-		if !fullyReplicated {
+		hasDesiredFaultTolerance, err := hasDesiredFaultTolerance(adminClient, cluster)
+		if err != nil {
+			return &Requeue{Error: err}
+		}
+
+		if !hasDesiredFaultTolerance {
 			return &Requeue{
-				Message: "Cluster is not fully replicated but is required for removals",
+				Message: "Cluster has degraded fault tolerance but is required for removals",
```
brownleej, line 74:
```suggestion
				Message: "Removals cannot proceed because cluster has degraded fault tolerance",
```

brownleej (changes requested):

controllers/cluster_controller.go
johscheuer: That's a good point, I just used the method name. I'll change that.

brownleej (approved):
👍

PR #919, opened 2021-09-13
Don't requeue in generate initial cluster file when no error occurs

### LINKED ISSUE #838: Adding message to requeue results where we have a possibly-nil error
When we refactored the return value for the subreconcilers recently, we wanted to ensure that requeues always came with a message, either in the form of an specific message or an error result. However, in the `GenerateInitialClusterFile` subreconciler, we are requeuing unconditionally with a potentially nil error at the end of the action. When the error is nil, we don't have any message in the requeue event. I think we should look for cases like this and make sure that we are always checking for the error being nil, and add a message in that case.

09harsh:
# Description

Fixes #838. Checked all the other occurrences and didn't find anything similar

## Type of change

# Discussion

Not needed

# Testing

Not needed

# Documentation
Not needed

# Follow-up

Not needed

### DIFF
--- a/controllers/generate_initial_cluster_file.go
+++ b/controllers/generate_initial_cluster_file.go
@@ -99,5 +99,8 @@ func (g GenerateInitialClusterFile) Reconcile(r *FoundationDBClusterReconciler, 
 	cluster.Status.ConnectionString = connectionString.String()
 
 	err = r.Status().Update(context, cluster)
-	return &Requeue{Error: err}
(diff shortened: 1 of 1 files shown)


johscheuer (changes requested):
Could you please change the commit message to reflect what you actually did?

Code LGTM.

johscheuer (approved):
LGTM 👍

PR #924, opened 2021-09-20
Add prefixes to our resource labels

### LINKED ISSUE #423: Update labels to contain FDB prefix
We should ensure that all labels set by the operator are prefixed with `foundationdb.org` some labels already go this prefix but there are still some labels without the prefix (not a complete list):

- `fdb-instance-id`
- `process-class`
- `fdb-cluster-name`

brownleej:
# Description

This adds a way to customizing the labels we use for process class and process group ID, and starts a transition to having `foundationdb.org` as a prefix for our default labels.

Resolves #423 

## Type of change

# Discussion

This requires the users to take on a multi-stage process to change the labels. I don't see any way around that because of the way we use the labels in the service selectors, since those selectors don't allow having two different sets of labels that are both accepted.

This puts in a latent change to add the new foundationdb.org labels to all clusters. We generally want to avoid putting in latent changes like this, so that users of the operator can control how changes get rolled out to their clusters. In this case, I think it's worth it because it will allow us to make the new labels the only ones used by default in the next major release without requiring the users to manually add the new labels first. Rolling out additional labels in a latent change should be fast and non-disruptive.

# Testing

Manual testing?*

I ran local tests where I created a cluster on the mainline code, added the labels with the new code, and modified the labels from there.

No.

# Documentation

Yes, I added documentation on how to configure the resource labels, and I added documentation on the different sets of labels we are now applying.

# Follow-up

 No

No, since this PR is already establishing the future defaults.


johscheuer (commented):
I take a closer look tomorrow. Do you think it makes sense to add a *Action required* section to the release notes for such changes?

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -2910,3 +2920,23 @@ func (cluster *FoundationDBCluster) GetEnforceFullReplicationForDeletion() bool
 
 	return *cluster.Spec.AutomationOptions.EnforceFullReplicationForDeletion
 }
+
+// GetProcessClassLabel provides the label that this cluster is using for the
+// process class when identifying resources.
+func (cluster *FoundationDBCluster) GetProcessClassLabel() string {
+	labels := cluster.Spec.LabelConfig.ProcessClassLabels
+	if len(labels) == 0 {
+		return FDBProcessClassLabel
```
johscheuer, line 2929: In order to not require current users to directly set the label we may want to use the old labels as default and before 1.0 we change that to the current labels.

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -2910,3 +2920,23 @@ func (cluster *FoundationDBCluster) GetEnforceFullReplicationForDeletion() bool
...
+	}
+	return labels[0]
+}
+
+// GetInstanceIDLabel provides the label that this cluster is using for the
+// instance ID when identifying resources.
+func (cluster *FoundationDBCluster) GetInstanceIDLabel() string {
+	labels := cluster.Spec.LabelConfig.InstanceIDLabels
+	if len(labels) == 0 {
+		return FDBInstanceIDLabel
```
johscheuer, line 2939: In order to not require current users to directly set the label we may want to use the old labels as default and before 1.0 we change that to the current labels.

johscheuer (changes requested):
In general the changes LGTM. I think with the new label we should directly move to `fdb-process-group-id` to prevent another round of label changes.

api/v1beta1/foundationdbcluster_types.go
brownleej: That shouldn't be a problem, because the cluster spec normalization provides the old labels as the defaults in this field if the field is empty.

johscheuer: True, that makes sense.

johscheuer (approved):
LGTM 👍

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -2775,6 +2775,16 @@ type LabelConfig struct {
 	// resources it creates.
 	ResourceLabels map[string]string `json:"resourceLabels,omitempty"`
 
+	// ProcessGroupIDLabels provides the labels that we use for the instance ID
+	// field. The first label will be used by the operator when filtering
+	// resources.
+	ProcessGroupIDLabels []string `json:"processGroupIDLabels,omitempty"`
```
ltsampros, line 2781: Since we are using only using one label for the filtering, what's the purpose of having a `[]string` ?

johscheuer: All labels will be added to the resource and this allows the user to move from one set of labels to another set: https://github.com/brownleej/fdb-kubernetes-operator/blob/label-prefix/docs/manual/customization.md#per-resource-labels

controllers/add_services.go
```diff
@@ -66,26 +73,55 @@ func (a AddServices) Reconcile(r *FoundationDBClusterReconciler, context ctx.Con
...
+
+// updateServices updates selected safe fields on a service based on a new
+// service definition.
+func updateService(r *FoundationDBClusterReconciler, context ctx.Context, cluster *fdbtypes.FoundationDBCluster, currentService *corev1.Service, newService *corev1.Service) error {
+	serviceLog := log.WithValues("namespace", cluster.Namespace, "cluster", cluster.Name, "service", currentService.Name)
+	originalSpec := currentService.Spec.DeepCopy()
+
+	currentService.Spec.Selector = newService.Spec.Selector
+
+	needsUpdate := !equality.Semantic.DeepEqual(currentService.Spec, *originalSpec)
```
ltsampros, line 113: shouldn't currentService be `newService` ?

controllers/add_services_test.go
```diff
@@ -86,6 +88,34 @@ var _ = Describe("add_services", func() {
 		It("should not create any services", func() {
 			Expect(newServices.Items).To(HaveLen(len(initialServices.Items)))
 		})
+
+		Context("with a change to the resource labels", func() {
```
ltsampros, line 92: nit: but this should read `match labels` since we are only modifying `cluster.Spec.LabelConfig.MatchLabels` right?

controllers/add_services.go
```diff
@@ -66,26 +73,55 @@ func (a AddServices) Reconcile(r *FoundationDBClusterReconciler, context ctx.Con
...
+	metadata := currentService.ObjectMeta
+	if mergeLabelsInMetadata(&metadata, newService.ObjectMeta) {
```
ltsampros, line 115: I fail to see how we remove a Label while we update the service. I think `mergeLabelsInMetadata` will just update the values of labels that exist in `newService.ObjectMeta`

controllers/add_services.go
brownleej: We want to compare with the currentService here because the currentService is what we're passing to the Update call. We're only updating selected fields on the existing service, because Kubernetes sets fields behind the scenes and we don't want to overwrite that.

ltsampros (approved):

Issue #916, opened 2021-09-12
Optimize the update Pod config routine


johscheuer:
Currently the [update Pod Config routine](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/controllers/update_pod_config.go#L58-L122) will run in a serial manner in addition to that we defined a wait time of `5s` and `10` retries in the [Pod client](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/83f97e8b3187f3649a4e1119cd8741b7fa582ab6/internal/pod_client.go#L165-L166). If multiple Pods are not reachable or not responding this can be very time consuming especially for clusters with many Pods. We should think about reducing the timeout, the number of retries and running the http calls in a go routine (or actually all checks). In addition to that we also have https://github.com/FoundationDB/fdb-kubernetes-operator/issues/590 which will solve the problem by decoupling the connection between the operator and the sidecars.

PR #930, opened 2021-09-23
Improve http client

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/916

## Type of change

# Discussion

Like described in the issue the previous settings had the drawback that they would block for a relative long time if one of the Pods is not reachable.

# Testing

Unit + local.

# Documentation

-

# Follow-up

-


brownleej (approved):
👍

PR #933, opened 2021-09-24
Use default retry policy

johscheuer:
# Description

Make use of the default retry policy instead of a custom one (https://pkg.go.dev/github.com/hashicorp/go-retryablehttp#DefaultRetryPolicy)

## Type of change

# Discussion

I'm not sure why we added the custom logic but looking at the sidecar I don't think that the custom check has any benefits.

# Testing

Unit + local.

# Documentation

None.

# Follow-up

-

### DIFF
--- a/internal/pod_client.go
+++ b/internal/pod_client.go
@@ -21,7 +21,6 @@
 package internal
 
 import (
-	"context"
 	"crypto/sha256"
 	"crypto/tls"
 	"crypto/x509"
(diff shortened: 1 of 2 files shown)


internal/pod_client.go
```diff
@@ -166,14 +165,6 @@ func (client *realFdbPodClient) makeRequest(method string, path string) (string,
 	retryClient.RetryWaitMax = 1 * time.Second
 	// Prevent logging
 	retryClient.Logger = nil
-	retryClient.CheckRetry = func(c context.Context, resp *http.Response, err error) (bool, error) {
```
ltsampros, line 169: I would prefer us to use `retryablehttp.ErrorPropagatedRetryPolicy()` so that we can log what happened after we exhaust our retries. Check the use here: https://github.com/hashicorp/go-retryablehttp/blob/master/client_test.go#L252

ltsampros (commented):
Minor changes needed.

internal/pod_client.go
johscheuer: That sounds reasonable and the `retryablehttp.ErrorPropagatedRetryPolicy` is a drop-in replacement.

ltsampros (approved):
approved with 1 nit: there is some code that with your change should not be reachable around line 210 in the same function, as exhausting the retries even in `>= http.StatusBadRequest` will set `err`.

PR #934, opened 2021-09-25
Ensure we only iterate over process groups

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/918 (Ensure in reconcilers that we iterate over ProcessGroups not Instances/Pods )

## Type of change

*Please select one of the options below.*

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)

# Discussion

- 

# Testing

Unit + local 

# Documentation

Adjusted.

# Follow-up

-

PR #939, opened 2021-09-29
Avoid updating metadata on pods that are pending removal

brownleej:
# Description

This prevents updating the metadata for pods that are pending removal. This behavior presented a problem in situations when making a change that required setting an annotation that can only be defined at pod creation time, which means it requires a migration to change the annotation value.

## Type of change

# Discussion

No.

# Testing

Manual testing?*

Manual testing.

No.

# Documentation

No.

# Follow-up

No

No.


johscheuer (commented):
The changes LGTM but we already have a fix for that in: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/934 (to be precise here: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/934/files#diff-3a97f91cbf5a278451049c185f23ea6e7704695afb004d09db47691936751dd1R53-R56). I'm not sure if it's worth merging this change and then the other PR (which would require a rebase).

brownleej:
Sounds good. I'll take a look at that PR today, and update / close this one accordingly.

brownleej:
@johscheuer  This PR also had some additional logging in a different area, so I rebased it and updated the description to reflect the new focus.

johscheuer (approved):
LGTM 👍

internal/pod_client.go
```diff
@@ -246,6 +246,9 @@ func (client *realFdbPodClient) GetVariableSubstitutions() (map[string]string, e
 	}
 	substitutions := map[string]string{}
 	err = json.Unmarshal([]byte(contents), &substitutions)
+	if err != nil {
+		log.Error(err, "Error deserializing pod substitutions", "responseBody", contents)
```
johscheuer, line 250: Shouldn't the caller method actually log the error? Otherwise we log the same error multiple times.

brownleej: It logs the error, but JSON parsing errors don't provide much context, so I think logging it twice will be helpful.

PR #941, opened 2021-10-01
Correct log level in UpdatePodConfig

johscheuer:
# Description

Corrects the log level to error for all error in `UpdatePodConfig`.

## Type of change

# Discussion
-

# Testing

local setup

# Documentation

-

# Follow-up

-

### DIFF
--- a/controllers/update_pod_config.go
+++ b/controllers/update_pod_config.go
@@ -74,21 +74,21 @@ func (u UpdatePodConfig) Reconcile(r *FoundationDBClusterReconciler, context ctx
 
 		serverPerPod, err := internal.GetStorageServersPerPodForPod(pod)
 		if err != nil {
-			curLogger.Info("Error when receiving storage server per Pod", "error", err)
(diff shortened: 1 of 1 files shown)


ltsampros (approved):
LGTM with one nit.

controllers/update_pod_config.go
```diff
@@ -137,9 +138,10 @@ func (u UpdatePodConfig) Reconcile(r *FoundationDBClusterReconciler, context ctx
 		}
 	}
 
-	// If any error has happened return the first error
+	// If any error has happened requeue.
+	// We don't provide an error here since we log all errors above.
 	if len(errs) > 0 {
-		return &Requeue{Error: errs[0]}
+		return &Requeue{Message: "errors occurred during update pod config reconcile"}
```
ltsampros, line 144: nit: I'd prefer to be logging all the errors here as it could help with filtering.

johscheuer: We already log all errors above when they occur why should we log the errors again?

Issue #926, opened 2021-09-21
Proposal: Support bin packing of fault domains


johscheuer:
In our current implementation the operator tries to distribute the process groups across as many fault domains as possible. If the number of fault domains is greater than the number of process groups we create (and we have enough resources) that means that every process group will be on a different fault domain. This might not be desirable in environments with a non-trivial number of process groups. One challenge is that most operational tasks are bound to fault domains since a human operator or the Kubernetes operator can only ensure that e.g. operating on a single fault domain reduces the risk of dataloss. A possible way to solve this is to allow a user to define a number of distinct logical fault domains. The operator would add an additional label to all Pods e.g. `foundationdb.org/distribution-key`, this key could be used to write a Pod Affinity and a Pod Anti Affinity. The value of the label would be the process group number modulo the number of desired fault domains. We could also extend this to bin pack all stateful processes.

I try to find some time this week to write a more formal proposal/design doc with additional examples and the limitations of this approach. From my current point of view I think this should be fairly simple to be supported in the operator and could have a huge benefit on the operational side for bigger clusters.

brownleej:
Yes, I think it would be great to get a proposal for this. I'm concerned that this kind of approach would lead to new scenarios where we fail to find capacity, but the right combination of soft constraints and hard constraints could limit that risk in practice.

PR #942, opened 2021-10-01
Initial design for better fault domain distribution

johscheuer:
# Description

Design for https://github.com/FoundationDB/fdb-kubernetes-operator/issues/926

## Type of change

- Documentation

# Discussion

-

# Testing

-

# Documentation

-
# Follow-up

Implement the design once approved.

### DIFF
--- a/docs/design/bin_pack_fault_domains.md
+++ b/docs/design/bin_pack_fault_domains.md
@@ -23,16 +23,24 @@ A better way would be to allow the operator to bin pack Pods in the same fault d
 
 The current implementation adds a `preferredDuringSchedulingIgnoredDuringExecution` `PodAntiAffinity` to every Pod [code](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/internal/pod_models.go#L308-L334).
 The label selector will contain the process class and the cluster name, to ensure processes with the same class are distributed across different fault domains is possible.
-A user can define additional `PodAntiAffinity` but can't prevent the operator from createn the default `PodAntiAffinity`.
-The only expection is the specival fault domain `foundationdb.org/none` and `foundationdb.org/kubernetes-cluster`.
+A user can define additional `PodAntiAffinity`s but can't prevent the operator from creating the default `PodAntiAffinity`.
+The only exception is the special fault domain `foundationdb.org/none` and `foundationdb.org/kubernetes-cluster`.
+
+In the design we will use the following terms:
+
+- `physical fault domain`: This refers to a fault domain that is available in Kubernetes e.g. a machine running the Kubernetes worker or a higher level fault domain like a rack, data hall etc.
+  This will reflect an underlying physical fault domain.
+- `logical fault domain`: The logical fault domain is not something that refers directly to a physical fault domain and can also be spread across multiple physical fault domains.
+  The idea is to have a way to group some processes of a FoundationDB cluster together with the same zone ID, this zone ID will be a logical value like "zone-1".
 
 ## Proposed Design
 
 The idea would be to use a custom label like `foundationdb.org/distribution-key`.
-As value we would use the process number modulo the `DesiredFaultDomains` and a prefix for the `BinPack` mode.
-In addition to that we have to change the `PodAntiAffinity` rule.
+As value we will use a prefix for the `BinPack` mode and a value defining in which zone the process should be running.
+We will also set the `FDB_ZONE_ID` environment variable on the Pod level, the default value of this field is currently the hostname.
+In addition to that we change the `PodAntiAffinity` rule.
 The following would be an example of the `affinity` term for a Pod.
-The example assumes that process number modulo `DesiredFaultDomains` is `0` and we have set `BinPackAll`:
+The example assumes that process should be running in zone `0` and we have set `BinPackAll`:
 
 ```yaml
   affinity:
@@ -43,35 +51,43 @@ The example assumes that process number modulo `DesiredFaultDomains` is `0` and 
           labelSelector:
             matchLabels:
               foundationdb.org/distribution-key: "all-0"
+              foundationdb.org/fdb-cluster-name: example-cluster
           topologyKey: fault_domain
         weight: 1
     # AntiAffinity to not schedule the Pod on the same fault domain where other Pods with a different distribution key are running
     podAntiAffinity:
+      # The required statement is per default only allowed on the hostname:
+      # https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#an-example-of-a-pod-that-uses-pod-affinity
       preferredDuringSchedulingIgnoredDuringExecution:
       - podAffinityTerm:
           labelSelector:
             matchExpressions:
-              key: "foundationdb.org/distribution-key"
+            - key: "foundationdb.org/distribution-key"
               operator: NotIn
               values:
               - "all-0"
+            - key: "foundationdb.org/fdb-cluster-name"
+              operator: In
+              values:
+              - "example-cluster"
           topologyKey: fault_domain
         weight: 1
 ```
 
-The additional label and the `affinity` term enable the operator to spread Pods across fault domains.
-If `Required` is set to false this will be on a best effort base and can be useful if a fault domain is full but the Pod should be scheduled anyway.
+The additional label and the `affinity` term enable the operator to spread Pods across physical fault domains.
 The following code snippet shows a possible implementation of the required structs:
 
 ```golang
 type BinPack string
 
 const (
-    // BinPackAll defines that processes independent of their class should be bin packed
+    // BinPackAll defines that processes independent of their class should be bin packed.
     BinPackAll BinPack = "All"
-    // BinPackStateful defines that stateful processes should be bin packed and additionally stateless processes
+    // BinPackStateful defines that stateful processes should be bin packed and additionally stateless processes.
+    // The prefix will be either "stateful" or "stateless".
     BinPackStateful BinPack = "Stateful"
-    // BinPackClass only processes of the same class will be bin packed
+    // BinPackClass only processes of the same class will be bin packed.
+    // As prefix we will use the according process class.
     BinPackClass BinPack = "Class"
 )
 
@@ -81,8 +97,8 @@ type DistributionConfig struct {
     // Default: false
     Enabled *bool
     // DesiredFaultDomains desfines the number of desired fault domain.
-    // Should be greater than 0 is fault domain distribution is enabled.
-    // Default: 0
+    // Must be greater than 0 if fault domain distribution is enabled.
+    // Default: 1
     DesiredFaultDomains *int
     // BinPack defines what processes should be bin packed to the fault domains.
     // Default: "Class"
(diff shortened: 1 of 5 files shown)


docs/design/bin_pack_fault_domains.md
```diff
@@ -0,0 +1,126 @@
...
+## Current Implementation
+
+The current implementation adds a `preferredDuringSchedulingIgnoredDuringExecution` `PodAntiAffinity` to every Pod [code](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/internal/pod_models.go#L308-L334).
+The label selector will contain the process class and the cluster name, to ensure processes with the same class are distributed across different fault domains is possible.
+A user can define additional `PodAntiAffinity` but can't prevent the operator from createn the default `PodAntiAffinity`.
+The only expection is the specival fault domain `foundationdb.org/none` and `foundationdb.org/kubernetes-cluster`.
+
+## Proposed Design
+
+The idea would be to use a custom label like `foundationdb.org/distribution-key`.
```
johscheuer, line 31: I think an additional label like `foundationdb.org/desird-fault-domain` could be useful for detecting changes in `DesiredFaultDomains` and triggering the replacement.

docs/design/bin_pack_fault_domains.md
```diff
@@ -0,0 +1,126 @@
...
+```golang
+type BinPack string
+
+const (
+    // BinPackAll defines that processes independent of their class should be bin packed
+    BinPackAll BinPack = "All"
+    // BinPackStateful defines that stateful processes should be bin packed and additionally stateless processes
+    BinPackStateful BinPack = "Stateful"
+    // BinPackClass only processes of the same class will be bin packed
+    BinPackClass BinPack = "Class"
```
johscheuer, line 75: I should add some additional context that we will use the process class as prefix

docs/design/bin_pack_fault_domains.md
```diff
@@ -0,0 +1,126 @@
...
+
+The current implementation adds a `preferredDuringSchedulingIgnoredDuringExecution` `PodAntiAffinity` to every Pod [code](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/internal/pod_models.go#L308-L334).
+The label selector will contain the process class and the cluster name, to ensure processes with the same class are distributed across different fault domains is possible.
+A user can define additional `PodAntiAffinity` but can't prevent the operator from createn the default `PodAntiAffinity`.
+The only expection is the specival fault domain `foundationdb.org/none` and `foundationdb.org/kubernetes-cluster`.
+
+## Proposed Design
+
+The idea would be to use a custom label like `foundationdb.org/distribution-key`.
+As value we would use the process number modulo the `DesiredFaultDomains` and a prefix for the `BinPack` mode.
```
brownleej, line 32: Could this produce scenarios where we had fewer logical fault domains than desired? For instance, if we had process groups with ids `storage-1`, `storage-2`, `storage-3`, with double replication and three fault domains desired, they would have the logical fault domains `1`, `2`, and `0`. If we then replaced `storage-2`, we would get the process groups `storage-1`, `storage-3`, `storage-4`, with logical fault domains `1`, `0`, and `1`.

docs/design/bin_pack_fault_domains.md
```diff
@@ -0,0 +1,126 @@
...
+
+// DistributionConfig
+type DistributionConfig struct {
+    // Enabled defines if the binpacking is enabled or not.
+    // Default: false
+    Enabled *bool
+    // DesiredFaultDomains desfines the number of desired fault domain.
+    // Should be greater than 0 is fault domain distribution is enabled.
+    // Default: 0
+    DesiredFaultDomains *int
```
brownleej, line 86: Why does this default to 0 if the 0 is an invalid value when the bin packing is enabled? Could we set it based on the replication mode by default?

docs/design/bin_pack_fault_domains.md
```diff
@@ -0,0 +1,126 @@
...
+    BinPack *BinPack
+    // Required defines if the affinity terms are required or preffered.
+    // Default: false
+    Required *bool
+}
+```
+
+The operator won't guarantee that the fault domains are equally used.
+Depending on the replacements for unreachable Pods we could use one fault domain more than another (or depending on the capacity),
+We also don't guarantee that we use exactly the number of `DesiredFaultDomains` e.g. when we spawn less processes than `DesiredFaultDomains` or when a fault domain is out of capacity and we are not using the `Required` with `true`.
```
brownleej, line 98: If the `Required` setting is true and a fault domain is out of capacity, I would expect that we would automatically spread to another fault domain. Is that correct?

docs/design/bin_pack_fault_domains.md
```diff
@@ -0,0 +1,126 @@
...
+We also don't guarantee that the operator replaces any Pods that violate any constraint and we are not using the distribution with `Required` with `true` (`Required` with `true` would only ensure that we are not hurting the constraint).
+
+A change to `DesiredFaultDomains` will lead to a migration of all existing Pods in order to honor the new affinity term.
```
brownleej, line 101: How will we detect that this migration is necessary?

docs/design/bin_pack_fault_domains.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Reducing the number of DesiredFaultDomains
+
+If we currently use `DesiredFaultDomains` and we have `4` fault domains `f0`, `f1`, `f2` and `f3` with a Pod in each fault domain and we reduce `DesiredFaultDomains` to `2` this will result in the following steps:
```
brownleej, line 105: I worry that the term "fault domain" is overloaded here. Is it referring to the physical fault domains (i.e. the things that are used as the topology key in the affinity rules) or the logical fault domains (i.e. the things used as the match expression in the affinity rules)?

docs/design/bin_pack_fault_domains.md
```diff
@@ -0,0 +1,126 @@
...
+* The user should be able to decide if the number of logical fault domains is a requirement (default should be optional)
+
+## Current Implementation
+
+The current implementation adds a `preferredDuringSchedulingIgnoredDuringExecution` `PodAntiAffinity` to every Pod [code](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/internal/pod_models.go#L308-L334).
+The label selector will contain the process class and the cluster name, to ensure processes with the same class are distributed across different fault domains is possible.
+A user can define additional `PodAntiAffinity` but can't prevent the operator from createn the default `PodAntiAffinity`.
+The only expection is the specival fault domain `foundationdb.org/none` and `foundationdb.org/kubernetes-cluster`.
+
+## Proposed Design
```
brownleej, line 29: This doesn't mention how the zone field is populated in the start commands. Is that intended to remain the same, or do we want to start using the logical fault domains as the zones for the database?

docs/design/bin_pack_fault_domains.md
brownleej: As an alternative, maybe we could look at how many current non-removed pods are in each logical fault domain, and assign each new pod to whichever fault domain has the fewest pods currently.

brownleej: That could also make the process of detecting changes to the desired number of fault domains easier, because we could look at fault domains that are overloaded compared to the desired distribution and trigger replacements for processes in those overloaded fault domains.

docs/design/bin_pack_fault_domains.md
brownleej: We might also want to talk about how this could replace of the special fault domains that we have today, `foundationdb.org/none` and `foundationdb.org/kubernetes-cluster`. That overlaps with the question of whether we want to use the logical fault domains for replication in the database, instead of the physical fault domains.

brownleej (changes requested):

docs/design/bin_pack_fault_domains.md
brownleej: I'd also like to see some more analysis of the possibility for deadlocks in the new design that aren't present in the old design. For instance, if a Kubernetes cluster has 6 physical fault domains and the FDB cluster is configured to desire 5, are we going to see replacements deadlock?

johscheuer: The idea was to maintain the zone id (same behaviour we use currently). I can add an additional section to more information on that. I can add some additional examples here. I don't see how that should result in a deadlock (depending if you set the `required` field or not).

docs/design/bin_pack_fault_domains.md
johscheuer: Yes, that could happen and is stated below `The operator won't guarantee that the fault domains are equally used.` which in that case would lead to a non-functional cluster. I would rather put that as a limitation for that design. I think about the proposal with the assignment of the fault domains and add some more information in the design.

docs/design/bin_pack_fault_domains.md
johscheuer: Using a different default from 0 based on the replication mode makes sense. I'll rephrase that and add more information. I didn't want to change the FDB zone ID in this design and would leave it to another design to change that. I think about that and if I get a good idea I'll add that here.

docs/design/bin_pack_fault_domains.md
johscheuer: That's not correct. In that case the Pod would stay in pending and is not able to be scheduled. Otherwise the required setting would be a little bit meaningless. The required setting only defines if we use a `preferredDuringSchedulingIgnoredDuringExecution` or a `requiredDuringSchedulingIgnoredDuringExecution` with the latter and a fault domain out of resources the Pod is stuck in Pending until resources will be available.

docs/design/bin_pack_fault_domains.md
johscheuer: By adding an additional label: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/942/files#r721072454

docs/design/bin_pack_fault_domains.md
johscheuer: In this case a logical fault domain == physical fault domain with the difference that a logical fault domain will be "bound" to a physical fault domain once the scheduler schedules the first Pod into the domain. I can add some more context on the difference between those in the design doc and maybe I come up with a better name.

docs/design/bin_pack_fault_domains.md
brownleej: Does the `required` setting affect the constraint type for the affinity toward the same logical fault domain, the anti-affinity against a different logical fault domain, or both?

docs/design/bin_pack_fault_domains.md
brownleej: Could that lead to scenarios where we are migrating more than we have to?

docs/design/bin_pack_fault_domains.md
brownleej: My assumption was that the `required` setting would only affect the anti-affinity rules, and that the affinity rules would always be preferred in order to prevent running out of capacity.

docs/design/bin_pack_fault_domains.md
brownleej: The deadlocks that you're alluding to when the `required` field is set are the kind of things I was talking about. If there's a potential for things to get stuck with some configurations, we should call that out.

docs/design/bin_pack_fault_domains.md
```diff
@@ -0,0 +1,126 @@
...
+
+```yaml
+  affinity:
+    # Affinity to schedule Pods with the same distribution key together
+    podAffinity:
+      preferredDuringSchedulingIgnoredDuringExecution:
+      - podAffinityTerm:
+          labelSelector:
+            matchLabels:
+              foundationdb.org/distribution-key: "all-0"
```
brownleej, line 45: Should these match labels should also include the general match labels for the cluster? i.e. `foundationdb.org/fdb-cluster-name: example-cluster`

docs/design/bin_pack_fault_domains.md
```diff
@@ -0,0 +1,126 @@
...
+          topologyKey: fault_domain
+        weight: 1
+    # AntiAffinity to not schedule the Pod on the same fault domain where other Pods with a different distribution key are running
+    podAntiAffinity:
+      preferredDuringSchedulingIgnoredDuringExecution:
```
brownleej, line 50: It may be worth calling out that `requiredDuringScheduling` anti-affinity is only allowed on the hostname topology key by default: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#an-example-of-a-pod-that-uses-pod-affinity.

docs/design/bin_pack_fault_domains.md
johscheuer: Updated.

docs/design/bin_pack_fault_domains.md
johscheuer: Yes, let me update that phrase and create an example.

docs/design/bin_pack_fault_domains.md
```diff
@@ -0,0 +1,197 @@
...
+            matchLabels:
+              foundationdb.org/distribution-key: "all-0"
+              foundationdb.org/fdb-cluster-name: example-cluster
+          topologyKey: fault_domain
+        weight: 1
+    # AntiAffinity to not schedule the Pod on the same fault domain where other Pods with a different distribution key are running
+    podAntiAffinity:
+      # The required statement is per default only allowed on the hostname:
+      # https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#an-example-of-a-pod-that-uses-pod-affinity
+      preferredDuringSchedulingIgnoredDuringExecution:
```
johscheuer, line 61: This must be a required.

docs/design/bin_pack_fault_domains.md
```diff
@@ -0,0 +1,197 @@
...
+    // DesiredFaultDomains desfines the number of desired fault domain.
+    // Must be greater than 0 if fault domain distribution is enabled.
+    // Default: 1
+    DesiredFaultDomains *int
+    // BinPack defines what processes should be bin packed to the fault domains.
+    // Default: "Class"
+    BinPack *BinPack
+    // Required defines if the affinity terms are required or preffered.
+    // Default: false
+    Required *bool
```
johscheuer, line 108: Remove this one.

docs/design/bin_pack_fault_domains.md
```diff
@@ -0,0 +1,197 @@
...
+)
+
+// DistributionConfig
+type DistributionConfig struct {
+    // Enabled defines if the binpacking is enabled or not.
+    // Default: false
+    Enabled *bool
+    // DesiredFaultDomains desfines the number of desired fault domain.
+    // Must be greater than 0 if fault domain distribution is enabled.
+    // Default: 1
```
brownleej, line 101: Based on the conversation below, it seemed like we wanted this default to be based on the replication mode. Is that still what you're thinking?

docs/design/bin_pack_fault_domains.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Scheduling and Eviction Policy
+
+For the implementation we need two additional steps:
+
+- The `AddProcessGroups` reconcile loop must be aware of the fault domains.
+- The `ReplaceMisconfiguredProcessGroups` reconcile loop must be aware of the fault domains.
+
+We will only count process groups that are not marked for removal.
+We can calculate the minimum and maximum number of Pods per fault domain with the following logic:
```
brownleej, line 127: Is this calculated separately for every process class? I'm a little confused about how that interacts with the different bin packing strategies.

docs/design/bin_pack_fault_domains.md
```diff
@@ -0,0 +1,197 @@
...
+              - "example-cluster"
+          topologyKey: fault_domain
+        weight: 1
+```
+
+The additional label and the `affinity` term enable the operator to spread Pods across physical fault domains.
+The following code snippet shows a possible implementation of the required structs:
+
+```golang
+type BinPack string
```
brownleej, line 81: Do we have use cases defined for different bin packing strategies? Today we're always bin packing by class, and I'm not clear on the advantages of different strategies.

docs/design/bin_pack_fault_domains.md
```diff
@@ -0,0 +1,197 @@
...
+- `physical fault domain`: This refers to a fault domain that is available in Kubernetes e.g. a machine running the Kubernetes worker or a higher level fault domain like a rack, data hall etc.
+  This will reflect an underlying physical fault domain.
+- `logical fault domain`: The logical fault domain is not something that refers directly to a physical fault domain and can also be spread across multiple physical fault domains.
+  The idea is to have a way to group some processes of a FoundationDB cluster together with the same zone ID, this zone ID will be a logical value like "zone-1".
+
+## Proposed Design
+
+The idea would be to use a custom label like `foundationdb.org/distribution-key`.
+As value we will use a prefix for the `BinPack` mode and a value defining in which zone the process should be running.
+We will also set the `FDB_ZONE_ID` environment variable on the Pod level, the default value of this field is currently the hostname.
```
brownleej, line 40: Is this implying that we will set the FDB_ZONE_ID environment variable _to the distribution key_? That's my assumption, but it may be good to make that explicit if so.

docs/design/bin_pack_fault_domains.md
```diff
@@ -0,0 +1,197 @@
...
+    // BinPackClass only processes of the same class will be bin packed.
+    // As prefix we will use the according process class.
+    BinPackClass BinPack = "Class"
+)
+
+// DistributionConfig
+type DistributionConfig struct {
+    // Enabled defines if the binpacking is enabled or not.
+    // Default: false
+    Enabled *bool
```
brownleej, line 98: In the long term, do you think that we want to default this to true, and possibly deprecate the non-bin-packed behavior entirely?

brownleej (changes requested):

docs/design/bin_pack_fault_domains.md
johscheuer: I think that makes sense and I'll update the doc.

docs/design/bin_pack_fault_domains.md
brownleej: Conclusion from discussion: We don't need different bin packing options with the current state of the design.

docs/design/bin_pack_fault_domains.md
```diff
@@ -0,0 +1,197 @@
...
+
+Initial state:
+
+<img src="./imgs/fault_domain_multi_host.svg">
+
+End state:
+
+<img src="./imgs/reduce_fault_domains.svg">
+
+### Increasing the number of DesiredFaultDomains
```
johscheuer, line 174: Add additional example for dead-locks.

johscheuer (commented):
Once thing that I forgot to mention is that we might want to add the fault domain to the current process group struct, but that would also be an implementation detail.

brownleej (approved):
👍

PR #946, opened 2021-10-03
Specify the complete registry

johscheuer:
# Description

Specify the complete registry to allow builds with container runtimes/builders other than `docker`.

## Type of change

# Discussion

Docker has hardcoded that search path but we should make it explicit to make it easier for other tools to use the Dockerfile.

# Testing

local.

# Documentation

-

# Follow-up

-

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,9 @@
-FROM foundationdb/foundationdb:6.2.30 as fdb62
-FROM foundationdb/foundationdb:6.1.13 as fdb61
-FROM foundationdb/foundationdb:6.3.10 as fdb63
+FROM docker.io/foundationdb/foundationdb:6.2.30 as fdb62
+FROM docker.io/foundationdb/foundationdb:6.1.13 as fdb61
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,9 +1,9 @@
-FROM foundationdb/foundationdb:6.2.30 as fdb62
-FROM foundationdb/foundationdb:6.1.13 as fdb61
-FROM foundationdb/foundationdb:6.3.10 as fdb63
+FROM docker.io/foundationdb/foundationdb:6.2.30 as fdb62
+FROM docker.io/foundationdb/foundationdb:6.1.13 as fdb61
+FROM docker.io/foundationdb/foundationdb:6.3.10 as fdb63
 
 # Build the manager binary
-FROM golang:1.16.8 as builder
+FROM docker.io/library/golang:1.16.8 as builder
```
johscheuer, line 6: This is an "official" image so it's inside `docker.io/library` most builder tools can resolve that but I think it makes sense to make it explicit here.

brownleej (approved):
👍

Issue #811, opened 2021-07-02
Automatic replacement shouldn't count terminating process groups


johscheuer:
Currently the automatic replacement logic counts how many process groups are in a `remove` state. I would argue that process groups that are marked as `remove` and in a `Terminating` state are should be ignored and not be counted against the current replacements since these resources are also in the tear down state and probably just a Kubernetes resource is slow.

brownleej:
It looks to me like it only counts process groups that flagged with `remove` and not flagged with `excluded`, which should capture the common cases of resources that are stuck in terminating. We might want to think about how we use the `excluded` flag and the `exclusionSkipped` flag when a process group is in the `instancesToRemoveWithoutExclusion` list.

johscheuer:
I think there are 2 things we have to add:

 1. Check for the `exclusionSkipped` flag if set ignore (same like the process group is excluded)
 2. Ignore all process groups that are marked for removal and don't have an address assigned (e.g. pending state)

Since the operator never marks a process group for removal if the process group dosen't have an address assigned this case is rather rare but could happen if a human operator marks that process group for removal without setting `exclusionSkipped`.

Issue #791, opened 2021-06-28
Allow to remove process groups without an address 


johscheuer:
We should add a setting to allow the controller to remove process groups without an address. In the current design of the controller a process group without an address means that the underlying Pod was never scheduled (so it should never receive any data) that means it should be safe to remove the process group without exclusion. There is a potential race condition where we want to remove the process group without an address and during that time the underlying Pod get's scheduled. To reduce this risk we could wait for a safety period to assume the Pod will never we scheduled additionally the duration should be relatively short (a few seconds) so that only few data should be replicated to that new process group. As an further safety check we could only remove process groups without an address if the cluster is in a fully replicated state to ensure we don't remove any data. That is only an additional layer of safety since an unscheduled process group without an address has never been running.

johscheuer:
Fixed in https://github.com/FoundationDB/fdb-kubernetes-operator/pull/947

PR #947, opened 2021-10-06
Allow to replace pg without address

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/811 + https://github.com/FoundationDB/fdb-kubernetes-operator/issues/791

## Type of change

# Discussion

We only replace process groups without an address if the cluster has the desired fault tolerance and is available otherwise we will skip that process group.

# Testing

Unit + local testing.

# Documentation

I just noticed we never mentioned in the docs that we skip process groups without an address so nothing to update here.

# Follow-up

-

### DIFF
--- a/controllers/replace_failed_process_groups.go
+++ b/controllers/replace_failed_process_groups.go
@@ -1,5 +1,5 @@
 /*
- * replace_failed_pods.go
+ * replace_failed_process_groups.go
  *
  * This source file is part of the FoundationDB open source project
  *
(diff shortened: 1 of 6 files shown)


controllers/replace_failed_process_groups.go
```diff
@@ -76,13 +85,26 @@ func chooseNewRemovals(cluster *fdbtypes.FoundationDBCluster) bool {
...
+				// and is available.
+				hasDesiredFaultTolerance, err := internal.HasDesiredFaultTolerance(adminClient, cluster)
+				if err != nil {
+					log.Error(err, "Could not fetch if cluster has desired fault tolerance")
+					continue
+				}
+
+				if !hasDesiredFaultTolerance {
+					log.Info(
+						"Replace instance with missing address",
```
brownleej, line 98: If I'm interpreting the code correctly, in this branch we're choosing not to replace the instance because of the fault tolerance issues. Can we make the log message more descriptive so that's clear to the users?

brownleej (changes requested):
Small suggestion on log messaging. Other than that this looks good to me.

controllers/replace_failed_process_groups.go
johscheuer: Yes, thanks for the catch 👍

Issue #400, opened 2020-11-26
Support rolling update of pods on migration


simenl:
A pod update that requires migration, e.g. any changes to volumes, requires a migration. A migration requires a full set of the updated pods to run concurrently with the previous set of pods. Often this requires double the cpu, memory, and storage.

In a resource constrained cluster, these updates may not be possible to perform, due to the increase in resource required.
An option to do a rolling update would make this possible, given that the cluster can support the surge of one extra pod.

brownleej:
Yeah, this seems like a reasonable feature. Maybe we could implement this as a limit on how many pods we can replace at once.

johscheuer:
Like proposed in #404 it makes sense to add `maxSurge` which would behave like [maxSurge](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#max-surge) of a `Deployment`.

Issue #805, opened 2021-07-01
Limiting the number of in-flight replacements across clusters


brownleej:
We have operations that require a migration, such as changing the spec for the volume claims. These operations are going to require a temporary spike in resource usage between the time when we add new pods and remove the old pods, which can take a while. If someone updates the specs for multiple clusters in a way that triggers this change, they could exhaust their resource quota or their physical resources, putting clusters in a bad state. It might be useful to have an option to limit the number of in-flight replacements across clusters, to avoid bursting the overall resource usage too high compared to the steady state.

johscheuer:
Related to: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/400

johscheuer:
I would favour to implement a solution for https://github.com/FoundationDB/fdb-kubernetes-operator/issues/400 and then the (human) operator is responsible for setting a sensible value there. That would be more or less the same like Deployments etc. work today (there is no global or namespace coordination between them). The whole problem get's more interesting if we think about the FDB operator operating in a global mode.

PR #952, opened 2021-10-16
Allow rolling replacements

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/400 this is a simple implementation to limit the number of concurrent replacements.

## Type of change

# Discussion

If someone has a better name I'm happy to change it :)

# Testing

Unit + local 

# Documentation

Added.

# Follow-up

We could add some more "complex" logic to ensure that all Pods are up or adding the same functionality a Deployment has today. I think for the current state that this would be a slight over kill :)


api/v1beta1/foundationdbcluster_types.go
```diff
@@ -1112,6 +1114,16 @@ type FoundationDBClusterAutomationOptions struct {
...
+
+	// MaxConcurrentMisconfiguredReplacements defines how many process groups can be concurrently
+	// replaced if they are misconfigured. If the value will be set to 0 this will block replacements
+	// and these misconfigured Pods must be replaced manually or by another process. For each reconcile
+	// loop the operator calculates the maximum number of possible replacements by taken this value as the
+	// upper limit and removes all ongoing replacements that have not finished. Which means if the value is
+	// set to 5 and we have 4 ongoing replacements (process groups marked with remove but not excluded) the
+	// operator is allowed to replace on further process group.
+	// +kubebuilder:validation:Minimum=0
+	MaxConcurrentMisconfiguredReplacements *int `json:"maxConcurrentMisconfiguredReplacements,omitempty"`
```
brownleej, line 1126: As you alluded to in the PR body, I think this name is going to be confusing to people. We could call them "discretionary" replacements, or maybe we could just call this `MaxConcurrentReplacements` and call out in the documentation that it only applies to replacements that the operator chooses to do.

internal/replacements/removals.go
```diff
@@ -0,0 +1,104 @@
...
+			removalCount++
+		}
+	}
+
+	return maxReplacements - removalCount
+}
+
+// ChooseNewRemovals flags failed processes groups for removal and returns an indicator
+// of whether any processes were thus flagged.
+func ChooseNewRemovals(log logr.Logger, cluster *fdbtypes.FoundationDBCluster, adminClient fdbadminclient.AdminClient) bool {
```
brownleej, line 51: It looks like this function is doing a different kind of selection than the ChooseRemovals action. Is that right? If so, we might want to rename one of them to avoid confusion.

brownleej: I'm also not clear on how we're using the term "replacement" and "removal" in these new functions. It seems like `ChooseNewRemovals` and `HasReplacements` are doing the same kind of thing, but for different root problems. It also seems odd to have one labelled `ChooseX` and the other `HasX` if both functions are marking things for removal.

brownleej (changes requested):

brownleej:
I don't think this will fix #805, since that issue specifically calls for a limit across clusters. I'll add more thoughts in that issue and we can discuss it more from there.

brownleej (approved):
👍

Issue #379, opened 2020-11-13
Unify naming in the operator


johscheuer:
Currently we have multiple different names for sometimes the same thing and sometimes for different things and also we don't use them consistently over the code base and the docs. Just to mention some of the naming that should be unified:

- Instance
- Pod
- Process

This is pretty confusing for newcomers and people that work on the code base. Also we should define what each name means.

Issue #927, opened 2021-09-21
Create a setting to use `process_group_id`


johscheuer:
We should have a setting to allow users to make use of `locality_process_group_id` instead of `locality_instance_id`. This allows a user to specify to use the new `locality_process_group_id` and allows a smooth transition before we want to move to `1.0` and remove all occurrences of `instance-id`.

Related code part: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/internal/monitor_conf.go#L120

PR #967, opened 2021-10-28
Unify naming

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/379

## Type of change

# Discussion

There are still some occurrences of `FDB_INSTANCE_ID` and the locality field (`FDBLocalityInstanceIDKey`)which will be refactored in another PR. 

# Testing

Unit tests + local test

# Documentation

Changed to match the refactoring.

# Follow-up

 https://github.com/FoundationDB/fdb-kubernetes-operator/issues/927


kubectl-fdb/cmd/remove.go
```diff
@@ -1,5 +1,5 @@
 /*
- * remove.go
```
brownleej, line 2: Since we're not renaming the file, I don't think we want to change this in the header.

brownleej (changes requested):
I have a couple of small suggestions. Otherwise LGTM

kubectl-fdb/cmd/remove.go
johscheuer: Thanks, that was the refactoring. Let me fix that.

brownleej (approved):
👍

ltsampros (approved):

PR #969, opened 2021-10-29
makefile: add a deps target to fetch all needed Go binaries

ltsampros:
# Description

A Makefile level macro to fetch needed go dependencies during the code process. As long as a sane version of go/gmake are installed in the path, this will fetch needed dependencies. 

If the files already exist, the makefile will skip those targets completely.

I have also modified the default make goal to `all` (it was previously unset so it was picking up the first target defined in the file).

## Type of change
- Other

# Discussion
After upgrading to go 1.17 we should change `go get` to `go install`

# Testing
Local testing.

# Documentation
Not needed but it does improve on the local development process.

# Follow-up
- After upgrading to go 1.17 we should change `go get` to `go install`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,8 @@
+.DEFAULT_GOAL=all
 
 # Image URL to use all building/pushing image targets
 IMG ?= fdb-kubernetes-operator:latest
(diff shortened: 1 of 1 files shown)


brownleej (approved):
👍

Issue #950, opened 2021-10-12
Using a PodNodeSelector causes endless pods to be created


benzim:
[This change](https://github.com/FoundationDB/fdb-kubernetes-operator/pull/467) to replace a pod when the nodeSelector is changed is causing the operator to create pods infinitely if a [PodNodeSelector](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#podnodeselector) is present. This is because the PodNodeSelector applies a nodeSelector to every pod before it is deployed so when the operator checks if the nodeSelector for the deployed pod matches the nodeSelector in the pod template they don't match. This causes the operator to mark every pod for replacement and create new pods which then also have a nodeSelector applied by the PodNodeSelector.

Steps to reproduce:
1. Create a namespace with the `scheduler.alpha.kubernetes.io/node-selector` annotation like the one below.
```
apiVersion: v1
kind: Namespace
metadata:
  annotations:
    scheduler.alpha.kubernetes.io/node-selector: node-role.kubernetes.io/worker=
  name: test-namespace
```
2. Deploy the operator and FoundationDBCluster to the namespace.
3. Operator keeps creating pods.

Is this something that could be fixed or will the user have to make sure that the nodeSelector in the FoundationDBCluster pod template matches the PodNodeSelector?

johscheuer:
> Do you think it would be possible to detect the bad state and replace the pod then instead of replacing on a nodeSelector change?

That should be possible if we would check the node the Pod is running on. so far we tried to not use any additional knowledge in the operator about nodes to prevent that the operator requires access to nodes.

benzim:
The issue says that it will leave the pod in a state where it can't schedule. Would it be possible to tell that from the pod status/conditions?

benzim:
@johscheuer I tested it and opened this PR. Could you review it? https://github.com/FoundationDB/fdb-kubernetes-operator/pull/973

johscheuer:
@benzim I think we can close this issue now? The solution might not be perfect but I think it's the best we can do currently.

benzim:
I agree. Closing this issue.

PR #973, opened 2021-11-05
Only replace on changed node selector if pod spec hash has changed

benzim:
# Description

This is a fix for https://github.com/FoundationDB/fdb-kubernetes-operator/issues/950.

## Type of change

# Discussion

This will trigger a replacement if any part of the pod template spec is changed if there is a [PodNodeSelector](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#podnodeselector). Which isn't ideal but is better than the previous behavior of the pods being replaced endlessly.

# Testing

I ran unit tests and manually deployed the operator and cluster in a namespace with a PodNodeSelector.

### DIFF
--- a/internal/replacements/replacements.go
+++ b/internal/replacements/replacements.go
@@ -190,9 +190,16 @@ func instanceNeedsRemoval(cluster *fdbtypes.FoundationDBCluster, pod *corev1.Pod
 
 	expectedNodeSelector := cluster.GetProcessSettings(processClass).PodTemplate.Spec.NodeSelector
 	if !equality.Semantic.DeepEqual(pod.Spec.NodeSelector, expectedNodeSelector) {
-		logger.Info("Replace instance",
-			"reason", fmt.Sprintf("nodeSelector has changed from %s to %s", pod.Spec.NodeSelector, expectedNodeSelector))
-		return true, nil
+		specHash, err := internal.GetPodSpecHash(cluster, processClass, idNum, nil)
+		if err != nil {
+			return false, err
+		}
+
+		if pod.ObjectMeta.Annotations[fdbtypes.LastSpecKey] != specHash {
+			logger.Info("Replace instance",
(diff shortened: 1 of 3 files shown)


internal/replacements/replacements.go
```diff
@@ -189,13 +189,8 @@ func instanceNeedsRemoval(cluster *fdbtypes.FoundationDBCluster, pod *corev1.Pod
...
 
 	expectedNodeSelector := cluster.GetProcessSettings(processClass).PodTemplate.Spec.NodeSelector
-	if !equality.Semantic.DeepEqual(pod.Spec.NodeSelector, expectedNodeSelector) {
-		logger.Info("Replace instance",
-			"reason", fmt.Sprintf("nodeSelector has changed from %s to %s", pod.Spec.NodeSelector, expectedNodeSelector))
-		return true, nil
-	}
 
-	if cluster.Spec.UpdatePodsByReplacement {
+	if cluster.Spec.UpdatePodsByReplacement || !equality.Semantic.DeepEqual(pod.Spec.NodeSelector, expectedNodeSelector) {
```
johscheuer, line 193: Can we still leave the node selector check in a different if statement? That makes it easier to see why the Pod was replaced with this change we will only see "specHash has changed from ... to ..." but we can be more precise by logging that the node selector doesn't match.

johscheuer (changes requested):

johscheuer (commented):
The changes look good to me, could you add an additional test case here: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/internal/replacements/replacements_test.go#L238-L255 where we change the nodeSelector of the Pod. That shouldn't result in a replacement with the new changes.

benzim:
@johscheuer Hopefully this isn't out of scope for this PR, but the replacements tests weren't running when I ran `make test` so I added some stuff to get the replacements tests running and passing.

johscheuer (approved):
LGTM 👍 Thanks for fixing the test suite.

Issue #971, opened 2021-10-29
For exclusions we should ignore timeout errors


johscheuer:
For any exclusion operation we should ignore the `Specified timeout reached -- exiting...` error. This can happen rather often when we try to exclude an instance with a non-trivial amount of data and we don't want that this exclusion blocks further execution. We should continue with the other sub reconcilers but set [delayedRequeue](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/controllers/controllers.go#L93) to true.

PR #974, opened 2021-11-08
Make exclusions not block in sub reconciler

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/971

## Type of change

# Discussion

Moving forward allows the operator to actually change coordinators if required or do any other operator and is not blocked on a single replacement/exclusion. Since the operator will only replace process groups that are excluded (or we skip the exclusion) it's safe to move forwards.

# Testing

Unit + local

# Documentation

-

# Follow-up

None


brownleej:
LGTM. What is our next step to move this out of draft?

brownleej (approved):
👍

PR #975, opened 2021-11-09
Add support for deletion mode

johscheuer:
# Description

Adding a deletion mode to allow the user to specify how Pods should be delete if they need an update.

## Type of change

# Discussion

-

# Testing

Unit + local.

# Documentation

Updated.

# Follow-up

As a next step we should add this mode to the removal (I create a radar for that)


controllers/update_pods.go
```diff
@@ -140,30 +144,84 @@ func (updatePods) reconcile(r *FoundationDBClusterReconciler, context ctx.Contex
...
-		hasLock, err := r.takeLock(cluster, "updating pods")
-		if !hasLock {
-			return &requeue{curError: err}
+	if deletionMode == fdbtypes.DeletionModeZone {
+		// Default case is zone
+		for zoneName, zoneInstances := range updates {
+			zone = zoneName
+			deletions = zoneInstances
+			// Fetch the first zone and stop
+			return zone, deletions, nil
```
johscheuer, line 186:
```suggestion
			// Fetch the first zone and stop
			return zoneName, zoneInstances, nil
```

controllers/update_pods.go
```diff
@@ -140,30 +144,84 @@ func (updatePods) reconcile(r *FoundationDBClusterReconciler, context ctx.Contex
...
+	if deletionMode == fdbtypes.DeletionModeProcessGroup {
+		for _, zoneInstances := range updates {
+			if len(zoneInstances) < 1 {
+				continue
+			}
+			pod := zoneInstances[0]
+			zone = pod.Name
+			deletions = append(deletions, pod)
+			// Fetch the first pod and delete it
+			return zone, deletions, nil
```
johscheuer, line 176:
```suggestion
			pod := zoneInstances[0]
			// Fetch the first pod and delete it
			return pod.Name, []*corev1.Pod{pod}, nil
```

controllers/update_pods.go
```diff
@@ -140,30 +144,84 @@ func (updatePods) reconcile(r *FoundationDBClusterReconciler, context ctx.Contex
...
+	if len(updates) == 0 {
+		return nil
+	}
+
+	return deletePodsForUpdates(context, r, cluster, adminClient, updates, logger)
+}
+
+func getPodsToDelete(deletionMode fdbtypes.DeletionMode, updates map[string][]*corev1.Pod) (string, []*corev1.Pod, error) {
+	var deletions []*corev1.Pod
+	var zone string
```
johscheuer, line 156: Those can be moved into the delete all clause.

controllers/update_pods.go
```diff
@@ -140,30 +144,78 @@ func (updatePods) reconcile(r *FoundationDBClusterReconciler, context ctx.Contex
...
-		logger.Info("Deleting pods", "zone", zone, "count", len(zoneInstances))
-		r.Recorder.Event(cluster, corev1.EventTypeNormal, "UpdatingPods", fmt.Sprintf("Recreating pods in zone %s", zone))
+	return "", nil, fmt.Errorf("unknown deletion mode: \"%s\"", deletionMode)
+}
 
-		err = r.PodLifecycleManager.UpdatePods(r, context, cluster, zoneInstances, false)
-		if err != nil {
+// deletePodsForUpdates will delete Pods with the specified deletion mode
+func deletePodsForUpdates(context context.Context, r *FoundationDBClusterReconciler, cluster *fdbtypes.FoundationDBCluster, adminClient fdbadminclient.AdminClient, updates map[string][]*corev1.Pod, logger logr.Logger) *requeue {
+	deletioNMode := r.PodLifecycleManager.GetDeletionMode(cluster)
```
brownleej, line 189:
```suggestion
	deletionMode := r.PodLifecycleManager.GetDeletionMode(cluster)
```

brownleej (approved):
I have a small suggestion on renaming a variable, but otherwise LGTM.

PR #976, opened 2021-11-10
Prevents the cluster creation from getting deadlocked when it cannot connect to the database before the processes have a cluster file.

brownleej:
# Description

This addresses a failure mode I've observed in cluster creation where an error in reconciliation at a certain phase will call cause the cluster to have a connection string in the FoundationDBCluster resource, but not in the config map or any of the pods.

## Type of change

# Discussion

We may want to be more resilient to failures to get the database status in general, rather than narrowly targeting this at cases where the database has not been configured.

# Testing

Manual testing?*

I had a live case of this problem in my test environment, and I verified that the fix allowed cluster creation to move forward.

No.

# Documentation

No.

# Follow-up

No.

No.

### DIFF
--- a/controllers/update_status.go
+++ b/controllers/update_status.go
@@ -82,7 +82,16 @@ func (updateStatus) reconcile(r *FoundationDBClusterReconciler, context ctx.Cont
 
 		databaseStatus, err = adminClient.GetStatus()
 		if err != nil {
-			return &requeue{curError: err}
+			if cluster.Status.Configured {
+				return &requeue{curError: err}
+			}
+			databaseStatus = &fdbtypes.FoundationDBStatus{
+				Cluster: fdbtypes.FoundationDBStatusClusterInfo{
+					Layers: fdbtypes.FoundationDBStatusLayerInfo{
+						Error: "configurationMissing",
+					},
+				},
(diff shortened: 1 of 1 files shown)


controllers/update_status.go
```diff
@@ -82,7 +82,16 @@ func (updateStatus) reconcile(r *FoundationDBClusterReconciler, context ctx.Cont
 
 		databaseStatus, err = adminClient.GetStatus()
 		if err != nil {
-			return &requeue{curError: err}
+			if cluster.Status.Configured {
+				return &requeue{curError: err}
+			}
+			databaseStatus = &fdbtypes.FoundationDBStatus{
```
johscheuer, line 88: I think that should be fine to assume the DB is not configured. One think that can happen, but shouldn't cause any issues, is that the DB get's configured again with the same settings.

Do you think it makes sense to move the creation of the status into a method to reduce the duplication?

johscheuer (approved):
LGTM 👍

controllers/update_status.go
brownleej: I think it's only a small amount of code, so it's fine to have a little duplication.

PR #980, opened 2021-11-12
moving FdbPodClient from internal to pkg

09harsh:
# Description

Moving the interface from `internal` to `pkg` module so that it can be referenced from other scopes as well. 

## Type of change

- Other

# Discussion

# Testing

Manual testing?*
not needed

no
# Documentation

not needed 

no

not needed

not needed 

# Follow-up

no

no

### DIFF
--- a/pkg/podclient/pod_client.go
+++ b/pkg/podclient/pod_client.go
@@ -1,0 +1,52 @@
+/*
+ * pod_client.go
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2018-2019 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


pkg/podclient/pod_client.go
```diff
@@ -0,0 +1,52 @@
+/*
+ * pod_client.go
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2018-2019 Apple Inc. and the FoundationDB project authors
```
johscheuer, line 6:
```suggestion
 * Copyright 2018-2021 Apple Inc. and the FoundationDB project authors
```

pkg/podclient/pod_client.go
```diff
@@ -0,0 +1,52 @@
...
+	corev1 "k8s.io/api/core/v1"
+)
+
+// FdbPodClient provides methods for working with a FoundationDB pod
+type FdbPodClient interface {
+	// GetCluster returns the cluster associated with a client
+	GetCluster() *fdbtypes.FoundationDBCluster
+
+	// GetPod returns the pod associated with a client
+	GetPod() *corev1.Pod
```
johscheuer, line 34: Before exposing we should remove those exposed methods. It looks like we use them only for the log statement, a better option would be to create the logger during the struct creation in the internal package.

johscheuer (commented):
I think we should remove the two methods that are not really used externally to keep the exposed methods to a minimum.

johscheuer (approved):
LGTM 👍

Issue #887, opened 2021-08-19
Deprecating custom images in the pod templates in favor of the new image configs


brownleej:
In the recent work on image configs (#878), we left in support for specifying images for built-in containers in the pod templates. In most cases, custom images in the pod templates can be directly translated into the image config model. This means that having images in the pod template is an alternative way of expressing the same thing, with less flexible semantics around upgrades. I think we should consider deprecating support for custom images in the templates and converge on a single model going forward.

The one area where image configs are not a direct equivalent is that the pod templates allow specifying different images for different process classes. I don't know if there is a real use case for this that is worth supporting with an alternate mechanism.

PR #981, opened 2021-11-13
Deprecate AllowTagOverride

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/887

## Type of change

# Discussion

One of the concerns in the issue was that the `ImageConfigs` are not allowing to change the image for only a specific process class. I think that this is out of scope and we currently don't want to support that. Once we identify the need we could also extend the `ImageConfigs` to include a process class.

# Testing

-

# Documentation

Was already documented.

# Follow-up

Remove the knob before 1.0


brownleej (approved):
👍

Issue #841, opened 2021-07-10
Allow the operator to only reconcile FoundationDBClusters with a given label


johscheuer:
We should add support for this by adding a `label-selector` field and test these changes. This should enable user to run multiple operators in the same namespace and each of them targeting a subset of clusters e.g. for reducing the blast radius and better scalability.

PR #984, opened 2021-11-15
Add support for label selector

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/841

## Type of change

# Discussion

The label selector is optional but allows users to operate multiple operators in the same namespace/cluster. 

# Testing

Unit + local.

# Documentation

Added.

# Follow-up

I create an issue to optimise the client cache.


brownleej (approved):
👍

PR #987, opened 2021-11-17
Support for the new FDB Kubernetes image

### LINKED ISSUE #923: Support for new FoundationDB image type
As part of the work on the [unified image](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/docs/design/unified_sidecar_and_process_monitor.md), we need to add support in the operator for the new image type.

brownleej:
# Description

This adds support for the new image that provides a Kubernetes-specific process monitor.

Resolves #923 

## Type of change

# Discussion

No.

# Testing

Manual testing?*

I've added unit tests, and ran tests locally with the new image types.

No.

# Documentation

I added information about configuring the image types to the user manual, as well as more technical detail about the different interaction models for the different image types.

# Follow-up

We need to support selecting an IP family with the new images, which will likely require changes in the image as well.

We need to support using the new image for the backup agents.

We need additional issues on moving to the new image as the default in the tests, and in real configs.

I think we should also follow up on eliminating the differences between the monitor conf for different storage server densities in the new image.

I think we should look for ways to improve the process of transitioning between user IDs, since the new images have a different user ID.


config/tests/base/cluster.yaml
```diff
@@ -3,7 +3,7 @@ kind: FoundationDBCluster
 metadata:
   name: test-cluster
 spec:
-  version: 6.2.30
+  version: 6.3.13
```
johscheuer, line 6: Should we update to `6.3.22`?

controllers/cluster_controller_test.go
```diff
@@ -268,6 +267,34 @@ var _ = Describe(string(fdbtypes.ProcessClassClusterController), func() {
...
+			It("should update the pods", func() {
+				pods := &corev1.PodList{}
+				err = k8sClient.List(context.TODO(), pods, getListOptions(cluster)...)
+				Expect(len(pods.Items)).To(Equal(len(originalPods.Items)))
+
+				for _, pod := range pods.Items {
+					Expect(pod.Spec.Containers[0].Name).To(Equal("foundationdb"))
+					Expect(pod.Spec.Containers[0].Image).To(Equal(fmt.Sprintf("foundationdb/foundationdb-kubernetes:%s", fdbtypes.Versions.Default)))
+					Expect(pod.Spec.Containers[1].Name).To(Equal("foundationdb-kubernetes-sidecar"))
+					Expect(pod.Spec.Containers[1].Image).To(Equal(fmt.Sprintf("foundationdb/foundationdb-kubernetes:%s", fdbtypes.Versions.Default)))
```
johscheuer, line 292: This image doesn't exist currently right?

controllers/update_pod_config.go
```diff
@@ -102,14 +105,16 @@ func (updatePodConfig) reconcile(r *FoundationDBClusterReconciler, context ctx.C
...
+				curLogger.Error(err, "Update Pod ConfigMap annotation")
+			}
 			if internal.IsNetworkError(err) {
 				processGroup.UpdateCondition(fdbtypes.SidecarUnreachable, true, cluster.Status.ProcessGroups, processGroup.ProcessGroupID)
 			} else {
 				processGroup.UpdateCondition(fdbtypes.IncorrectConfigMap, true, cluster.Status.ProcessGroups, processGroup.ProcessGroupID)
 			}
 
-			pod.ObjectMeta.Annotations[fdbtypes.OutdatedConfigMapKey] = time.Now().Format(time.RFC3339)
+			pod.ObjectMeta.Annotations[fdbtypes.OutdatedConfigMapKey] = fmt.Sprintf("%d", time.Now().Unix())
```
johscheuer, line 117: What is the benefit of this change? I would prefer the previous format to have it in aa human-readable format.

docs/manual/technical_design.md
```diff
@@ -331,6 +333,58 @@ The restore reconciler runs the following subreconcilers:
 
 The `StartRestore` subreconciler starts a restore. If there is no active restore, this will run a `start` command in `fdbrestore`.
 
+## Interaction Between the Operator and the Pods
+
+The operator communicates with processes running inside of the FoundationDB pods at multiple stages in the reconciliation flow. The exact flow will depend on whether you are using split images (which is the current default) or unified images.
+
+### Split Image
```
johscheuer, line 340: Should we rather name that "Sidecar Image" I know that this is ambiguous with the sidecar pattern but it's a little bit more specific than split image.

docs/manual/technical_design.md
```diff
@@ -331,6 +333,58 @@ The restore reconciler runs the following subreconcilers:
...
+
+When using split images, the `foundationdb` container runs an `fdbmonitor` process, which is responsible for starting `fdbserver` processes. `fdbmonitor` receives its configuration in the form of a monitor conf file, which contains the start command and arguments for the fdbserver process. This configuration can vary based on dynamic information like the node where the pod is running, so the operator provides a templated configuration file, which contains placeholders that are filled in based on the environment variables. That templating process is handled by the sidecar process in the `foundationdb-kubernetes-sidecar` container. The sidecar also provides an HTTP API for getting information about the state of the configuration in the pod. The sidecar mounts the config map containing dynamic conf as its input directory, and shares an `emptyDir` volume with the `foundationdb` container where it can put its output.
+
+The flow for updating the fdbserver processes has the following steps:
+
+1. The operator updates the monitor conf template in the `sample-cluster-config` config map. There is one monitor conf template for every process class the cluster uses.
```
johscheuer, line 346:
```suggestion
1. The operator updates the monitor conf template in the config map. There is one monitor conf template for every process class the cluster uses.
```

docs/manual/technical_design.md
```diff
@@ -331,6 +333,58 @@ The restore reconciler runs the following subreconcilers:
...
+
+### Unified Image
+
+**NOTE**: The unified image is still experimental, and is not recommended outside of development environments.
+
+When using the unified image, the `foundationdb` container runs a `fdb-kubernetes-monitor` process, which is responsible for starting `fdbserver` processes. `fdb-kubernetes-monitor` receives its configuration in the form of a JSON file, which provides the command-line arguments in a structured form. These arguments can reference environment variables, which will be filled in by `fdb-kubernetes-monitor`. They can also reference the process number, which allows `fdb-kubernetes-monitor` to start multiple `fdbserver` processes that use different ports and different data directories.
+
+The flow for updating the monitor conf file has the following steps:
+
+1. The operator updates the monitor conf template in the `sample-cluster-config` config map. There is one monitor conf file for every process class the cluster uses.
```
johscheuer, line 369:
```suggestion
1. The operator updates the monitor conf template in the config map. There is one monitor conf file for every process class the cluster uses.
```

internal/configmap_helper_test.go
```diff
@@ -0,0 +1,390 @@
...
+				Expect(err).NotTo(HaveOccurred())
+
+				Expect(configMap.Data[ClusterFileKey]).To(Equal("operator-test:asdfasf@127.0.0.1:4501"))
+				Expect(configMap.Data["fdbmonitor-conf-storage"]).To(Equal(expectedConf))
+				Expect(configMap.Data["running-version"]).To(Equal(fdbtypes.Versions.Default.String()))
+				Expect(configMap.Data["sidecar-conf"]).To(Equal(""))
+			})
+		})
+
+		When("the only unified image is enabled", func() {
```
johscheuer, line 86:
```suggestion
		When("only the unified image is enabled", func() {
```

internal/configmap_helper_test.go
```diff
@@ -0,0 +1,390 @@
...
+				Expect(config).To(Equal(expectedConfig))
+			})
+
+			It("does not include the data for the split monitor conf", func() {
+				_, present := configMap.Data["fdbmonitor-conf-storage"]
+				Expect(present).To(BeFalse())
+			})
+		})
+
+		When("the only split image is enabled", func() {
```
johscheuer, line 108:
```suggestion
		When("only the split image is enabled", func() {
```

internal/deprecations.go
```diff
@@ -51,6 +50,11 @@ type DeprecationOptions struct {
 // future-proof form, by applying any implicit defaults and moving configuration
 // from deprecated fields into fully-supported fields.
 func NormalizeClusterSpec(cluster *fdbtypes.FoundationDBCluster, options DeprecationOptions) error {
+	useUnifiedImage := false
+	if cluster.Spec.UseUnifiedImage != nil {
+		useUnifiedImage = *cluster.Spec.UseUnifiedImage
+	}
```
johscheuer, line 56:
```suggestion
	useUnifiedImage := pointer.BoolDeref(cluster.Spec.UseUnifiedImage, false)
```

internal/deprecations.go
```diff
@@ -278,8 +282,16 @@ func NormalizeClusterSpec(cluster *fdbtypes.FoundationDBCluster, options Depreca
 			cluster.Spec.AutomationOptions.Replacements.FailureDetectionTimeSeconds = pointer.Int(1800)
 		}
 
-		cluster.Spec.MainContainer.ImageConfigs = append(cluster.Spec.MainContainer.ImageConfigs, fdbtypes.ImageConfig{BaseImage: "foundationdb/foundationdb"})
-		cluster.Spec.SidecarContainer.ImageConfigs = append(cluster.Spec.SidecarContainer.ImageConfigs, fdbtypes.ImageConfig{BaseImage: "foundationdb/foundationdb-kubernetes-sidecar", TagSuffix: "-1"})
+		if cluster.Spec.UseUnifiedImage == nil {
+			cluster.Spec.UseUnifiedImage = pointer.Bool(useUnifiedImage)
+		}
```
johscheuer, line 287:
```suggestion
		cluster.Spec.UseUnifiedImage = useUnifiedImage
```

`useUnifiedImage` will be already `false` if `cluster.Spec.UseUnifiedImage` is not set otherwise we just overwrite `cluster.Spec.UseUnifiedImage` with the same value again, so that simplifies things.

internal/deprecations.go
```diff
@@ -278,8 +282,16 @@ func NormalizeClusterSpec(cluster *fdbtypes.FoundationDBCluster, options Depreca
...
+
+		if *cluster.Spec.UseUnifiedImage {
```
johscheuer, line 289:
```suggestion
		if useUnifiedImage {
```

internal/pod_models.go
```diff
@@ -148,6 +148,7 @@ func GetImage(image string, configs []fdbtypes.ImageConfig, versionString string
 func GetPodSpec(cluster *fdbtypes.FoundationDBCluster, processClass fdbtypes.ProcessClass, idNum int) (*corev1.PodSpec, error) {
 	processSettings := cluster.GetProcessSettings(processClass)
 	podSpec := processSettings.PodTemplate.Spec.DeepCopy()
+	useUnifiedImages := cluster.Spec.UseUnifiedImage != nil && *cluster.Spec.UseUnifiedImage
```
johscheuer, line 151:
```suggestion
	useUnifiedImages :=  pointer.BoolDeref(cluster.Spec.UseUnifiedImage, false)
```

johscheuer (changes requested):
A few comments from my side, I will take another look at it next week. Could you also add the doc string for the exported methods?

controllers/cluster_controller_test.go
brownleej: That's right, the image has not been published. I've been testing it with a locally-built image.

controllers/update_pod_config.go
brownleej: The kubernetes monitor is reading this annotation and parsing its contents, and it is currently expecting an integer timestamp. If you think it's worthwhile, we can modify that to expect a different format.

docs/manual/technical_design.md
brownleej: Is there a problem with including the name of the config map? I thought it would be helpful to be more specific about the config map we're discussing.

docs/manual/technical_design.md
brownleej: I thought that "split" would be a better parallel to "unified", which is the way we've been describing the new image.

internal/monitor_conf.go
```diff
@@ -104,70 +190,124 @@ func getStartCommandLines(cluster *fdbtypes.FoundationDBCluster, processClass fd
...
+// buildIPArgument builds an argument that takes an IP address from an environment variable
+func buildIPArgument(parameter string, environmentVariable string, imageType FDBImageType, sampleAddresses []fdbtypes.ProcessAddress) []monitorapi.Argument {
+	var leftIPWrap string
+	var rightIPWrap string
+	if imageType == FDBImageTypeUnified {
+		leftIPWrap = "["
+		rightIPWrap = "]"
+	} else {
+		leftIPWrap = ""
+		rightIPWrap = ""
```
johscheuer, line 282: I'm not totally sure if I understand that, why are we adding `[]` for the unified image but not for the split image? I thought that depends in the IP family we want to use and not the image type? Otherwise this will generate the wrong format for IPv4/IPv6?

internal/monitor_conf.go
```diff
@@ -104,70 +190,124 @@ func getStartCommandLines(cluster *fdbtypes.FoundationDBCluster, processClass fd
...
+		flags := make([]string, 0, len(address.Flags))
+		for flag, set := range address.Flags {
+			if set {
+				flags = append(flags, flag)
+			}
+		}
+
+		sort.Slice(flags, func(i int, j int) bool {
+			return flags[i] < flags[j]
+		})
```
johscheuer, line 306: I think we can move this piece into a common method for the `ProcessAddress` e.g. `SortedFlags()` and make use of it here and in https://github.com/brownleej/fdb-kubernetes-operator/blob/new-image-type/api/v1beta1/foundationdbcluster_types.go#L1947-L1960.

internal/pod_client.go
```diff
@@ -32,24 +32,47 @@ import (
...
 	"github.com/FoundationDB/fdb-kubernetes-operator/pkg/podclient"
+	monitorapi "github.com/apple/foundationdb/fdbkubernetesmonitor/api"
 	"github.com/go-logr/logr"
 	"github.com/hashicorp/go-retryablehttp"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/utils/pointer"
 )
 
+// FDBImageType describes a type of image a pod or cluster is using.
+type FDBImageType string
```
johscheuer, line 49: I think it's a good time to actually split up this file to have a dedicated file for the mock, sidecar and the annotation client.

internal/pod_client.go
```diff
@@ -107,11 +151,21 @@ func NewFdbPodClient(cluster *fdbtypes.FoundationDBCluster, pod *corev1.Pod) (po
...
+}
+
+// GetCluster returns the cluster associated with a client
+func (client *realFdbPodSidecarClient) GetCluster() *fdbtypes.FoundationDBCluster {
+	return client.Cluster
+}
+
+// GetPod returns the pod associated with a client
+func (client *realFdbPodSidecarClient) GetPod() *corev1.Pod {
+	return client.Pod
```
johscheuer, line 164: We are not using these methods, we should remove them.

internal/pod_client.go
```diff
@@ -279,15 +310,103 @@ func UpdateDynamicFiles(client podclient.FdbPodClient, filename string, contents
...
-	present, err := client.IsPresent(filename)
+// GetCluster returns the cluster associated with a client
+func (client *realFdbPodAnnotationClient) GetCluster() *fdbtypes.FoundationDBCluster {
+	return client.Cluster
+}
+
+// GetPod returns the pod associated with a client
+func (client *realFdbPodAnnotationClient) GetPod() *corev1.Pod {
+	return client.Pod
+}
```
johscheuer, line 321: We are not using these methods, we should remove them.

internal/pod_client.go
```diff
@@ -279,15 +310,103 @@ func UpdateDynamicFiles(client podclient.FdbPodClient, filename string, contents
...
+
+// GetCluster returns the cluster associated with a client
+func (client *mockFdbPodClient) GetCluster() *fdbtypes.FoundationDBCluster {
+	return client.Cluster
+}
+
+// GetPod returns the pod associated with a client
+func (client *mockFdbPodClient) GetPod() *corev1.Pod {
+	return client.Pod
+}
```
johscheuer, line 400: We are not using these methods, we should remove them.

internal/pod_client.go
brownleej: I don't think the file is too large now that we've moved some other things out, and hopefully it will get smaller in the future once we're able to deprecate and then remove the sidecar-based client. There's also some common functions like `GetImageType` that aren't necessarily specific to one type of client, but feel right in a shared file for multiple types of clients.

config/tests/unified_image/images.yaml
```diff
@@ -0,0 +1,10 @@
+- op: add
```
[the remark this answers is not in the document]
brownleej: It looks like the top-level object in the JSON patch files is supposed to be a list, so I don't think there is a header or resource kind to define.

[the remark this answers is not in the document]
johscheuer: That sounds right, I have on my list to go over all examples and ensure we are setting he headers at the right level. The last kubectl complained.

internal/pod_client.go
johscheuer: That's fair, I guess he diff view made it look bigger.

controllers/update_status.go
```diff
@@ -291,8 +292,12 @@ func (updateStatus) reconcile(r *FoundationDBClusterReconciler, context ctx.Cont
 		status.Locks.DenyList = denyList
 	}
 
-	// Sort the storage servers per disk to prevent a reordering to issue a new reconcile loop.
+	// Sort slices that are assembled based on pods to prevent a reordering from
+	// issuing a new reconcile loop.
 	sort.Ints(status.StorageServersPerDisk)
+	sort.Strings(status.ImageTypes)
+
+	//
```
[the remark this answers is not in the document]
johscheuer: Ping :)

internal/monitor_conf.go
johscheuer: Could you add some context here (preferable as a comment).

johscheuer (approved):
LGTM 👍

Issue #988, opened 2021-11-18
Pass knobs to fdbbackup and fdbrestore


johscheuer:
If someone specifies a know like `knob_http_verbose_level=3` on the backup resource we should pass that knob down to the `fdbbackup` command that our operator calls.

PR #996, opened 2021-11-25
Add support for passing custom parameters to the commandline

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/988

## Type of change

# Discussion

We could also filter some knobs but I think so far that doesn't make sense. I did some additional refactoring to make the solution cleaner.

# Testing

Unit + local.

# Documentation

-

# Follow-up

-


controllers/backup_controller.go
```diff
@@ -120,7 +120,14 @@ func (r *FoundationDBBackupReconciler) adminClientForBackup(context ctx.Context,
 		return nil, err
 	}
 
-	return r.getDatabaseClientProvider().GetAdminClient(cluster, r)
+	adminClient, err := r.getDatabaseClientProvider().GetAdminClient(cluster, r)
+	if err != nil {
+		return nil, err
+	}
+
+	adminClient.SetKnobs(backup.Spec.CustomParameters.GetKnobsForCLI())
```
brownleej, line 128: Are we sure that it's safe to apply the same knobs to the backup commands that we are setting on the backup agents?

brownleej: The knobs are also specific to the operation that the admin client is running at a given time (i.e. whether it is running fdbcli or fdbbackup), so storing them on the client itself seems like it could cause trouble.

brownleej (changes requested):

controllers/backup_controller.go
johscheuer: For the first part: The knobs that I looked at where save and could/must be used in both setups e.g. when using it as a CLI and using it as an agent. Maybe @bnamasivayam has some additional information about that since he worked on the fdbbackup support that requires the new knobs.

I see, we could run into an issue if we create one admin client and then call different cmd e.g. with `fdbcli` and `fdbbackup`. We could store those knobs into a map and only set it for `fdbbackup` and `fdbrestore`. I would like to prevent to have another parameter for the knobs and then have to pass the knobs in every call.

bnamasivayam: The backup knobs that I added recently `http_request_aws_v4_header` and `blobstore_encryption_type` should be present in both the fdbbackup and backup_agent commands. 

fdbcli doesn't require these knobs. So I don't know how it behaves if these knobs are passed to it. Although I don't think it should cause any trouble.

brownleej: Having the knobs on the admin client scoped to the specific binary seems like a good way to go. It sounds like there's no need to distinguish the knobs for backup_agent from the knobs for fdbbackup, so I'm OK with using the same CustomParameters field in the spec to represent both.

brownleej (approved):
👍

Issue #982, opened 2021-11-13
Change ProcessGroupStatus remove and excluded to timestamp


johscheuer:
We should add two new fields for the `ProcessGroupStatus`: `deletionTimestamp` and `exclusionTimestamp` those fields will supersede the `remove` and `excluded` field. The benefit of a timestamp is to see when the operator changed that status without looking into the logs. We should implement that with one release transition phase (I don't assume that many people use the status field). For the transition phase we will have the operator set always both fields and will check both fields.

PR #997, opened 2021-11-26
Add timestamp for remove and exclude

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/982

This allows to see when a process group was removed or excluded without looking into the logs.

## Type of change

# Discussion

I added this as an addition and in another release we can drop the support for the boolean flags.

# Testing

Unit + local testing:

```yaml
 - addresses:
    - 10.42.0.118
    processClass: storage
    processGroupID: storage-3
    remove: true
    removeTimestamp: "2021-11-26T17:52:41Z"
```

# Documentation

I can update the operation manual with the new variable.

# Follow-up

-

### DIFF
--- a/api/v1beta1/foundationdbcluster_types.go
+++ b/api/v1beta1/foundationdbcluster_types.go
@@ -484,15 +484,17 @@ type ProcessGroupStatus struct {
 	// Addresses represents the list of addresses the process group has been known to have.
 	Addresses []string `json:"addresses,omitempty"`
 	// Remove defines if the process group is marked for removal.
+	// Deprecated: Use RemovalTimestamp instead.
 	Remove bool `json:"remove,omitempty"`
 	// RemoveTimestamp if not empty defines when the process group was marked for removal.
-	RemoveTimestamp *metav1.Time `json:"removeTimestamp,omitempty"`
+	RemovalTimestamp *metav1.Time `json:"removalTimestamp,omitempty"`
 	// Excluded defines if the process group has been fully excluded.
 	// This is only used within the reconciliation process, and should not be considered authoritative.
+	// Deprecated: Use ExclusionTimestamp instead.
 	Excluded bool `json:"excluded,omitempty"`
 	// ExcludedTimestamp defines when the process group has been fully excluded.
 	// This is only used within the reconciliation process, and should not be considered authoritative.
-	ExcludedTimestamp *metav1.Time `json:"excludedTimestamp,omitempty"`
+	ExclusionTimestamp *metav1.Time `json:"exclusionTimestamp,omitempty"`
 	// ExclusionSkipped determines if exclusion has been skipped for a process, which will allow the process group to be removed without exclusion.
 	ExclusionSkipped bool `json:"exclusionSkipped,omitempty"`
 	// ProcessGroupConditions represents a list of degraded conditions that the process group is in.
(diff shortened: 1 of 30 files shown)


api/v1beta1/foundationdbcluster_types.go
```diff
@@ -471,15 +471,42 @@ type ProcessGroupStatus struct {
...
 	// Remove defines if the process group is marked for removal.
 	Remove bool `json:"remove,omitempty"`
+	// RemoveTimestamp if not empty defines when the process group was marked for removal.
+	RemoveTimestamp metav1.Time `json:"removeTimestamp,omitempty"`
 	// Excluded defines if the process group has been fully excluded.
 	// This is only used within the reconciliation process, and should not be considered authoritative.
 	Excluded bool `json:"excluded,omitempty"`
+	// ExcludedTimestamp defines when the process group has been fully excluded.
+	// This is only used within the reconciliation process, and should not be considered authoritative.
+	ExcludedTimestamp metav1.Time `json:"excludedTimestamp,omitempty"`
```
johscheuer, line 481: There is an issue in Kubernetes version < 1.20 (see: https://github.com/kubernetes/kubernetes/issues/86811). Since 1.19 is not "supported" anymore we might think about dropping the test case for 1.19 (https://kubernetes.io/releases/). Another option would be to make this `*metav1.Time` with all the additional `nil` checks which should be fine since we never use that field directly (only over the according methods).

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -471,15 +471,42 @@ type ProcessGroupStatus struct {
...
+	// This is only used within the reconciliation process, and should not be considered authoritative.
+	ExcludedTimestamp *metav1.Time `json:"excludedTimestamp,omitempty"`
 	// ExclusionSkipped determines if exclusion has been skipped for a process, which will allow the process group to be removed without exclusion.
 	ExclusionSkipped bool `json:"exclusionSkipped,omitempty"`
 	// ProcessGroupConditions represents a list of degraded conditions that the process group is in.
 	ProcessGroupConditions []*ProcessGroupCondition `json:"processGroupConditions,omitempty"`
 }
 
+// IsExcluded returns if a process group is excluded
+func (processGroupStatus *ProcessGroupStatus) IsExcluded() bool {
```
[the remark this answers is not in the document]
brownleej: For the fields themselves, we may want to mirror the timestamp fields in the resource metadata, which seem to follow a convention of `${noun}Timestamp`. e.g. creationTimestamp, deletionTimestamp.

brownleej (changes requested):

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -471,15 +471,42 @@ type ProcessGroupStatus struct {
 	Addresses []string `json:"addresses,omitempty"`
 	// Remove defines if the process group is marked for removal.
 	Remove bool `json:"remove,omitempty"`
+	// RemoveTimestamp if not empty defines when the process group was marked for removal.
+	RemoveTimestamp *metav1.Time `json:"removeTimestamp,omitempty"`
```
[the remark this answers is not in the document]
johscheuer: I will deprecate them, that's a good point and I just forgot to set the deprecation note.

api/v1beta1/foundationdbcluster_types.go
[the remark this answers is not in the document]
johscheuer: Initially I had `IsMarkedAsExcluded` and `IsMarkedForRemoval` those felt a little bit clumsy but I'm happy to change that back. I can change `excludedTimestamp` to `exclusionTimestamp` and `removeTimestamp` to `removalTimestamp`.

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -470,16 +470,45 @@ type ProcessGroupStatus struct {
...
 	Remove bool `json:"remove,omitempty"`
+	// RemoveTimestamp if not empty defines when the process group was marked for removal.
+	RemovalTimestamp *metav1.Time `json:"removalTimestamp,omitempty"`
 	// Excluded defines if the process group has been fully excluded.
 	// This is only used within the reconciliation process, and should not be considered authoritative.
+	// Deprecated: Use ExclusionTimestamp instead.
 	Excluded bool `json:"excluded,omitempty"`
+	// ExcludedTimestamp defines when the process group has been fully excluded.
+	// This is only used within the reconciliation process, and should not be considered authoritative.
+	ExclusionTimestamp *metav1.Time `json:"exclusionTimestamp,omitempty"`
```
johscheuer, line 483: Thinking more about it I actually think we can remove the `exclude` state completely. We have a few places where we check for that state e.g. in the replace failed process group sub-reconciler which could also use the `ResourcesTerminating` state to see if the process group was already excluded.

brownleej:
Since this is adding deprecations to the schema, we should talk about how we want to stage this with regards to the 1.0 release.

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -470,16 +470,45 @@ type ProcessGroupStatus struct {
...
+	processGroupStatus.ExclusionTimestamp = &metav1.Time{Time: time.Now()}
+}
+
+// IsMarkedForRemoval returns if a process group is marked for removal
+func (processGroupStatus *ProcessGroupStatus) IsMarkedForRemoval() bool {
+	return processGroupStatus.Remove || (processGroupStatus.RemovalTimestamp != nil && !processGroupStatus.RemovalTimestamp.IsZero())
+}
+
+// SetRemove marks a process group for removal
+func (processGroupStatus *ProcessGroupStatus) SetRemove() {
```
brownleej, line 507: Maybe we should rename this to `MarkForRemoval` to better match with `IsMarkedForRemoval`.

api/v1beta1/foundationdbcluster_types.go
brownleej: We're also using it to track metrics on exclusions, but maybe it's better to look at that through the database status.

brownleej (changes requested):

brownleej (approved):
👍

PR #1004, opened 2021-12-03
add go mod tidy

ammolitor:
this is a test of the emergency broadcasting system, remember this is only a test 😱

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -53,6 +53,7 @@ COPY internal/ internal/
 COPY pkg/ pkg/
 
 # Build
+RUN go mod download && go mod tidy
 RUN CGO_ENABLED=1 GOOS=linux GOARCH=amd64 GO111MODULE=on make manager
 
 # Create user and group here since we don't have the tools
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -40,7 +40,7 @@ COPY go.mod go.mod
 COPY go.sum go.sum
 # cache deps before building and copying source so that we don't need to re-download as much
 # and so that source changes don't invalidate our downloaded layer
-RUN go mod download
+RUN go mod download && go mod tidy
```
johscheuer, line 43:
```suggestion
RUN go mod download
```
The tidy command here will remove everything since we don't have any go files copied into the container (those are the next steps).

Dockerfile
```diff
@@ -53,6 +53,7 @@ COPY internal/ internal/
 COPY pkg/ pkg/
 
 # Build
+RUN go mod download && go mod tidy
```
johscheuer, line 56: To make the tidy command work we have to add thee following line above:

```
COPY mock-kubernetes-client/ mock-kubernetes-client/
```

Otherwise this will be missing. I think we should add that anyway. I open a PR for that., I guess we just missed it because we only build the manager and never run the tests inside the container. I understand that this is only for testing but in general we don't want to run `go mod tidy` in this step.

johscheuer: Fixed in: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1005

johscheuer (commented):
Should we close this one?

PR #1005, opened 2021-12-04
Add missing go pkg to copy

johscheuer:
# Description

I don' see a harm here since we only copy the files into the build container. On the other hand we remove the surprises for user that want to change something in the Dockerfile or if they want to run tests in the (build) container.

## Type of change

# Discussion

-

# Testing

Local.

# Documentation

-

# Follow-up

Wee might consider in the future to load everything into the container and work with a `.dockerignore` file .


brownleej (approved):
👍

PR #1007, opened 2021-12-04
Remove reference to old context package and move context as first parameter

johscheuer:
# Description

Remove the references to the old context package: https://pkg.go.dev/golang.org/x/net/context and move the context to be the first parameter from the referenced docs:

```
Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx:
```

## Type of change

# Discussion

# Testing

Local + unit.

# Documentation

-

# Follow-up

-

### DIFF
--- a/controllers/add_pods.go
+++ b/controllers/add_pods.go
@@ -38,17 +38,17 @@ import (
 type addPods struct{}
 
 // reconcile runs the reconciler's work.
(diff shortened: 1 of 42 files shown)


brownleej (approved):
👍

Issue #351, opened 2020-10-20
Safety checks to prevent deleting too many pods at once


brownleej:
Once the operator believes that an exclusion has completed, it will delete all of the excluded pods and PVCs at once. This presents a risk of losing a large number of volumes if the operator has a bug that causes it to mistakenly mark exclusions complete too soon. To mitigate this, we may want to throttle the deletion of pods. By default, we could only delete one pod at a time in the remove_pods step. We could also add a safety check to refuse to delete any pods if the operator is already at degraded fault tolerance.

rbtcollins:
See also #354

PR #1011, opened 2021-12-07
Add removal mode

johscheuer:
# Description

This PR adds a removal mode similar to the deletion mode (which maybe should be called update mode?). In addition to that I made the PVC creation unblocking e.g. if a user hit's a resource quota it should not block any further actions.

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/351

## Type of change

# Discussion

We probably also want to have and "exclusionMode" that defines how we want to exclude processes (or we assume we want to exclude all processes always). I thought a little bit about how we can make these operations safer in general and I will try to create a design doc for that this week.

# Testing

Local + unit tests.

# Documentation

Will be updated.

# Follow-up

-


api/v1beta1/foundationdbcluster_types.go
```diff
@@ -1187,6 +1187,15 @@ type FoundationDBClusterAutomationOptions struct {
...
 	DeletionMode DeletionMode `json:"deletionMode,omitempty"`
+
+	// RemovalMode defines the removal mode for this cluster. This can be
+	// DeletionModeAll, DeletionModeZone or DeletionModeProcessGroup. The
+	// RemovalMode defines how process groups are deleted in order when they
+	// are marked for removal.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum=All;Zone;ProcessGroup
+	// +kubebuilder:default:=Zone
+	RemovalMode DeletionMode `json:"removalMode,omitempty"`
```
johscheuer, line 1198: Actually `DeletionMode` is perhaps not the best name but I didn't came up with a better name.

controllers/remove_process_groups.go
```diff
@@ -77,7 +80,26 @@ func (u removeProcessGroups) reconcile(r *FoundationDBClusterReconciler, context
...
+	}
+
+	zone, zoneRemovals, err := removals.GetProcessGroupsToRemove(cluster.GetRemovalMode(), zonedRemovals)
+	if err != nil {
+		return &requeue{curError: err}
+	}
+
+	logger.Info("Removing process groups", "zone", zone, "count", len(zoneRemovals), "deletionMode", string(cluster.Spec.AutomationOptions.DeletionMode))
+
+	removedProcessGroups := r.removeProcessGroups(context, cluster, zoneRemovals)
```
johscheuer, line 102: In theory we should requeue here if we have some zones left. In practice we will requeue anyway because of the generation change in the Pod, PVC, service.

I will add a test case for resources that hang in a terminating state to ensure we exclude those.

api/v1beta1/foundationdbcluster_types.go
brownleej: I like the idea of renaming the type to `UpdateMode`, and coming up with a better name for the current `DeletionMode` field on the spec. Maybe `PodUpdateMode` would cover it, since there's a general understanding that "updating" a pod involves deleting and recreating it, and `RecreationMode` doesn't look right.

controllers/remove_process_groups.go
brownleej: I think we can have a reasonable heuristic based around the deletionTimestamp on the pod. If it's in the last 10 minutes or so, we can treat it as an active deletion and wait before deleting any further resources. If it's older than that, it's likely a stuck termination and we can ignore it and continue deleting further resources.

internal/error_helper.go
```diff
@@ -51,3 +53,15 @@ func IsTimeoutError(err error) bool {
 
 	return false
 }
+
+// IsQuotaExceeded returns tru if the error returned by the Kubernetes API is a forbidden error with the error message
```
johscheuer, line 57:
```suggestion
// IsQuotaExceeded returns true if the error returned by the Kubernetes API is a forbidden error with the error message
```

internal/removals/remove.go
```diff
@@ -0,0 +1,173 @@
...
+		if !ok {
+			zoneMap[UnknownZone] = append(zoneMap[UnknownZone], pg.ProcessGroupID)
+			continue
+		}
+
+		zone := p.Locality[fdbtypes.FDBLocalityZoneIDKey]
+		zoneMap[zone] = append(zoneMap[zone], pg.ProcessGroupID)
+	}
+
+	// TODO return last time since removal
```
johscheuer, line 118: Can be dropped.

internal/removals/remove.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	return "", nil, fmt.Errorf("unknown deletion mode: \"%s\"", removalMode)
+}
+
+// GetZonedRemovals returns a map with the zone as key and a list of process groups IDs to be removed.
+// If the process group has not an associated process in the cluster status the zone will be UnknownZone.
+// if the process group has the ResourcesTerminating condition the zone will be TerminatingZone.
+func GetZonedRemovals(status *fdbtypes.FoundationDBStatus, processGroupsToRemove []*fdbtypes.ProcessGroupStatus) (map[string][]string, int64, error) {
+	var lastRemovalTimestamp int64
```
johscheuer, line 87: I think that should rather be phrased `latest...`

internal/removals/remove.go
```diff
@@ -0,0 +1,173 @@
...
+		remainingMap[address.String()] = false
+	}
+	for _, address := range remaining {
+		remainingMap[address.String()] = true
+	}
+
+	return remainingMap, nil
+}
+
+// RemovalAllowed returns if we are allowed to remove the process group of if we have to wait to ensure a safer deletion.
```
johscheuer, line 165:
```suggestion
// RemovalAllowed returns if we are allowed to remove the process group or if we have to wait to ensure a safe deletion.
```

internal/removals/remove_test.go
```diff
@@ -0,0 +1,214 @@
+/*
+ * remove.go
```
johscheuer, line 2:
```suggestion
 * remove_test.go
```

controllers/update_pods.go
```diff
@@ -156,8 +156,8 @@ func (updatePods) reconcile(ctx context.Context, r *FoundationDBClusterReconcile
 	return deletePodsForUpdates(ctx, r, cluster, adminClient, updates, logger)
 }
 
-func getPodsToDelete(deletionMode fdbtypes.DeletionMode, updates map[string][]*corev1.Pod) (string, []*corev1.Pod, error) {
-	if deletionMode == fdbtypes.DeletionModeAll {
+func getPodsToDelete(deletionMode fdbtypes.PodUpdateMode, updates map[string][]*corev1.Pod) (string, []*corev1.Pod, error) {
```
johscheuer, line 159: As a follow up we should move this method into the internal package and try to reuse the same mechanism that we are using for the process groups.

controllers/remove_process_groups.go
```diff
@@ -327,24 +316,28 @@ func (r *FoundationDBClusterReconciler) getProcessGroupsToRemove(cluster *fdbtyp
...
 	for _, id := range processGroupsToRemove {
 		err := removeProcessGroup(r, context, cluster, id)
 		if err != nil {
 			logger.Error(err, "Error during remove Pod", "processGroupID", id)
 			continue
 		}
+	}
 
+	removedProcessGroups := make(map[string]bool)
+	// We have to check for the currently removed process group if they are completely removed
```
johscheuer, line 338:
```suggestion
	// We have to check if the removed process groups are completely removed
```

controllers/update_pods.go
```diff
@@ -131,7 +131,7 @@ func (updatePods) reconcile(ctx context.Context, r *FoundationDBClusterReconcile
 			return &requeue{message: "Requeueing reconciliation to replace pods"}
 		}
 
-		if !pointer.BoolDeref(cluster.Spec.AutomationOptions.DeletePods, true) {
+		if !pointer.BoolDeref(cluster.Spec.AutomationOptions.DeletePods, true) || r.PodLifecycleManager.GetDeletionMode(cluster) == fdbtypes.PodUpdateModeNone {
```
brownleej, line 134: If the field in AutomationOptions is deprecated, do we need to read it here, or can we take care of it in NormalizeClusterSpec?

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -1184,13 +1185,30 @@ type FoundationDBClusterAutomationOptions struct {
...
+	// +kubebuilder:default:=Zone
+	RemovalMode PodUpdateMode `json:"removalMode,omitempty"`
+
+	// WaitTimeBetweenRemovals defines how long to wait between the last removal and the next removal. This is only an
+	// upper limit if the process group and the according resources are deleted faster then the provided duration the
+	// operator will move on with the next removal. The idea is to prevent a race condition were the operator deletes
+	// a resource but the Kubernetes API is slower to trigger the actual deletion and we are running into a situation
+	// where the fault tolerance check still includes the already deleted processes.
+	// Defaults to 60s.
+	WaitTimeBetweenRemovals time.Duration `json:"waitTimeBetweenRemovals,omitempty"`
```
brownleej, line 1211: Should we include the unit in the field name? The fact that it's parsed into a duration tells us how it's interpreted on the go side, but it could be ambiguous from the perspective of clients.

brownleej (changes requested):

api/v1beta1/foundationdbcluster_types.go
johscheuer: That's a good idea 👍

controllers/update_pods.go
johscheuer: I think we can take care of that in the `NormalizeClusterSpec`.

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -1231,13 +1231,13 @@ type FoundationDBClusterAutomationOptions struct {
...
-	// WaitTimeBetweenRemovals defines how long to wait between the last removal and the next removal. This is only an
-	// upper limit if the process group and the according resources are deleted faster then the provided duration the
+	// WaitDurationBetweenRemovals defines how long to wait between the last removal and the next removal. This is only an
+	// upper limit if the process group and the according resources are deleted faster than the provided duration the
 	// operator will move on with the next removal. The idea is to prevent a race condition were the operator deletes
 	// a resource but the Kubernetes API is slower to trigger the actual deletion and we are running into a situation
 	// where the fault tolerance check still includes the already deleted processes.
 	// Defaults to 60s.
-	WaitTimeBetweenRemovals time.Duration `json:"waitTimeBetweenRemovals,omitempty"`
+	WaitDurationBetweenRemovals time.Duration `json:"waitDurationBetweenRemovals,omitempty"`
```
brownleej, line 1240: I think it would be more in line with the Kubernetes idioms if we called it something like `waitDurationBetweenRemovalsNanoseconds`, similar to fields like `terminationGracePeriodSeconds` and `activeDeadlineSeconds` in the pod spec.

brownleej (changes requested):

api/v1beta1/foundationdbcluster_types.go
johscheuer: The difference is that those referenced fields are `int64` and not a `time.Duration`. Currently we already have the `ignorePendingPodsDuration` that also takes a `time.Duration` as input which means that the user has to specify the time unit e.g. `30s` or `5m` in the request.

brownleej (approved):
👍

PR #1013, opened 2021-12-08
[DO NOT MERGE] whitespace change for codebuild test PR

ammolitor:
The CodeBuild PR builder is timing out at 4 hours and the build logs are a mess. This PR (and branch) is for us to fix the remaining issues with the CodeBuild Project

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,7 @@
 FROM docker.io/foundationdb/foundationdb:6.2.30 as fdb62
+
 FROM docker.io/foundationdb/foundationdb:6.1.13 as fdb61
+
 FROM docker.io/foundationdb/foundationdb:6.3.22 as fdb63
 
 # Build the manager binary


ammolitor:
@johscheuer and @ltsampros please review the CodeBuild logs from this build when they land. I'm hoping they'll help us determine why we're failing/timing out.

johscheuer:
The build uses the old broken image.

johscheuer:
```
...
 ---> Running in a6eacb3c2d1b
time="2021-12-08T07:30:32.157189313Z" level=info msg="shim containerd-shim started" address="/containerd-shim/moby/a6eacb3c2d1b74b089c0b5ae64b541dafc6060c56d88903b58001d76b2fffc12/shim.sock" debug=false pid=6426 
go build -ldflags="-s -w -X github.com/FoundationDB/fdb-kubernetes-operator/setup.operatorVersion=latest" -o bin/manager main.go
[91mcontrollers/backup_controller.go:33:2: missing go.sum entry for module providing package golang.org/x/net/context; to add:
	go mod download golang.org/x/net
/go/pkg/mod/golang.org/x/oauth2@v0.0.0-20200107190931-bf48bf16ab8d/internal/token.go:23:2: missing go.sum entry for module providing package golang.org/x/net/context/ctxhttp; to add:
	go mod download golang.org/x/net
/go/pkg/mod/k8s.io/apimachinery@v0.21.3/pkg/util/net/http.go:40:2: missing go.sum entry for module providing package golang.org/x/net/http2; to add:
	go mod download golang.org/x/net
[0m[91mmake: *** [Makefile:100: bin/manager] Error 1
```

johscheuer:
I'll rebase that branch

johscheuer:
```
Summarizing 3 Failures:
  [FAIL] Operator Migrating a cluster from gp2 to gp3 [BeforeEach] should migrate the cluster
  /codebuild/output/src931170019/src/github.com/FoundationDB/fdb-kubernetes-tests/tests/test_operator/operator_test.go:862
  [FAIL] Operator Replacing a Pod with PVC stuck in Terminating state [It] should replace the pvc stuck in Terminating state
  /codebuild/output/src931170019/src/github.com/FoundationDB/fdb-kubernetes-tests/tests/test_operator/operator_test.go:927
  [FAIL] [AfterSuite] 
  /codebuild/output/src931170019/src/github.com/FoundationDB/fdb-kubernetes-tests/tests/common/fixtures.go:39

Ran 22 of 26 Specs in 8888.479 seconds
FAIL! -- 20 Passed | 2 Failed | 4 Pending | 0 Skipped
--- FAIL: TestKubernetes (8888.48s)
FAIL
```

johscheuer:
If we could remove the `constructing many client instances from the same exec auth config can cause performance problems during cert rotation and can exhaust available network connections;` from the output that would be nice.

ammolitor:
> If we could remove the `constructing many client instances from the same exec auth config can cause performance problems during cert rotation and can exhaust available network connections;` from the output that would be nice.

https://github.com/kubernetes/kubernetes/issues/91913#issuecomment-840100077

I can upgrade EKS if we think that is the right thing to do. I feel like having the K8s version and the client version out of sync (i.e. only changing the client version) might introduce other problems.

johscheuer:
> > If we could remove the `constructing many client instances from the same exec auth config can cause performance problems during cert rotation and can exhaust available network connections;` from the output that would be nice.
> 
> [kubernetes/kubernetes#91913 (comment)](https://github.com/kubernetes/kubernetes/issues/91913#issuecomment-840100077)
> 
> I can upgrade EKS if we think that is the right thing to do. I feel like having the K8s version and the client version out of sync (i.e. only changing the client version) might introduce other problems.

A quick workaround would bee a grep statement 😇

Issue #1016, opened 2021-12-09
Variable substitution with custom parameters is broken on master 


johscheuer:
We recently added some changes for the new image format. With the that we added a new way how we model the expected command line flags. This introduced a new bug when someone sets a custom parameter with a variables e.g.:

```
expected: ... --locality_disk_id=$FDB_PUBLIC_IP ...
got: ...  --locality_disk_id=1.2.3.4 ...
```

The place where we are missing the substitution is here:  https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/internal/monitor_conf.go#L251-L260

To be clear the issue is in the operator itself and the Pod/Sidecar does the substitution correctly but we end up in a state where the operator always thinks that the process args are wrong.

PR #1017, opened 2021-12-13
Add support for env variables in custom parameters

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1016

## Type of change

- 
# Discussion

-

# Testing

Unit + local

# Documentation

We should document in another PR that only env variables that are provided as `substitutes` will work.

# Follow-up

-


brownleej:
I'd like to follow up on having a new custom parameters structure that supports the full argument syntax from fdb-kubernetes-monitor, but I think this makes sense as a quicker fix.

brownleej (changes requested):
I suggested some changes to the test descriptions. Otherwise LGTM

johscheuer (commented):
Ups, thanks for finding that.

PR #1021, opened 2021-12-17
Fix handling of custom parameters with substitution variables.

### LINKED ISSUE #1016: Variable substitution with custom parameters is broken on master

brownleej:
# Description

This fixes the handling of custom parameters with substitution variables.

Fixes #1016 

## Type of change

# Discussion

This will only work in old image with the separate sidecar image; it won't work with the unified image. I think that supporting it properly in the new image will require a more complex solution with structured custom parameters, but I wanted to get a fix in for the regression on this feature in the old image.

@johscheuer took a different approach in #1017, but closed that PR without merging it. One difference in the approaches is that this PR will also support custom parameters that have a substitution variable embedded partway through the value, like `locality_custom_id=process-$FDB_INSTANCE_ID`. 

# Testing

Manual testing?*

I added unit tests, and did manual tests in my dev environment.

No.

# Documentation

No.

# Follow-up

We should find a way to make this more structured.

No.

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -260,7 +260,7 @@ func (r *FoundationDBClusterReconciler) updatePodDynamicConf(cluster *fdbtypes.F
 
 	imageType := internal.GetImageType(pod)
 	if imageType == internal.FDBImageTypeUnified {
-		config, err := internal.GetMonitorProcessConfiguration(cluster, processClass, serversPerPod, imageType)
+		config, err := internal.GetMonitorProcessConfiguration(cluster, processClass, serversPerPod, imageType, nil)
 		if err != nil {
 			return false, err
 		}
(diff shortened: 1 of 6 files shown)


JoshuaMcManus (approved):
👍

PR #1023, opened 2022-01-05
Download FDB assets from GitHub rather than the FDB website.

brownleej:
# Description

This moves the download source for FDB assets to GitHub rather than the FDB website. This is necessary in the short term while the website is under maintenance, but I think we can continue to use GitHub as the source of these assets in the long term. 

## Type of change

# Discussion

No.

# Testing

Manual testing?*

I ran a local build of the operator with the new dockerfile.

We'll need to make sure that the GitHub actions work with the new locations, and we should probably cut a patch release to make sure that process works and get a fully supported release on the 0.49 code.

# Documentation

N/A

# Follow-up

No.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,18 +7,18 @@ FROM docker.io/library/golang:1.16.11 as builder
 
 # Install FDB
 ARG FDB_VERSION=6.2.30
-ARG FDB_WEBSITE=https://www.foundationdb.org
+ARG FDB_WEBSITE=https://github.com/apple/foundationdb/releases/download
 ARG TAG="latest"
 
 RUN set -eux && \
-	curl --fail ${FDB_WEBSITE}/downloads/${FDB_VERSION}/ubuntu/installers/foundationdb-clients_${FDB_VERSION}-1_amd64.deb -o fdb.deb && \
+	curl --fail -L ${FDB_WEBSITE}/${FDB_VERSION}/foundationdb-clients_${FDB_VERSION}-1_amd64.deb -o fdb.deb && \
 	dpkg -i fdb.deb && rm fdb.deb && \
-	curl --fail $FDB_WEBSITE/downloads/$FDB_VERSION/linux/fdb_$FDB_VERSION.tar.gz -o fdb_$FDB_VERSION.tar.gz && \
-	tar -xzf fdb_$FDB_VERSION.tar.gz --strip-components=1 && \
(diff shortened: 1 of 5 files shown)


JoshuaMcManus (commented):
We should consider leaving the FDB website as an option to fetch the files from in the event github goes down. A lot of engineers learned that the hard way late last year that github isn't up 100% ;)

We could simply use a flag, or check if the status from the curl is 2xx and use the backup site otherwise.

Dockerfile
```diff
@@ -7,18 +7,18 @@ FROM docker.io/library/golang:1.16.11 as builder
...
-	curl --fail ${FDB_WEBSITE}/downloads/${FDB_VERSION}/ubuntu/installers/foundationdb-clients_${FDB_VERSION}-1_amd64.deb -o fdb.deb && \
+	curl --fail -L ${FDB_WEBSITE}/${FDB_VERSION}/foundationdb-clients_${FDB_VERSION}-1_amd64.deb -o fdb.deb && \
 	dpkg -i fdb.deb && rm fdb.deb && \
-	curl --fail $FDB_WEBSITE/downloads/$FDB_VERSION/linux/fdb_$FDB_VERSION.tar.gz -o fdb_$FDB_VERSION.tar.gz && \
-	tar -xzf fdb_$FDB_VERSION.tar.gz --strip-components=1 && \
-	rm fdb_$FDB_VERSION.tar.gz && \
-	chmod u+x fdbbackup fdbcli fdbdr fdbmonitor fdbrestore fdbserver backup_agent dr_agent && \
 	mkdir -p /usr/bin/fdb/${FDB_VERSION%.*} && \
-	mv fdbbackup fdbcli fdbdr fdbmonitor fdbrestore fdbserver backup_agent dr_agent /usr/bin/fdb/${FDB_VERSION%.*} && \
+	for binary in fdbbackup fdbcli; do \
```
JoshuaMcManus, line 17: It looks like the old code downloaded a bunch of the other agents and this update only downloads the cli and backup client. I assume this is intentional to reduce the dockerfile size?

brownleej:
I took a look at the CodeBuild error, and it seems to be unrelated to this PR.

Dockerfile
brownleej: Yes, I don't think we use the other binaries anywhere.

ammolitor (approved):

JoshuaMcManus (approved):

PR #1024, opened 2022-01-12
Remove minimal e2e tests

johscheuer:
# Description

With the integration of the e2e test suite for this repo we can remove the small e2e test.

## Type of change

- Other

# Discussion

I don't see a huge value in that test when we already run a suite of e2e tests.

# Testing

Local.

# Documentation

-

# Follow-up

-


brownleej (approved):
👍

Issue #1006, opened 2021-12-04
Upgrade go version to 1.17


johscheuer:
Upgrade the current go version to the latest 1.17 version. Release notes: https://tip.golang.org/doc/go1.17

I don't see any breaking changes or deprecation but we should run a test with the new version just to be sure.

PR #1025, opened 2022-01-12
Update go version

johscheuer:
# Description

fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1006

Update to the latest go 1.17.6 version (which is also used by Kubernetes).

## Type of change

# Discussion

-

# Testing

Local testing.

# Documentation

-

# Follow-up

-


go.mod
```diff
@@ -26,3 +26,83 @@ require (
 	sigs.k8s.io/controller-runtime v0.9.6
 	sigs.k8s.io/yaml v1.2.0
 )
+
+require (
```
johscheuer, line 30: As a reference why we have two require blocks: https://go.dev/doc/go1.17#go-command (that's new with 1.17)

brownleej (approved):
👍

PR #1026, opened 2022-01-12
Correct the env setting for race detection

johscheuer:
# Description

Set the env variable at the correct place.

## Type of change

# Discussion

-

# Testing

-

# Documentation

-

# Follow-up

-


Makefile
```diff
@@ -59,7 +59,7 @@ GO_ALL=${GO_SRC} ${GENERATED_GO}
 MANIFESTS=config/crd/bases/apps.foundationdb.org_foundationdbbackups.yaml config/crd/bases/apps.foundationdb.org_foundationdbclusters.yaml config/crd/bases/apps.foundationdb.org_foundationdbrestores.yaml
 
 ifeq "$(TEST_RACE_CONDITIONS)" "1"
-	go_test_flags := $(go_test_flags) -race
+	go_test_flags := $(go_test_flags) -race -timeout=30m
```
johscheuer, line 62: Adding the race detection slows down the testing and the GitHub action runner seems to be pretty slow so we hit the default 10m timeout.

brownleej (approved):
👍

PR #1027, opened 2022-01-13
Add validation for storage engine and redundancy mode

johscheuer:
# Description

Add validation to prevent errors based on typos.

## Type of change

# Discussion

The drawback of this is that the CRD must be adjusted when we want to support new redundancy modes or storage engines but I think that's a fair trade off and I don't suspect to many changes in this field (and the adjustment is trivial).

# Testing

Local.

# Documentation

-

# Follow-up

-

### DIFF
--- a/api/v1beta1/foundationdbcluster_types.go
+++ b/api/v1beta1/foundationdbcluster_types.go
@@ -2094,6 +2094,7 @@ type FoundationDBClusterFaultDomain struct {
 }
 
 // RedundancyMode defines the core replication factor for the database
+// +kubebuilder:validation:MaxLength=100
 type RedundancyMode string
 
 const (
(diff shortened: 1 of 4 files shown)


api/v1beta1/foundationdbcluster_types.go
```diff
@@ -2107,13 +2108,34 @@ const (
...
 	// RedundancyMode defines the core replication factor for the database.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum=single;double;triple
+	// +kubebuilder:default:double
 	RedundancyMode RedundancyMode `json:"redundancy_mode,omitempty"`
 
 	// StorageEngine defines the storage engine the database uses.
-	StorageEngine string `json:"storage_engine,omitempty"`
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum=ssd;ssd-1;ssd-2;memory;memory-1;memory-2;ssd-redwood-1-experimental;ssd-rocksdb-experimental;memory-radixtree-beta;custom
```
johscheuer, line 2136: I took the fields from: https://github.com/apple/foundationdb/blob/master/fdbclient/DatabaseConfiguration.cpp#L290-L312

brownleej (approved):
👍

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -2107,13 +2108,34 @@ const (
...
+
+const (
+	// StorageEngineSSD defines the storage engine ssd.
+	StorageEngineSSD StorageEngine = "ssd"
+	// StorageEngineSSD2 defines the storage engine ssd-2.
+	StorageEngineSSD2 StorageEngine = "ssd-2"
+	// StorageEngineMemory defines the storage engine memory.
+	StorageEngineMemory StorageEngine = "memory"
+	// StorageEngineMemory2 defines the storage engine memory-2.
+	StorageEngineMemory2 StorageEngine = "memory-2"
```
xumengpanda, line 2123: how about rocksdb engine for master branch?

xumengpanda (commented):
How was this PR tested?

api/v1beta1/foundationdbcluster_types.go
johscheuer: These are only constant used in the code the actual validation fields are here and include the rocksdb engine: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/api/v1beta1/foundationdbcluster_types.go#L2136

johscheuer:
> How was this PR tested?

We tested that locally to ensure only the specified values in https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/api/v1beta1/foundationdbcluster_types.go#L2136 are allowed. What is your precise concern/

xumengpanda:
I was interested in:
* Does operator repo have the regression/CI test for each PR as fdb src repo does?
* Is the storage engine type included there?

Issue #995, opened 2021-11-25
Resource leak in update status


johscheuer:
I discovered today a scenario where the operator did not cleanup old pending Pods. The process group of that Pod were not being stored in the cluster status from Kubernetes under `processGroups`. In addition to that the Pods were in a Pending state and therefore are not part of the cluster status from the FDB cluster itself. I think the easiest solution for that is to change the `AddPods` method to check if there are Pods that have no process group and if so delete them, we should to the same for PVCs and services.

If I find some additional information I will provide it here.

PR #1028, opened 2022-01-13
Add plugin support to find Pods that are not part of the cluster status spec

johscheuer:
# Description

As part of https://github.com/FoundationDB/fdb-kubernetes-operator/issues/995 it would be useful if the plugin discovers this misbehaviour and reports it in the `analyse` command.

## Type of change

# Discussion

-

# Testing

Local + unit

# Documentation

-

# Follow-up

-


johscheuer (commented):
In another PR we might want to extend this to also find PVCs and services. With the latest change this shouldn't happen but I think having those checks in place doesn't hurt.

JoshuaMcManus (approved):

Issue #1009, opened 2021-12-04
Refactoring: rename context parameter to ctx and ensure we pass it as the first one


johscheuer:
Initial work: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1007 we have to verify all our methods and ensure we change all occurrences. Here is a first (not complete) list:

```
controllers/toggle_backup_paused.go
34:func (s toggleBackupPaused) reconcile(r *FoundationDBBackupReconciler, context ctx.Context, backup *fdbtypes.FoundationDBBackup) *requeue {

controllers/add_services.go
109:func updateService(r *FoundationDBClusterReconciler, context ctx.Context, cluster *fdbtypes.FoundationDBCluster, currentService *corev1.Service, newService *corev1.Service) error {

controllers/update_backup_agents.go
44:func (u updateBackupAgents) reconcile(r *FoundationDBBackupReconciler, context ctx.Context, backup *fdbtypes.FoundationDBBackup) *requeue {

controllers/start_restore.go
35:func (s startRestore) reconcile(r *FoundationDBRestoreReconciler, context ctx.Context, restore *fdbtypes.FoundationDBRestore) *requeue {

controllers/restore_controller.go
99:func (r *FoundationDBRestoreReconciler) adminClientForRestore(context ctx.Context, restore *fdbtypes.FoundationDBRestore) (fdbadminclient.AdminClient, error) {

controllers/remove_process_groups.go
89:func removeProcessGroup(r *FoundationDBClusterReconciler, context ctx.Context, cluster *fdbtypes.FoundationDBCluster, processGroupID string) error {
137:func confirmRemoval(r *FoundationDBClusterReconciler, context ctx.Context, cluster *fdbtypes.FoundationDBCluster, processGroupID string) (bool, bool, error) {
196:func includeProcessGroup(r *FoundationDBClusterReconciler, context ctx.Context, cluster *fdbtypes.FoundationDBCluster, removedProcessGroups map[string]bool) error {
336:func (r *FoundationDBClusterReconciler) removeProcessGroups(context ctx.Context, cluster *fdbtypes.FoundationDBCluster, processGroupsToRemove []string) map[string]bool {

controllers/update_status.go
[...]

PR #1040, opened 2022-01-16
Make context the first argument for all methods

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1009

## Type of change

# Discussion

This will break some interfaces so that's a good change to do before we cut 1.0

# Testing

local.

# Documentation

-

# Follow-up

-

### DIFF
--- a/controllers/add_pods.go
+++ b/controllers/add_pods.go
@@ -21,7 +21,7 @@
 package controllers
 
 import (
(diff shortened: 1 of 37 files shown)


brownleej (approved):
👍

PR #1043, opened 2022-01-16
Refresh process group status and that Pods, PVCs and Services machting selector into account

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/995

## Type of change

# Discussion

We already added process groups for PVCs and Services, I think we should also do that for Pods to ensure we are not leaking resources.

# Testing

Local.

# Documentation

-

# Follow-up

-


brownleej (approved):
👍

Issue #659, opened 2021-05-03
WATCH_NAMESPACE should be a command line flag


johscheuer:
Currently we can define the behaviour of the operator with `WATCH_NAMESPACE` to define if the operator should only watch a specific namespace. Instead of using an environment variable we should add a command line flag to define the watched namespace.

PR #1046, opened 2022-01-18
manager: add new cli flag `-watch-namespace`

ltsampros:
# Description

Add new cli flag `-watch-namespace` with same functionality as the `WATCH_NAMESPACE` env variable. For compatibility, I let the env var take precedence if set. 

Addresses issue: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/659

## Type of change

### DIFF
--- a/config/samples/deployment.yaml
+++ b/config/samples/deployment.yaml
@@ -122,6 +122,9 @@ spec:
       containers:
       - command:
         - /manager
+        args:
+        - "-watch-namespace"
+        - "$(WATCH_NAMESPACE)"
         env:
         - name: FDB_NETWORK_OPTION_EXTERNAL_CLIENT_DIRECTORY
           value: /usr/bin/fdb
--- a/docs/manual/customization.md
+++ b/docs/manual/customization.md
@@ -255,7 +255,7 @@ To run the controller in single-namespace mode, you will need to configure the f
 
 * A service account for the controller
 * The serviceAccountName field in the controller's pod spec
-* A `WATCH_NAMESPACE` environment variable defined in the controller's pod spec or in the arguments of the st
+* A `WATCH_NAMESPACE` environment variable defined in the controller's pod spec or in the arguments of the container command
 * A Role that grants access to the necessary permissions to all of the resources that the controller manages. See the [sample role](https://raw.githubusercontent.com/FoundationDB/fdb-kubernetes-operator/master/config/samples/deployment/rbac_role.yaml) for the list of those permissions.
 * A RoleBinding that binds that role to the service account for the controller
 
--- a/docs/manual/customization.md
+++ b/docs/manual/customization.md
@@ -247,7 +247,7 @@ Our [sample deployment](https://raw.githubusercontent.com/foundationdb/fdb-kuber
 
 ### Single-Namespace Mode
 
-To use single-namespace mode, set the `WATCH_NAMESPACE` environment variable or the command-line option `-watch-namespace` for the controller to be the namespace where your FDB clusters will run. It does not have to be the same namespace where the operator is running, though this is generally the simplest way to configure it. When you are running in single-namespace mode, the controller will ignore any clusters you try to create in namespaces other than the one you give it. If both options are defined, the environment variable will take precedence.
+To use single-namespace mode, set the `WATCH_NAMESPACE` environment variable or the command-line option `-watch-namespace` for the controller to be the namespace where your FDB clusters will run. It does not have to be the same namespace where the operator is running, though this is generally the simplest way to configure it. When you are running in single-namespace mode, the controller will ignore any clusters you try to create in namespaces other than the one you give it. If both options are defined, the command line argument will be used.
 
 The advantage of single-namespace mode is that it allows owners of different namespaces to run the operator themselves without needing access to other namespaces that may be managed by other tenants. The only cluster-level configuration it requires is the installation of the CRD. The disadvantage of single-namespace mode is that if you are running multiple namespaces for a single team, each namespace will need its own installation of the controller, which can make it more operationally challenging.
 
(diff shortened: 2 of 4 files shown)


docs/manual/customization.md
```diff
@@ -247,23 +247,23 @@ Our [sample deployment](https://raw.githubusercontent.com/foundationdb/fdb-kuber
 
 ### Single-Namespace Mode
 
-To use single-namespace mode, set the `WATCH_NAMESPACE` environment variable for the controller to be the namespace where your FDB clusters will run. It does not have to be the same namespace where the operator is running, though this is generally the simplest way to configure it. When you are running in single-namespace mode, the controller will ignore any clusters you try to create in namespaces other than the one you give it.
+To use single-namespace mode, set the `WATCH_NAMESPACE` environment variable or the command-line option `-watch-namespace` for the controller to be the namespace where your FDB clusters will run. It does not have to be the same namespace where the operator is running, though this is generally the simplest way to configure it. When you are running in single-namespace mode, the controller will ignore any clusters you try to create in namespaces other than the one you give it. If both options are defined, the environment variable will take precedence.
```
johscheuer, line 250: I would remove the env variable option in the docs and only mention the new cli flag.

setup/setup.go
```diff
@@ -146,7 +148,10 @@ func StartManager(
 
 	namespace := os.Getenv("WATCH_NAMESPACE")
 	if namespace != "" {
+		setupLog.Info("WATCH_NAMESPACE env variable will be deprecated.\nPlease use the command line option -watch-namespace instead")
 		options.Namespace = namespace
+	} else if operatorOpts.WatchNamespace != "" {
+		options.Namespace = operatorOpts.WatchNamespace
```
johscheuer, line 154: I think implementing this as a fallback is the better option e.g. if the cli argument is not set take the value from the env variable.

johscheuer (changes requested):
As part of this PR we should also change all the examples to use the new cli flag instead of the env variable (e.g. https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/config/samples/deployment.yaml#L128-L131)

docs/manual/customization.md
brownleej: The environment variable is useful because you can set it to specify the namespace the operator is running in via the downward API, which AFAIK is not possible for command-line arguments.

docs/manual/customization.md
```diff
@@ -247,23 +247,23 @@ Our [sample deployment](https://raw.githubusercontent.com/foundationdb/fdb-kuber
...
 
 The advantage of single-namespace mode is that it allows owners of different namespaces to run the operator themselves without needing access to other namespaces that may be managed by other tenants. The only cluster-level configuration it requires is the installation of the CRD. The disadvantage of single-namespace mode is that if you are running multiple namespaces for a single team, each namespace will need its own installation of the controller, which can make it more operationally challenging.
 
 To run the controller in single-namespace mode, you will need to configure the following things:
 
 * A service account for the controller
 * The serviceAccountName field in the controller's pod spec
-* A `WATCH_NAMESPACE` environment variable defined in the controller's pod spec
+* A `WATCH_NAMESPACE` environment variable defined in the controller's pod spec or in the arguments of the st
```
brownleej, line 258: Is this supposed to have more words?

docs/manual/customization.md
```diff
@@ -276,7 +276,7 @@ To run the controller in global mode, you will need to configure the following t
 
 You can build this kind of configuration easily from the sample deployment by changing the following things:
 
-* Delete the configuration for the `WATCH_NAMESPACE` variable
+* Delete the configuration for the `WATCH_NAMESPACE` variable or remove the command line option `-watch-namespace`.
```
brownleej, line 279: Since this is referring specifically to the sample deployment, I don't think we need to change this section of the docs until we change the sample deployment.

brownleej (changes requested):

docs/manual/customization.md
johscheuer: That makes sense and is a valid concern. One option would be to have the env variable `WATCH_NAMESPACE` as the default value for the `--watch-namespace` flag? That would support both the usage of the env variable and the specific passing of the new cli flag?

ltsampros: That makes sense to me.

johscheuer (commented):
Congratulations @ltsampros for the first passing pipeline :) 🎉

config/samples/deployment.yaml
```diff
@@ -122,6 +122,9 @@ spec:
       containers:
       - command:
         - /manager
+        args:
+        - "-watch-namespace"
+        - "$(WATCH_NAMESPACE)"
```
brownleej, line 127: I didn't know Kubernetes had first-class support for variable substitution in the args. That's neat!

docs/manual/customization.md
```diff
@@ -247,23 +247,23 @@ Our [sample deployment](https://raw.githubusercontent.com/foundationdb/fdb-kuber
...
+* A `WATCH_NAMESPACE` environment variable defined in the controller's pod spec or in the arguments of the container command
 * A Role that grants access to the necessary permissions to all of the resources that the controller manages. See the [sample role](https://raw.githubusercontent.com/FoundationDB/fdb-kubernetes-operator/master/config/samples/deployment/rbac_role.yaml) for the list of those permissions.
 * A RoleBinding that binds that role to the service account for the controller
 
 The sample deployment provides all of this configuration.
 
 ### Global Mode
 
-To use global mode, omit the `WATCH_NAMESPACE` environment variable for the controller. When you are running in global mode, the controller will watch for changes to FDB clusters in all namespaces, and will manage them all through a single instance of the controller.
+To use global mode, omit the `WATCH_NAMESPACE` environment variable or the `-watch-namespace` command line flag for the controller. When you are running in global mode, the controller will watch for changes to FDB clusters in all namespaces, and will manage them all through a single instance of the controller.
```
brownleej, line 266: Nitpick: I think this should say you have to omit the variable *and* the flag, rather than the variable *or* the flag.

brownleej (approved):
Other than a nitpick on wording, this looks good to me.

johscheuer (approved):
LGTM 👍

docs/manual/customization.md
ltsampros: Addressed.

Issue #735, opened 2021-06-02
Provide an issue template


johscheuer:
We should create an issue template to make it easier to file issues in that repository one rough idea would be adding some categories (could be used for automatic labels) like `bug`, `feature` and then the component `sidecar`, `plugin`, `cluster-controller`, etc. I'll write down a template this week as a starting point.

PR #1053, opened 2022-01-23
Add issue templates

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/735

## Type of change

- Documentation

# Discussion

I think those two templates are a good start and we can add more templates if we identify the need for it.

# Testing

-

# Documentation

-

# Follow-up

-


brownleej (approved):
👍

Issue #883, opened 2021-08-14
Allow in the analyze command to ignore given conditions


johscheuer:
I think it would be useful to have an option to ignore some conditions when running the analyze command. During an upgrade on a larger cluster it can get hard to see the process groups actually causing issue when all process groups have the `IncorrectCommandline` condition. To make that a little bit more precise the idea would be to just skip these conditions and don't do any further actions or print them.

PR #1054, opened 2022-01-24
Allow plugin to filter and ignore conditions

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/883

## Type of change

# Discussion

This feature is pretty useful to find the process group that blocks an operation e.g. during upgrades.

# Testing

local + unit

# Documentation

Added in the cli command.

# Follow-up

-

### DIFF
--- a/kubectl-fdb/cmd/analyze.go
+++ b/kubectl-fdb/cmd/analyze.go
@@ -70,6 +70,11 @@ func newAnalyzeCmd(streams genericclioptions.IOStreams) *cobra.Command {
 				return err
 			}
 
+			ignoreConditions, err := cmd.Flags().GetStringArray("ignore-condition")
+			if err != nil {
+				return err
+			}
+
 			if flagNoColor {
 				color.NoColor = true
 			}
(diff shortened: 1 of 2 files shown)


kubectl-fdb/cmd/analyze.go
```diff
@@ -70,6 +70,11 @@ func newAnalyzeCmd(streams genericclioptions.IOStreams) *cobra.Command {
 				return err
 			}
 
+			ignoreConditions, err := cmd.Flags().GetStringArray("ignore-condition")
```
JoshuaMcManus, line 73: Should we add a validation to the arguments such that it's within ProcessGroupConditionType.AllProcessGroupConditionTypes()?

JoshuaMcManus (commented):
One other suggestion that _might_ be useful is if we were to report how many conditions were ignored as one line in the output.

kubectl-fdb/cmd/analyze.go
johscheuer: I think that's a good idea and we should fail early if a condition is passed that is unknown.

brownleej (approved):
👍

JoshuaMcManus (approved):

Issue #968, opened 2021-10-28
Using DNS names in cluster files


brownleej:
There's a feature in development in FoundationDB to allow using DNS names in the cluster: https://github.com/apple/foundationdb/issues/354. This will be a big improvement on Kubernetes, due to the risk of coordinators losing their IPs when the pods are deleted. This issue will track the work to add basic support for this feature in the operator.

brownleej:
I think that the DNS name support is far enough along that we can close this.

PR #1055, opened 2022-01-25
Initial support for DNS names

brownleej:
# Description

This is the first pass at supporting the upcoming feature to use DNS namespace in the cluster file. This is not fully functional in this form, but since we're expecting more rounds of back-and-forth on the feature, I wanted to get the first pass merged in to unblock further testing and limit how many merge conflicts we accumulate.

This addresses part of #968, but we should discuss whether we want to keep that open or break it down into more manageable batches of work. 

## Type of change

# Discussion

During the testing for this, I hit a problem with the operator not being able to get the coordinator status through the client library. It worked fine through the CLI, which is why I added a workaround to get the status json through the CLI. That workaround is turned off in this build, but we can turn it back on for subsequent testing, until we are able to get the client library working properly.

# Testing

Manual testing?*

I was able to create a cluster in my dev environment using DNS names.

Not right away, but once we have this feature working on the main branch for both FDB and the operator, we'll want to add it to our integration suite.

# Documentation

I added it to the documentation for the cluster spec, but I don't think we need further documentation with the feature in its current state.

# Follow-up

We still need to resolve some bugs, as well as adding support for changing coordinators when using DNS names.

In some future release, we will likely want the DNS name feature to be on by default, but that will require further discussion.


api/v1beta1/foundationdbcluster_types.go
```diff
@@ -2893,6 +2901,17 @@ type RoutingConfig struct {
...
+	// UseDNSInClusterFile determines whether to use DNS names rather than IP
+	// addresses to identify coordinators in the cluster file.
+	// NOTE: This is an experimental feature, and is not supported in the
+	// latest stable version of FoundationDB.
+	UseDNSInClusterFile *bool `json:"useDNSInClusterFile,omitempty"`
+
+	// DNSSuffix defines the cluster domain used in a DNS name generated for a
+	// service.
+	// The default is `cluster.local`.
+	DNSDomain *string `json:"dnsSuffix,omitempty"`
```
johscheuer, line 2914: Can we add some validation marker here like the default, a max length e.g.:
```suggestion
	DNSDomain *string `json:"domain,omitempty"`
```

We can also call if `dnsDomain`.

```
+kubebuilder:validation:MinLength=1
+kubebuilder:validation:MaxLength=253 
```

and perhaps we want to add a `+kubebuilder:validation:Pattern="..."`we should take the DNS validation regex from Kubernetes to match the expectations: https://github.com/kubernetes/apimachinery/blob/release-1.23/pkg/util/validation/validation.go#199-L200 the cluster behaviour. We might also want to use some of those methods in our code later, if we have to do some validation at a place.

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -3172,3 +3191,21 @@ func (configuration *DatabaseConfiguration) FailOver() DatabaseConfiguration {
...
+	return cluster.UseDNSInClusterFile() || pointer.BoolDeref(cluster.Spec.Routing.HeadlessService, false)
+}
+
+// UseDNSInClusterFile determines whether we need to use DNS entries in the
+// clsuter file for this cluster.
+func (cluster *FoundationDBCluster) UseDNSInClusterFile() bool {
+	return pointer.BoolDeref(cluster.Spec.Routing.UseDNSInClusterFile, false)
+}
+
+// GetDNSDomain gets the suffix used when forming DNS names generated for a
```
johscheuer, line 3207:
```suggestion
// GetDNSDomain gets the domain used when forming DNS names generated for a
```

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -3172,3 +3191,21 @@ func (configuration *DatabaseConfiguration) FailOver() DatabaseConfiguration {
...
 }
+
+// NeedsHeadlessService determines whether we need to create a headless service
+// for this cluster.
+func (cluster *FoundationDBCluster) NeedsHeadlessService() bool {
+	return cluster.UseDNSInClusterFile() || pointer.BoolDeref(cluster.Spec.Routing.HeadlessService, false)
+}
+
+// UseDNSInClusterFile determines whether we need to use DNS entries in the
+// clsuter file for this cluster.
```
johscheuer, line 3202:
```suggestion
// cluster file for this cluster.
```

api/v1beta1/foundationdbcluster_types_test.go
```diff
@@ -2329,15 +2329,28 @@ var _ = Describe("[api] FoundationDBCluster", func() {
 					input: "127.0.0.1:bad",
 					err:   fmt.Errorf("strconv.Atoi: parsing \"bad\": invalid syntax"),
 				}),
-			Entry("IPv6 with invalid address",
+			Entry("host name",
 				testCase{
-					input: "[::1:]:4500",
```
johscheuer, line 2334: Can we leave those tests in place? Those are still invalid IP addresses and invalid hostnames.

api/v1beta1/foundationdbcluster_types_test.go
```diff
@@ -2329,15 +2329,28 @@ var _ = Describe("[api] FoundationDBCluster", func() {
...
-					input: "127.0.0.A:4500",
-					err:   fmt.Errorf("invalid address: 127.0.0.A:4500"),
+					input: "127.0.0.1:4501(fromHostname)",
+					expectedAddr: ProcessAddress{
+						IPAddress:    net.ParseIP("127.0.0.1"),
+						Port:         4501,
+						Flags:        nil,
+						FromHostname: true,
+					},
+					expectedStr: "127.0.0.1:4501(fromHostname)",
```
johscheuer, line 2352: If the provided input is an IP address port pair, why do we add the `(fromHostname)`? I would expect to either get an error or that the `(fromHostname)` suffix is removed/absent.

api/v1beta1/foundationdbcluster_types_test.go
```diff
@@ -3584,4 +3597,48 @@ var _ = Describe("[api] FoundationDBCluster", func() {
...
+		BeforeEach(func() {
+			cluster = &FoundationDBCluster{}
+		})
+
+		When("checking whether we need a headless service", func() {
+			It("respects the value in the flag", func() {
+				Expect(cluster.NeedsHeadlessService()).To(BeFalse())
+
+				var enable = true
+				cluster.Spec.Routing.HeadlessService = &enable
```
johscheuer, line 3612:
```suggestion
				cluster.Spec.Routing.HeadlessService = pointer.BoolPtr(true)
```

api/v1beta1/foundationdbcluster_types_test.go
```diff
@@ -3584,4 +3597,48 @@ var _ = Describe("[api] FoundationDBCluster", func() {
...
+				Expect(cluster.NeedsHeadlessService()).To(BeTrue())
+			})
+
+			It("can be overridden by the DNS setting", func() {
+				enableHeadless := false
+				enableDNS := true
+				cluster.Spec.Routing.HeadlessService = &enableHeadless
+				cluster.Spec.Routing.UseDNSInClusterFile = &enableDNS
```
johscheuer, line 3620:
```suggestion
				cluster.Spec.Routing.HeadlessService = pointer.BoolPtr(false)
				cluster.Spec.Routing.UseDNSInClusterFile = pointer.BoolPtr(true)
```

api/v1beta1/foundationdbcluster_types_test.go
```diff
@@ -3584,4 +3597,48 @@ var _ = Describe("[api] FoundationDBCluster", func() {
...
 	})
+
+	Describe("routing configuration", func() {
+		var cluster *FoundationDBCluster
+		BeforeEach(func() {
+			cluster = &FoundationDBCluster{}
+		})
+
+		When("checking whether we need a headless service", func() {
+			It("respects the value in the flag", func() {
```
johscheuer, line 3608:
```suggestion
			It("respects the headless service setting in the routing config", func() {
```

That is a little bit more verbose than before but also more precise instead of saying "flag".

api/v1beta1/foundationdbcluster_types_test.go
```diff
@@ -3584,4 +3597,48 @@ var _ = Describe("[api] FoundationDBCluster", func() {
...
+				Expect(cluster.NeedsHeadlessService()).To(BeTrue())
+			})
+		})
+
+		When("checking whether we use DNS in the cluster file", func() {
+			It("respects the value in the flag", func() {
+				Expect(cluster.UseDNSInClusterFile()).To(BeFalse())
+
+				enable := true
+				cluster.Spec.Routing.UseDNSInClusterFile = &enable
```
johscheuer, line 3630:
```suggestion
				cluster.Spec.Routing.UseDNSInClusterFile = pointer.BoolPtr(true)
```

api/v1beta1/foundationdbcluster_types_test.go
```diff
@@ -3584,4 +3597,48 @@ var _ = Describe("[api] FoundationDBCluster", func() {
...
+				Expect(cluster.UseDNSInClusterFile()).To(BeTrue())
+			})
+		})
+
+		When("getting the DNS domain", func() {
+			It("allows overrides in the spec", func() {
+				Expect(cluster.GetDNSDomain()).To(Equal("cluster.local"))
+				suffix := "cluster.example"
+				cluster.Spec.Routing.DNSDomain = &suffix
+				Expect(cluster.GetDNSDomain()).To(Equal("cluster.example"))
```
johscheuer, line 3640:
```suggestion
				suffix := "cluster.example"
				cluster.Spec.Routing.DNSDomain = &suffix
				Expect(cluster.GetDNSDomain()).To(Equal(suffix))
```

config/tests/dns/client.yaml
```diff
@@ -0,0 +1,90 @@
...
+apiVersion: v1
+kind: Service
+metadata:
+  name: test-cluster-client
+  labels:
+    app: test-cluster-client
+spec:
+  selector:
+    app: test-cluster-client
+  type: LoadBalancer
```
johscheuer, line 87: Do we want to expose this service with the LB type? I'm open to it but that would mean that in the background a LoadBalancer will be created.

config/tests/dns/dns.yaml
```diff
@@ -0,0 +1,8 @@
+- op: add
+  path: "/spec/version"
+  value: 7.1.0
```
johscheuer, line 3: We might want to make a note here that this version is yet not released and doesn't exist.

config/tests/dns/kustomization.yaml
```diff
@@ -0,0 +1,12 @@
+# This file provides a test cluster with TLS enabled.
```
johscheuer, line 1:
```suggestion
# This file provides a test cluster that uses hostnames in the cluster file.
```

controllers/change_coordinators.go
```diff
@@ -185,3 +186,18 @@ func selectCoordinators(cluster *fdbtypes.FoundationDBCluster, status *fdbtypes.
 
 	return coordinators, nil
 }
+
+func getCoordinatorAddress(cluster *fdbtypes.FoundationDBCluster, info localityInfo) fdbtypes.ProcessAddress {
```
johscheuer, line 190:
```suggestion
func getCoordinatorAddress(cluster *fdbtypes.FoundationDBCluster, info localityInfo) fdbtypes.ProcessAddress {
```

locality might be a better name here than `info` but I'm open to both names.

controllers/cluster_controller.go
```diff
@@ -640,31 +642,60 @@ func checkCoordinatorValidity(cluster *fdbtypes.FoundationDBCluster, status *fdb
...
-		if isCoordinator && !process.Excluded && !pendingRemoval {
-			coordinatorStatus[address] = true
+		coordinatorAddress := ""
+		_, isCoordinatorWithIP := coordinatorStatus[ipAddress.String()]
+		if isCoordinatorWithIP {
+			coordinatorAddress = ipAddress.String()
 		}
 
-		if isCoordinator {
+		dnsName := process.Locality["dns_name"]
```
johscheuer, line 659:
```suggestion
		dnsName := process.Locality["dns_name"]
```

Can we use the constant `FDBLocalityDNSNameKey` here?

controllers/cluster_controller.go
```diff
@@ -640,31 +642,60 @@ func checkCoordinatorValidity(cluster *fdbtypes.FoundationDBCluster, status *fdb
...
+			StringAddress: dnsName,
+			Port:          ipAddress.Port,
+			Flags:         ipAddress.Flags,
+		}
+		_, isCoordinatorWithDNS := coordinatorStatus[dnsAddress.String()]
+
+		if !isCoordinatorWithDNS {
+			dnsAddress = ipAddress
+			dnsAddress.FromHostname = true
+			_, isCoordinatorWithDNS = coordinatorStatus[dnsAddress.String()]
```
johscheuer, line 670: Is there a reason that we need the `127.0.0.1:4500(fromHostname)`? IMHO it doesn't make sense to have an IP address and define `fromHostname`.

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -1893,6 +1894,10 @@ func (address ProcessAddress) MarshalJSON() ([]byte, error) {
 // representation.
 func ParseProcessAddress(address string) (ProcessAddress, error) {
 	result := ProcessAddress{}
+	if strings.HasSuffix(address, "(fromHostname)") {
+		address = strings.TrimSuffix(address, "(fromHostname)")
+		result.FromHostname = true
```
johscheuer, line 1899: Do we really need this flag or should we just try to autodetect if the address is a DNS name or an IP address?

controllers/cluster_controller.go
```diff
@@ -640,31 +642,60 @@ func checkCoordinatorValidity(cluster *fdbtypes.FoundationDBCluster, status *fdb
...
+			coordinatorStatus[coordinatorAddress] = true
+		}
+
+		if coordinatorAddress != "" {
 			coordinatorZones[process.Locality[fdbtypes.FDBLocalityZoneIDKey]]++
 			coordinatorDCs[process.Locality[fdbtypes.FDBLocalityDCIDKey]]++
 
 			if !cluster.IsEligibleAsCandidate(process.ProcessClass) {
-				pLogger.Info("Process class of process is not eligible as coordinator", "class", process.ProcessClass, "address", address)
+				pLogger.Info("Process class of process is not eligible as coordinator", "class", process.ProcessClass, "address", ipAddress)
```
johscheuer, line 686:
```suggestion
				pLogger.Info("Process class of process is not eligible as coordinator", "class", process.ProcessClass, "address", coordinatorAddress)
```

controllers/cluster_controller.go
```diff
@@ -640,31 +642,60 @@ func checkCoordinatorValidity(cluster *fdbtypes.FoundationDBCluster, status *fdb
...
 				allEligible = false
 			}
+
+			useDNS := cluster.UseDNSInClusterFile() && dnsName != ""
+			if (isCoordinatorWithIP && useDNS) || (isCoordinatorWithDNS && !useDNS) {
+				pLogger.Info("Coordinator is not using the correct address", "coordinatorList", coordinatorStatus)
```
johscheuer, line 692:
```suggestion
				pLogger.Info("Coordinator is not using the correct address type", "coordinatorList", coordinatorStatus)
```

We might want to add thee expected address "type" to the log statement.

controllers/cluster_controller_test.go
```diff
@@ -3676,6 +3696,58 @@ var _ = Describe("cluster_controller", func() {
...
+					}
+
+					coordinatorsValid, addressesValid, err := checkCoordinatorValidity(cluster, status, coordinatorStatus)
+					Expect(err).NotTo(HaveOccurred())
+					Expect(coordinatorsValid).To(BeTrue())
+					Expect(addressesValid).To(BeTrue())
+				})
+			})
+
+			Context("when the pods have DNS named assigned", func() {
```
johscheuer, line 3719:
```suggestion
			Context("when the pods have DNS names assigned", func() {
```

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -2893,6 +2901,17 @@ type RoutingConfig struct {
...
 	// dual-stack support in your Kubernetes environment.
 	PodIPFamily *int `json:"podIPFamily,omitempty"`
+
+	// UseDNSInClusterFile determines whether to use DNS names rather than IP
+	// addresses to identify coordinators in the cluster file.
+	// NOTE: This is an experimental feature, and is not supported in the
+	// latest stable version of FoundationDB.
+	UseDNSInClusterFile *bool `json:"useDNSInClusterFile,omitempty"`
+
+	// DNSSuffix defines the cluster domain used in a DNS name generated for a
```
johscheuer, line 2911:
```suggestion
	// DNSDomain defines the cluster domain used in a DNS name generated for a
```

fdbclient/admin_client.go
```diff
@@ -217,9 +227,27 @@ func (client *cliAdminClient) runCommand(command cliCommand) (string, error) {
...
+	}
+	contents, err := client.runCommand(cliCommand{command: "status json"})
+	if err != nil {
+		return nil, err
+	}
+	log.V(1).Info("Fetched status JSON", "contents", contents)
+	startIndex := strings.Index(contents, "{")
+	if startIndex == -1 {
+		return nil, fmt.Errorf("Could not parse status JSON output")
+	}
```
johscheuer, line 244: You might want to use `removeWarningsInJSON` here instead.

internal/monitor_conf.go
```diff
@@ -270,6 +270,13 @@ func GetMonitorProcessConfiguration(cluster *fdbtypes.FoundationDBCluster, proce
 		configuration.Arguments = append(configuration.Arguments, monitorapi.Argument{Value: fmt.Sprintf("--locality_data_hall=%s", cluster.Spec.DataHall)})
 	}
 
+	if cluster.UseDNSInClusterFile() {
+		configuration.Arguments = append(configuration.Arguments, monitorapi.Argument{ArgumentType: monitorapi.ConcatenateArgumentType, Values: []monitorapi.Argument{
+			{Value: "--locality_dns_name="},
```
johscheuer, line 275: Is this an official locality that is required by FDB to make the DNS resolution work or is that only used by the operator?

internal/monitor_conf_test.go
```diff
@@ -570,18 +574,19 @@ var _ = Describe("pod_models", func() {
 			It("should substitute the variables in the start command", func() {
 				podClient, err := NewMockFdbPodClient(cluster, pod)
 				Expect(err).NotTo(HaveOccurred())
-				command, err = GetStartCommand(cluster, processClass, podClient, 1, 2)
+				command, err = GetStartCommand(cluster, "storage", podClient, 1, 2)
```
johscheuer, line 577:
```suggestion
				command, err = GetStartCommand(cluster, string(fdbtypes.ProcessClassStorage), podClient, 1, 2)
```

internal/monitor_conf_test.go
```diff
@@ -550,6 +550,10 @@ var _ = Describe("pod_models", func() {
 				command, err = GetStartCommand(cluster, processClass, podClient, 1, 1)
 				Expect(err).NotTo(HaveOccurred())
 
+				Expect(err).NotTo(HaveOccurred())
+				command, err = GetStartCommand(cluster, "storage", podClient, 1, 1)
```
johscheuer, line 554:
```suggestion
				command, err = GetStartCommand(cluster, string(fdbtypes.ProcessClassStorage), podClient, 1, 1)
```

internal/pod_models.go
```diff
@@ -554,6 +554,14 @@ func configureSidecarContainer(container *corev1.Container, initMode bool, proce
...
 		sidecarEnv = append(sidecarEnv, getEnvForMonitorConfigSubstitution(cluster, processGroupID)...)
 
+		if cluster.UseDNSInClusterFile() {
+			sidecarArgs = append(sidecarArgs, "--substitute-variable", "FDB_DNS_NAME")
+		}
+
+		if cluster.UseDNSInClusterFile() {
+			sidecarEnv = append(sidecarEnv, corev1.EnvVar{Name: "FDB_DNS_NAME", Value: GetPodDNSName(cluster, podName)})
+		}
+
```
johscheuer, line 564:
```suggestion
		if cluster.UseDNSInClusterFile() {
			sidecarArgs = append(sidecarArgs, "--substitute-variable", "FDB_DNS_NAME")
			sidecarEnv = append(sidecarEnv, corev1.EnvVar{Name: "FDB_DNS_NAME", Value: GetPodDNSName(cluster, podName)})
		}

```

In a follow up PR wee probably want to move all of those env strings into constants.

internal/pod_models.go
```diff
@@ -1007,3 +1015,8 @@ func GetObjectMetadata(cluster *fdbtypes.FoundationDBCluster, base *metav1.Objec
 
 	return *metadata
 }
+
+// GetPodDNSName determines the fully qualified DNS name for a pod.
+func GetPodDNSName(cluster *fdbtypes.FoundationDBCluster, podName string) string {
+	return fmt.Sprintf("%s.%s.%s.svc.%s", podName, cluster.Name, cluster.Namespace, cluster.GetDNSDomain())
```
johscheuer, line 1021: I wonder if we run in some possible issues here for longer Pod or cluster names (https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names) but I would guess we should be fine.

johscheuer (changes requested):

api/v1beta1/foundationdbcluster_types_test.go
brownleej: I think there were some issues with trying to validate the IP addresses when it needs to be able to accept host names as well, but I'll take another look at it.

api/v1beta1/foundationdbcluster_types_test.go
brownleej: This is a reflection of how the addresses are represented in the JSON status. @RenxuanW may be able to shed more light on the details there.

config/tests/dns/client.yaml
brownleej: For local testing purposes, having a load balancer service makes it easier to test that the service is behaving correctly, because we can make requests to it from the Mac side.

internal/monitor_conf.go
brownleej: It is only used by the operator.

api/v1beta1/foundationdbcluster_types.go
brownleej: I see the benefit in adding the validation on the pattern, but the pattern required is pretty messy, so I'd rather not embed it in the CRD. We may want to consider adding that logic in the code if it can make errors with the DNS Domain format more visible.

api/v1beta1/foundationdbcluster_types_test.go
brownleej: Because the process address can contain a DNS name in some contexts, we can't apply the same validations. We can introduce new validations to ensure it's a valid DNS name, but I don't know if that's worth the complexity.

config/tests/dns/dns.yaml
brownleej: I've added a note on that to the kustomization file.

api/v1beta1/foundationdbcluster_types_test.go
johscheuer: I think we don't want to have that complexity in our code we might want to leverage some validation features from the kube api-machinery for that.

api/v1beta1/foundationdbcluster_types_test.go
johscheuer: Okay, that makes sense 👍

config/tests/dns/client.yaml
johscheuer: 👍

config/tests/dns/dns.yaml
johscheuer: Thanks!

config/crd/bases/apps.foundationdb.org_foundationdbclusters.yaml
```diff
@@ -7923,12 +7923,19 @@ spec:
                   type: object
                 routing:
                   properties:
+                    dnsDomain:
+                      maxLength: 253
+                      minLength: 1
+                      pattern: '...'
```
johscheuer, line 7929: Is that pattern here expected?

internal/monitor_conf.go
johscheuer: I think we should document all "special" localities that the operator sets. That would be useful for others to e.g. query the status output.

johscheuer (approved):
LGTM 👍

config/crd/bases/apps.foundationdb.org_foundationdbclusters.yaml
brownleej: No, this was a missing change from the last commit.

Issue #743, opened 2021-06-05
Drop support for 6.1


johscheuer:
Given that the 6.1 release is now ~ 2years old with our any further releases/bug fixes (https://github.com/apple/foundationdb/tree/release-6.1) and 6.3 as the current (?) stable (?) version I would propose to drop the support for 6.1. Any concerns/vote against that?

I will cross post this in the forum and wait for the next release (2 weeks). If no concerns are raised I would vote for removal in the upcoming release (in 4 weeks for `v0.38.0`).

CC: @brownleej

brownleej:
I think this would be a good thing to do as part of the 1.0 release of the operator.

PR #1056, opened 2022-02-02
Only support 6.3.20+

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/743

## Type of change

# Discussion

I added the new compatibility to our docs. We might want to consider to remove some older `6.2` versions to reduce our if branching.

# Testing

local.

# Documentation

Added.

# Follow-up

-


api/v1beta1/foundationdb_version.go
```diff
@@ -208,8 +202,7 @@ var Versions = struct {
 	WithBinariesFromMainContainer:        FdbVersion{Major: 6, Minor: 2, Patch: 15},
 	WithoutBinariesFromMainContainer:     FdbVersion{Major: 6, Minor: 2, Patch: 11},
 	WithRatekeeperRole:                   FdbVersion{Major: 6, Minor: 2, Patch: 15},
-	WithoutRatekeeperRole:                FdbVersion{Major: 6, Minor: 1, Patch: 12},
 	WithSidecarCrashOnEmpty:              FdbVersion{Major: 6, Minor: 2, Patch: 20},
 	WithoutSidecarCrashOnEmpty:           FdbVersion{Major: 6, Minor: 2, Patch: 15},
-	MinimumVersion:                       FdbVersion{Major: 6, Minor: 1, Patch: 12},
+	MinimumVersion:                       FdbVersion{Major: 6, Minor: 2, Patch: 0},
```
johscheuer, line 207: I think a good minimum version would be `6.2.20` that would allow us to remove nearly all of the version specific behaviour.

docs/compatibility.md
```diff
@@ -23,9 +23,10 @@ versions of related services that each operator version is compatible with. The
 "Most Recent Version" column shows the last version of the operator that was
 published for each major version.
 
-| Operator Version  | Most Recent Version | Supported Cluster Models  | Supported FDB Versions  | Supported Kubernetes Versions |
-| ----------------- | ------------------- | ------------------------- | ----------------------- | ----------------------------- |
-| 0.x               | 0.51.0              | v1beta1                   | 6.1.12+                 | 1.15.0+                       |
+| Operator Version | Most Recent Version | Supported Cluster Models  | Supported FDB Versions | Supported Kubernetes Versions |
+|------------------|---------------------| ------------------------- |-----------------------|-------------------------------|
+| 0.* - 0.51.0     | 0.51.0              | v1beta1                   | 6.1.12+               | 1.15.0+                       |
+| 0.52.0+          | -                   | v1beta1                   | 6.2.0+                | 1.19.0+                       |
```
johscheuer, line 29: I decided to increase the Kubernetes version to `1.19` since our tests are not testing against `1.15` so we might not support that version.

api/v1beta1/foundationdb_version.go
johscheuer: The minimum version should be 6.3.22, try to find the documentation in the forum and document it.

docs/compatibility.md
```diff
@@ -23,9 +23,10 @@ versions of related services that each operator version is compatible with. The
 "Most Recent Version" column shows the last version of the operator that was
 published for each major version.
 
-| Operator Version  | Most Recent Version | Supported Cluster Models  | Supported FDB Versions  | Supported Kubernetes Versions |
-| ----------------- | ------------------- | ------------------------- | ----------------------- | ----------------------------- |
-| 0.x               | 0.51.0              | v1beta1                   | 6.1.12+                 | 1.15.0+                       |
```
brownleej, line 28: I think that we should keep this table with one entry per major version of the operator, in order to keep things simple for users of the operator. That means that we need to commit to making the changes in this PR part of a major release and retain compatibility with older versions of FDB in the 0.x family.

internal/monitor_conf_test.go
```diff
@@ -1113,8 +1086,8 @@ var _ = Describe("monitor_conf", func() {
 
 		Context("with a version that can use binaries from the main container", func() {
 			BeforeEach(func() {
-				cluster.Spec.Version = fdbtypes.Versions.WithBinariesFromMainContainer.String()
```
brownleej, line 1116: Do we still need a special test case for this if it is the only supported behavior?

internal/pod_models_test.go
```diff
@@ -2079,7 +2079,7 @@ var _ = Describe("pod_models", func() {
 
 		Context("with command line arguments for the sidecar", func() {
 			BeforeEach(func() {
-				cluster.Spec.Version = fdbtypes.Versions.WithCommandLineVariablesForSidecar.String()
+				cluster.Spec.Version = fdbtypes.Versions.Default.String()
```
brownleej, line 2082: Do we still need a special test case for this if it is the only supported behavior?

internal/pod_models_test.go
```diff
@@ -3144,7 +3081,7 @@ var _ = Describe("pod_models", func() {
 
 		Context("with the sidecar require-not-empty field", func() {
 			BeforeEach(func() {
-				backup.Spec.Version = fdbtypes.Versions.WithSidecarCrashOnEmpty.String()
+				backup.Spec.Version = fdbtypes.Versions.Default.String()
```
brownleej, line 3084: Another test case we may not need.

brownleej (changes requested):

PR #1057, opened 2022-02-04
Correct selector for all Pods

johscheuer:
# Description

This change ensures we only select Pods with the expected set of labels even if we pass an empty string, we still ensure that the label key is present.

## Type of change

# Discussion

-

# Testing

Local.

# Documentation

-

# Follow-up

-

### DIFF
--- a/internal/pod_helper.go
+++ b/internal/pod_helper.go
@@ -25,6 +25,8 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
 	"net"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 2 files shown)


internal/pod_helper.go
```diff
@@ -127,44 +130,78 @@ func GetJSONHash(object interface{}) (string, error) {
...
+	return matchLabels
+}
+
+func generateSelector(cluster *fdbtypes.FoundationDBCluster, processClass fdbtypes.ProcessClass, processGroupID string) labels.Selector {
+	matchLabels := GetPodMatchLabels(cluster, processClass, processGroupID)
+	s := labels.SelectorFromSet(matchLabels)
+
+	// If the process class is not specified we still want to ensure that the Pod has
+	// a label with the that key.
+	if _, ok := matchLabels[cluster.GetProcessClassLabel()]; !ok {
```
brownleej, line 179: This feels like it's accomplishing something indirect, and I wonder if it is better to achieve this by the user setting the match labels differently.

johscheuer: Since the `GetPodMatchLabels` returns a map I'm not sure if we could do something better here but I also don't see an issue with the current approach. The problem would be if we add the according label with an empty value the selector will expect an empty value too.

johscheuer: Sounds like a valid approach to me 👍

Issue #856, opened 2021-07-15
FDB upgrades should allow to migrate transaction system


johscheuer:
In the current implementation of upgrade we delete and recreate all Pods after the initial upgrade to ensure all Pods have the expected image. This has the drawback that this process will cause multiple recoveries if a Pod from the transaction system is deleted and recreated. Normally these recoveries shouldn't cause issues and the recovery should be very fast. One way to reduce the number of recoveries would be by migrating the transaction system that means once the cluster has done the initial upgrade we spin up a new transaction system or a batch (should be configurable) of new Pods and then exclude and delete the old once. This has the benefit that we only cause "number of batches" + 1 recoveries.

brownleej:
I think there's now a few different places where we are configuring different scenarios where we do a migration vs a rolling bounce, and we should think about how to unify those in a way that makes it more discoverable and flexible.

PR #1058, opened 2022-02-04
Replace transaction system

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/856

## Type of change

# Discussion

This setting will affect all PodSpec changes and not only upgrades but I think this is a more consistent behaviour.

# Testing

Local testing.

# Documentation

Updated.

# Follow-up

-

### DIFF
--- a/api/v1beta1/foundationdbcluster_types.go
+++ b/api/v1beta1/foundationdbcluster_types.go
@@ -3026,7 +3026,7 @@ func (pClass ProcessClass) IsStateful() bool {
 
 // IsTransaction determines whether a process class is part of the transaction system.
 func (pClass ProcessClass) IsTransaction() bool {
-	return pClass == ProcessClassStateless || pClass == ProcessClassLog || pClass == ProcessClassTransaction
+	return pClass != ProcessClassStorage && pClass != ProcessClassGeneral
 }
 
 // AddStorageServerPerDisk adds serverPerDisk to the status field to keep track which ConfigMaps should be kept
--- a/api/v1beta1/foundationdbcluster_types.go
+++ b/api/v1beta1/foundationdbcluster_types.go
@@ -1216,7 +1216,7 @@ type FoundationDBClusterAutomationOptions struct {
 	MaxConcurrentReplacements *int `json:"maxConcurrentReplacements,omitempty"`
 
 	// DeletionMode defines the deletion mode for this cluster. This can be
-	// PodUpdateModeAll, PodUpdateModeZone or PodUpdateModeProcessGroup. The
+	// PodUpdateModeNone, PodUpdateModeAll, PodUpdateModeZone or PodUpdateModeProcessGroup. The
 	// DeletionMode defines how Pods are deleted in order to update them or
 	// when they are removed.
 	// +kubebuilder:validation:Optional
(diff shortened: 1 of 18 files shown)


controllers/cluster_controller_test.go
```diff
@@ -2218,7 +2218,119 @@ var _ = Describe("cluster_controller", func() {
...
+
+					currentNames := make([]string, 0, len(originalPods.Items))
+					for _, pod := range pods.Items {
+						class := pod.Labels[fdbtypes.FDBProcessClassLabel]
+						if !fdbtypes.ProcessClass(class).IsTransaction() {
+							continue
+						}
+
+						currentNames = append(currentNames, pod.Name)
+					}
```
johscheuer, line 2278: I will add an additional test statement to check that set storage Pods are the same.

brownleej (changes requested):

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -194,11 +194,12 @@ type FoundationDBClusterSpec struct {
 	// This must be a valid Kubernetes label value. See
 	// https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set
 	// for more details on that.
-	// +kubebuilder:validation:MaxLength=32
+	// +kubebuilder:validation:MaxLength=64
```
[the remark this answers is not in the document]
johscheuer: Oh, that is an unrelated issue and should be in another PR. I'll move that out.

internal/deprecations.go
```diff
@@ -403,6 +403,10 @@ func NormalizeClusterSpec(cluster *fdbtypes.FoundationDBCluster, options Depreca
 		cluster.Spec.ProcessGroupIDPrefix = cluster.Spec.InstanceIDPrefix
 	}
 
+	if cluster.Spec.UpdatePodsByReplacement {
```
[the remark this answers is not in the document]
johscheuer: This is a boolean value so unsetting the value would be setting it to false?

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -3016,6 +3024,11 @@ func (pClass ProcessClass) IsStateful() bool {
 	return pClass == ProcessClassStorage || pClass == ProcessClassLog || pClass == ProcessClassTransaction
 }
 
+// IsTransaction determines whether a process class is part of the transaction system.
+func (pClass ProcessClass) IsTransaction() bool {
+	return pClass == ProcessClassStateless || pClass == ProcessClassLog || pClass == ProcessClassTransaction
```
johscheuer, line 3029: @brownleej do you think it make sense to invert this logic to have something like: `pClass == ProcessClassStorage || pClass == ProcessClassGeneral`? Otherwise we might not replacing a Pods that services a transaction system process, like the cluster controller. Another option is to review the selection logic from FDB and add all possible roles here (except for storage).

brownleej: Yes, I think that makes sense.

brownleej (approved):
👍

PR #1064, opened 2022-02-15
Check if processes are excluded before run exclude command in CanSafelyRemove

johscheuer:
# Description

This change adds support for only checking with the `exclude` command for already excluded addresses if they are fully excluded and if they can be removed from the cluster. Before this change it could happen that we invoked the `CanSafelyRemove` method before an address was excluded in the exclude sub reconciler (which has some additional age checks). In order to be more restrictive and be safer I added a check to only run the exclude command for processes that are marked as excluded in the cluster status.

## Type of change

# Discussion

# Testing

local testing and I will add some more unit tests.

# Documentation

-

# Follow-up

I will create a design doc for some ideas how to improve the exclusion handling in the operator.


brownleej (approved):
👍

PR #1070, opened 2022-02-22
Ignoring terminating pods after ignore limit

09harsh:
# Description

Added a new setting which will ignore the Terminating pods after the ignore limit. This will make the operator to proceed further and don't block the reconciliation.

## Type of change

[x]New feature (non-breaking change which adds functionality)
- Documentation
- Other

# Discussion

No

# Testing

Manual testing?*
Added Unit Tests

No

# Documentation

Yes

# Follow-up

No 
No


controllers/update_pods.go
```diff
@@ -154,6 +156,11 @@ func (updatePods) reconcile(ctx context.Context, r *FoundationDBClusterReconcile
 	return deletePodsForUpdates(ctx, r, cluster, adminClient, updates, logger)
 }
 
+func shouldRequeueDueToTerminatingPod(pod *corev1.Pod, cluster *fdbtypes.FoundationDBCluster) bool {
```
johscheuer, line 159: In general I would say that this method is too small to be factored out into something separate. My assumption is you only factored it out to have a unit test for this, if so I would prefer to move the `cluster.ProcessGroupIsBeingRemoved(processGroup.ProcessGroupID)` check into this method.

johscheuer (changes requested):
Please take a look after the test failures.

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -3099,6 +3104,15 @@ func (cluster *FoundationDBCluster) GetIgnorePendingPodsDuration() time.Duration
 	return cluster.Spec.AutomationOptions.IgnorePendingPodsDuration
 }
 
+// GetIgnoreTerminatingPodsDuration returns the value of IgnoreTerminatingPodsDuration or 5 minutes if unset.
```
ltsampros, line 3107: I see 10 minutes set returned as the default below.

docs/cluster_spec.md
```diff
@@ -189,7 +189,8 @@ FoundationDBClusterAutomationOptions provides flags for enabling or disabling op
 | killProcesses | KillProcesses defines whether the operator is allowed to bounce fdbserver processes. | *bool | false |
 | deletePods | DeletePods defines whether the operator is allowed to delete pods in order to recreate them. **Deprecated: Use DeletionMode with PodUpdateModeNone to prevent the operator from deleting Pods.** | *bool | false |
 | replacements | Replacements contains options for automatically replacing failed processes. | [AutomaticReplacementOptions](#automaticreplacementoptions) | false |
-| ignorePendingPodsDuration | IgnorePendingPodsDuration defines how long a Pod has to be in the Pending Phase before ignore it during reconciliation. This prevents Pod that are stuck in Pending to block further reconciliation. | time.Duration | false |
+| ignorePendingPodsDuration | IgnorePendingPodsDuration defines how long a Pod has to be in the Pending Phase before we ignore it during reconciliation. This prevents Pod that are stuck in Pending to block further reconciliation. | time.Duration | false |
```
ltsampros, line 192: I think the defaults should be different.

ltsampros (commented):
only doc updates

controllers/update_pods.go
09harsh: I have moved this in the `shouldRequeueDueToTerminatingPod()` function although I didn't added any additional test cases for that as we already have test cases to validate `cluster.ProcessGroupIsBeingRemoved`

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -1189,10 +1189,15 @@ type FoundationDBClusterAutomationOptions struct {
...
 	// IgnorePendingPodsDuration defines how long a Pod has to be in the Pending Phase before
-	// ignore it during reconciliation. This prevents Pod that are stuck in Pending to block
+	// we ignore it during reconciliation. This prevents Pod that are stuck in Pending to block
 	// further reconciliation.
 	IgnorePendingPodsDuration time.Duration `json:"ignorePendingPodsDuration,omitempty"`
 
+	// IgnoreTerminatingPodsDuration defines how long a Pod has to be in the Terminating Phase before
+	// we ignore it during reconciliation. This prevents Pod that are stuck in Terminating to block
+	// further reconciliation.
+	IgnoreTerminatingPodsDuration *int `json:"ignoreTerminatingPodsDuration,omitempty"`
```
johscheuer, line 1199:
```suggestion
	IgnoreTerminatingPodsSeconds *int `json:"ignoreTerminatingPodsSeconds,omitempty"`
```

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -3099,6 +3104,11 @@ func (cluster *FoundationDBCluster) GetIgnorePendingPodsDuration() time.Duration
 	return cluster.Spec.AutomationOptions.IgnorePendingPodsDuration
 }
 
+// GetIgnoreTerminatingPodsDuration returns the value of IgnoreTerminatingPodsDuration or defaults to 10 minutes.
+func (cluster *FoundationDBCluster) GetIgnoreTerminatingPodsDuration() int {
+	return pointer.IntDeref(cluster.Spec.AutomationOptions.IgnoreTerminatingPodsDuration, int(10*time.Minute.Nanoseconds()))
```
johscheuer, line 3109: Is there a reason we use nanoseconds here?

controllers/update_pods.go
```diff
@@ -154,6 +155,12 @@ func (updatePods) reconcile(ctx context.Context, r *FoundationDBClusterReconcile
 	return deletePodsForUpdates(ctx, r, cluster, adminClient, updates, logger)
 }
 
+func shouldRequeueDueToTerminatingPod(pod *corev1.Pod, cluster *fdbtypes.FoundationDBCluster, processGroup string) bool {
```
johscheuer, line 158:
```suggestion
func shouldRequeueDueToTerminatingPod(pod *corev1.Pod, cluster *fdbtypes.FoundationDBCluster, processGroupID string) bool {
```

johscheuer (changes requested):
Could we change the wording of the setting like initially proposed? And it seems there is still a race condition in one of the tests: https://github.com/FoundationDB/fdb-kubernetes-operator/runs/5374739914?check_suite_focus=true

api/v1beta1/foundationdbcluster_types.go
09harsh: I misinterpreted your previous comment and again added nanosecond here, will make the change.

09harsh:
test this please

controllers/update_pods.go
```diff
@@ -154,6 +155,12 @@ func (updatePods) reconcile(ctx context.Context, r *FoundationDBClusterReconcile
 	return deletePodsForUpdates(ctx, r, cluster, adminClient, updates, logger)
 }
 
+func shouldRequeueDueToTerminatingPod(pod *corev1.Pod, cluster *fdbtypes.FoundationDBCluster, processGroupID string) bool {
+	return pod.DeletionTimestamp != nil &&
+		pod.DeletionTimestamp.Add(time.Duration(cluster.GetIgnoreTerminatingPodsSeconds())).After(time.Now()) &&
```
johscheuer, line 160:
```suggestion
		pod.DeletionTimestamp.Add(cluster.GetIgnoreTerminatingPodsSeconds() * time.Second).After(time.Now()) &&
```

controllers/update_pods_test.go
```diff
@@ -106,4 +107,72 @@ var _ = Describe("update_pods", func() {
 				}),
 		)
 	})
+
+	Context("Validating isPendingDeletion", func() {
+		var ignoreTerminatingPodsSeconds = int(5 * time.Minute.Nanoseconds())
```
johscheuer, line 112: We use this value only once. We should code that value directly.

johscheuer (changes requested):

controllers/update_pods.go
09harsh: I don't think this change will work, as `.Add()` expects `time.Duration` and multiplying it with `time.Second` doesn't change the type. What is the issue with the current approach. IMHO it seems more readable.

controllers/update_pods.go
```diff
@@ -154,6 +155,12 @@ func (updatePods) reconcile(ctx context.Context, r *FoundationDBClusterReconcile
 	return deletePodsForUpdates(ctx, r, cluster, adminClient, updates, logger)
 }
 
+func shouldRequeueDueToTerminatingPod(pod *corev1.Pod, cluster *fdbtypes.FoundationDBCluster, processGroupID string) bool {
+	return pod.DeletionTimestamp != nil &&
+		pod.DeletionTimestamp.Add(time.Duration(cluster.GetIgnoreTerminatingPodsSeconds())*time.Second).After(time.Now()) &&
```
johscheuer, line 160:
```suggestion
		pod.DeletionTimestamp.Add(cluster.GetIgnoreTerminatingPodsSeconds()*time.Second).After(time.Now()) &&
```

This will already return a `time.Duration`

controllers/update_pods_test.go
johscheuer: Using `pointer.Int()` will solve that for you.

johscheuer (commented):
One comment otherwise it looks good.

johscheuer (approved):

johscheuer (approved):
Please run `goimports`-ed` and push those changes once the pipeline passes feel free to merge.

brownleej (approved):
👍

09harsh:
test this please

Issue #1067, opened 2022-02-17
Create a debug image of the operator


johscheuer:
### What would you like to be added/changed?

I would like to create and push an additional image for debugging the operator. This image should make use of a base image with a bash that allows to exec into the container to run some debugging steps.

brownleej:
We should also use this image variant by default in the local testing environment. I've found myself having to change the Dockerfile temporarily during testing in order to debug issues.

PR #1079, opened 2022-03-04
Add slim image

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1067 IO also fixed a bug that the latest image was not published anymore when we changed from `master` to `main`.

## Type of change

# Discussion

This adds a new slim image (which was the default previously) and the new default image will have centos8 as base image that allows people easier to debug any issues.

# Testing

I currently have some tests running in GitHub and I ensured that the new image runs with centos8

# Documentation

-

# Follow-up

-

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -105,7 +105,6 @@ jobs:
         make test
   build_images:
     if: github.ref != 'refs/heads/main'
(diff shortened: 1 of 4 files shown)


09harsh (approved):
👍

Issue #1071, opened 2022-02-22
Investigate e2e test failures 


johscheuer:
### What happened?

Currently we see a lot of failing e2e tests in our PRs we have to look into our e2e pipeline and setup to see what is happening and why those tests are failing.

### What did you expect to happen?

That our tests are not that flaky.

### How can we reproduce it (as minimally and precisely as possible)?

Create a PR and wait for the test result.

### FDB Kubernetes operator

-

### Kubernetes version

-

### Cloud provider

-

johscheuer:
I think I found the issue. It's a bug In the new exclusion code.

PR #1090, opened 2022-03-07
Fix the handling of missing processes in status for safe removes

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1071

## Type of change

# Discussion

In the new code was a race condition, when a Pod was deleted and the process was absent from the cluster status we added it to the `notExcludedAddresses` instead we should add missing processes to the `markedAsExcluded` to ensure that those processes data are replicated. This leaves a small risk that a Pod changes it's address and then the process get's excluded even when the operator didn't trigger the exclusion but I think that risk is fairly small and still better than the previous behaviour were we always excluded all processes when we called the `CanSafelyRemove` method.

# Testing

Unit tests + e2e test run.

# Documentation

I create an issue to better document the exclusion process in the operator.

# Follow-up

-


09harsh (approved):

johscheuer:
Result of my "local" run:

```bash
Ran 22 of 26 Specs in 5257.540 seconds
SUCCESS! -- 22 Passed | 0 Failed | 4 Pending | 0 Skipped
PASS

Ginkgo ran 1 suite in 1h27m43.086131493s
Test Suite Passed
```

johscheuer:
It looks like the namespace of the PR run was deleted by the one that finished earlier.

PR #1094, opened 2022-03-08
Remove the watch namespace flag from the sample deployment.

### LINKED ISSUE #1081: Current deployment main is broken
### What happened?

When installing the operator with `kubectl apply -f https://raw.githubusercontent.com/foundationdb/fdb-kubernetes-operator/master/config/samples/deployment.yaml` the operator deployment will be in a crash loop:

```bash
kubectl logs -f fdb-kubernetes-operator-controller-manager-b7cd98c84-f2grt
flag provided but not defined: -watch-namespace
Usage of /manager:
  -cleanup-old-cli-logs
        Defines if the operator should delete old fdbcli log files. (default true)
  -cli-timeout int
        The timeout to use for CLI commands. (default 10)
  -compress
        Defines whether the rotated log files should be compressed using gzip or not.
  -enable-leader-election
        Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager. (default true)
  -kubeconfig string
        Paths to a kubeconfig. Only required if out-of-cluster.
  -label-selector string
        Defines a label-selector that will be used to select resources.
  -leader-election-id string
        LeaderElectionID determines the name of the resource that leader election will use for holding the leader lock. (default "fdb-kubernetes-operator")
  -log-file string
        The path to a file to write logs to.
  -log-file-max-age int
        Defines the maximum age to retain old operator log file in number of days. (default 28)
  -log-file-max-size int
        Defines the maximum size in megabytes of the operator log file before it gets rotated. (default 250)
  -log-file-min-age duration
        Defines the minimum age of fdbcli log files before removing when "--cleanup-old-cli-logs" is set. (default 5m0s)
  -max-concurrent-reconciles int
        Defines the maximum number of concurrent reconciles for all controllers. (default 1)
  -max-old-log-files int
        Defines the maximum number of old operator log files to retain. (default 3)
  -metrics-addr string
        The address the metric endpoint binds to. (default ":8080")
  -use-future-defaults
        Apply defaults from the next major version of the operator. This is only intended for use in development.
  -version
        Prints the version of the operator and exits.
  -zap-devel
        Development Mode defaults(encoder=consoleEncoder,logLevel=Debug,stackTraceLevel=Warn). Production Mode defaults(encoder=jsonEncoder,logLevel=Info,stackTraceLevel=Error)
  -zap-encoder value
        Zap log encoding (one of 'json' or 'console')
  -zap-log-level value
        Zap Level to configure the verbosity of logging. Can be one of 'debug', 'info', 'error', or any integer value > 0 which corresponds to custom debug levels of increasing verbosity
  -zap-stacktrace-level value
        Zap Level at and above which stacktraces are captured (one of 'info', 'error', 'panic').
```

The flag was introduced in https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1046 and all deployments were. adjusted but the issue is that those deployments refer to a tag version of the operator. 

I think it would be more consistent to use `latest` in our deployments from the main branch and only use a specific version in our release tags. Otherwise we potentially install a newer CRD with features that are not supported in that operator deployment.

That the following steps bring up the operator in a running state:

```bash
kubectl apply -f https://raw.githubusercontent.com/FoundationDB/fdb-kubernetes-operator/master/config/crd/bases/apps.foundationdb.org_foundationdbclusters.yaml
kubectl apply -f https://raw.githubusercontent.com/FoundationDB/fdb-kubernetes-operator/master/config/crd/bases/apps.foundationdb.org_foundationdbbackups.yaml
kubectl apply -f https://raw.githubusercontent.com/FoundationDB/fdb-kubernetes-operator/master/config/crd/bases/apps.foundationdb.org_foundationdbrestores.yaml
kubectl apply -f https://raw.githubusercontent.com/foundationdb/fdb-kubernetes-operator/master/config/samples/deployment.yaml
```

Run:

```bash
kubectl apply -f https://raw.githubusercontent.com/FoundationDB/fdb-kubernetes-operator/master/config/crd/bases/apps.foundationdb.org_foundationdbclusters.yaml
kubectl apply -f https://raw.githubusercontent.com/FoundationDB/fdb-kubernetes-operator/master/config/crd/bases/apps.foundationdb.org_foundationdbbackups.yaml
kubectl apply -f https://raw.githubusercontent.com/FoundationDB/fdb-kubernetes-operator/master/config/crd/bases/apps.foundationdb.org_foundationdbrestores.yaml
kubectl apply -f https://raw.githubusercontent.com/foundationdb/fdb-kubernetes-operator/master/config/samples/deployment.yaml
```

-

### FDB Kubernetes operator

-

### Kubernetes version

-

### Cloud provider

-

brownleej:
# Description

This removes a flag from our sample deployments that isn't supported in the last stable release. This flag doesn't have any effect in the context where it was being used, because it was being set to the environment variable that is used as the default when the flag is omitted.

Resolves #1081 

## Type of change

# Discussion

I considered changing the version of the operator in the deployment to use the latest tag, but we would want to revert that in the future, and I don't think we need this flag in the sample deployment.

# Testing

Manual testing?*

I manually deployed the config in my local environment.

No.

# Documentation

No.

# Follow-up

No.

No.

### DIFF
--- a/config/samples/deployment.yaml
+++ b/config/samples/deployment.yaml
@@ -120,10 +120,7 @@ spec:
         control-plane: controller-manager
     spec:
       containers:
-      - args:
-        - -watch-namespace
-        - $(WATCH_NAMESPACE)
-        command:
+      - command:
         - /manager
         env:
         - name: WATCH_NAMESPACE
--- a/config/samples/deployment/manager.yaml
+++ b/config/samples/deployment/manager.yaml
@@ -66,9 +66,6 @@ spec:
       containers:
       - command:
         - /manager
-        args:
-        - "-watch-namespace"
-        - "$(WATCH_NAMESPACE)"
         image: foundationdb/fdb-kubernetes-operator:v0.51.0
         name: manager
         env:


ltsampros (approved):

johscheuer (approved):
Thanks for doing that 👍 Could you open an issue to add this flag once 1.0 is released? Otherwise we might forget about it.

johscheuer:
Issue for reference: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1095

PR #1099, opened 2022-03-10
Fix the namespace parsing in flags

johscheuer:
# Description

Using `os.Getenv` is not supported in the flag package I noticed this during testing. The operator always starts in the global mode even when we set that env variable.

## Type of change

# Discussion

-

# Testing

Local test setup.

# Documentation

-

# Follow-up

-


johscheuer:
Just noticed that the env variable was vanished in my deployment.

PR #1111, opened 2022-03-13
Pass configured logger into fdbclient

johscheuer:
# Description

This removes the creation of a new (not configured) logger in the fdbclient package instead we pass down the configured logger from our setup method. Now the status printout is correctly identified as debug log:

```
{"level":"info","ts":1647165605.8607554,"logger":"fdbclient","msg":"Fetch status from FDB","namespace":"default","cluster":"test-cluster"}
{"level":"debug","ts":1647165605.8626332,"logger":"controller-runtime.manager.events","msg":"Normal","object":{"kind":"FoundationDBCluster","namespace":"default","name":"test-cluster","uid":"876e80a3-d0ac-4174-9b8b-0f101401e9d4","apiVersion":"apps.foundationdb.org/v1beta2","resourceVersion":"662619"},"reason":"RemovingProcesses","message":"Removing pods: []"}
{"level":"debug","ts":1647165606.1626394,"logger":"fdbclient","msg":"Retrieved JSON status","raw":"{\"client\": ... 
```

Before all logs where handled as `info` and printed our irrespective of their actual level. This will reduce the noise for the `info` level.

## Type of change

# Discussion

-

# Testing

Local.

# Documentation

-

# Follow-up

https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1112

### DIFF
--- a/fdbclient/admin_client.go
+++ b/fdbclient/admin_client.go
@@ -32,6 +32,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/go-logr/logr"
+
 	fdbv1beta2 "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta2"
 	"github.com/FoundationDB/fdb-kubernetes-operator/internal"
 	"github.com/FoundationDB/fdb-kubernetes-operator/pkg/fdbadminclient"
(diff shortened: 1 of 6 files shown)


fdbclient/admin_client.go
```diff
@@ -229,13 +235,13 @@ func (client *cliAdminClient) GetStatus() (*fdbv1beta2.FoundationDBStatus, error
...
 		// This will call directly the database and fetch the status information
 		// from the system key space.
-		return getStatusFromDB(client.Cluster)
+		return getStatusFromDB(client.Cluster, client.log)
 	}
 	contents, err := client.runCommand(cliCommand{command: "status json"})
 	if err != nil {
 		return nil, err
 	}
-	log.V(1).Info("Fetched status JSON", "contents", contents)
```
ltsampros, line 238: Could we get rid of this or increase the level to debug?

ltsampros (approved):

fdbclient/admin_client.go
johscheuer: This was already at `debug` level, it was just ignored because we didn't passed a configured logger (see the example log output).

PR #1120, opened 2022-03-16
Recreating pods even if they are marked for removal.

aasmundeldhuset:
Is it as simple as this?


aasmundeldhuset:
Apologies; I meant to create this in my fork, not in upstream.

Issue #1078, opened 2022-03-04
Investigate unit test flakiness in "remove instances command"


09harsh:
### What happened?

After raising [PR](https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1070) for some other component this test failed. On rerunning the checks the test passed on its own. We need to investigate the root cause for this flakiness. 

```
• Failure [0.013 seconds]
66
[plugin] remove instances command
67
/home/runner/work/fdb-kubernetes-operator/fdb-kubernetes-operator/kubectl-fdb/cmd/remove_process_group_test.go:40
68
  when running remove instances command
69
  /home/runner/work/fdb-kubernetes-operator/fdb-kubernetes-operator/kubectl-fdb/cmd/remove_process_group_test.go:60
70
    when removing instances from a cluster
71
    /home/runner/work/fdb-kubernetes-operator/fdb-kubernetes-operator/kubectl-fdb/cmd/remove_process_group_test.go:133
72
      should cordon all targeted processes
73
      /home/runner/work/fdb-kubernetes-operator/fdb-kubernetes-operator/kubectl-fdb/cmd/remove_process_group_test.go:260
74
        Remove all failed instances [It]
75
        /home/runner/work/fdb-kubernetes-operator/fdb-kubernetes-operator/kubectl-fdb/cmd/remove_process_group_test.go:330
76

77
        Expected
78
            <[]string | len:1, cap:1>: ["failed"]
79
        to contain elements
80
            <[]string | len:2, cap:2>: ["failed", "failed"]
81
        the missing elements were
82
            <[]string | len:1, cap:1>: ["failed"]
83

84
        /home/runner/work/fdb-kubernetes-operator/fdb-kubernetes-operator/kubectl-fdb/cmd/remove_process_group_test.go:276
85
------------------------------
86
•••••••••••••••••••••••••••••••••••••••2022/03/03 16:26:33 Could not find process for coordinator IP 127.0.0.1:4501
87
••••
88

89
Summarizing 1 Failure:
90

91
[Fail] [plugin] remove instances command when running remove instances command when removing instances from a cluster should cordon all targeted processes [It] Remove all failed instances 
92
/home/runner/work/fdb-kubernetes-operator/fdb-kubernetes-operator/kubectl-fdb/cmd/remove_process_group_test.go:276
93
``` 

### What did you expect to happen?

All unit test running without failure. 

### How can we reproduce it (as minimally and precisely as possible)?

[...]

PR #1129, opened 2022-03-19
Fix flaky unit test and bug in process group removal

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1078

## Type of change

# Discussion

There was actual a bug in the old implementation in a way that we didn't added a processes group to the removal list without exclusions if the process was already marked as removal.

# Testing

Local + unit.

# Documentation

-

# Follow-up

-

### DIFF
--- a/api/v1beta2/foundationdbcluster_types.go
+++ b/api/v1beta2/foundationdbcluster_types.go
@@ -1994,3 +1994,39 @@ func (cluster *FoundationDBCluster) GetUseUnifiedImage() bool {
 func (cluster *FoundationDBCluster) GetIgnoreTerminatingPodsSeconds() int {
 	return pointer.IntDeref(cluster.Spec.AutomationOptions.IgnoreTerminatingPodsSeconds, int((10 * time.Minute).Seconds()))
 }
+
+// AddProcessGroupsToRemovalList adds the provided process group IDs to the remove list.
+// If a process group ID is already present on that list it won't be added a second time.
+func (cluster *FoundationDBCluster) AddProcessGroupsToRemovalList(processGroupIDs []string) {
+	removals := map[string]None{}
+
+	for _, id := range cluster.Spec.ProcessGroupsToRemove {
+		removals[id] = None{}
+	}
+
+	for _, processGroupID := range processGroupIDs {
+		if _, ok := removals[processGroupID]; ok {
+			continue
+		}
+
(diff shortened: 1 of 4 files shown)


api/v1beta2/foundationdbcluster_types.go
```diff
@@ -1994,3 +1994,39 @@ func (cluster *FoundationDBCluster) GetUseUnifiedImage() bool {
 func (cluster *FoundationDBCluster) GetIgnoreTerminatingPodsSeconds() int {
 	return pointer.IntDeref(cluster.Spec.AutomationOptions.IgnoreTerminatingPodsSeconds, int((10 * time.Minute).Seconds()))
 }
+
+// AddProcessGroupsToRemovalList adds the provided process group IDs to the remove list.
+// If a process group ID is already present on that list it won't be added a second time.
+func (cluster *FoundationDBCluster) AddProcessGroupsToRemovalList(processGroupIDs []string) {
+	removals := map[string]None{}
+
+	for _, id := range cluster.Spec.ProcessGroupsToRemove {
```
johscheuer, line 2003: In theory would could check if a process is already on the `ProcessGroupsToRemoveWithoutExclusion` and if so skip it, because if the process group is already on the `ProcessGroupsToRemoveWithoutExclusion` it will be removed anyway without waiting for the exclusion. I didn't implemented this because I think the benefit will not be huge.

kubectl-fdb/cmd/remove_process_group.go
```diff
@@ -222,12 +214,10 @@ func replaceProcessGroups(kubeClient client.Client, clusterName string, processG
...
 	}
 
-	processGroups = filterAlreadyMarkedProcessGroups(cluster, processGroups)
-
 	if withExclusion {
-		cluster.Spec.ProcessGroupsToRemove = append(cluster.Spec.ProcessGroupsToRemove, processGroups...)
+		cluster.AddProcessGroupsToRemovalList(processGroups)
 	} else {
-		cluster.Spec.ProcessGroupsToRemoveWithoutExclusion = append(cluster.Spec.ProcessGroupsToRemoveWithoutExclusion, processGroups...)
+		cluster.AddProcessGroupsToRemovalWithoutExclusionList(processGroups)
```
johscheuer, line 220: TODO: We should add test cases to ensure that we can add a process group to the removal without exclusion list even when it's already marked for deletion.

09harsh (approved):
👍 LGTM

PR #1131, opened 2022-03-22
Document failover with plugin and allow to update config

johscheuer:
# Description

Document the configuration sub command in the kubectl plugin and allow the plugin to update a cluster directly.

## Type of change

## Discussion

-

## Testing

locally.

## Documentation

Added.

## Follow-up

-

### DIFF
--- a/kubectl-fdb/cmd/configuration.go
+++ b/kubectl-fdb/cmd/configuration.go
@@ -21,6 +21,9 @@
 package cmd
 
 import (
+	ctx "context"
+	"fmt"
+
 	"github.com/spf13/cobra"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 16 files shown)


kubectl-fdb/cmd/configuration.go
```diff
@@ -34,11 +37,21 @@ func newConfigurationCmd(streams genericclioptions.IOStreams) *cobra.Command {
 		Short: "Get the configuration string based on the database configuration of the cluster spec.",
 		Long:  "Get the configuration string based on the database configuration of the cluster spec.",
 		RunE: func(cmd *cobra.Command, args []string) error {
+			force, err := cmd.Root().Flags().GetBool("force")
```
JoshuaMcManus, line 40: Can we use a term other than "force", I feel like that describes something that would be done to configuration command itself. Maybe change it to something like "--no-wait" or "--no-check"?

johscheuer: I think that's a great idea and I would choose `--no-wait` or `--no-confirmation` to better represent what this flag actually does and nothing is forced 👍

JoshuaMcManus (approved):

PR #1133, opened 2022-03-23
Experimental command for the plugin to show the state of exclusions

johscheuer:
# Description

Add an experimental plugin command to show the state of exclusions. This command is not finished and has some todos open but already offers some value.

## Type of change

## Discussion

I will check if the `stored_bytes` is the best indicator for how long it will take until the process is fully excluded.

## Testing

Local.

## Documentation

None, will be added once the command is in a stable command.

## Follow-up

Finish the to-dos.

### DIFF
--- a/api/v1beta2/foundationdb_status.go
+++ b/api/v1beta2/foundationdb_status.go
@@ -133,6 +133,8 @@ type FoundationDBStatusProcessInfo struct {
 type FoundationDBStatusProcessRoleInfo struct {
 	// Role defines the role a process currently has
 	Role string `json:"role,omitempty"`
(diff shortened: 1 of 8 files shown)


JoshuaMcManus (approved):
This looks good. A couple suggestions for improvements in the future:
* Add a '-w' option which uses the looping behaviour, rather than always doing an infinite loop
* Add an option to check the exclusion status of a specific instance

johscheuer:
> This looks good. A couple suggestions for improvements in the future:
> 
> * Add a '-w' option which uses the looping behaviour, rather than always doing an infinite loop
> * Add an option to check the exclusion status of a specific instance

I'll create issues for those improvements 👍

Issue #990, opened 2021-11-23
CanSafelyRemove should make use of in_progress_exclusion


johscheuer:
The current implementation of [CanSafelyRemove](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/fdbclient/admin_client.go#L296-L326) has the side effect, that we are actually issuing an `exclude` command. With the `7.0.0` we are able to read the special key range `(\xff\xff/management/in_progress_exclusion/, \xff\xff/management/in_progress_exclusion0)` instead to see which processes are still in progress. In order to see if a process is safe to remove we have to check if the process is marked a excluded in the DB and is not part of the  `in_progress_exclusion` key range.

The related issue: https://github.com/apple/foundationdb/issues/5871#issuecomment-975721874

PR #1136, opened 2022-03-25
Recreating pods that are not fully excluded.

### LINKED ISSUE #970: Operator should recreate Pods if they are not fully excluded
Currently when a process group is marked for removal and someone deletes the Pod the Pod won't be recreated. Instead of only checking if the Pod is removed we should check if the Pod is marked for removal and is fully excluded (or exclusion is skipped).

Code part: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/controllers/add_pods.go#L68

That could help in cases where someone bring the cluster in a bad state and marks to many Pods with removal.

aasmundeldhuset:
# Description

This fixes #970, which is an issue that can happen when some nodes crash (or are accidentally deleted, as happened in our case) while a cluster is in the process of being scaled down. If one or more of the nodes that crashed had been marked for removal but weren't fully drained before they crashed, we need to bring them back up again so that they can be drained before actually being shut down cleanly.

## Type of change

Bug fix (non-breaking change which fixes an issue).

# Discussion

This implementation of the fix relies on a previous run of the `removeProcessGroups` subreconciler having called `SetExcluded()` on process groups that FDB has confirmed to be fully excluded from the cluster (and saved that status to Kubernetes).

# Testing

- Unit tests (including new tests in `controllers/add_pods_test.go`)
- Initiating a scale-down in a dev FDB cluster and then deleting all the pods, and verifying that all the original storage pods get recreated so that they can be drained before they are actually excluded

# Documentation

Updated the "AddPods" section in `docs/manual/technical_design.md`.

# Follow-up

When #990 has been implemented (and we have upgraded FDB to version 7), we could simply query for the current exclusion status rather than relying on the previous run of the `removeProcessGroups` subreconciler having used the `exclude` command to determine which process groups are actually fully excluded.

### DIFF
--- a/controllers/add_pods.go
+++ b/controllers/add_pods.go
@@ -24,9 +24,8 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/FoundationDB/fdb-kubernetes-operator/pkg/podmanager"
-
 	"github.com/FoundationDB/fdb-kubernetes-operator/internal"
+	"github.com/FoundationDB/fdb-kubernetes-operator/pkg/podmanager"
 
 	fdbv1beta2 "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta2"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


johscheuer (approved):
Changes look good from my side. A user that doesn't care about the data or wants to prevent the operator from recreating the Pod (for whatever reason) could still use the remove without exclusion list.

brownleej (approved):
👍

PR #1152, opened 2022-04-01
Add support for ssd-rocksdb-v1

bnamasivayam:


controllers/cluster_controller_test.go
```diff
@@ -2535,6 +2535,60 @@ var _ = Describe("cluster_controller", func() {
...
+				})
+				It("should pass with 7.1.0", func() {
+					cluster.Spec.DatabaseConfiguration.StorageEngine = "ssd-rocksdb-v1"
+					cluster.Spec.Version = "7.1.0"
+					err := k8sClient.Update(context.TODO(), cluster)
+					Expect(err).NotTo(HaveOccurred())
+					_, err = reconcileCluster(cluster)
+					Expect(err).ToNot(HaveOccurred())
+				})
+				It("should pass with 7.1.0-rc4", func() {
```
johscheuer, line 2559: Can we add a test for `7.1.0-rc3` that should fail.

controllers/cluster_controller_test.go
```diff
@@ -2535,6 +2535,60 @@ var _ = Describe("cluster_controller", func() {
...
+					generationGap = 0
+				})
+				It("should fail with 6.3.26", func() {
+					cluster.Spec.DatabaseConfiguration.StorageEngine = "ssd-rocksdb-v1"
+					cluster.Spec.Version = "6.3.26"
+					err := k8sClient.Update(context.TODO(), cluster)
+					Expect(err).NotTo(HaveOccurred())
+					_, err = reconcileCluster(cluster)
+					Expect(err).To(HaveOccurred())
+				})
```
johscheuer, line 2550: We already have the reconcile logic here: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/controllers/cluster_controller_test.go#L133-L149 that will be always executed. I think the following changes should work:

```
		When("using FDB 6.3.26", func() {
			BeforeEach(func() {
				cluster.Spec.DatabaseConfiguration.StorageEngine = "ssd-rocksdb-v1"
				cluster.Spec.Version = "6.3.26"
				err := k8sClient.Update(context.TODO(), cluster)
				shouldCompleteReconciliation = false
			})

		        // It check that generations are not matching.
		})
```

Issue #1126, opened 2022-03-19
Update ginkgo to v2


johscheuer:
### What would you like to be added/changed?

Ginkgo v2 was released (https://onsi.github.io/ginkgo/) and we should look into updating our code to use the new version.

PR #1156, opened 2022-04-01
Update ginkgo version to v2

johscheuer:
# Description

## Type of change

## Discussion

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1126

## Testing

Locally.

## Documentation

-

## Follow-up

-

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,8 +35,6 @@ GOLANGCI_LINT_PKG=github.com/golangci/golangci-lint/cmd/golangci-lint@v1.42.1
 GOLANGCI_LINT=$(GOBIN)/golangci-lint
 GORELEASER_PKG=github.com/goreleaser/goreleaser@v1.6.3
 GORELEASER=$(GOBIN)/goreleaser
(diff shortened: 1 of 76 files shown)


Makefile
```diff
@@ -89,9 +91,9 @@ test_if_changed: cover.out
 # Run tests
 cover.out: ${GO_ALL} ${MANIFESTS}
 
-test:
+test: $(GINKGO)
 ifneq "$(SKIP_TEST)" "1"
-	go test ${go_test_flags} ./... -coverprofile cover.out --ginkgo.parallel.total $(GINKGO_PARALLEL)
+	$(GINKGO) -p --slow-spec-threshold=10s -keep-going ${go_test_flags} ./... -coverprofile cover.out
```
johscheuer, line 96: Using `go test` has the drawback that we can't use the parallelise feature of ginkgo but it might be worth to not force everyone to install ginkgo.

johscheuer (commented):
With `ginkgo` v2 and running specs in parallel it roughly took `17min` let's see how long it takes to run those tests with `go test`.

ltsampros (approved):

Issue #1151, opened 2022-03-31
GRV and commit proxies are not working as expected


johscheuer:
### What happened?

I created a cluster with the following configuration:

```yaml
databaseConfiguration:
    commit_proxies: 26
    grv_proxies: 4
    redundancy_mode: triple
    storage_engine: ssd
```

The cluster never reconciled with the following error message:

```text
"current":{"redundancy_mode":"triple","storage_engine":"ssd-2","usable_regions":1,"logs":3,"proxies":3,"commit_proxies":2,"grv_proxies":1,"resolvers":1,"log_routers":-1,"remote_logs":-1}
"desired":{"redundancy_mode":"triple","storage_engine":"ssd-2","usable_regions":1,"logs":3,"proxies":3,"commit_proxies":26,"grv_proxies":4,"resolvers":1,"log_routers":-1,"remote_logs":-1}
```

### What did you expect to happen?

The cluster is properly configured and reconciles.

### How can we reproduce it (as minimally and precisely as possible)?

Create a cluster with the following config;

```yaml
databaseConfiguration:
    commit_proxies: 26
    grv_proxies: 4
    redundancy_mode: triple
    storage_engine: ssd
```

### FDB Kubernetes operator

<details>

```console
$ kubectl fdb version
# paste output here
v1.1.1
```

</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Cloud provider

<details>

</details>

johscheuer:
Fixe by the reference PR.

PR #1172, opened 2022-04-14
role count fix

### LINKED ISSUE #1150: Proxies setting doesn't work as expected for 7.0+
### What happened?

I created a cluster with a different number of proxies than the default and the cluster was never able to reconcile. the cluster never reconciled with the error message that a configuration change is expected:

This was the result from the FDB status:

```json
{"redundancy_mode":"triple","storage_engine":"ssd-2","usable_regions":1,"logs":3,"proxies":30,"commit_proxies":26,"grv_proxies":4,"resolvers":1,"log_routers":-1,"remote_logs":-1}
```

This is what the operator expected:

```json
{"redundancy_mode":"triple","storage_engine":"ssd-2","usable_regions":1,"logs":3,"proxies":30,"commit_proxies":2,"grv_proxies":1,"resolvers":1,"log_routers":-1,"remote_logs":-1}}
```

The cluster should be able to reconcile.

Create a cluster with the following spec:

```yaml
apiVersion: apps.foundationdb.org/v1beta2
kind: FoundationDBCluster
metadata:
  name: test-cluster
spec:
  processCounts:
    proxies: 30
  version: 7.1.0.-rc3
```

_No response_

### FDB Kubernetes operator

<details>

```console
$ kubectl fdb version
# paste output here
v1.1.1
```

</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Cloud provider

<details>

</details>

ltsampros:
# Description

Fixes  #1150 #1151. 
If proxies are set to 0, then grv_proxies/commit_proxies will be used in the configure string correctly and vice versa. Defaults have been set for all cases and added some extra logic in normalization of the status to account for that so that reconciliation completes.

## Type of change

## Testing
Unit tests and local dev environment.

### DIFF
--- a/api/v1beta2/foundationdb_database_configuration.go
+++ b/api/v1beta2/foundationdb_database_configuration.go
@@ -150,18 +150,8 @@ func (configuration *DatabaseConfiguration) FailOver() DatabaseConfiguration {
 //
 // This will fill in defaults of -1 for some fields that have a default of 0,
 // and will ensure that the region configuration is ordered consistently.
-func (configuration DatabaseConfiguration) NormalizeConfiguration(version string, areSeparatedProxiesConfigured bool) DatabaseConfiguration {
+func (configuration DatabaseConfiguration) NormalizeConfiguration() DatabaseConfiguration {
 	result := configuration.DeepCopy()
-
-	parsedVersion, _ := ParseFdbVersion(version)
-	if parsedVersion.HasSeparatedProxies() {
-		if !areSeparatedProxiesConfigured {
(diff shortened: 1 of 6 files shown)


brownleej (changes requested):

brownleej (approved):
👍

johscheuer (commented):
I know that these changes are already merged but I still have some questions: Those changes imply that proxies and grv/commit proxies can't be set at the same time correct? Is this documented somewhere outside of the code e.g. in our docs?

PR #1168, opened 2022-04-09
Add sidecar image and config for 7.1.2

johscheuer:
# Description

Add config for 7.1.0.

## Type of change

*Please select one of the options below.*

- New feature (non-breaking change which adds functionality)

## Discussion

-

## Testing

Locally.

## Documentation

-

## Follow-up

-

PR #1175, opened 2022-04-20
Remove release candidates  and substitute with latest release from sidecars

ltsampros:
# Description

## Type of change

- Documentation
- Other

## Discussion

## Testing

Manual testing?*

## Documentation

## Follow-up

### DIFF
--- a/config/development/images.yaml
+++ b/config/development/images.yaml
@@ -6,26 +6,10 @@ spec:
   template:
     spec:
       initContainers:
-        # Install client libraries for unreleased versions of FDB
-        - name: foundationdb-kubernetes-init-7-1
-          image: foundationdb/foundationdb-kubernetes-sidecar:7.1.0-rc2-1
-          args:
-            - "--copy-binary"
-            - "fdbcli"
-            - "--copy-binary"
-            - "fdbbackup"
-            - "--copy-binary"
-            - "fdbrestore"
-            - "--output-dir"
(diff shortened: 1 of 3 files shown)


config/development/images.yaml
```diff
@@ -6,26 +6,10 @@ spec:
   template:
     spec:
       initContainers:
-        # Install client libraries for unreleased versions of FDB
```
brownleej, line 9: I believe a block like this is still necessary, because the `foundationdb-kubernetes-init-7-1-primary` block doesn't install any of the CLIs.

brownleej (changes requested):

config/development/images.yaml
johscheuer: I think we can get rid of all the custom changes here by just adding the init container to the opreator in the [manager.yml](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/config/deployment/manager.yaml#L32-L48).

johscheuer (commented):
just noted that there is an overlap with: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1168 I'll finish my PR today to move forward with it (it should include all required changes that are made in this PR). I just have to revert the default FDB build version back until we have a working 7.1 deb package.

johscheuer (commented):
Implemented in: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1168

PR #1182, opened 2022-04-27
Check exclusion status of fully excluded and in progress exclusions in two steps

johscheuer:
# Description

The idea here is to run the exclusion check in two steps:

1. Check all processes that are already fully excluded (don't serve any roles), this is only an additional safety check and shouldn't be required actually.
2. Check all exclusions that are "in progress"

The benefit of this approach is that a slow exclusion doesn't block the removal of other process groups that are already fully excluded.

## Type of change

## Discussion

-

## Testing

Unit tests + locally.

## Documentation

Like commented here: https://github.com/apple/foundationdb/issues/6854#issuecomment-1110608724 it would actually be nice if FDB would offer this functionality. if I get some spare time I might look into creating this in the c bindings. 

## Follow-up

### DIFF
--- a/fdbclient/admin_client.go
+++ b/fdbclient/admin_client.go
@@ -332,17 +332,19 @@ func (client *cliAdminClient) GetExclusions() ([]fdbv1beta2.ProcessAddress, erro
 
 // getRemainingAndExcludedFromStatus checks which processes of the input address list are excluded in the cluster and which are not.
 // The first list return all addresses that are excluded in the cluster and the exclude command can be used to verify if it's safe to remove this address.
-// The second list contains all addresses that are part of the input list and are not marked as excluded in the cluster, those addresses are not safe to remove.
-func getRemainingAndExcludedFromStatus(status *fdbv1beta2.FoundationDBStatus, addresses []fdbv1beta2.ProcessAddress) ([]fdbv1beta2.ProcessAddress, []fdbv1beta2.ProcessAddress) {
-	notExcluded := map[string]fdbv1beta2.None{}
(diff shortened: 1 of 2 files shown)


fdbclient/admin_client.go
```diff
@@ -332,17 +332,19 @@ func (client *cliAdminClient) GetExclusions() ([]fdbv1beta2.ProcessAddress, erro
...
 // The first list return all addresses that are excluded in the cluster and the exclude command can be used to verify if it's safe to remove this address.
-// The second list contains all addresses that are part of the input list and are not marked as excluded in the cluster, those addresses are not safe to remove.
-func getRemainingAndExcludedFromStatus(status *fdbv1beta2.FoundationDBStatus, addresses []fdbv1beta2.ProcessAddress) ([]fdbv1beta2.ProcessAddress, []fdbv1beta2.ProcessAddress) {
-	notExcluded := map[string]fdbv1beta2.None{}
+// The second list returns all addresses that are excluded and don't have any roles assigned, this is a sign that the process is "fully" excluded and safe to remove.
+// The third list contains all addresses that are part of the input list and are not marked as excluded in the cluster, those addresses are not safe to remove.
+func getRemainingAndExcludedFromStatus(status *fdbv1beta2.FoundationDBStatus, addresses []fdbv1beta2.ProcessAddress) ([]fdbv1beta2.ProcessAddress, []fdbv1beta2.ProcessAddress, []fdbv1beta2.ProcessAddress) {
```
brownleej, line 337: Since we're now adding a third return value with the same type, it may be worth defining a new type to capture the returned address lists, to make it more explicit in the code what each one means.

brownleej (approved):
👍

Issue #1039, opened 2022-01-15
Add support for exclude/include based on localities for 7.0.0


johscheuer:
In the new 7.0.0 release `fdbcli` added support for excluding and including processes based on localities: https://github.com/apple/foundationdb/blob/main/documentation/sphinx/source/release-notes/release-notes-700.rst#features. We Should add support for that feature as part of the changes or 7.0.0

PR #1187, opened 2022-04-28
supporting exclude using locality

09harsh:
# Description
Fixes: #1039

Added support in the operator to perform exclude using localities. 

## Type of change

## Discussion

No

## Testing

Manual testing?*
Added Unit tests

No 

## Documentation

## Follow-up

No


controllers/remove_process_groups.go
```diff
@@ -277,6 +268,30 @@ func includeProcessGroup(ctx context.Context, r *FoundationDBClusterReconciler,
...
+	fdbVersion, err := fdbv1beta2.ParseFdbVersion(cluster.Spec.Version)
+	if err != nil {
+		return false, err
+	}
+
+	hasStatusUpdate := false
+	for _, processGroup := range cluster.Status.ProcessGroups {
+		if processGroup.IsMarkedForRemoval() && removedProcessGroups[processGroup.ProcessGroupID] {
+			if fdbVersion.IsAtLeast(fdbv1beta2.Versions.LatestFdbVersion) {
+				*fdbProcessesToInclude = append(*fdbProcessesToInclude, fdbv1beta2.ProcessAddress{StringAddress: processGroup.GetExclusionString()})
```
johscheuer, line 281: We should also add in this case the IP addresses of the process group otherwise we might run in the following case:

1. Operator excludes process group
2. Upgrade to 7.0.0+ is started (and locality based exclusion is enabled)
3. Process group get's excluded again with the locality string
4. Process group gets removed and we only include the locality string and not the IPs (leaving them excluded in the cluster).

controllers/remove_process_groups_test.go
```diff
@@ -279,3 +279,112 @@ var _ = Describe("remove_process_groups", func() {
 		k8sClient.Clear()
 	})
 })
+
+var _ = Describe("validating getIncludedProcesses", func() {
```
johscheuer, line 283: Any reason why those tests are not nested in. the above describe statement?

controllers/remove_process_groups_test.go
```diff
@@ -279,3 +279,112 @@ var _ = Describe("remove_process_groups", func() {
...
+					{ProcessGroupID: "log-4", ProcessClass: "log", Addresses: []string{"1.1.2.7"}},
+				},
+			},
+		}
+		fdbProcessesToInclude = make([]fdbv1beta2.ProcessAddress, 0)
+		removedProcessGroups = make(map[string]bool)
+		processGroups = make([]*fdbv1beta2.ProcessGroupStatus, 0)
+	})
+
+	Context("fdb version is below 7.0.1", func() {
```
johscheuer, line 319: Why 7.0.1?

controllers/remove_process_groups_test.go
```diff
@@ -279,3 +279,112 @@ var _ = Describe("remove_process_groups", func() {
...
+				Expect(err).To(BeNil())
+				Expect(hasStatusUpdate).To(BeTrue())
+				Expect(len(fdbProcessesToInclude)).To(Equal(1))
+				Expect(fdbv1beta2.ProcessAddressesString(fdbProcessesToInclude, " ")).To(Equal("1.1.1.1"))
+				Expect(len(processGroups)).To(Equal(15))
+			})
+		})
+	})
+
+	Context("fdb version is above 7.0.1", func() {
```
johscheuer, line 355: Same here: Why 7.0.1?

fdbclient/admin_client.go
```diff
@@ -318,13 +318,21 @@ func (client *cliAdminClient) GetExclusions() ([]fdbv1beta2.ProcessAddress, erro
 	lines := strings.Split(output, "\n")
 	exclusions := make([]fdbv1beta2.ProcessAddress, 0, len(lines))
 	for _, line := range lines {
+		line = strings.TrimSpace(line)
+		if len(strings.Split(line, " ")) != 1 {
```
johscheuer, line 322: What is the reasoning behind this? Could you add a comment above that?

fdbclient/admin_client.go
```diff
@@ -318,13 +318,21 @@ func (client *cliAdminClient) GetExclusions() ([]fdbv1beta2.ProcessAddress, erro
...
 		exclusionMatch := exclusionLinePattern.FindStringSubmatch(line)
 		if exclusionMatch != nil {
 			pAddr, err := fdbv1beta2.ParseProcessAddress(exclusionMatch[1])
 			if err != nil {
 				return nil, err
 			}
 			exclusions = append(exclusions, pAddr)
+			continue
+		}
+		if strings.HasPrefix(line, "locality") {
```
johscheuer, line 334: I think a better way to do is querying the FDB status and returning all excluded processes that are under `excluded_servers` this way we don't need to use a regex (and I would expect that the regex from above actually matches the locality string). Have you tested this locally?

In another step( not your PR) we could query the mgmt API directly.

internal/removals/remove.go
```diff
@@ -132,8 +137,12 @@ func GetRemainingMap(logger logr.Logger, adminClient fdbadminclient.AdminClient,
 			continue
 		}
 
-		for _, pAddr := range processGroup.Addresses {
-			addresses = append(addresses, fdbv1beta2.ProcessAddress{IPAddress: net.ParseIP(pAddr)})
+		if fdbVersion.IsAtLeast(fdbv1beta2.Versions.LatestFdbVersion) {
+			addresses = append(addresses, fdbv1beta2.ProcessAddress{StringAddress: processGroup.GetExclusionString()})
```
johscheuer, line 141: Should we also add the IP addresses here? See my comment above for the background.

johscheuer (changes requested):

fdbclient/admin_client.go
09harsh: Mainly I was trying to remove the introductory line that 
```There are currently x servers or localities being excluded from the database: ```
But on talking another look I realised that this will be skipped in the later part of the code as it will not match the criteria.

File: controllers/exclude_processes_test.go [hunk 8]
```diff
@@ -136,6 +138,223 @@ var _ = Describe("exclude_processes", func() {
 	})
 })
 
+var _ = Describe("validating getProcessesToExclude", func() {
```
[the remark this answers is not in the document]
09harsh: Added it to the existing one

controllers/remove_process_groups_test.go
09harsh: typo

internal/removals/remove.go
```diff
@@ -132,8 +137,12 @@ func GetRemainingMap(logger logr.Logger, adminClient fdbadminclient.AdminClient,
 			continue
 		}
 
+		if fdbVersion.IsAtLeast(fdbv1beta2.Versions.NextMajorVersion) && cluster.GetUseLocalitiesForExclusion() {
+			addresses = append(addresses, fdbv1beta2.ProcessAddress{StringAddress: processGroup.GetExclusionString()})
+		}
 		for _, pAddr := range processGroup.Addresses {
 			addresses = append(addresses, fdbv1beta2.ProcessAddress{IPAddress: net.ParseIP(pAddr)})
+
```
johscheuer, line 145: Newline can bee removed.

johscheuer (changes requested):
Could you please run a local test to ensure those changes are working our e2e pipeline won't use this feature.

controllers/remove_process_groups.go
```diff
@@ -277,6 +268,32 @@ func includeProcessGroup(ctx context.Context, r *FoundationDBClusterReconciler,
 	return nil
 }
 
+func getIncludedProcesses(cluster *fdbv1beta2.FoundationDBCluster, removedProcessGroups map[string]bool) ([]fdbv1beta2.ProcessAddress, []*fdbv1beta2.ProcessGroupStatus, bool, error) {
+	fdbProcessesToInclude := make([]fdbv1beta2.ProcessAddress, 0)
+	processGroups := make([]*fdbv1beta2.ProcessGroupStatus, 0, len(cluster.Status.ProcessGroups))
```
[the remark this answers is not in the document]
09harsh: created an issue  #1196

docs/cluster_spec.md
```diff
@@ -155,6 +155,7 @@ FoundationDBClusterAutomationOptions provides flags for enabling or disabling op
 | replacements | Replacements contains options for automatically replacing failed processes. | [AutomaticReplacementOptions](#automaticreplacementoptions) | false |
 | ignorePendingPodsDuration | IgnorePendingPodsDuration defines how long a Pod has to be in the Pending Phase before ignore it during reconciliation. This prevents Pod that are stuck in Pending to block further reconciliation. | time.Duration | false |
 | useNonBlockingExcludes | UseNonBlockingExcludes defines whether the operator is allowed to use non blocking exclude commands. The default is false. | *bool | false |
+| useLocalitiesForExclusion | UseLocalitiesForExclusion defines whether the exclusions are done using localities instead of IP addresses. The default is false. | *bool | false |
```
johscheuer, line 158: The documentation is not updated and must be regenerated by running `make clean all` or `make docs/cluster_spec.md`.

johscheuer (changes requested):

api/v1beta2/foundationdbcluster_types.go
```diff
@@ -1753,6 +1762,11 @@ func (cluster *FoundationDBCluster) GetUseNonBlockingExcludes() bool {
 	return *cluster.Spec.AutomationOptions.UseNonBlockingExcludes
 }
 
+// UseLocalitiesForExclusion returns the value of UseLocalitiesForExclusion or false if unset.
+func (cluster *FoundationDBCluster) UseLocalitiesForExclusion(fdbVersion Version) bool {
+	return fdbVersion.IsAtLeast(Versions.NextMajorVersion) && pointer.BoolDeref(cluster.Spec.AutomationOptions.UseLocalitiesForExclusion, false)
```
johscheuer, line 1767:
```suggestion
	version, err := fdbv1beta2.ParseFdbVersion(cluster.Spec.Version)
	if err != nil {
		// Fall back to use exclusions with IP if we can't parse the version.
		// This should never happen since the version is validated in earlier steps.
		return false
	}
	
	return version.IsAtLeast(Versions.NextMajorVersion) && pointer.BoolDeref(cluster.Spec.AutomationOptions.UseLocalitiesForExclusion, false)
```

I actually meant to use a function like this. This way a user just calls the method and doesn't have to parse the version.

johscheuer (changes requested):
Please fix the failing CI. You have to regenerate the CRD files since you changed the API types.

api/v1beta2/foundationdb_database_configuration.go
```diff
@@ -49,6 +49,9 @@ type DatabaseConfiguration struct {
 	// Regions defines the regions that the database can replicate in.
 	Regions []Region `json:"regions,omitempty"`
 
+	// ExcludedServers defines the list  of excluded servers form the database.
+	ExcludedServers []ExcludedServers `json:"excluded_servers,omitempty"`
```
johscheuer, line 53: We have to ignore this slice when we compare the current and the desired configuration otherwise the operator will compare the currently excluded servers in this comparison.

johscheuer (changes requested):

johscheuer (approved):
LGTM 👍 Please wait until the build pipeline has finished with merging.

Issue #1179, opened 2022-04-27
FDB 7.1 when both proxies and grv/commit proxies are defined grv and commit proxies have no effect


johscheuer:
### What happened?

When creating a cluster with the following database configuration (version `7.1.2`):

```yaml
  databaseConfiguration:
     commit_proxies: 24
     grv_proxies: 2
     proxies: 26
```

This resulted in the following configuration string: 

```
configure triple ... usable_regions=1 logs=.. resolvers=... log_routers=-1 remote_logs=-1 proxies=26 regions=[]
```

### What did you expect to happen?

I would expect that the grv and commit proxy count are preferred over the proxies count. There was a change in that logic in: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1172 from a user perspective I would expect that the operator takes the commit and grv proxies or complains about an invalid configuration if commit and grv proxies are defined with proxies.

### How can we reproduce it (as minimally and precisely as possible)?

Create a cluster with the above configuration and see that the `proxies` count is used instead of the more specific grv and commit proxy count.

### Anything else we need to know?

-

### FDB Kubernetes operator

<details>

```console
$ kubectl fdb version
# paste output here
v1.3.1
```

</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Cloud provider

<details>

</details>

PR #1191, opened 2022-04-28
For versions that support grv and commit use grv and commit proxy count if defined

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1179 this change makes upgrades easier since all three proxies can be defined before upgrading and after the upgrade the more precise definition (grv and commit) of proxies will be used.

## Type of change

## Discussion

-

## Testing

Local + unit test

## Documentation

I'll add that in a separate PR to document the upgrade process in general.

## Follow-up

-


brownleej (approved):
👍

Issue #1178, opened 2022-04-26
Race condition when recreating a pod during an upgrade


brownleej:
### What happened?

During an upgrade, we perform the following steps:

1. Update the image version for the sidecar container in each pod to inject new binaries
2. Bounce the fdbserver processes so they start running the new binaries
3. Update the running version in the resource status to reflect that the cluster is running the new version
4. Recreate the pods so the main container is running the new version

There can be a delay between steps 1 and 2, if one of the pods is failing to update. This is more likely in a multi-DC configuration, because we need the pods to be updated in all of the DCs. If a pod gets deleted between steps 1 and 2, the operator will recreate it with the init container and main container at the old version, and the sidecar container at the new version. The init container will generate a monitor conf referencing the binary in /usr/bin/fdbserver, which is the old version. Then in parallel we will start the main container and the sidecar container. The main container may load the monitor conf from the init container referencing the old version, or it may load the monitor conf from the sidecar referencing the new version, depending on the timing of the two container launches.

### What did you expect to happen?

The desired behavior here isn't obvious because it's ambiguous what the desired version for the binary is. At a minimum we should make it consistently produce a single version depending on the operator's understanding of the state of the cluster.

### How can we reproduce it (as minimally and precisely as possible)?

1. Start a multi-DC cluster on FDB 6.3
2. Start an upgrade in DC1 to FDB 7.1, and leave the others on 6.3
3. Delete all the pods in DC1
4. Observe the distribution of fdbserver versions that are running in the pods.

### FDB Kubernetes operator

<details>

1.3.0

</details>

### Kubernetes version

<details>

v1.19.13
</details>

### Cloud provider

<details>
N/A
</details>

johscheuer:
I would expect that if a Pod get's deleted between step 1-2 that the operator is able to converge the state of this Pod to the new version again. We should always try. to reconcile the state based on the version defined in the cluster spec.

brownleej:
> We should always try. to reconcile the state based on the version defined in the cluster spec.
[...]

PR #1200, opened 2022-05-04
If a Pod gets deleted during an upgrade recreate it with the new version

johscheuer:
# Description

fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1178

## Type of change

## Discussion

Like discussed in the issue we will recreate Pods that are deleted during an upgrade.

## Testing

Unit testing, local test and e2e test.

## Documentation

Will be updated.

## Follow-up

-

### DIFF
--- a/controllers/bounce_processes.go
+++ b/controllers/bounce_processes.go
@@ -67,6 +67,7 @@ func (bounceProcesses) reconcile(ctx context.Context, r *FoundationDBClusterReco
 
 	processesToBounce := fdbv1beta2.FilterByConditions(cluster.Status.ProcessGroups, map[fdbv1beta2.ProcessGroupConditionType]bool{
 		fdbv1beta2.IncorrectCommandLine: true,
+		fdbv1beta2.IncorrectPodSpec:     false,
 	}, true)
 
 	addresses := make([]fdbv1beta2.ProcessAddress, 0, len(processesToBounce))
--- a/controllers/bounce_processes.go
+++ b/controllers/bounce_processes.go
@@ -67,7 +67,6 @@ func (bounceProcesses) reconcile(ctx context.Context, r *FoundationDBClusterReco
 
 	processesToBounce := fdbv1beta2.FilterByConditions(cluster.Status.ProcessGroups, map[fdbv1beta2.ProcessGroupConditionType]bool{
 		fdbv1beta2.IncorrectCommandLine: true,
-		fdbv1beta2.IncorrectPodSpec:     false,
 	}, true)
 
 	addresses := make([]fdbv1beta2.ProcessAddress, 0, len(processesToBounce))
@@ -110,7 +109,7 @@ func (bounceProcesses) reconcile(ctx context.Context, r *FoundationDBClusterReco
 		return &requeue{message: "Waiting for config map to sync to all pods"}
 	}
 
-	upgrading := cluster.Status.RunningVersion != cluster.Spec.Version
+	upgrading := cluster.IsBeingUpgraded()
 
 	if len(addresses) > 0 {
 		if !pointer.BoolDeref(cluster.Spec.AutomationOptions.KillProcesses, true) {
--- a/controllers/bounce_processes.go
+++ b/controllers/bounce_processes.go
@@ -110,8 +110,6 @@ func (bounceProcesses) reconcile(ctx context.Context, r *FoundationDBClusterReco
 		return &requeue{message: "Waiting for config map to sync to all pods"}
 	}
 
-	upgrading := cluster.IsBeingUpgraded()
-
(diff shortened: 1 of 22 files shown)


controllers/bounce_processes.go
```diff
@@ -67,7 +67,6 @@ func (bounceProcesses) reconcile(ctx context.Context, r *FoundationDBClusterReco
 
 	processesToBounce := fdbv1beta2.FilterByConditions(cluster.Status.ProcessGroups, map[fdbv1beta2.ProcessGroupConditionType]bool{
 		fdbv1beta2.IncorrectCommandLine: true,
-		fdbv1beta2.IncorrectPodSpec:     false,
```
johscheuer, line 70: We have to keep this here and update the logic for how we calculate the incorrect pod spec.

pkg/podmanager/fdb_process_group.go
```diff
@@ -34,7 +34,7 @@ import (
 var processIDRegex = regexp.MustCompile(`^([\w-]+-\d)-\d$`)
 
 // ParseProcessGroupID extracts the components of a process group ID.
-func ParseProcessGroupID(id string) (fdbv1beta2.ProcessClass, int, error) {
+func ParseProcessGroupID(id string) (int, error) {
```
johscheuer, line 37: I know that's a breaking change but the underlying method returned actually the prefix and not the process class, which means the signature was wrong. We also never used the prefix somewhere and only make use of the ID number.

johscheuer:
I ran a test for single cluster upgrades (one stable and one with chaos killing randomly a Pod):

internal/pod_models.go
```diff
@@ -37,6 +37,17 @@ import (
...
 
+const (
+	// SidecarContainer is the name of the sidecar container
+	SidecarContainer string = "foundationdb-kubernetes-sidecar"
+	// MainContainer is the name of main container that runs FoundationDB
+	MainContainer string = "foundationdb"
+	// InitContainer is the name of the init container
+	InitContainer string = "foundationdb-kubernetes-init"
+	// SidecarDefaultImage is the name of the default sidecar and init container
+	SidecarDefaultImage string = "foundationdb/foundationdb-kubernetes-sidecar"
```
johscheuer, line 48: We could also think about exposing this information to allow adopters to make use of these constants.

09harsh (approved):
LGTM 👍

pkg/podmanager/fdb_process_group.go
johscheuer: We should change this back and make a note at the top that the process class is actually a string.

johscheuer (commented):
We should document a possible failure and how to recover from it.

controllers/update_status.go
```diff
@@ -514,27 +502,20 @@ func validateProcessGroup(ctx context.Context, r *FoundationDBClusterReconciler,
...
-
-	specHash, err := internal.GetPodSpecHash(cluster, processGroupStatus.ProcessClass, idNum, nil)
+	correctPodSpec, err := internal.PodHasCorrectSpec(cluster, processGroupStatus.ProcessClass, processGroupStatus.ProcessGroupID, *pod)
 	if err != nil {
 		return err
 	}
 
-	incorrectPod := !metadataMatches(pod.ObjectMeta, internal.GetPodMetadata(cluster, processGroupStatus.ProcessClass, processGroupStatus.ProcessGroupID, specHash))
-	if !incorrectPod {
+	if correctPodSpec && !cluster.IsBeingUpgraded() {
```
brownleej, line 510: Is the implication of this that we will not set the `IncorrectPodSpec` condition until after the processes are running on the desired version? If so, is the reasoning that we will have a different condition set on the process groups when they're running at the wrong version? That seems like subtle logic, and it might be good to document.

internal/pod_models_test.go
```diff
@@ -3181,4 +3181,114 @@ var _ = Describe("pod_models", func() {
...
+				})
+			})
+
+		})
+	})
 })
+
+/*
+
+add tests:
```
brownleej, line 3285: Does this need additional action, or should we remove it?

pkg/podmanager/fdb_process_group.go
brownleej: We should deprecate it as well, if we have no need identified for the version that returns a process class or prefix.

brownleej (changes requested):

internal/pod_models_test.go
johscheuer: Oh, I forgot to remove that.

brownleej (approved):
👍

johscheuer:
We won't follow this path and we will be focusing instead of the unified image approach.

PR #1203, opened 2022-05-05
updating the logic for check more than 1 value

09harsh:
# Description

In `remove_process_group.go` we are throwing an error for multiple occurrences of an instance using `len(x) > 0`. This should be greater than 1. 

## Type of change

## Discussion

-

## Testing
Did unit testing for the same. 

## Documentation

-

## Follow-up

-


johscheuer (changes requested):
You missed the check for pvcs. Could you please specify what kin of testing you did.

johscheuer (approved):
LGTM :+1:

PR #1216, opened 2022-05-13
Reviewing technical design

### LINKED ISSUE #1134: Review technical design

Review our technical design to see if it still represents the current state: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/docs/manual/technical_design.md#technical-design

09harsh:
# Description

Fixes #1134

## Type of change

- Documentation

## Discussion

No

## Testing

Not needed

## Documentation

-

## Follow-up

-

### DIFF
--- a/api/v1beta2/foundationdbcluster_types.go
+++ b/api/v1beta2/foundationdbcluster_types.go
@@ -565,11 +565,11 @@ func FilterByCondition(processGroupStatus []*ProcessGroupStatus, conditionType P
 // groups with that condition will be returned. If a condition is mapped to
 // false in the conditionRules map, only process groups without that condition
 // will be returned.
-func FilterByConditions(processGroupStatus []*ProcessGroupStatus, conditionRules map[ProcessGroupConditionType]bool, ignoreRemoved bool) []string {
+func FilterByConditions(processGroups []*ProcessGroupStatus, conditionRules map[ProcessGroupConditionType]bool, ignoreRemoved bool) []string {
 	result := make([]string, 0)
 
-	for _, groupStatus := range processGroupStatus {
-		if ignoreRemoved && groupStatus.IsMarkedForRemoval() {
+	for _, processGroup := range processGroups {
+		if ignoreRemoved && processGroup.IsMarkedForRemoval() {
 			continue
 		}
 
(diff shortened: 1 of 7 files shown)


api/v1beta2/foundationdbcluster_types.go
```diff
@@ -565,25 +565,25 @@ func FilterByCondition(processGroupStatus []*ProcessGroupStatus, conditionType P
 // groups with that condition will be returned. If a condition is mapped to
 // false in the conditionRules map, only process groups without that condition
 // will be returned.
-func FilterByConditions(processGroupStatus []*ProcessGroupStatus, conditionRules map[ProcessGroupConditionType]bool, ignoreRemoved bool) []string {
+func FilterByConditions(processGroups []*ProcessGroupStatus, conditionRules map[ProcessGroupConditionType]bool, ignoreRemoved bool) []string {
 	result := make([]string, 0)
```
09harsh, line 569: Changes here are mainly done to increase readability.

controllers/check_client_compatibility.go
```diff
@@ -42,12 +42,6 @@ func (c checkClientCompatibility) reconcile(_ context.Context, r *FoundationDBCl
 		return nil
 	}
 
-	adminClient, err := r.getDatabaseClientProvider().GetAdminClient(cluster, r)
-	if err != nil {
-		return &requeue{curError: err}
-	}
-	defer adminClient.Close()
-
```
09harsh, line 50: Moving it at a later stage where it actually needed. This can help in unnecessary fetching of the admin client if there are error before its actual use,.

api/v1beta2/foundationdbcluster_types.go
johscheuer: We use the name `processGroupStatus` in multiple places what is the benefit of changing it only in this place? If we see a benefit in changing the name of the variable we should do it consistently in all methods. I would prefer having a dedicated PR for that, since those changes are out-of-scope for the referenced issue.

johscheuer (changes requested):
When you revert the changes in `foundationdbcluster_types.go` the PR looks good and can be merged.

johscheuer (approved):

Issue #966, opened 2021-10-28
Enable logging in fault tolerance check


johscheuer:
Currently our fault tolerance check has logging disabled: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/internal/fault_tolerance.go#L50-L64 we should make the changes to enable it again. The idea would be that we pass the logger down to that method that has the benefit that the logger can already include the values for the namespace and cluster.

PR #1217, opened 2022-05-13
enabling logger in fault tolerance check

09harsh:
# Description

Fixes: #966

## Type of change

- Other

## Discussion

no

## Testing

Ran make command to verify everything is working

## Documentation

Not needed

## Follow-up

No

### DIFF
--- a/pkg/podmanager/pod_lifecycle_manager.go
+++ b/pkg/podmanager/pod_lifecycle_manager.go
@@ -26,6 +26,7 @@ import (
 	fdbv1beta2 "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta2"
 	"github.com/FoundationDB/fdb-kubernetes-operator/internal"
 	"github.com/FoundationDB/fdb-kubernetes-operator/pkg/fdbadminclient"
+	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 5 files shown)


pkg/podmanager/pod_lifecycle_manager.go
```diff
@@ -43,7 +44,7 @@ type PodLifecycleManager interface {
 	DeletePod(context.Context, client.Client, *corev1.Pod) error
 
 	// CanDeletePods checks whether it is safe to delete pods.
-	CanDeletePods(context.Context, fdbadminclient.AdminClient, *fdbv1beta2.FoundationDBCluster) (bool, error)
+	CanDeletePods(context.Context, logr.Logger, fdbadminclient.AdminClient, *fdbv1beta2.FoundationDBCluster) (bool, error)
```
johscheuer, line 47: We can't change the interface here. That would be a breaking change and requires a new major version and adding a new method will also break the interface in an incompatible way. We can discuss some ideas how we can move forward with this.

johscheuer (changes requested):

pkg/podmanager/pod_lifecycle_manager.go
johscheuer: Like discussed offline we will make use of https://pkg.go.dev/github.com/go-logr/logr#NewContext and https://pkg.go.dev/github.com/go-logr/logr#FromContext

johscheuer (changes requested):

johscheuer (approved):

Issue #1201, opened 2022-05-05
Crash looping transaction system doesn't work with default ReplaceTransactionSubsystem


johscheuer:
### What happened?

When setting some Pods from the transaction system (stateless or transaction) into crash loop state it will be replaced instead of being set into crash loop:

```yaml
spec:
  buggify:
    crashLoop:
       - id-1
       - id-2
```

### What did you expect to happen?

I would have expected that the hosts are not replaced but just set into crash loop and stay in this state.

### How can we reproduce it (as minimally and precisely as possible)?

Create a cluster and set one Pod from the transaction system into crash loop state.

### FDB Kubernetes operator

<details>

```console
$ kubectl fdb version
# paste output here
v1.3.1
```

</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Cloud provider

<details>

</details>

PR #1219, opened 2022-05-15
Don't replace process groups if they are set into crash loop

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1201

## Type of change

## Discussion

-

## Testing

Unit tests + local testing.

## Documentation

-

## Follow-up

-

### DIFF
--- a/controllers/delete_pods_for_buggification.go
+++ b/controllers/delete_pods_for_buggification.go
@@ -45,15 +45,7 @@ func (d deletePodsForBuggification) reconcile(ctx context.Context, r *Foundation
 	podMap := internal.CreatePodMap(cluster, pods)
 	updates := make([]*corev1.Pod, 0)
 
-	crashLoopPods := make(map[string]bool, len(cluster.Spec.Buggify.CrashLoop))
-	crashLoopAll := false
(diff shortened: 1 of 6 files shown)


AnnigeriShambu (approved):

controllers/delete_pods_for_buggification.go
```diff
@@ -76,7 +68,10 @@ func (d deletePodsForBuggification) reconcile(ctx context.Context, r *Foundation
 			}
 		}
 
-		shouldCrashLoop := crashLoopAll || crashLoopPods[processGroup.ProcessGroupID]
+		shouldCrashLoop := crashLoopAll
+		if !shouldCrashLoop {
+			_, shouldCrashLoop = crashLoopPods[processGroup.ProcessGroupID]
+		}
```
09harsh, line 74: What benefit are we getting by doing this ?
If `crashLoopAll` is true we will not check `crashLoopPods[processGroup.ProcessGroupID]` else we will. The current implementation added one extra check with `if` condition.

09harsh (approved):
LGTM

Issue #1213, opened 2022-05-12
Automatically replacing processes with ResourcesTerminating condition


brownleej:
### What would you like to be added/changed?

If a pod gets deleted but is partitioned, it can stay in the `ResourcesTerminating` state indefinitely. Normally when a pod is deleted, we wait for it to get deleted and bring up a new pod for the same process group, so that we can re-use some of the other resources and recover more quickly. However, if a process is taking a long time to terminate, I think it would be better to replace it automatically, so we can recover the lost capacity. We could accomplish this by adding `ResourcesTerminating` to the list of conditions that are eligible for automatic replacement.

This is especially salient because when a pod is in terminating, we clear all of the other conditions for the process group. This means that even if the process group would normally get replaced because of the `MissingProcesses` condition, the operator will miss this when it is in terminating.

johscheuer:
We only set the `ResourcesTerminating ` when a process group is marked fro removal already (see: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/controllers/update_status.go#L413-L425). Adding the condition to the conditions that trigger a replacement won't have an effect since the process group must be already marked for removal to get the `ResourcesTerminating` condition. I think the correct approach is to add the `MissingPod` condition to the replacement conditions.

PR #1220, opened 2022-05-15
Add more conditions for automatic replacement

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1213

## Type of change

## Discussion

This should catch the case where a Pod is deleted (or in a deleted state) but the process is still running.

## Testing

Unit testing.

## Documentation

-

## Follow-up

-


brownleej (approved):
👍

Issue #1215, opened 2022-05-13
Provide a metric for how many process groups per process class type are expected.


johscheuer:
### What would you like to be added/changed?

We should provide a metric for how many process groups are desired for each process class.

PR #1221, opened 2022-05-15
Add metrics to reflect the desired number of processes

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1215

## Type of change

## Discussion

-

## Testing

Unit + local:

```bash
$ curl -s localhost:8080/metrics | rg 'fdb_operator_desired'
# HELP fdb_operator_desired_process_group_total the count of the desired Fdb process groups
# TYPE fdb_operator_desired_process_group_total gauge
fdb_operator_desired_process_group_total{name="test-cluster",namespace="default",process_class="cluster_controller"} 1
fdb_operator_desired_process_group_total{name="test-cluster",namespace="default",process_class="log"} 4
fdb_operator_desired_process_group_total{name="test-cluster",namespace="default",process_class="storage"} 3
```

## Documentation

-

## Follow-up

-

### DIFF
--- a/controllers/metrics.go
+++ b/controllers/metrics.go
@@ -61,7 +61,7 @@ var (
 
 	descProcessGroupsToRemoveWithoutExclusion = prometheus.NewDesc(
 		"fdb_operator_process_group_to_remove_without_exclusion_total",
-		"the count of rocess groups that should be removed from the cluster without excluding.",
+		"the count of process groups that should be removed from the cluster without excluding.",
 		descClusterDefaultLabels,
 		nil,
 	)
@@ -90,6 +90,13 @@ var (
 	descProcessGroupMarkedExcluded = prometheus.NewDesc(
 		"fdb_operator_process_group_marked_excluded",
 		"the count of Fdb process groups that are marked as excluded.",
+		append(descClusterDefaultLabels, "process_class"),
+		nil,
+	)
+
+	desDesiredProcessGroups = prometheus.NewDesc(
+		"fdb_operator_desired_process_group_total",
+		"the count of the desired Fdb process groups",
 		append(descClusterDefaultLabels, "process_class"),
 		nil,
 	)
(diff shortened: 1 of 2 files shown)


AnnigeriShambu (approved):

controllers/metrics.go
```diff
@@ -151,6 +158,15 @@ func collectMetrics(ch chan<- prometheus.Metric, cluster *fdbv1beta2.FoundationD
 		addGauge(descProcessGroupMarkedRemoval, float64(removals[pclass]), string(pclass))
 		addGauge(descProcessGroupMarkedExcluded, float64(exclusions[pclass]), string(pclass))
 	}
+
+	counts, err := cluster.GetProcessCountsWithDefaults()
+	if err != nil {
+		return
```
09harsh, line 164: Can we add a log statement here. Just to know the error and for bookkeeping that we hadn't calculated the metrics for desired number of processes.

controllers/metrics.go
```diff
@@ -151,6 +158,15 @@ func collectMetrics(ch chan<- prometheus.Metric, cluster *fdbv1beta2.FoundationD
 		addGauge(descProcessGroupMarkedRemoval, float64(removals[pclass]), string(pclass))
 		addGauge(descProcessGroupMarkedExcluded, float64(exclusions[pclass]), string(pclass))
 	}
+
+	counts, err := cluster.GetProcessCountsWithDefaults()
```
09harsh, line 162: Can we add a single line comment here which specifies this metric as the last metric to be calculated in the function. As it includes a return statement when it encounters error. If in future, we define new metrics that are defined below this, on encountering error we will not calculate those as well. Another approach could be putting this in another function to avoid such complexities.

09harsh (approved):
I left few comments but rest LGTM 👍

controllers/metrics.go
johscheuer: I wouldn't add a comment here, if someone touches the code the person has to see that the return will stop the calculation of any metrics that follow after that. In the whole code base we don't comment on the last return value and I'm not sure how valuable this comment would be.

Moving it into a dedicated method sounds like a good idea.

controllers/metrics.go
johscheuer: Adding a logger here requires some refactoring. I can open an issue for that to add a logger to the metrics collector.

controllers/metrics.go
johscheuer: I just looked over it and moving this code part into a dedicated method is not that trivial otherwise we have to replicate the addGauge method and everything. I'll rase an issue to rework the structure later.

Issue #849, opened 2021-07-13
Create a debug log when we create resources


johscheuer:
We should emit a debug log when we create new resources or delete resources e.g. like Pods, Services, Process Groups.

PR #1222, opened 2022-05-16
adding log statements

09harsh:
# Description
Addresses: #849 

## Type of change

- Other

## Discussion

-

## Testing
-

## Documentation

-

## Follow-up

-

### DIFF
--- a/controllers/add_pvcs.go
+++ b/controllers/add_pvcs.go
@@ -66,6 +66,7 @@ func (a addPVCs) reconcile(ctx context.Context, r *FoundationDBClusterReconciler
 
 			owner := internal.BuildOwnerReference(cluster.TypeMeta, cluster.ObjectMeta)
 			pvc.ObjectMeta.OwnerReferences = owner
+			log.Info("Creating PVC", "name", pvc.Name)
 			err = r.Create(ctx, pvc)
 			if err != nil {
 				if internal.IsQuotaExceeded(err) {
--- a/controllers/add_services.go
+++ b/controllers/add_services.go
@@ -56,6 +56,7 @@ func (a addServices) reconcile(ctx context.Context, r *FoundationDBClusterReconc
 			}
 			owner := internal.BuildOwnerReference(cluster.TypeMeta, cluster.ObjectMeta)
 			service.ObjectMeta.OwnerReferences = owner
+			log.Info("Creating service", "name", service.Name)
 			err = r.Create(ctx, service)
 			if err != nil {
 				return &requeue{curError: err, delayedRequeue: true}
@@ -89,7 +90,7 @@ func (a addServices) reconcile(ctx context.Context, r *FoundationDBClusterReconc
 					return &requeue{curError: err}
 				}
 			} else if k8serrors.IsNotFound(err) {
-				// Create a new service
+				log.Info("Creating service", "name", service.Name)
 				err = r.Create(ctx, service)
 				if err != nil {
 					if internal.IsQuotaExceeded(err) {
--- a/controllers/remove_process_groups.go
+++ b/controllers/remove_process_groups.go
@@ -150,6 +150,7 @@ func removeProcessGroup(ctx context.Context, r *FoundationDBClusterReconciler, c
 		return err
 	}
 	if len(pvcs.Items) == 1 && pvcs.Items[0].DeletionTimestamp.IsZero() {
+		log.Info("Deleting pvc", "name", pvcs.Items[0].Name)
 		err = r.Delete(ctx, &pvcs.Items[0])
 		if err != nil {
 			return err
@@ -164,6 +165,7 @@ func removeProcessGroup(ctx context.Context, r *FoundationDBClusterReconciler, c
 		return err
 	}
 	if len(services.Items) == 1 && services.Items[0].DeletionTimestamp.IsZero() {
+		log.Info("Deleting service", "name", services.Items[0].Name)
 		err = r.Delete(ctx, &services.Items[0])
 		if err != nil {
 			return err
--- a/controllers/remove_services.go
+++ b/controllers/remove_services.go
@@ -48,6 +48,7 @@ func (u removeServices) reconcile(ctx context.Context, r *FoundationDBClusterRec
 		return &requeue{curError: err}
 	}
 
+	log.Info("Deleting service", "name", existingService.Name)
 	err = r.Delete(ctx, existingService)
 	if err != nil {
 		return &requeue{curError: err}
--- a/controllers/update_backup_agents.go
+++ b/controllers/update_backup_agents.go
@@ -66,6 +66,7 @@ func (u updateBackupAgents) reconcile(ctx context.Context, r *FoundationDBBackup
 	}
 
 	if needCreation && deployment != nil {
+		log.Info("Creating deployment", "name", deployment.Name)
 		err = r.Create(ctx, deployment)
 		if err != nil {
 			return &requeue{curError: err}
@@ -84,6 +85,7 @@ func (u updateBackupAgents) reconcile(ctx context.Context, r *FoundationDBBackup
 	}
 
 	if !needCreation && deployment == nil {
+		log.Info("Deleting deployment", "name", existingDeployment.Name)
 		err = r.Delete(ctx, existingDeployment)
 		if err != nil {
 			return &requeue{curError: err}
--- a/controllers/update_config_map.go
+++ b/controllers/update_config_map.go
@@ -48,7 +48,7 @@ func (u updateConfigMap) reconcile(ctx context.Context, r *FoundationDBClusterRe
 	existing := &corev1.ConfigMap{}
 	err = r.Get(ctx, types.NamespacedName{Namespace: configMap.Namespace, Name: configMap.Name}, existing)
 	if err != nil && k8serrors.IsNotFound(err) {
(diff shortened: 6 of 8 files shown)


controllers/add_pvcs.go
```diff
@@ -66,6 +66,7 @@ func (a addPVCs) reconcile(ctx context.Context, r *FoundationDBClusterReconciler
 
 			owner := internal.BuildOwnerReference(cluster.TypeMeta, cluster.ObjectMeta)
 			pvc.ObjectMeta.OwnerReferences = owner
+			log.Info("Creating PVC", "name", pvc.Name)
```
johscheuer, line 69:
```suggestion
			log.V(1).Info("Creating PVC", "name", pvc.Name)
```

That should be a debug statement not an info statement. I would prefer if we also log the corresponding namespace and cluster name.

controllers/add_services.go
```diff
@@ -56,6 +56,7 @@ func (a addServices) reconcile(ctx context.Context, r *FoundationDBClusterReconc
 			}
 			owner := internal.BuildOwnerReference(cluster.TypeMeta, cluster.ObjectMeta)
 			service.ObjectMeta.OwnerReferences = owner
+			log.Info("Creating service", "name", service.Name)
```
johscheuer, line 59: That should be a debug statement not an info statement. I would prefer if we also log the corresponding namespace and cluster name.

controllers/add_services.go
```diff
@@ -89,7 +90,7 @@ func (a addServices) reconcile(ctx context.Context, r *FoundationDBClusterReconc
 					return &requeue{curError: err}
 				}
 			} else if k8serrors.IsNotFound(err) {
-				// Create a new service
+				log.Info("Creating service", "name", service.Name)
```
johscheuer, line 93: That should be a debug statement not an info statement. I would prefer if we also log the corresponding namespace and cluster name.

controllers/remove_process_groups.go
```diff
@@ -150,6 +150,7 @@ func removeProcessGroup(ctx context.Context, r *FoundationDBClusterReconciler, c
 		return err
 	}
 	if len(pvcs.Items) == 1 && pvcs.Items[0].DeletionTimestamp.IsZero() {
+		log.Info("Deleting pvc", "name", pvcs.Items[0].Name)
```
johscheuer, line 153: That should be a debug statement not an info statement. I would prefer if we also log the corresponding namespace and cluster name.

controllers/remove_process_groups.go
```diff
@@ -164,6 +165,7 @@ func removeProcessGroup(ctx context.Context, r *FoundationDBClusterReconciler, c
 		return err
 	}
 	if len(services.Items) == 1 && services.Items[0].DeletionTimestamp.IsZero() {
+		log.Info("Deleting service", "name", services.Items[0].Name)
```
johscheuer, line 168: That should be a debug statement not an info statement. I would prefer if we also log the corresponding namespace and cluster name.

controllers/remove_services.go
```diff
@@ -48,6 +48,7 @@ func (u removeServices) reconcile(ctx context.Context, r *FoundationDBClusterRec
 		return &requeue{curError: err}
 	}
 
+	log.Info("Deleting service", "name", existingService.Name)
```
johscheuer, line 51: That should be a debug statement not an info statement. I would prefer if we also log the corresponding namespace and cluster name.

controllers/update_backup_agents.go
```diff
@@ -66,6 +66,7 @@ func (u updateBackupAgents) reconcile(ctx context.Context, r *FoundationDBBackup
 	}
 
 	if needCreation && deployment != nil {
+		log.Info("Creating deployment", "name", deployment.Name)
```
johscheuer, line 69: That should be a debug statement not an info statement. I would prefer if we also log the corresponding namespace and cluster name.

controllers/update_backup_agents.go
```diff
@@ -84,6 +85,7 @@ func (u updateBackupAgents) reconcile(ctx context.Context, r *FoundationDBBackup
 	}
 
 	if !needCreation && deployment == nil {
+		log.Info("Deleting deployment", "name", existingDeployment.Name)
```
johscheuer, line 88: That should be a debug statement not an info statement. I would prefer if we also log the corresponding namespace and cluster name.

controllers/update_config_map.go
```diff
@@ -48,7 +48,7 @@ func (u updateConfigMap) reconcile(ctx context.Context, r *FoundationDBClusterRe
 	existing := &corev1.ConfigMap{}
 	err = r.Get(ctx, types.NamespacedName{Namespace: configMap.Namespace, Name: configMap.Name}, existing)
 	if err != nil && k8serrors.IsNotFound(err) {
-		logger.Info("Creating config map")
+		logger.Info("Creating config map", "name", configMap.Name)
```
johscheuer, line 51: That should be a debug statement not an info statement. I would prefer if we also log the corresponding namespace and cluster name.

mock-kubernetes-client/client/client.go
```diff
@@ -660,6 +661,7 @@ func (client *MockClient) Status() ctrlClient.StatusWriter {
 }
 
 func (client *MockClient) createEvent(event *corev1.Event) {
+	log.Log.Info("Creating event", "name", event.Name)
```
johscheuer, line 664: That should be a debug statement not an info statement. I would prefer if we also log the corresponding namespace and cluster name.

controllers/add_pods.go
```diff
@@ -117,7 +118,7 @@ func (a addPods) reconcile(ctx context.Context, r *FoundationDBClusterReconciler
 			pod.Annotations[fdbv1beta2.PublicIPAnnotation] = ip
 		}
 
-		err = r.PodLifecycleManager.CreatePod(ctx, r, pod)
+		err = r.PodLifecycleManager.CreatePod(logr.NewContext(ctx, logger), r, pod)
```
johscheuer, line 121: I think it makes sense to create the new context with the logger in the cluster controller e.g. here https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/controllers/cluster_controller.go#L96 and just assign that new context to the `ctx` variable that get's passed down into all `reconcile` methods. That way we are able to reduce the `log.WithValues` calls to only add the reconciler information.

This should be adde in another PR (I can create an issue later to track those efforts). Just wanted to. discuss that idea.

controllers/remove_process_groups.go
```diff
@@ -150,6 +151,7 @@ func removeProcessGroup(ctx context.Context, r *FoundationDBClusterReconciler, c
 		return err
 	}
 	if len(pvcs.Items) == 1 && pvcs.Items[0].DeletionTimestamp.IsZero() {
+		logr.FromContextOrDiscard(ctx).V(1).Info("Deleting pvc", "name", pvcs.Items[0].Name)
```
johscheuer, line 154: You missed the same log message above for deleting a Pod.

go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/apple/foundationdb/bindings/go v0.0.0-20190724023245-90ba203c166c
 	github.com/apple/foundationdb/fdbkubernetesmonitor v0.0.0-20211117195619-01c37a053a62
 	github.com/fatih/color v1.10.0
+	github.com/go-kit/kit v0.9.0
```
johscheuer, line 9: Was that pulled in by the `logr` usage?

mock-kubernetes-client/client/client.go
```diff
@@ -660,6 +661,7 @@ func (client *MockClient) Status() ctrlClient.StatusWriter {
 }
 
 func (client *MockClient) createEvent(event *corev1.Event) {
+	log.Log.V(1).Info("namespace", event.Namespace, "cluster", event.ClusterName, "Creating event", "name", event.Name)
```
johscheuer, line 664: I don't think we want to. have. those logs in the mock client.

pkg/podmanager/pod_lifecycle_manager.go
```diff
@@ -116,6 +119,7 @@ func (manager StandardPodLifecycleManager) CanDeletePods(_ context.Context, admi
 // UpdatePods updates a list of Pods to match the latest specs.
 func (manager StandardPodLifecycleManager) UpdatePods(ctx context.Context, r client.Client, _ *fdbv1beta2.FoundationDBCluster, pods []*corev1.Pod, _ bool) error {
 	for _, pod := range pods {
+		logr.FromContextOrDiscard(ctx).V(1).Info("Deleting pod", "name", pod.Name)
```
johscheuer, line 122: I would prefer to pull the `logr.FromContextOrDiscard(ctx)` call out an assign it to a variable instead of doing the same work multiple times in a loop.

johscheuer (changes requested):

controllers/remove_process_groups.go
09harsh: The above method is calling `DeletePod` and I already added logger there https://github.com/FoundationDB/fdb-kubernetes-operator/blob/9a2fe7f545f7b4e524baea0072fc05868bffeb74/pkg/podmanager/pod_lifecycle_manager.go#L110

johscheuer (approved):

Issue #814, opened 2021-07-04
Make use of RetryOnConflict


johscheuer:
The operator sometimes failed to update the status of a resource because another system has modified it. Instead of aborting and restarting the reconcile loop we should make use of the [RetryOnConflict](https://pkg.go.dev/k8s.io/client-go/util/retry?utm_source=godoc#RetryOnConflict) method provided by the Kubernetes client. That allows to retry an operation if a conflict was encountered. Since we only update the status of the `FoundationDBCluster` a retry with fetching the latest "version" of the spec should be fine.

PR #1241, opened 2022-06-09
Make use of retry method to prevent conflicts

johscheuer:
# Description

Fixes https://github.com/FoundationDB/fdb-kubernetes-operator/issues/814 and reduce the number of conflicts the operator sees.

## Type of change

## Discussion

There might be some other places for the backup and restore controller where we want to add this in the future. From my testing this improves the reconciliation speed and I didn't find any issues running the e2e test suite.

## Testing

Local testing.

## Documentation

-

## Follow-up

-

### DIFF
--- a/controllers/controllers.go
+++ b/controllers/controllers.go
@@ -104,7 +104,7 @@ func processRequeue(requeue *requeue, subReconciler interface{}, object runtime.
 	if err != nil && k8serrors.IsConflict(err) {
 		err = nil
 		if requeue.delay == time.Duration(0) {
-			requeue.delay = time.Minute
+			requeue.delay = 2 * time.Second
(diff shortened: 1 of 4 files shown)


controllers/controllers.go
```diff
@@ -104,7 +104,7 @@ func processRequeue(requeue *requeue, subReconciler interface{}, object runtime.
 	if err != nil && k8serrors.IsConflict(err) {
 		err = nil
 		if requeue.delay == time.Duration(0) {
-			requeue.delay = time.Minute
+			requeue.delay = 2 * time.Second
```
johscheuer, line 107: I think decreasing this time should be enough and we can remove the other retries.

PR #1242, opened 2022-06-15
remove duplicate updateStatus reconciler

bzd111:
remove duplicate updateStatus reconciler 

Signed-off-by: bzd111 <zxc@bzd111.me>

# Description
why clusterSubReconciler slice have duplicate updateStatus struct? I remove the first updateStatus{}.

https://github.com/FoundationDB/fdb-kubernetes-operator/blob/84406f793cf5fb85f14ed2756f33110204b1c42f/controllers/cluster_controller.go#L133

https://github.com/FoundationDB/fdb-kubernetes-operator/blob/84406f793cf5fb85f14ed2756f33110204b1c42f/controllers/cluster_controller.go#L156

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -130,7 +130,6 @@ func (r *FoundationDBClusterReconciler) Reconcile(ctx context.Context, request c
 	}
 
 	subReconcilers := []clusterSubReconciler{
-		updateStatus{},
 		updateLockConfiguration{},
 		updateConfigMap{},
 		checkClientCompatibility{},


johscheuer (approved):
It's not well documented but this is intended, see: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/docs/manual/technical_design.md#updatestatus

The idea is that the first `updateStatus` calculates what needs to be changed by adding information in the status field(s) and the final call is to check if the cluster is reconciled after all the previous sub reconcilers have acted.

johscheuer (changes requested):
Wrong selection

PR #1275, opened 2022-07-07
Add a new reconciler to put the cluster into maintenance mode during …

bnamasivayam:
…rolling upgrade

x New feature (non-breaking change which adds functionality)
- Documentation
- Other

## Discussion

## Testing

Manual testing?*
Unit tests and manual testing on AWS

## Documentation

## Follow-up


controllers/maintenance_mode_checker.go
```diff
@@ -0,0 +1,121 @@
...
+	// FDB Cluster is in maintenance mode due to this operator actions
+	if cluster.Status.MaintenanceModeInfo.StartTimestamp.Add(time.Duration(podUpCheckDelayInSeconds) * time.Second).After(time.Now()) {
+		return &requeue{message: "Waiting for delay to expire for checking if pods under maintenance mode have been bounced", delayedRequeue: true}
+	}
+	status, err := adminClient.GetStatus()
+	if err != nil {
+		return &requeue{curError: err}
+	}
+	processGroupsToCheck := make(map[string]struct{})
+	for _, id := range cluster.Status.MaintenanceModeInfo.ProcessGroups {
```
johscheuer, line 95: This has a race condition and we probably want to check the uptime of the processes otherwise we could run into the issue that we set the maintenance mode, the operator triggers the deletion of the Pods, which might be delayed by Kubernetes and the Pods which are still in deletion still have the processes running and now we would remove the maintenance mode before those Pods are actually recreated. Which would result in data distribution.

controllers/update_pods.go
```diff
@@ -47,6 +48,7 @@ func (updatePods) reconcile(ctx context.Context, r *FoundationDBClusterReconcile
 	}
 
 	updates := make(map[string][]*corev1.Pod)
+	zoneProcessGroupMap := make(map[string][]string)
```
johscheuer, line 51: I don't think we need this additional struct, all Pods will have the important information.

controllers/update_pods.go
```diff
@@ -226,6 +240,24 @@ func deletePodsForUpdates(ctx context.Context, r *FoundationDBClusterReconciler,
 		}
 	}
 
+	if deletionMode == fdbv1beta2.PodUpdateModeZone {
```
johscheuer, line 243: What is the benefit of moving this here?

johscheuer (changes requested):
I think we should add an additional setting to disable and enable maintenance mode if there are any issues with the implementation or the maintenance mode.

fdbclient/admin_client.go
```diff
@@ -41,7 +41,8 @@ import (
 )
 
 const (
-	fdbcliStr = "fdbcli"
+	fdbcliStr                        = "fdbcli"
+	maintenanceModeDurationInSeconds = 600
```
johscheuer, line 45: Shouldn't this be configurable?

johscheuer (changes requested):
Could you please answer my initial comments? I still think we should put this behind a feature flag/setting to enable/disable it.

fdbclient/admin_client.go
bnamasivayam: Yes. I will put this behind a flag and also make the duration configurable.

controllers/maintenance_mode_checker.go
```diff
@@ -0,0 +1,121 @@
...
+	defer adminClient.Close()
+
+	maintenanceZone, err := adminClient.GetMaintenanceZone()
+	if err != nil {
+		return &requeue{curError: err}
+	}
+	// TODO
+	logger.Info("maintenance mode zone", "zone", maintenanceZone)
+	// Cluster is not in maintenance mode
+	if maintenanceZone == "" {
```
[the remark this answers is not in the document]
brownleej: I believe that it will return an error if a different fault domain is already in maintenance mode. I agree that it would be best to have this gated with a lock.

controllers/maintenance_mode_checker.go
brownleej: +1

brownleej (changes requested):

controllers/maintenance_mode_checker.go
bnamasivayam: I am using a 60 second delay as in this variable podUpCheckDelayInSeconds and then check if the pod is present in the status to account for the race condition. I agree that using uptime is a better approach.

controllers/maintenance_mode_checker.go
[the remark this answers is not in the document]
bnamasivayam: I can gate it with a lock

controllers/maintenance_mode_checker.go
```diff
@@ -0,0 +1,114 @@
+/*
+ * maintenance_mode_.go
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2019-2021 Apple Inc. and the FoundationDB project authors
```
johscheuer, line 6:
```suggestion
 * Copyright 2022 Apple Inc. and the FoundationDB project authors
```

controllers/maintenance_mode_checker.go
```diff
@@ -0,0 +1,114 @@
+/*
+ * maintenance_mode_.go
```
johscheuer, line 2:
```suggestion
 * maintenance_mode_checker.go
```

controllers/maintenance_mode_checker.go
```diff
@@ -0,0 +1,114 @@
...
+	status, err := adminClient.GetStatus()
+	if err != nil {
+		return &requeue{curError: err}
+	}
+	processGroupsToCheck := make(map[string]struct{})
+	for _, id := range cluster.Status.MaintenanceModeInfo.ProcessGroups {
+		processGroupsToCheck[id] = struct{}{}
+	}
+	for _, process := range status.Cluster.Processes {
+		if _, ok := processGroupsToCheck[process.Locality["instance_id"]]; !ok {
```
johscheuer, line 85:
```suggestion
		if _, ok := processGroupsToCheck[process.Locality[fdbv1beta2.FDBLocalityInstanceIDKey]]; !ok {
```

controllers/maintenance_mode_checker.go
```diff
@@ -0,0 +1,114 @@
...
+			continue
+		}
+		if process.UptimeSeconds < time.Now().Sub(cluster.Status.MaintenanceModeInfo.StartTimestamp.Time).Seconds() {
+			delete(processGroupsToCheck, process.Locality["instance_id"])
+		} else {
+			return &requeue{message: fmt.Sprintf("Waiting for pod %s to be updated", process.Locality["instance_id"]), delayedRequeue: true}
```
johscheuer, line 91:
```suggestion
			return &requeue{message: fmt.Sprintf("Waiting for Pod %s to be updated", process.Locality[fdbv1beta2.FDBLocalityInstanceIDKey]), delayedRequeue: true}
```

I would prefer to just printout a log and not to return. This has the benefit that all processes will be logged and not only one. The requeue will be issued after the for loop anyway since `len(processGroupsToCheck) != 0` will be true.

controllers/maintenance_mode_checker.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	logger.Info("Cluster in maintenance mode", "zone", maintenanceZone)
+	// FDB Cluster is in maintenance mode due to this operator actions
+	status, err := adminClient.GetStatus()
+	if err != nil {
+		return &requeue{curError: err}
+	}
+	processGroupsToCheck := make(map[string]struct{})
+	for _, id := range cluster.Status.MaintenanceModeInfo.ProcessGroups {
+		processGroupsToCheck[id] = struct{}{}
```
johscheuer, line 82:
```suggestion
		processGroupsToCheck[id] = fdbv1beta2.None{}
```

controllers/maintenance_mode_checker_test.go
```diff
@@ -0,0 +1,151 @@
+/*
+ * remove_process_groups_test.go
```
johscheuer, line 2:
```suggestion
 * maintenance_mode_checker_test.go
 ```

File: controllers/maintenance_mode_checker_test.go [hunk 12]
```diff
@@ -0,0 +1,151 @@
...
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2019-2021 Apple Inc. and the FoundationDB project authors
```
johscheuer, line 6:
```suggestion
 * Copyright 2022 Apple Inc. and the FoundationDB project authors
```

controllers/update_pods.go
johscheuer: Any comment on this? It seems a little bit wasteful to generate this for all process groups instead of only the ones that are selected for removal. I'm okay with this as an initial implementation be we should have a follow up to improve this.

controllers/update_pods.go
```diff
@@ -226,6 +230,22 @@ func deletePodsForUpdates(ctx context.Context, r *FoundationDBClusterReconciler,
 		}
 	}
 
+	if deletionMode == fdbv1beta2.PodUpdateModeZone && cluster.UseMaintenaceMode() {
+		logger.Info("Setting maintenance mode", "zone", zone)
+		cluster.Status.MaintenanceModeInfo = fdbv1beta2.MaintenanceModeInfo{}
+		cluster.Status.MaintenanceModeInfo.StartTimestamp = &metav1.Time{Time: time.Now()}
+		cluster.Status.MaintenanceModeInfo.ZoneID = zone
+		cluster.Status.MaintenanceModeInfo.ProcessGroups = zoneProcessGroupMap[zone]
```
johscheuer, line 238:
```suggestion
		cluster.Status.MaintenanceModeInfo = fdbv1beta2.MaintenanceModeInfo{
			StartTimestamp: &metav1.Time{Time: time.Now()},
			ZoneID: zone,
			ProcessGroups: zoneProcessGroupMap[zone]	
		}
```

johscheuer (changes requested):
Could you please rebase your changes? I'm not sure why the GitHub actions are not running but I hesitate to approve a PR without having our CI checks running.

api/v1beta2/foundationdbcluster_types.go
```diff
@@ -1875,6 +1903,16 @@ func (cluster *FoundationDBCluster) GetWaitBetweenRemovalsSeconds() int {
 	return duration
 }
 
+// UseMaintenaceMode returns true if UseMaintenanceModeChecker is set.
+func (cluster *FoundationDBCluster) UseMaintenaceMode() bool {
+	return pointer.BoolDeref(cluster.Spec.AutomationOptions.MaintenanceModeOptions.UseMaintenanceModeChecker, true)
```
johscheuer, line 1908: I think we should disable this feature per default to preserve the current behaviour. Once we are doing a new release we can think of enabling it per default.

controllers/maintenance_mode_checker.go
```diff
@@ -0,0 +1,114 @@
...
+		processGroupsToCheck[id] = fdbv1beta2.None{}
+	}
+	for _, process := range status.Cluster.Processes {
+		if _, ok := processGroupsToCheck[process.Locality[fdbv1beta2.FDBLocalityInstanceIDKey]]; !ok {
+			continue
+		}
+		if process.UptimeSeconds < time.Since(cluster.Status.MaintenanceModeInfo.StartTimestamp.Time).Seconds() {
+			delete(processGroupsToCheck, process.Locality[fdbv1beta2.FDBLocalityInstanceIDKey])
+		} else {
+			return &requeue{message: fmt.Sprintf("Waiting for pod %s to be updated", process.Locality[fdbv1beta2.FDBLocalityInstanceIDKey]), delayedRequeue: true}
```
johscheuer, line 91: This is probably fine but I would have preferred to actually query the Pod deletion timestamp. With this implementation we have the potential risk that a restart/crash of a fdbserver process gives the wrong signal.

controllers/maintenance_mode_checker.go
```diff
@@ -0,0 +1,114 @@
...
+	if !hasLock {
+		return &requeue{curError: err}
+	}
+	logger.Info("Switching off maintenance mode", "zone", maintenanceZone)
+	err = adminClient.ResetMaintenanceMode()
+	if err != nil {
+		return &requeue{curError: err}
+	}
+	cluster.Status.MaintenanceModeInfo = fdbv1beta2.MaintenanceModeInfo{}
+	err = r.Status().Update(ctx, cluster)
```
johscheuer, line 109:
```suggestion
	err = r.updateOrApply(ctx, cluster)
```

This command was added in https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1319 to support server side apply.

controllers/update_pods.go
```diff
@@ -227,7 +228,29 @@ func deletePodsForUpdates(ctx context.Context, r *FoundationDBClusterReconciler,
...
+			ProcessGroups:  processGroups,
+		}
+		err = r.Status().Update(ctx, cluster)
+		if err != nil {
+			return &requeue{curError: err}
+		}
+		err = adminClient.SetMaintenanceZone(zone, cluster.GetMaintenaceModeTimeoutSeconds())
+		if err != nil {
+			return &requeue{curError: err}
+		}
```
johscheuer, line 250: Should this whole step be guarded by a lock? To prevent concurrent requests to set the maintenance zone.

controllers/maintenance_mode_checker.go
```diff
@@ -0,0 +1,114 @@
...
+
+	maintenanceZone, err := adminClient.GetMaintenanceZone()
+	if err != nil {
+		return &requeue{curError: err}
+	}
+	// Cluster is not in maintenance mode
+	if maintenanceZone == "" {
+		if cluster.Status.MaintenanceModeInfo.ZoneID != "" {
+			cluster.Status.MaintenanceModeInfo = fdbv1beta2.MaintenanceModeInfo{}
+			err = r.Status().Update(ctx, cluster)
```
johscheuer, line 56:
```suggestion
			err = r.updateOrApply(ctx, cluster)
```

controllers/update_pods.go
```diff
@@ -227,7 +228,29 @@ func deletePodsForUpdates(ctx context.Context, r *FoundationDBClusterReconciler,
...
+			processGroups = append(processGroups, pod.Labels[cluster.GetProcessGroupIDLabel()])
+		}
+
+		logger.Info("Setting maintenance mode", "zone", zone)
+		cluster.Status.MaintenanceModeInfo = fdbv1beta2.MaintenanceModeInfo{
+			StartTimestamp: &metav1.Time{Time: time.Now()},
+			ZoneID:         zone,
+			ProcessGroups:  processGroups,
+		}
+		err = r.Status().Update(ctx, cluster)
```
johscheuer, line 243:
```suggestion
		err = r.updateOrApply(ctx, cluster)
```

fdbclient/admin_client.go
```diff
@@ -345,6 +345,34 @@ func (client *cliAdminClient) ConfigureDatabase(configuration fdbv1beta2.Databas
 	return err
 }
 
+// GetMaintenanceZone gets current maintenance zone, if any. Returns empty string if maintenance mode is off
+func (client *cliAdminClient) GetMaintenanceZone() (string, error) {
+	status, err := client.GetStatus()
```
johscheuer, line 350: Do you know if there's a special key that could be used to read this information? Fetching the whole status only for one key seems sub-optimal.

bnamasivayam: I think there is one.

controllers/update_pods.go
bnamasivayam: It is already guarded by taking a lock a few lines above.

fdbclient/admin_client.go
bnamasivayam: I will add this as a TODO for future improvements. Since it is going to be off by default, we can address it before it is enabled.

controllers/maintenance_mode_checker.go
bnamasivayam: I will add this as a TODO

api/v1beta2/foundationdbcluster_types.go
bnamasivayam: Ok

johscheuer (commented):
Changes LGTM 👍 Could you look at the failing tests?

johscheuer (approved):
LGTM 👍

Issue #1281, opened 2022-07-08
Major/minor version upgrade can get stuck when operator is killed


johscheuer:
### What happened?

Currently there is a rather rare case that can happen during major/minor version upgrades of FDB. Those versions are version incompatible, which means older fdbservers are not able to communicate with the newer processes. During the upgrade process the operator issues a `kill` command to ensure that the current processes are getting started with the new fdbserver version. When the operator is getting deleted (for a different set of reasons) during the kill command this can leave the cluster in an undefined state, where the operator might not being able to recover from. In general the kill command is not reliable and we should use multiple kills to ensure we have all processes:

```bash
kill; kill <list>; sleep 1; kill <list> ; sleep 1; kill <list>;
```

In a second step we have to think about how we can recover the cluster if the operator is killed in between the kill command. A manual workaround currently is to just restart all fdbserver processes by hand (or by using the kubectl plugin) we might have to consider to to something similar in the operator.

### What did you expect to happen?

That the operator converges the cluster to the new version.

### How can we reproduce it (as minimally and precisely as possible)?

Run a HA upgrade test and let the operator be killed in some interval.

### Anything else we need to know?

-

### FDB Kubernetes operator

<details>

```console
$ kubectl fdb version
latest
```

</details>

### Kubernetes version

<details>

```console
$ kubectl version
-
```

</details>

### Cloud provider

<details>

</details>

halfprice:
kill command in 7.1 is slightly more reliable which checks remote before sending out the kill, but if the operator is killed in the middle, we may not be able to send the kill command.

I think we may need a mode that sends a kill without checking with the cluster controller to see if a worker is in the cluster or not. This can be the very last way to kill a process.

PR #1283, opened 2022-07-09
Initial changes to make upgrades more reliable

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1281
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1178

## Type of change

## Discussion

Those changes make the upgrade process more reliable and ensures we converge to the desired state.

## Testing

Locally + e2e tests

## Documentation

Updated.

## Follow-up

-


johscheuer:
CC @halfprince

controllers/restart_incompatible_processes.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	if err != nil {
+		return &requeue{curError: err, delay: 15 * time.Second}
+	}
+
+	// When we restarted the incompatible fdbserver processes we have to wait until they report to the cluster again to
+	// prevent to restart them in a loop.
+	if hasRestart {
+		return &requeue{message: "wait until processes are restarted", delay: 15 * time.Second}
```
johscheuer, line 101: I think we actually don't need to requeue here otherwise we might be in an endless loop.

controllers/cluster_controller.go
```diff
@@ -747,3 +756,32 @@ func (r *FoundationDBClusterReconciler) getCoordinatorSet(cluster *fdbv1beta2.Fo
...
+func (r *FoundationDBClusterReconciler) restartFdbserverProcess(name string, namespace string) error {
+	execReq := r.RestClient.
+		Post().
+		Namespace(namespace).
+		Resource("pods").
+		Name(name).
+		SubResource("exec").
+		VersionedParams(&corev1.PodExecOptions{
+			Container: "foundationdb",
+			Command:   []string{"sh", "-c", "pkill fdbserver"},
```
johscheuer, line 770: Validate that we are only killing the fdbserver if it points to the default version.

johscheuer:
The current upgrade tests fails because we have to add the following changes to the RBAC setup to allow the operator to restart the old processes:

```yaml
- apiGroups:
  - ""
  resources:
  - pods/exec
```

controllers/restart_incompatible_processes_test.go
```diff
@@ -0,0 +1,166 @@
...
+
+	"github.com/go-logr/logr"
+
+	fdbv1beta2 "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta2"
+	"github.com/FoundationDB/fdb-kubernetes-operator/internal"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("restart_incompatible_pods", func() {
```
xumengpanda, line 34: How will the test verify the `processIncompatibleProcesses` work with other reconcilers and ensure all processes will be upgraded?

Shall we have a FDB cluster test -- spin up a cluster, create transient network partition to some pods during the upgrade, and the cluster should always be able to upgrade all processes?

controllers/suite_test.go
```diff
@@ -186,5 +191,16 @@ func createTestClusterReconciler() *FoundationDBClusterReconciler {
 		PodLifecycleManager:    podmanager.StandardPodLifecycleManager{},
 		PodClientProvider:      internal.NewMockFdbPodClient,
 		DatabaseClientProvider: mockDatabaseClientProvider{},
+		RestClient:             &fake.RESTClient{Resp: &http.Response{StatusCode: http.StatusOK}},
+		RestConfig: &rest.Config{
+			Host: "https://notused.co.uk:443",
```
brownleej, line 196: In general, I would recommend that we use a `.example` domain for a URL that we never intend to be called, so that we don't rely on unspecified behavior like a domain not existing at the moment.

brownleej (changes requested):

controllers/bounce_processes.go
```diff
@@ -187,13 +187,13 @@ func (bounceProcesses) reconcile(ctx context.Context, r *FoundationDBClusterReco
...
-		cluster.Status.RunningVersion = cluster.Spec.Version
-		err = r.Status().Update(ctx, cluster)
-		if err != nil {
-			return &requeue{curError: err}
+		// If the cluster was upgraded we will requeue and let the update_status command set the correct version.
+		// Updating the version in this method has the drawback that we upgrade the version independent of the success
+		// of the kill command. The kill command is not reliable, which means that some kill request might not be
+		// delivered and the return value will still not contain any error.
+		if upgrading {
+			return &requeue{message: "fetch latest status after upgrade"}
```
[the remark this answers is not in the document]
johscheuer: One idea is to read the version from the coordinators.

johscheuer, line 196: Taking a not here for myself, to get. the current running version (during version compatible upgrades we could count the processes by version and assume the version with the most processes running is the current "running" version.

fdbclient/admin_client.go
```diff
@@ -226,37 +252,88 @@ func (client *cliAdminClient) runCommand(command cliCommand) (string, error) {
...
 }
 
-// GetStatus gets the database's status
-func (client *cliAdminClient) GetStatus() (*fdbv1beta2.FoundationDBStatus, error) {
-	adminClientMutex.Lock()
-	defer adminClientMutex.Unlock()
+// runCommandWithBackoff is a wrapper around runCommand which allows retrying commands if they hit a timeout.
+func (client *cliAdminClient) runCommandWithBackoff(command string) (string, error) {
+	maxTimeoutInSeconds := 40
+	currentTimeoutInSeconds := DefaultCLITimeout
```
brownleej, line 258: At some point we might want to make this constant also expressed as a duration, but I don't think that's necessary for this PR.

brownleej (approved):
👍

Issue #1291, opened 2022-07-15
Plugin should not fetch all Pods to query process groups


johscheuer:
### What would you like to be added/changed?

Currently our [getProcessGroupIDsFromPod](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/kubectl-fdb/cmd/remove_process_group.go#L124) method fetches the cluster and then all Pods to get the process group ID. I think we should reconsider this process to either compute this information from the cluster spec or only query specific Pods instead of fetching the whole. list. For larger cluster this will make a significant impact.

Issue #761, opened 2021-06-17
Shrink in the plugin doesn't work was expected


johscheuer:
Currently the `--shrink` flag in the kubectl plugin doesn't work as expected since we query for all Pods of the cluster: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/master/kubectl-fdb/cmd/remove_instances.go#L186-L202 instead we should read the expected process counts directly from the cluster without reading the Pod count which is possible error prone and can lead to creating more Pods than expected (e.g. during migrations).

PR #1295, opened 2022-07-16
Reduce API call to Kubernetes to fetch process group ID information

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/761 (by removing this feature)
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1291

## Type of change

## Discussion

This change reduces the required API calls to Kubernetes to fetch the process group information, this has a benefit in larger Kubernetes clusters. I decided to remove the shrink logic in the remove process groups subcommand, I'm not aware of a use case where we used this and I don't think the additional complexity is worth it for an unused feature.

The new approach has one minor drawback, which I think is okay: Since the process group ID is now calculated based on the cluster spec we could run in a race condition where a user changes the prefix and the cluster is not yet fully reconciled and the user would use the plugin to remove a process group (which still has the old ID). In this case the workaround would be to not use the Pod and rather specify the process group ID directly.

## Testing

Unit tests + locally.

## Documentation

-

## Follow-up

-

### DIFF
--- a/kubectl-fdb/cmd/remove_process_group.go
+++ b/kubectl-fdb/cmd/remove_process_group.go
@@ -26,9 +26,7 @@ import (
 	"log"
 
 	fdbv1beta2 "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta2"
-	"github.com/FoundationDB/fdb-kubernetes-operator/internal"
 	"github.com/spf13/cobra"
-	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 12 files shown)


kubectl-fdb/cmd/remove_process_group.go
```diff
@@ -132,26 +121,12 @@ func replaceProcessGroups(kubeClient client.Client, clusterName string, processG
...
-
-	if withShrink {
-		var pods corev1.PodList
-		err := kubeClient.List(ctx.Background(), &pods,
-			client.InNamespace(namespace),
-			client.MatchingLabels(cluster.GetMatchLabels()),
-		)
+	// In this case the user has Pod name specified
+	if !useProcessGroupID {
+		processGroups, err = getProcessGroupIDsFromPodName(cluster, processGroups)
```
09harsh, line 126:
```suggestion
		processGroupIDs, err = getProcessGroupIDsFromPodName(cluster, processGroups)
```

09harsh (approved):
LGTM 👍 
just few NIT changes

PR #1301, opened 2022-07-28
Add a separate setting for enabling DNS names in pod locality

### LINKED ISSUE #1230: Setting DNS configuration on pods independently of using DNS in the cluster file

When we enable DNS names in the cluster file, we do a rolling bounce to recreate the pods in order to add the necessary DNS information. This slows down the process, and slows down our testing. It should be harmless to add all of the DNS metadata before someone tries to convert the cluster to use DNS names, and that may be desireable in the transition path, because actually using the DNS names requires an FDB upgrade and may require additional testing in the infrastructure. I think we should have an option to enable this DNS configuration independently of using DNS in the cluster file.

As an alternative, if we can find a way to avoid this configuration, such as having the DNS locality field be built off of static configuration and existing dynamic fields, that would help serve the same purpose.

brownleej:
# Description

This allows us to configure the pod spec aspects of the new DNS name support long before updating the cluster file itself, which should provide a smoother transition and easier testing of the conversion process.

Fixes #1230 

## Discussion

No.

## Testing

Manual testing?*

Manual testing.

No.

## Documentation

Yes.

No, I don't think so. It's fairly minor.

## Follow-up

We may want to make this new setting default to true in the next major release.


johscheuer (approved):

johscheuer (commented):
I wonder if those failures are a race condition want we should add an additional "kubectl wait"  to ensure the CRD is properly updated.

ltsampros (approved):
LGTM

AnnigeriShambu (approved):

johscheuer (commented):
Do you think you have some time to rebase those changes?

johscheuer (commented):
@brownleej what do you think are the next steps for this PR?

brownleej:
I think this just got lost in the noise when we shifted our focus onto stabilization last fall. I'll get this PR into a mergeable state.

brownleej:
I've opened a separate PR to try to fix the integration checks.

.github/workflows/pull_request.yml
```diff
@@ -95,6 +95,7 @@ jobs:
         # Ensure we can upgrade the CRD with the current changes
         make install
         kubectl wait --for condition="established" crd --all
+        sleep 60
```
johscheuer, line 98: Can wee make a comment here that this is a workaround and we should investigate how to solve this without a sleep?

api/v1beta2/foundationdbcluster_types.go
```diff
@@ -1718,6 +1718,12 @@ type RoutingConfig struct {
 	// latest stable version of FoundationDB.
 	UseDNSInClusterFile *bool `json:"useDNSInClusterFile,omitempty"`
 
+	// DefineDNSLocalityFields determines whether to define pod DNS names on pod
```
johscheuer, line 1721:
```suggestion
	// DefineDNSLocalityFields determines whether to define Pod DNS names on Pod
```

api/v1beta2/foundationdbcluster_types.go
```diff
@@ -1997,6 +2003,12 @@ func (cluster *FoundationDBCluster) UseDNSInClusterFile() bool {
 	return pointer.BoolDeref(cluster.Spec.Routing.UseDNSInClusterFile, false)
 }
 
+// DefineDNSLocalityFields determines whether we need to put DNS entries in the
+// pod spec and process locality.
```
johscheuer, line 2007:
```suggestion
// Pod spec and process locality.
```

johscheuer (commented):
Changes LGTM.

api/v1beta2/foundationdbcluster_types.go
brownleej: We may need a breakout session on whether "pod" is a proper noun for style purposes. 😁

johscheuer: Sounds reasonable, even the official docs are inconsistent 🙄 https://kubernetes.io/docs/concepts/workloads/pods/

johscheuer:
> * Commit ID: [655df4c](https://github.com/FoundationDB/fdb-kubernetes-operator/commit/655df4cf564dcae16ccbc9420b78eb127b75620b)
> * Duration 3:12:30
> * Result: ❌ **FAILED**
> * [Build Log](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator/fdb-kubernetes-operator-pr/4747ffc2-8bd9-40dd-b425-23618a582bf0.gz) terminal output (available for 30 days)
> * [Build Workspace](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator-pr/4747ffc2-8bd9-40dd-b425-23618a582bf0/workspace.zip) zip file of the working directory (available for 30 days)

All tests passed except for one test, that failed with:

```text
ERROR: This exclude may cause the total free space in the cluster to drop below 10%.\nType `exclude FORCE <ADDRESS...>' to exclude without checking free space.\n
```

This is a known issue that should be fixed soon.

johscheuer:
Thanks!

PR #1311, opened 2022-08-03
Correct error check for timeout

johscheuer:
# Description

Otherwise the timeout check doesn't work.

## Type of change

## Discussion

-

## Testing

Locally.

## Documentation

-

## Follow-up

-


fdbclient/admin_client.go
```diff
@@ -49,7 +49,7 @@ type timeoutError struct {
 }
 
 func (timeoutErr timeoutError) Error() string {
-	return fmt.Sprintf("timeout: %s", timeoutErr.err.Error())
+	return fmt.Sprintf("fdbcli timeout: %s", timeoutErr.err.Error())
```
johscheuer, line 52: Looking at this again, I think we should move the timeout error into the API and do an actual type check instead of checking a string.

internal/error_helper.go
```diff
@@ -44,11 +46,11 @@ func IsNetworkError(err error) bool {
 // IsTimeoutError returns true if the observed error was a timeout error
 func IsTimeoutError(err error) bool {
 	for err != nil {
-		if strings.Contains(err.Error(), "Specified timeout reached") {
+		// See: https://apple.github.io/foundationdb/api-error-codes.html
+		// 1031: Operation aborted because the transaction timed out
+		if fdbError, ok := err.(fdb.Error); ok && fdbError.Code == 1031 {
```
johscheuer, line 51: Our own Timeout error is probably the better choice otherwise we have to import the FDB client lib.

brownleej: Yes, I think that it's best to avoid having the FDB client library dependency in the internal package, because so many other things will depend on that. Maybe we could have the fdbclient package catch FDB timeout errors and wrap them in something that's easier to check for in other parts of the code.

brownleej (changes requested):

brownleej (approved):
👍

PR #1313, opened 2022-08-04
Ignore process groups if they have the sidecar unreachable condition in the bounce sub reconciler

johscheuer:
# Description

Make the rollout process for config changes more reliable to process groups that are in an unreachable status.

## Type of change

## Discussion

I think it's safe to ignore process groups if their sidecar is unreachable otherwise one single process group might block the whole reconciliation.

## Testing

Locally. Will add an additional e2e test before merging.

## Documentation

-

## Follow-up

-

### DIFF
--- a/controllers/bounce_processes.go
+++ b/controllers/bounce_processes.go
@@ -58,7 +58,7 @@ func (bounceProcesses) reconcile(ctx context.Context, r *FoundationDBClusterReco
 	minimumUptime := math.Inf(1)
 	addressMap := make(map[string][]fdbv1beta2.ProcessAddress, len(status.Cluster.Processes))
 	for _, process := range status.Cluster.Processes {
-		addressMap[process.Locality["instance_id"]] = append(addressMap[process.Locality["instance_id"]], process.Address)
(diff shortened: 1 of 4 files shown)


controllers/bounce_processes.go
```diff
@@ -68,6 +68,7 @@ func (bounceProcesses) reconcile(ctx context.Context, r *FoundationDBClusterReco
 	processesToBounce := fdbv1beta2.FilterByConditions(cluster.Status.ProcessGroups, map[fdbv1beta2.ProcessGroupConditionType]bool{
 		fdbv1beta2.IncorrectCommandLine: true,
 		fdbv1beta2.IncorrectPodSpec:     false,
+		fdbv1beta2.SidecarUnreachable:   false, // ignore all Process groups that are not reachable and therefor not get any config map updates.
```
brownleej, line 71:
```suggestion
		fdbv1beta2.SidecarUnreachable:   false, // ignore all Process groups that are not reachable and therefore not get any config map updates.
```

brownleej (approved):
👍

PR #1317, opened 2022-08-05
Add better logging for fdb client calls

johscheuer:
# Description

Adds a logging call when we read a value from FDB with the fdb client and add an additional log when the call ended to see how long the call took.

## Type of change

## Discussion

-

## Testing

Locally.

## Documentation

-

## Follow-up

-


brownleej (approved):
👍

Issue #1278, opened 2022-07-08
Make use of server side apply to reduce conflicts


johscheuer:
### What would you like to be added/changed?

We should look into using server side apply to reduce the conflicts that we see when updating the status of our resources. This means we have to change the `Update` status calls to use the appropriate `Patch` calls. We want to run some additional tests before merging this and we want to have a feature flag on the operator to enable/disable this feature (default should be false to match the current behaviour). We probably also have to adjust our Kubernetes mock to support the `Patch` call.

johscheuer:
Docs: https://kubernetes.io/docs/reference/using-api/server-side-apply/

PR #1319, opened 2022-08-15
Initial support for server side apply

johscheuer:
# Description

Add support for server side apply.
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1278

## Type of change

## Discussion

SSA is per default disabled and has to to enabled to use it.

## Testing

I tested it in a dev cluster and ran the e2e test suite we have.

## Documentation

Added in the help method of the operator.

## Follow-up

Once the support for SSA in the controller-runtime is more mature we might change some things.


AnnigeriShambu (approved):

johscheuer:
Discussed some concerns with @ltsampros offline and we agreed that the current setup is fine.

Issue #1323, opened 2022-08-17
Cleanup Fake KubeClient Construction


jleach4:
### What would you like to be added/changed?

Remove some repetitive calls and setup (nit).

johscheuer:
We refactored the implementation of the fake kubeclient.

PR #1324, opened 2022-08-17
Cleanup Fake KubeClient Construction #1323

jleach4:
Cleaning up repetitive calls in construction of fake KubeClient.


johscheuer (commented):
Thanks, changes look good. Could you run `make fmt lint` and fix the resulting issues?

PR #1330, opened 2022-08-18
Fix issue 816

### LINKED ISSUE #816: Add wait time between actions
Currently the kubectl fdb plugin doesn't support any wait time between actions e.g. restarting processes or deleting Pods. There are some use cases where a wait time would be useful to prevent a bigger impact on the cluster. The wait time should be configured over a flag and the plugin should wait between sequential operations for the defined time. Per default we should either wait a short time period or disable it.

manfontan:
# Description

Fixes #816. This PR adds a `sleep` flag to the root command. It also implements the sleep flag for the `remove_process_group` and `restart` commands. The default sleep time between consecutive operations is zero.

## Type of change

## Discussion

The issue mentions the restart and delete commands. So I focused on those if there are other commands that may benefit from the sleep flag we can implement them as well.

## Testing

Manual testing?*
Existing unit test should cover this changes. Not sure how to test this since this is my first PR for the plugin. Beside building and manually testing the plugin. Which I didn't.

This should work for any cluster.

## Documentation

Nope

N/A

N/A

## Follow-up

Yes the default sleep time for the `kubectl fdb` would be zero seconds.


johscheuer (approved):
Thanks! LGTM 👍

brownleej (approved):
👍

PR #1332, opened 2022-08-24
Update the documentation on per-pod service IPs to reflect where the feature stands.

brownleej:
# Description

Given that the feature has remained in an experimental state for over a year, and we passed the 1.0 release without making it the default, I think it's worth changing our stance.

## Type of change

- Documentation

## Discussion

No.

## Testing

Manual testing?*

None.

No.

## Documentation

Yes.

## Follow-up

We might need additional discussion on how to move it out of the experimental phase, in one direction or the other.

### DIFF
--- a/docs/manual/customization.md
+++ b/docs/manual/customization.md
@@ -230,7 +230,7 @@ Using pod IPs can present several challenges:
 
 ### Service IPs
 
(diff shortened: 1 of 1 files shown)


johscheuer (approved):
LGTM 👍

PR #1333, opened 2022-08-25
Ignore excluded processes from minimum uptime calculation when doing a rolling bounce

ltsampros:
# Description

Ignore processes that are excluded when calculating the minimum uptime.

## Type of change

## Testing
Unit tests

### DIFF
--- a/controllers/bounce_processes.go
+++ b/controllers/bounce_processes.go
@@ -57,10 +57,14 @@ func (bounceProcesses) reconcile(ctx context.Context, r *FoundationDBClusterReco
 
 	minimumUptime := math.Inf(1)
 	addressMap := make(map[string][]fdbv1beta2.ProcessAddress, len(status.Cluster.Processes))
+
 	for _, process := range status.Cluster.Processes {
+		if process.Excluded {
+			continue
+		}
 		addressMap[process.Locality[fdbv1beta2.FDBLocalityInstanceIDKey]] = append(addressMap[process.Locality[fdbv1beta2.FDBLocalityInstanceIDKey]], process.Address)
 
-		if process.UptimeSeconds < minimumUptime {
+		if process.UptimeSeconds < minimumUptime && !process.Excluded {
 			minimumUptime = process.UptimeSeconds
 		}
 	}
--- a/controllers/bounce_processes.go
+++ b/controllers/bounce_processes.go
@@ -61,7 +61,6 @@ func (bounceProcesses) reconcile(ctx context.Context, r *FoundationDBClusterReco
 	for _, process := range status.Cluster.Processes {
 		addressMap[process.Locality[fdbv1beta2.FDBLocalityInstanceIDKey]] = append(addressMap[process.Locality[fdbv1beta2.FDBLocalityInstanceIDKey]], process.Address)
 
-		// comment to trigger PR builder
 		if process.UptimeSeconds < minimumUptime && !process.Excluded {
 			minimumUptime = process.UptimeSeconds
 		}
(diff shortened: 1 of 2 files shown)


johscheuer (approved):
LGTM 👍

controllers/bounce_processes.go
```diff
@@ -57,10 +57,14 @@ func (bounceProcesses) reconcile(ctx context.Context, r *FoundationDBClusterReco
...
 	addressMap := make(map[string][]fdbv1beta2.ProcessAddress, len(status.Cluster.Processes))
+
 	for _, process := range status.Cluster.Processes {
+		if process.Excluded {
+			continue
+		}
 		addressMap[process.Locality[fdbv1beta2.FDBLocalityInstanceIDKey]] = append(addressMap[process.Locality[fdbv1beta2.FDBLocalityInstanceIDKey]], process.Address)
 
-		if process.UptimeSeconds < minimumUptime {
+		if process.UptimeSeconds < minimumUptime && !process.Excluded {
```
johscheuer, line 67: What is the benefit of checking here that the process is not excluded? If the process is excluded we will continue the loop anyway?

johscheuer (changes requested):
Sorry for the early approval had an outdated view of the code.

controllers/bounce_processes.go
ltsampros: I think we need the process addresses in the `addressMap` otherwise we'll hit https://github.com/FoundationDB/fdb-kubernetes-operator/blob/c02da4a05c0808c17604ebab2b13ebcdc47dbf2c/controllers/bounce_processes.go#L110

ltsampros:
> Sorry for the early approval had an outdated view of the code.

Apologies for that, I read the code more after submitting and realised the other approach is better.

controllers/bounce_processes.go
johscheuer: My comment was actually referring to the statement in line 62-64 where we check if the process is excluded and if so we will continue the for loop (which is correct) but in line 68 we check again if the process is not excluded which is a case we can't hit based on the current code.

I think there is still one case we have to solve: If a process was manually excluded (but never included again) we potentially hit the case that `len(missingAddress) > 0` is true and we requeue forever until the process is included again or marked as removal. I believe the correct solution to this is to remove the if statement in line 62-64 and having the statement `process.UptimeSeconds < minimumUptime && !process.Excluded` should be enough for our use case. Could you add a unit test to ensure we handle this case properly (I hope I explained it well enough).

johscheuer (changes requested):
JFYI: You can trigger the e2e pipeline by closing and reopening the PR.

controllers/bounce_processes.go
ltsampros: It was supposed to be removed. Amended.

controllers/bounce_processes.go
```diff
@@ -57,10 +57,12 @@ func (bounceProcesses) reconcile(ctx context.Context, r *FoundationDBClusterReco
 
 	minimumUptime := math.Inf(1)
 	addressMap := make(map[string][]fdbv1beta2.ProcessAddress, len(status.Cluster.Processes))
+
 	for _, process := range status.Cluster.Processes {
 		addressMap[process.Locality[fdbv1beta2.FDBLocalityInstanceIDKey]] = append(addressMap[process.Locality[fdbv1beta2.FDBLocalityInstanceIDKey]], process.Address)
 
-		if process.UptimeSeconds < minimumUptime {
+		// comment to trigger PR builder
```
johscheuer, line 64: Can we remove his comment again?

johscheuer (approved):
LGTM 👍 Please remove the comment before merging.

PR #1353, opened 2022-09-15
Add Localities to FoundationDBClusterSpec

manfontan:
# Description

This PR implements the first step towards supporting three data hall configuration in the operator.

As discussed in:
https://github.com/FoundationDB/fdb-kubernetes-operator/issues/348

## Type of change

New feature (non-breaking change which adds functionality)

## Discussion

The changes in this PR are based on 
https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/docs/design/three_datahall.md

## Testing

Manual testing?*

```
docker run --rm --entrypoint=/bin/bash -ti --platform="linux/amd64" -v $(pwd):/work -w /work docker.io/library/golang:1.18.5 ./scripts/setup_container.sh
```

```
root@f7a6ffa5d70d:/work# make manifests
...
/work
/go/bin/controller-gen "crd:maxDescLen=0,crdVersions=v1,generateEmbeddedObjectMeta=true" rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
# Per default controller-gen will generate a ClusterRole for our example we want to use a Role and the namespace marker doesn't
# work since it requires a namespace and kustomize doesn't support to change the Kind.
make: Warning: File 'config/crd/bases/apps.foundationdb.org_foundationdbclusters.yaml' has modification time 0.35 s in the future
make: warning:  Clock skew detected.  Your build may be incomplete.
```
```
root@f7a6ffa5d70d:/work# make test
go test  ./... -coverprofile cover.out
?       github.com/FoundationDB/fdb-kubernetes-operator [no test files]
ok      github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta1     0.388s  coverage: 41.2% of statements
ok      github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta2     0.516s  coverage: 46.8% of statements
?       github.com/FoundationDB/fdb-kubernetes-operator/cmd/po-docgen   [no test files]
...
```

This changes should not have any effect so additional testing is not required I believe.

## Documentation

No

Once the three data hall functionality is fully implemented it should be documented indeed.

N/A

N/A

## Follow-up

No


api/v1beta1/foundationdbcluster_types.go
```diff
@@ -2278,3 +2282,20 @@ func (cluster *FoundationDBCluster) GetProcessClassLabels() []string {
...
+	TopologyKey string `json:"topologyKey,omitempty"`
+}
+
+// GetLocalities returns the cluster localities
+func (cluster *FoundationDBCluster) GetLocalities() []Locality {
+	result := make([]Locality, 0)
+
+	result = append(result, cluster.Spec.Localities...)
+
+	return result
```
johscheuer, line 2300:
```suggestion
	localities := make([]Locality, len(cluster.Spec.Localities))
	copy(localities, cluster.Spec.Localities)

	return localities
```

Any reason we want to return a copy here?

johscheuer (commented):
Is the plan to add the functionality in another PR?

manfontan:
> Is the plan to add the functionality in another PR?

I can continue working on this PR no problem. If that is your preference.

api/v1beta1/foundationdbcluster_types.go
manfontan: No strong feelings about it.

johscheuer:
> > Is the plan to add the functionality in another PR?
> 
> I can continue working on this PR no problem. If that is your preference.

Personally I would prefer to implement them in the same PR otherwise we have new fields on the CRD that have no effect.

manfontan:
It probably needs some work but I believe most of the initial implementation should be covered.

To keep track of the Process Group Locality I have updated the Cluster Status to add a new field `ProcessGroupLocalityZoneID` which will be updated by the status update sub-reconciler.

The idea is to set the Pod nodeSelector depending on the distribution of the ProcessGroups. Picking the zone with less PG to add new pods.

The zone is passed down to the Monitor configuration to set the `locality_data_hall` argument using the `FDB_ZONE_ID` variable.

When removing pods we can follow the same approach. In order to unbalance the cluster. We can pick the zone with more pods and remove a process from it. (This is not yet implemented but I added some comments where I believe the logic should go).

controllers/cluster_controller.go
```diff
@@ -370,6 +373,381 @@ type clusterSubReconciler interface {
 	reconcile(ctx context.Context, r *FoundationDBClusterReconciler, cluster *fdbv1beta2.FoundationDBCluster) *requeue
 }
 
+// localityInfo captures information about a process for the purposes of
+// choosing diverse locality.
+type localityInfo struct {
```
johscheuer, line 378: It seems like those methods are reintroduced after the rebase.

internal/monitor_conf.go
```diff
@@ -256,6 +256,10 @@ func GetMonitorProcessConfiguration(cluster *fdbv1beta2.FoundationDBCluster, pro
 		configuration.Arguments = append(configuration.Arguments, monitorapi.Argument{Value: fmt.Sprintf("--locality_data_hall=%s", cluster.Spec.DataHall)})
 	}
 
+	if cluster.Spec.DatabaseConfiguration.RedundancyMode == fdbv1beta2.RedundancyModeThreeDataHall {
+		configuration.Arguments = append(configuration.Arguments, monitorapi.Argument{Value: fmt.Sprintf("--locality_data_hall=%s", "FDB_ZONE_ID")})
```
johscheuer, line 260: We used earlier the `FDB_ZONE_ID` for the zone locality, with this change the data_hall locality and the zoneid will be the same.

internal/pod_models.go
```diff
@@ -113,16 +113,30 @@ func GetService(cluster *fdbv1beta2.FoundationDBCluster, processClass fdbv1beta2
 }
 
 // GetPod builds a pod for a new process group
-func GetPod(cluster *fdbv1beta2.FoundationDBCluster, processClass fdbv1beta2.ProcessClass, idNum int) (*corev1.Pod, error) {
+func GetPod(cluster *fdbv1beta2.FoundationDBCluster, processClass fdbv1beta2.ProcessClass, idNum int, status *fdbv1beta2.FoundationDBStatus) (*corev1.Pod, error) {
 	name, id := GetProcessGroupID(cluster, processClass, idNum)
 
+	var tdhLocality fdbv1beta2.Locality
```
johscheuer, line 119:
```suggestion
	var podLocality fdbv1beta2.Locality
```

internal/pod_models.go
```diff
@@ -137,6 +151,31 @@ func GetPod(cluster *fdbv1beta2.FoundationDBCluster, processClass fdbv1beta2.Pro
 	}, nil
 }
 
+func getPodLocality(cluster *fdbv1beta2.FoundationDBCluster, processClass fdbv1beta2.ProcessClass, status *fdbv1beta2.FoundationDBStatus) (fdbv1beta2.Locality, error) {
```
johscheuer, line 154: This method should use the data hall and not the zone ID correct?

internal/replacements/replacements.go
```diff
@@ -188,22 +188,27 @@ func processGroupNeedsRemoval(cluster *fdbv1beta2.FoundationDBCluster, pod *core
 		}
 	}
 
-	expectedNodeSelector := cluster.GetProcessSettings(processClass).PodTemplate.Spec.NodeSelector
-	if !equality.Semantic.DeepEqual(pod.Spec.NodeSelector, expectedNodeSelector) {
-		specHash, err := internal.GetPodSpecHash(cluster, processClass, idNum, nil)
-		if err != nil {
-			return false, err
-		}
+	// When three data hall redundancy is enabled. The node selecctor is not provided by the spec.
```
johscheuer, line 191: What if a user wants to define a nodeSelector and wants to use three data hall? e.g. a user wants to ensure that Pods are running on a node labels `foo=bar`.

internal/replacements/replacements.go
```diff
@@ -188,22 +188,27 @@ func processGroupNeedsRemoval(cluster *fdbv1beta2.FoundationDBCluster, pod *core
...
-				"reason", fmt.Sprintf("nodeSelector has changed from %s to %s", pod.Spec.NodeSelector, expectedNodeSelector))
-			return true, nil
+			if pod.ObjectMeta.Annotations[fdbv1beta2.LastSpecKey] != specHash {
+				logger.Info("Replace process group",
+					"reason", fmt.Sprintf("nodeSelector has changed from %s to %s", pod.Spec.NodeSelector, expectedNodeSelector))
+				return true, nil
+			}
 		}
 	}
+	//TODO(manuel.fontan) for three data hall if Localities Node Selector changes we should replace the affected nodes.
```
johscheuer, line 208:
```suggestion
	//TODO(manuel.fontan) for three data hall if Localities Node Selector changes we should replace the affected Pods.
```

api/v1beta2/foundationdbcluster_types.go
```diff
@@ -2181,6 +2196,33 @@ func (cluster *FoundationDBCluster) GetCrashLoopProcessGroups() (map[string]None
...
 
+// Locality represents a locality for the cluster processes.
+type Locality struct {
+	//The key identifying the locality
+	Key string `json:"key,omitempty"`
+	//The value of the locality
+	Value string `json:"value,omitempty"`
+	//The topology key (ex. topology.kubernetes.io/zone)
+	TopologyKey string `json:"topologyKey,omitempty"`
+	//The node selector map
```
johscheuer, line 2207: Can we add some more docs here how the different fields interact?

johscheuer (commented):
I'm not sure about the changes/used of the ZONE_ID, I would think for data hall we want to have a dedicated env variable and not reuse the zone id (since the zone ID is still used in the three data hall mode).

Could you add some more tests round the pod locality changes?

manfontan:
Hi sorry for the delay. 

Writing the tests is still on my backlog. These are the ones that I have in mind. Still need to get familiar with Ginkgo and figure out the best way to get them up and running though:

-  `When three data hall is enabled` 
  > `it should spread coordinators across data_halls`
  -  `When adding new process groups` 
      > `it should balance process groups across data_halls`
      > `it should add the process group to the smallest data_hall`
  - `When replacing process groups` > `it should place the new process group in the previous data_hall`
  - `When removing process groups` > `it should remove the process group from the largest data_hall`
  
  Once I have the tests running I will fix any issues with the draft logic I have implemented.

manfontan:
@johscheuer I have updated the PR with your feedback and added some unit tests (I can add a few more if required). 
I have been looking at e2e testing but I am not sure how long it would take me to get this PR e2e tested. 
If you think this can be done in a reasonable time I would be happy to do so.

controllers/update_status_test.go
```diff
@@ -112,6 +113,53 @@ var _ = Describe("update_status", func() {
 				processGroup := processGroupStatus[len(processGroupStatus)-4]
 				Expect(processGroup.ProcessGroupID).To(Equal(fdbv1beta2.ProcessGroupID("storage-1")))
 				Expect(len(processGroupStatus[0].ProcessGroupConditions)).To(Equal(0))
+				Expect(processGroup.LocalityDataHall).To(Equal(""))
+			})
+		})
+
+		When("a process group is fine and data hall is replication enabled", func() {
```
manfontan, line 120: TODO(manuel.fontan) fix syntax

api/v1beta2/foundationdbcluster_types.go
```diff
@@ -2267,6 +2278,40 @@ func (cluster *FoundationDBCluster) GetCrashLoopProcessGroups() (map[ProcessGrou
...
+	//The value of the locality
+	// +kubebuilder:validation:Optional
+	Value string `json:"value,omitempty"`
+	//The topology key (ex. topology.kubernetes.io/zone).
+	// The topology key is used by the operator to set the podAntiAffinity rules. This field requires the sidecar to read the topology labels
+	// from the pod which is not supported in the current version of the sidecar. See https://github.com/apple/foundationdb/pull/8506
+	TopologyKey string `json:"topologyKey,omitempty"`
+	//The node selectors for the locality
+	// When three data hall replication is enabled the node selectors are used to evenly distribute the processes across the data halls.
+	// Requiring at least thrree node selectors, each mapping to a different data hall.
```
manfontan, line 2295: TODO(manuel.fontan): fix typo

controllers/update_status_test.go
manfontan: a process group is fine and data hall ~is~ replication *is* enabled

johscheuer (commented):
Thanks for the PR, I block some time this week to review it.

manfontan:
Duplicates https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1651

This implementation is for a single FoundationDBCluster but since it has been idle for such a long time and `three_data_hall` initial support is on the way I will close the PR for now.

PR #1372, opened 2022-10-12
Initial changes to remove update Pod config in bounce reconciler

johscheuer:
# Description

## Type of change

## Discussion

## Testing

Manual testing?*

## Documentation

## Follow-up

### DIFF
--- a/controllers/bounce_processes.go
+++ b/controllers/bounce_processes.go
@@ -30,7 +30,6 @@ import (
 	fdbv1beta2 "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta2"
 	"github.com/FoundationDB/fdb-kubernetes-operator/internal"
 	"github.com/FoundationDB/fdb-kubernetes-operator/pkg/fdbadminclient"
-	"github.com/FoundationDB/fdb-kubernetes-operator/pkg/podmanager"
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/utils/pointer"
@@ -63,7 +62,7 @@ func (bounceProcesses) reconcile(ctx context.Context, r *FoundationDBClusterReco
 		return &requeue{curError: err}
 	}
 
-	addresses, req := getProcessesReadyForRestart(ctx, logger, r, cluster, addressMap)
+	addresses, req := getProcessesReadyForRestart(logger, cluster, addressMap)
 	if req != nil {
 		return req
 	}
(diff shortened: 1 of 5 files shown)


controllers/bounce_processes.go
```diff
@@ -92,28 +88,18 @@ func (bounceProcesses) reconcile(ctx context.Context, r *FoundationDBClusterReco
 
 		addresses = append(addresses, addressMap[process]...)
 
-		processGroupID := podmanager.GetProcessGroupIDFromProcessID(process)
-		pod, err := r.PodLifecycleManager.GetPods(ctx, r, cluster, internal.GetSinglePodListOptions(cluster, processGroupID)...)
-		if err != nil {
-			return &requeue{curError: err}
-		}
-		if len(pod) == 0 {
```
johscheuer, line 100: In order to catch this case we could check for `MissingPod` or `PendingPod` on the other hand a missing/pending Pod should not block any restarts, we can assume that the newly created Pod will get the new configuration directly.

johscheuer (commented):
I believe the test failure is because of some limitations in our mock client and not the logic itself. I will dig into that.

johscheuer (commented):
Some more tests are breaking now, probably because of some wrong assumptions of the mock client back then.

johscheuer (commented):
Updated the PR to have the latest changes in main. This should make the pipeline succeed.

johscheuer:
Ping @saintstack @sears @jleach4

controllers/bounce_processes.go
```diff
@@ -148,7 +147,7 @@ func (bounceProcesses) reconcile(ctx context.Context, r *FoundationDBClusterReco
 
 // getProcessesReadyForRestart returns a slice of process addresses that can be restarted. If addresses are missing or not all processes
 // have the latest configuration this method will return a requeue struct with more details.
-func getProcessesReadyForRestart(ctx context.Context, logger logr.Logger, r *FoundationDBClusterReconciler, cluster *fdbv1beta2.FoundationDBCluster, addressMap map[string][]fdbv1beta2.ProcessAddress) ([]fdbv1beta2.ProcessAddress, *requeue) {
+func getProcessesReadyForRestart(logger logr.Logger, cluster *fdbv1beta2.FoundationDBCluster, addressMap map[string][]fdbv1beta2.ProcessAddress) ([]fdbv1beta2.ProcessAddress, *requeue) {
```
johscheuer, line 150: With those changes it will be easier to move this out into the internal package and write some additional unit tests around the logic.

controllers/bounce_processes.go
```diff
@@ -181,28 +181,18 @@ func getProcessesReadyForRestart(ctx context.Context, logger logr.Logger, r *Fou
...
-		processGroupID := podmanager.GetProcessGroupIDFromProcessID(process)
-		pod, err := r.PodLifecycleManager.GetPods(ctx, r, cluster, internal.GetSinglePodListOptions(cluster, processGroupID)...)
-		if err != nil {
-			return nil, &requeue{curError: err}
-		}
-		if len(pod) == 0 {
-			return nil, &requeue{message: fmt.Sprintf("No pod defined for process group ID: \"%s\"", processGroupID), delay: podSchedulingDelayDuration}
-		}
-
-		synced, err := r.updatePodDynamicConf(cluster, pod[0])
```
sbodagala, line 193: I thought you said during the review session that "update_status" controller already calls "updatePodDynamicConf()" and so we don't need to call it here - is that correct? But I see "update_config_pod" reconciler calling this method, and not "update_status"? Anyway, we don't need to call this here because "update_config_pod" reconciler is already calling it on each pod?

sbodagala (approved):

controllers/bounce_processes.go
johscheuer: What I said in the review session was that in the update_status we set the `fdbv1beta2.IncorrectConfigMap` and the actual call of updatePodDynamicConf will happen in `update_config_pod`. I probably was a little bit quick going over those two files.

vishesh:
@johscheuer Feel free to merge it. I don't have write access in this repository, so can't do myself.

PR #1374, opened 2022-10-18
Use deterministic name for cluster file

johscheuer:
# Description

This removes a resource leak. Before that we never removed the cluster files when we use lock client or if we used the client libraries.

## Type of change

## Discussion

This change will add a `Close()` method to the lock client to ensure the files are removed once the lock client is not required anymore.

## Testing

Local testing + unit tests.

## Documentation

-

## Follow-up

-

### DIFF
--- a/fdbclient/common.go
+++ b/fdbclient/common.go
@@ -23,7 +23,9 @@ package fdbclient
 import (
 	"errors"
 	"fmt"
+	"io/fs"
 	"os"
+	"path"
 	"time"
 
 	fdbv1beta2 "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta2"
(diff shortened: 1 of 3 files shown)


fdbclient/common.go
```diff
@@ -42,43 +42,55 @@ var DefaultCLITimeout = 10 * time.Second
...
 	}()
-	database, err := getFDBDatabase(cluster)
+	database, clusterFilePath, err := getFDBDatabase(cluster)
+	// Ensure we delete the temporary file
+	defer func() {
+		if clusterFilePath == "" {
+			return
+		}
+
+		err = os.Remove(clusterFilePath)
```
brownleej, line 88: Does removing the file clean up the client cache maintained by the bindings, or will we still be leaking all of those resources? I think that cache is keyed off of the cluster file path, so having the cluster file deleted and potentially reused later could produce additional undesired behavior.

brownleej (changes requested):

fdbclient/common.go
brownleej: reference on the caching in the go bindings: https://github.com/apple/foundationdb/blob/main/bindings/go/src/fdb/fdb.go#L283

brownleej:
We might also want to pursue a solution where we can use an in-memory connection string rather than using a file, as discussed in this issue on FDB core: https://github.com/apple/foundationdb/issues/567

fdbclient/common.go
brownleej: I think leaking the cache maintained by the bindings while removing the file is more dangerous than the current behavior. Without changes in the FDB client itself, I think the best we can do is to make the cluster file path deterministic. We can base it on the UUID of the FoundationDBCluster object so we can gracefully handle clusters being deleted and recreated with the same cluster name but different connection strings and volumes.

go.sum
```diff
@@ -167,7 +167,6 @@ github.com/getkin/kin-openapi v0.76.0/go.mod h1:660oXbgy5JFMKreazJaQTw7o+X00qeSy
 github.com/getsentry/raven-go v0.2.0/go.mod h1:KungGk8q33+aIAZUIVWZDr2OfAEBsO49PX4NzFV5kcQ=
 github.com/ghodss/yaml v0.0.0-20150909031657-73d445a93680/go.mod h1:4dBDuWmgqj2HViK6kFavaiC9ZROes6MMH2rRYeMEF04=
 github.com/ghodss/yaml v1.0.0/go.mod h1:4dBDuWmgqj2HViK6kFavaiC9ZROes6MMH2rRYeMEF04=
-github.com/ghodss/yaml v1.0.0/go.mod h1:4dBDuWmgqj2HViK6kFavaiC9ZROes6MMH2rRYeMEF04=
```
jleach4, line 170: Where these removals related to the code?

jleach4 (changes requested):
Can you explain the package removals?  Related to the code changes?

go.sum
johscheuer: Reverted these changes. They where removed when I tried another approach and ran `go mod tidy`.

fdbclient/common.go
```diff
@@ -40,27 +43,41 @@ const (
...
+		return clusterFileName, nil
 	}
-	err = clusterFile.Close()
+
+	// The content doesn't match, so we have to write the new content to the cluster file.
+	return clusterFileName, os.WriteFile(clusterFileName, []byte(connectionString), 0777)
+}
+
+// getFDBDatabase opens an FDB database. The result will be cached for
+// subsequent calls, based on the cluster namespace and name.
```
johscheuer, line 73:
```suggestion
// getFDBDatabase opens an FDB database.
```

We don't do the caching here but the FDB go bindings will do the caching for us implicitly.

sbodagala (approved):

vishesh (approved):

johscheuer:
Once the pipeline is happy I'm going to merge this one.

johscheuer:
I looked at the test failure and this is one of the new tests, probably something we have to make more reliable. I spend some time to debug the failure.

PR #1375, opened 2022-10-19
Ensure we skip all work in the bounce reconciler if we disable restarts

johscheuer:
# Description

If we disable to restart FDB processes we want to skip any further work and just skip the whole reconciler.

## Type of change

## Discussion

-

## Testing

Local testing + unit tests.

## Documentation

-

## Follow-up

-

### DIFF
--- a/api/v1beta2/foundationdbcluster_types.go
+++ b/api/v1beta2/foundationdbcluster_types.go
@@ -1210,7 +1210,6 @@ func (cluster *FoundationDBCluster) CheckReconciliation(log logr.Logger) (bool, 
 				if condition.ProcessGroupConditionType == IncorrectCommandLine && cluster.Status.Generations.NeedsBounce == 0 {
 					logger.Info("Pending restart of fdbserver processes", "state", "NeedsBounce")
 					cluster.Status.Generations.NeedsBounce = cluster.ObjectMeta.Generation
-					reconciled = false
 				}
 				conditions = append(conditions, condition.ProcessGroupConditionType)
 			}
--- a/api/v1beta2/foundationdbcluster_types.go
+++ b/api/v1beta2/foundationdbcluster_types.go
@@ -1199,6 +1199,7 @@ func (cluster *FoundationDBCluster) CheckReconciliation(log logr.Logger) (bool, 
 	cluster.Status.DesiredProcessGroups = desiredCounts.Total()
 	cluster.Status.ReconciledProcessGroups = 0
 
+	var needsRestart bool
 	for _, processGroup := range cluster.Status.ProcessGroups {
 		if processGroup.IsMarkedForRemoval() {
 			continue
@@ -1207,6 +1208,9 @@ func (cluster *FoundationDBCluster) CheckReconciliation(log logr.Logger) (bool, 
 		if len(processGroup.ProcessGroupConditions) > 0 {
 			conditions := make([]ProcessGroupConditionType, 0, len(processGroup.ProcessGroupConditions))
 			for _, condition := range processGroup.ProcessGroupConditions {
+				if condition.ProcessGroupConditionType == IncorrectCommandLine {
+					needsRestart = true
+				}
 				conditions = append(conditions, condition.ProcessGroupConditionType)
 			}
 
@@ -1251,6 +1255,12 @@ func (cluster *FoundationDBCluster) CheckReconciliation(log logr.Logger) (bool, 
 	if cluster.Status.NeedsNewCoordinators {
 		logger.Info("Pending coordinator change", "state", "NeedsNewCoordinators")
 		cluster.Status.Generations.NeedsCoordinatorChange = cluster.ObjectMeta.Generation
+		reconciled = false
+	}
+
+	if needsRestart {
+		logger.Info("Pending restart of fdbserver processes", "state", "NeedsBounce")
+		cluster.Status.Generations.NeedsBounce = cluster.ObjectMeta.Generation
 		reconciled = false
 	}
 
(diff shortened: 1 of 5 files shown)


controllers/bounce_processes.go
```diff
@@ -60,10 +62,98 @@ func (bounceProcesses) reconcile(ctx context.Context, r *FoundationDBClusterReco
...
+	useLocks := cluster.ShouldUseLocks()
+	if useLocks {
+		lockClient, err = r.getLockClient(cluster)
+		if err != nil {
+			return &requeue{curError: err}
+		}
+	}
+	version, err := fdbv1beta2.ParseFdbVersion(cluster.Spec.Version)
+	if err != nil {
+		return &requeue{curError: err}
```
saintstack, line 100: Does getting the lockClient set a lock? If so, does returning here leave the lock in place (It will expire?)

controllers/bounce_processes.go
```diff
@@ -60,10 +62,98 @@ func (bounceProcesses) reconcile(ctx context.Context, r *FoundationDBClusterReco
...
+			return req
+		}
+		if addresses == nil {
+			return &requeue{curError: fmt.Errorf("unknown error when getting addresses that are ready for upgrade")}
+		}
+	}
+
+	logger.Info("Bouncing processes", "addresses", addresses, "upgrading", upgrading)
+	r.Recorder.Event(cluster, corev1.EventTypeNormal, "BouncingProcesses", fmt.Sprintf("Bouncing processes: %v", addresses))
+	err = adminClient.KillProcesses(addresses)
```
saintstack, line 134: What kind of failures do you think we could see here? Kill of some pods only? Requeue will get them on the next go around? If some were 'killed' (bounced), we'll leave these out on the next go around or will they be bounced again?

controllers/bounce_processes.go
```diff
@@ -60,10 +62,98 @@ func (bounceProcesses) reconcile(ctx context.Context, r *FoundationDBClusterReco
...
+	if err != nil {
+		return &requeue{curError: err}
+	}
+
+	// If the cluster was upgraded we will requeue and let the update_status command set the correct version.
+	// Updating the version in this method has the drawback that we upgrade the version independent of the success
+	// of the kill command. The kill command is not reliable, which means that some kill request might not be
+	// delivered and the return value will still not contain any error.
+	if upgrading {
+		return &requeue{message: "fetch latest status after upgrade"}
```
saintstack, line 144: What prevents us re-bouncing next time through?

controllers/bounce_processes.go
```diff
@@ -43,6 +41,10 @@ type bounceProcesses struct{}
 
 // reconcile runs the reconciler's work.
 func (bounceProcesses) reconcile(ctx context.Context, r *FoundationDBClusterReconciler, cluster *fdbv1beta2.FoundationDBCluster) *requeue {
+	if !pointer.BoolDeref(cluster.Spec.AutomationOptions.KillProcesses, true) {
+		return nil
```
saintstack, line 45: You don't want to record the old message here? It used to add an event with "Spec require a bounce of some processes, but killing processes is disabled"  and log an error?

api/v1beta2/foundationdbcluster_types.go
```diff
@@ -1254,6 +1258,12 @@ func (cluster *FoundationDBCluster) CheckReconciliation(log logr.Logger) (bool,
 		reconciled = false
 	}
 
+	if needsRestart {
```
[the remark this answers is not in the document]
johscheuer: That's a good point. Actually I can move the whole block up and remove the bool variable.

api/v1beta2/foundationdbcluster_types_test.go
```diff
@@ -3392,6 +3393,7 @@ var _ = Describe("[api] FoundationDBCluster", func() {
 				Expect(cluster.Status.Generations).To(Equal(ClusterGenerationStatus{
 					Reconciled:          1,
 					HasUnhealthyProcess: 2,
+					NeedsBounce:         2,
```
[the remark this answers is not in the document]
johscheuer: The `2` is the generation of the cluster that has a process group that is `Unhealthy` (that means at lease one process group as one or more conditions) and `NeedsBounce` means at least one process must be restarted (has the `IncorrectCommandLine` condition).

controllers/bounce_processes.go
johscheuer: Good, point. I'll add that back.

controllers/bounce_processes.go
johscheuer: The lock will expire and an error at this step should mean that the operator instance was not able to get a lock.

controllers/bounce_processes.go
johscheuer: We definitely can see that only some fdbserver processes are restarted and the operator will restart them in the next reconcile cycle (after waiting the minimum uptime). For upgrades with incompatible versions the answer is a little bit more complex and we can discuss that offline if you want.

controllers/bounce_processes.go
johscheuer: Before we try another restart of those processes we will fetch the current status of all fdbservers and at that point they should already be updated. Another "safeguard" is the minimum uptime.

api/v1beta2/foundationdbcluster_types_test.go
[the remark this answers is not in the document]
saintstack: nod.

saintstack (approved):

johscheuer:
I would wait for this PR: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1376 to be merged and retry the e2e tests.

johscheuer (commented):
I would wait for this PR: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1376 to be merged and retry the e2e tests.

PR #1376, opened 2022-10-19
Make remove incompatible processes more pessimistic

johscheuer:
# Description

*Please include a summary of the change and which issue is addressed. If this change resolves an issue, please include the issue number in the description.*

## Type of change

*Please select one of the options below.*

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)
- Documentation
- Other

## Discussion

*Are there any design details that you would like to discuss further?*

## Testing

*Please describe the tests that you ran to verify your changes. Unit tests?
Manual testing?*

*Do we need to perform additional testing once this is merged, or perform in a larger testing environment?*

## Documentation

*Did you update relevant documentation within this repository?*

*If this change is adding new functionality, do we need to describe it in our user manual?*

*If this change is adding or removing subreconcilers, have we updated the core technical design doc to reflect that?*

*If this change is adding new safety checks or new potential failure modes, have we documented and how to debug potential issues?*

## Follow-up

*Are there any follow-up issues that we should pursue in the future?*

*Does this introduce new defaults that we should re-evaluate in the future?*

PR #1416, opened 2022-10-26
Remove lock for exclusions

johscheuer:
# Description

Remove the lock mechanism from the exclude reconciler.

## Type of change

- Other

## Discussion

The initial idea of the lock was to reduce the risk of successive recoveries. This is only true if the exclude is happening in a different Kubernetes cluster/namespace not managed by this operator instance. Exclusions in the same namespace can still trigger successive recoveries. A drawback of this approach is that an exclusion in another Kubernetes cluster/namespace might be delay until this operator instance can get a lock.

If we want to use some safeguards around the exclusion we could think of using the minimum uptime, similar to the bounce reconciler.

## Testing

unit tests.

## Documentation

-

## Follow-up

-

### DIFF
--- a/controllers/exclude_processes.go
+++ b/controllers/exclude_processes.go
@@ -80,11 +80,6 @@ func (e excludeProcesses) reconcile(_ context.Context, r *FoundationDBClusterRec
 			}
 		}
 
-		hasLock, err := r.takeLock(cluster, fmt.Sprintf("excluding processes: %v", fdbProcessesToExclude))
-		if !hasLock {
-			return &requeue{curError: err}
-		}
-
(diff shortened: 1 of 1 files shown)


johscheuer (commented):
Test failure seems to be related to the fixes in https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1376

brownleej (approved):
👍

johscheuer:
The failure in the velocity test is unrelated and needs some additional investigation + improvements for robustness.

jleach4 (dismissed):
Test?

saintstack (approved):
I read the 'why' and fine by me. Try it. (Might be hard to write a test for this I'd imagine)

johscheuer:
> Test?

I can add a test tat verifies that no lock is taken, but I don't think that is a really useful test. We should rather trust our e2e tests (happy to add a dedicated test for his).

brownleej (approved):
👍

PR #1418, opened 2022-10-27
Refactor Pod models and move strings to constants

johscheuer:
# Description

Break the `GetPodSpec` method into smaller pieces and move repeated strings into a constant.

## Type of change

- Other (refactoring)

## Discussion

We should continue moving variable into constants that can be used everywhere instead of using magic strings, e.g. the container images.

## Testing

Local testing + unit tests.

## Documentation

-

## Follow-up

-


jleach4 (approved):
Small nit, but maybe add a comment to the new functions.

saintstack (approved):
The refactor of pod models is hard to follow. We have good coverage in here? Stuff will break if we put the containers together wrong? If so, +1.  The renames using constants look good.

ammolitor (approved):

PR #1421, opened 2022-10-28
Make use of client library to get the connection string

johscheuer:
# Description

The multi-version client library should be able to automatically detect the correct version to communicate with the FDB cluster.

## Type of change

## Discussion

I believe this is in general the right direction to use the client library where possible instead of having our code try multiple different configurations.

## Testing

Local testing and unit tests.

## Documentation

-

## Follow-up

-

### DIFF
--- a/controllers/update_status.go
+++ b/controllers/update_status.go
@@ -302,21 +302,19 @@ func optionList(options ...string) []string {
 	return values
 }
 
-// tryConnectionOptions attempts to connect with all the combinations of
-// versions and connection strings for this cluster and returns the set that
-// allow connecting to the cluster.
+// tryConnectionOptions attempts to connect with all the connection strings for this cluster and
+// returns the connection string that allows connecting to the cluster.
 func tryConnectionOptions(cluster *fdbv1beta2.FoundationDBCluster, r *FoundationDBClusterReconciler) (string, error) {
 	logger := log.WithValues("namespace", cluster.Namespace, "cluster", cluster.Name, "reconciler", "updateStatus")
 	connectionStrings := optionList(cluster.Status.ConnectionString, cluster.Spec.SeedConnectionString)
 
-	originalConnectionString := cluster.Status.ConnectionString
-
 	if len(connectionStrings) == 1 {
-		return originalConnectionString, nil
+		return cluster.Status.ConnectionString, nil
 	}
 
 	logger.Info("Trying connection options", "connectionString", connectionStrings)
 
+	originalConnectionString := cluster.Status.ConnectionString
 	defer func() { cluster.Status.ConnectionString = originalConnectionString }()
 
 	for _, connectionString := range connectionStrings {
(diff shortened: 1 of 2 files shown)


johscheuer (commented):
The next step to this would be to remove the `getConnectionString`  request (or at least deprecate it and don't use it) and make use of the connection string in the status. In all cases we fetch the status in addition to the connection string, by combining those requests we can reduce the number of FDB calls.

controllers/update_status.go
```diff
@@ -306,49 +305,44 @@ func optionList(options ...string) []string {
...
 	defer func() { cluster.Status.ConnectionString = originalConnectionString }()
 
-	for _, version := range versions {
-		for _, connectionString := range connectionStrings {
-			logger.Info("Attempting to get connection string from cluster",
-				"version", version, "connectionString", connectionString)
-			cluster.Status.RunningVersion = version
-			cluster.Status.ConnectionString = connectionString
-			adminClient, clientErr := r.getDatabaseClientProvider().GetAdminClient(cluster, r)
+	for _, connectionString := range connectionStrings {
```
jleach4, line 322: 1 at a time?

jleach4 (approved):

controllers/update_status.go
johscheuer: We could run them in a go routine but in the current implementation of the admin client we would work over the list in a serial manner anyway (and I'm not sure how go routines and the FDB bindings will work out together). In general we should have two different connection strings only for a short amount of time.

johscheuer:
The failure is related to the velocity test. I opened an issue to investigate that and I try to fix the test case today.

sears (dismissed):
This makes some parts of #1423 redundant.

johscheuer (commented):
Ping @ammolitor, could you check why @sears and @saintstack are not able to "approve" PRs?

controllers/update_status.go
```diff
@@ -306,49 +305,44 @@ func optionList(options ...string) []string {
...
-					"version", version, "connectionString", activeConnectionString)
-				return version, activeConnectionString, nil
-			}
-			logger.Error(err, "Error getting connection string from cluster",
-				"version", version, "connectionString", connectionString)
+		activeConnectionString, err := adminClient.GetConnectionString()
+
+		clientErr = adminClient.Close()
+		if clientErr != nil {
+			logger.V(1).Info("Cloud not close admin client")
```
saintstack, line 335: spelling

controllers/update_status.go
```diff
@@ -306,49 +305,44 @@ func optionList(options ...string) []string {
 // tryConnectionOptions attempts to connect with all the combinations of
 // versions and connection strings for this cluster and returns the set that
 // allow connecting to the cluster.
```
saintstack, line 307: Does comment need updating now this method works differently?

saintstack (approved):
Some nits but otherwise LGTM.

brownleej (commented):
How will this interact with the way the bindings cache the database connection?

saintstack (approved):

johscheuer:
> How will this interact with the way the bindings cache the database connection?

We probably have the same limitations like before. I have to double check the c/go bindings what happens if we reuse a database struct to ensure the cluster file content will be read again. Otherwise we don't add more limitations than before, so I would go ahead and merge this PR. I still will validate/read the c/go bindings for my own education (and curiosity).

PR #1423, opened 2022-11-02
update status cleanup

sears:
# Description

This PR helps reduce improve code complexity in the update status path.  It includes two functional changes:
 - Bypass the multi-version client when checking to see if the operator's FDB server version is correct, and getting an updated connection string.  Without this change, the operator would incorrectly infer the version the FDB cluster was running at.
 - Always use the client library when invoking `status json`.  I added a simple 1-second retry loop, under the assumption that the FDB multi-version client will eventually get lucky and connect to the version of the cluster that is currently available.  (If not, this is a serious FDB bug, and should be addressed there.)

## Type of change

- Documentation
- Other

## Discussion

## Testing

`make test` passes.  We should improve upgrade testing, in general.

## Documentation

Added inline comments

No

N/A

N/A

## Follow-up

The mock admin client does not inject errors, so we're not getting meaningful test coverage by invoking it.

This switches the default value of useClientLibrary from false to true, and removes support for false.  I added a deprecation warning to the docs.

### DIFF
--- a/controllers/admin_client_mock.go
+++ b/controllers/admin_client_mock.go
@@ -532,6 +532,10 @@ func (client *mockAdminClient) GetConnectionString() (string, error) {
 	return client.Cluster.Status.ConnectionString, nil
 }
 
+func (client *mockAdminClient) GetConnectionStringFromCLI() (string, error) {
+	return client.GetConnectionString()
+}
+
 // VersionSupported reports whether we can support a cluster with a given
 // version.
 func (client *mockAdminClient) VersionSupported(versionString string) (bool, error) {
--- a/controllers/controllers.go
+++ b/controllers/controllers.go
@@ -42,6 +42,17 @@ const (
 	podSchedulingDelayDuration = 15 * time.Second
 )
 
+// containsAll determines if one map contains all the keys and matching values
+// from another map.
+func containsAll(current map[string]string, desired map[string]string) bool {
+	for key, value := range desired {
+		if current[key] != value {
+			return false
+		}
+	}
+	return true
+}
+
 // metadataMatches determines if the current metadata on an object matches the
 // metadata specified by the cluster spec.
 func metadataMatches(currentMetadata metav1.ObjectMeta, desiredMetadata metav1.ObjectMeta) bool {
--- a/controllers/update_status.go
+++ b/controllers/update_status.go
@@ -154,23 +154,17 @@ func (updateStatus) reconcile(ctx context.Context, r *FoundationDBClusterReconci
 		return &requeue{curError: err}
 	}
 
-	status.ProcessGroups = make([]*fdbv1beta2.ProcessGroupStatus, 0, len(cluster.Status.ProcessGroups))
-	for _, processGroup := range cluster.Status.ProcessGroups {
-		if processGroup != nil && processGroup.ProcessGroupID != "" {
-			status.ProcessGroups = append(status.ProcessGroups, processGroup)
-		}
-	}
-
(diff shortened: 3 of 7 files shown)


controllers/controllers.go
```diff
@@ -42,6 +42,17 @@ const (
 	podSchedulingDelayDuration = 15 * time.Second
 )
 
+// containsAll determines if one map contains all the keys and matching values
+// from another map.
+func containsAll(current map[string]string, desired map[string]string) bool {
```
johscheuer, line 47: I believe a better place would be the internal package maybe a new `map_helper` file?

controllers/update_status.go
```diff
@@ -333,15 +315,21 @@ func tryConnectionOptions(cluster *fdbv1beta2.FoundationDBCluster, r *Foundation
 			adminClient, clientErr := r.getDatabaseClientProvider().GetAdminClient(cluster, r)
 
 			if clientErr != nil {
+				// If this fails, it means we couldn't instantaite a local object.  GetAdminClient does
+				// not attempt to connect to the FDB server.
```
johscheuer, line 319:
```suggestion
				// If this fails, it means we couldn't instantiate a local object.  GetAdminClient does
				// not attempt to connect to the FDB server.
```

fdbclient/admin_client.go
```diff
@@ -75,10 +75,6 @@ type cliAdminClient struct {
 	// custom parameters that should be set.
 	knobs []string
 
-	// Whether the admin client should be able to run operations through the
-	// client library rather than the CLI.
-	useClientLibrary bool
```
johscheuer, line 80: I think in the past @brownleej added this for doing some tests with DNS, but I don't remember the specifics and I'm happy to remove this code if we don't need it anymore.

fdbclient/admin_client.go
```diff
@@ -313,21 +296,24 @@ func (client *cliAdminClient) GetStatus() (*fdbv1beta2.FoundationDBStatus, error
 	adminClientMutex.Lock()
 	defer adminClientMutex.Unlock()
 
-	status, err := client.getStatus(client.useClientLibrary)
-	if err != nil {
-		return nil, err
-	}
+	for {
```
johscheuer, line 299: Should we limit the number of retries?

controllers/admin_client_mock.go
```diff
@@ -532,6 +532,10 @@ func (client *mockAdminClient) GetConnectionString() (string, error) {
 	return client.Cluster.Status.ConnectionString, nil
 }
 
+func (client *mockAdminClient) GetConnectionStringFromCLI() (string, error) {
```
johscheuer, line 535: Do you think we still need this one, once we get https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1421 merged? I think a better approach to using the fdbcli to determine the running version is to use the information from the status that we fetch in the next step in updateStatus.

For reference, this is the place where we check the running version: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/controllers/update_status.go#L99-L115 with a simple heuristic (just checking what version the actual fdbserver processes are running).

johscheuer (changes requested):
Could you run `make fmt lint` and fix the resulting issues? That should solve the lint issues.

fdbclient/admin_client.go
sears: I don't really know of a good way to test this change, but (at least in theory), we can delete it.

johscheuer: I think we can drop this feature since it was only used once for some manual testing.

PR #1425, opened 2022-11-08
Fix nil pointer error

johscheuer:
# Description

If the operator is not able to communicate with the sidecar HTTP server an error will be returned and the response will be nil. In the current implementation this would lead to a nil pointer exception: `panic: runtime error: invalid memory address or nil pointer dereference`

## Type of change

## Discussion

## Testing

Local testing + e2e tests.

## Documentation

-

## Follow-up

-

### DIFF
--- a/internal/pod_client_test.go
+++ b/internal/pod_client_test.go
@@ -22,8 +22,12 @@ package internal
 
 import (
 	fdbv1beta2 "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta2"
+	"github.com/hashicorp/go-retryablehttp"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"net/http"
+	"net/url"
+	"time"
 )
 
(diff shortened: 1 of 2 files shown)


internal/pod_client_test.go
```diff
@@ -58,4 +62,47 @@ var _ = Describe("pod_client", func() {
...
+				req, err := generateRequest(retryClient, target.String(), http.MethodGet, getTimeout, postTimeout)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(req.Method).To(Equal(http.MethodGet))
+				Expect(retryClient.HTTPClient.Timeout).To(Equal(getTimeout))
+			})
+		})
+
+		When("generating a http post request", func() {
+			It("should generate the request", func() {
+				req, err := generateRequest(retryClient, target.String(), http.MethodPost, getTimeout, postTimeout)
```
jleach4, line 92: generateRequestSafe function that incorporates all the checks would trim down all the test code a bit?

jleach4 (approved):

PR #1432, opened 2022-11-11
Refactor admin client

johscheuer:
# Description

Some initial refactoring to make the admin client package testable. For some features to test we have to spin up a FDB cluster and provide the binaries at the correct place, I'll open another issue for this to ensure we can work on this and make this package testable.

## Type of change

- Other

## Discussion

-

## Testing

Unit tests.

## Documentation

-

## Follow-up

Create an issue and add integration tests that interact with a FoundationDB cluster.


jleach4 (approved):
looks a bit better...

fdbclient/admin_client.go
```diff
@@ -123,19 +130,23 @@ type cliCommand struct {
 
 // hasTimeoutArg determines whether a command accepts a timeout argument.
 func (command cliCommand) hasTimeoutArg() bool {
-	return command.binary == "" || command.binary == fdbcliStr
+	return command.isFdbCli()
```
saintstack, line 133: hasTimeoutArg function gets fdbcli process name? Is this right?

fdbclient/admin_client.go
```diff
@@ -123,19 +130,23 @@ type cliCommand struct {
...
 }
 
-// hasDashInLogDir determines whether a command has a log-dir argument or a
-// logdir argument.
-func (command cliCommand) hasDashInLogDir() bool {
-	return command.binary == "" || command.binary == fdbcliStr
+// getLogDirParameter returns the log dir parameter for a command, depending on the binary the log dir parameter
+// has a dash or not.
```
saintstack, line 137: ugh

fdbclient/admin_client.go
```diff
@@ -151,75 +162,90 @@ func (command cliCommand) getTimeout() time.Duration {
 	return DefaultCLITimeout
 }
 
+// getVersion returns the versions defined in the command or if not present returns the running version of the
+// cluster.
+func (command cliCommand) getVersion(cluster *fdbv1beta2.FoundationDBCluster) string {
```
saintstack, line 167: This looks like we return two different answers out of a single method?

johscheuer:
> looks a bit better...

The next step at least for CLI calls would be to have a runner interface to check that the calls have the right format. For the transaction requests I believe we have to use a FDB cluster as integration tests.

fdbclient/admin_client.go
johscheuer: Yeah, fun times: https://github.com/apple/foundationdb/issues/6340 I have to reserve some time to actual do that.

fdbclient/admin_client.go
johscheuer: Not really. The logic is the follow:

- If the command has a version defined use that version (e.g. backup commands).
- Otherwise use the running version of the cluster (all calls in the cluster controller).

johscheuer (commented):
I'm spending some time tomorrow debugging the failure.

fdbclient/command_runner.go
```diff
@@ -0,0 +1,64 @@
...
+package fdbclient
+
+import (
+	"context"
+	"github.com/go-logr/logr"
+	"os/exec"
+)
+
+// commandRunner is an interface to run commands.
+type commandRunner interface {
```
johscheuer, line 30: This setup allows us to "run" whole commands and check if they behave like expected based on the mocked output. For everything that talks directly with FDB. without using a binary we have to implement a similar approach.

PR #1433, opened 2022-11-15
If configure database is set to false skip all work

johscheuer:
# Description

This change ensures that the operator is not performing any actions if we set `configureDatabase` to false. The idea is to reduce the load on the FDB cluster.

## Type of change

- Other

## Discussion

-

## Testing

Unit tests and e2e tests.

## Documentation

-

## Follow-up

-

### DIFF
--- a/controllers/update_database_configuration.go
+++ b/controllers/update_database_configuration.go
@@ -23,12 +23,11 @@ package controllers
 import (
 	"context"
 	"fmt"
-
-	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/utils/pointer"
 
 	fdbtypes "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta2"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 )
 
 // updateDatabaseConfiguration provides a reconciliation step for changing the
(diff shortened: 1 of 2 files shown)


controllers/update_database_configuration.go
```diff
@@ -37,6 +36,10 @@ type updateDatabaseConfiguration struct{}
 
 // reconcile runs the reconciler's work.
 func (u updateDatabaseConfiguration) reconcile(ctx context.Context, r *FoundationDBClusterReconciler, cluster *fdbtypes.FoundationDBCluster) *requeue {
+	if !pointer.BoolDeref(cluster.Spec.AutomationOptions.ConfigureDatabase, true) {
```
sbodagala, line 39: Do you want to log an event here, the way it used to do (on line 88 below)?

sbodagala: Also, why is it not applying the cluster status and doing a require the way it used to do before this change?

sbodagala (approved):

controllers/update_database_configuration.go
johscheuer: We can't log an event here since we skip all work in this reconciler. If we want to keep logging an event we would need to still perform the same operations, otherwise the reconciler can't know if the current configuration matches the desired configuration (or not). By require you mean a `requeue`? With the new approach there is no need for this, since we skip all work in this reconciler.

The new approach is an optimisation to reduce the load in the operator and on the FDB cluster.

PR #1442, opened 2022-11-24
Expose mock admin client and lock client

johscheuer:
# Description

This will expose the mock implementations for the AdminClient and the LockClient. The idea is to make it available for users of the fdbclient package for unit tests.

## Type of change

## Discussion

Since we are moving things around this will be a breaking change but my exception is that this allows users to use the mock client without breaking to many things in the future.

## Testing

Local testing with another test repository.

## Documentation

-

## Follow-up

-

### DIFF
--- a/pkg/fdbadminclient/mock/admin_client_mock.go
+++ b/pkg/fdbadminclient/mock/admin_client_mock.go
@@ -3,7 +3,7 @@
  *
  * This source file is part of the FoundationDB open source project
  *
- * Copyright 2020-2021 Apple Inc. and the FoundationDB project authors
+ * Copyright 2020-2022 Apple Inc. and the FoundationDB project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,7 +18,7 @@
  * limitations under the License.
  */
 
-package controllers
+package mock
 
 import (
 	"context"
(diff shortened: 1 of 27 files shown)


johscheuer:
Once this PR is ready to be merged we can work on https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1445

pkg/fdbadminclient/mock/admin_client_mock.go
```diff
@@ -28,58 +28,59 @@ import (
...
 	missingProcessGroups                     map[string]bool
 	additionalProcesses                      []fdbv1beta2.ProcessGroupStatus
 	localityInfo                             map[string]map[string]string
 	incorrectCommandLines                    map[string]bool
-	maxZoneFailuresWithoutLosingData         *int
-	maxZoneFailuresWithoutLosingAvailability *int
-	knobs                                    []string
-	maintenanceZone                          string
+	MaxZoneFailuresWithoutLosingData         *int
+	MaxZoneFailuresWithoutLosingAvailability *int
```
sbodagala, line 57: What is the convention here? When to do the capitalization and when not to?

sbodagala (approved):

pkg/fdbadminclient/mock/admin_client_mock.go
johscheuer: In go the capitalization is used to make a variable/method publicly exposed. All variables that are accesses from the outside have a capital letter.

PR #1446, opened 2022-11-29
If the cluster is upgraded wait for all processes to be ready

johscheuer:
# Description

Ensure we wait for all processes before doing the upgrade and don't skip processes in the `SidecarUnreachable` state. I believe in the long term we want to implement some kind of a ratio to ignore some processes before doing the upgrade (instead of blocking until the issue is resolved).

## Type of change

## Discussion

See above, we probably want to rethink the upgrade prerequisites to not require all processes being ready to be upgraded but rather a % or fixed number of processes.

## Testing

Unit tests + I tested this with the e2e tests to ensure the upgrade is blocked until all processes are "ready" to be upgraded.

## Documentation

-

## Follow-up

-

### DIFF
--- a/controllers/bounce_processes.go
+++ b/controllers/bounce_processes.go
@@ -25,6 +25,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/FoundationDB/fdb-kubernetes-operator/internal/restarts"
+
 	fdbv1beta2 "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta2"
 	"github.com/FoundationDB/fdb-kubernetes-operator/internal"
 	"github.com/FoundationDB/fdb-kubernetes-operator/pkg/fdbadminclient"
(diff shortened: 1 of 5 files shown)


controllers/bounce_processes.go
```diff
@@ -149,12 +151,7 @@ func (bounceProcesses) reconcile(ctx context.Context, r *FoundationDBClusterReco
 // getProcessesReadyForRestart returns a slice of process addresses that can be restarted. If addresses are missing or not all processes
 // have the latest configuration this method will return a requeue struct with more details.
 func getProcessesReadyForRestart(ctx context.Context, logger logr.Logger, r *FoundationDBClusterReconciler, cluster *fdbv1beta2.FoundationDBCluster, addressMap map[string][]fdbv1beta2.ProcessAddress) ([]fdbv1beta2.ProcessAddress, *requeue) {
-	processesToBounce := fdbv1beta2.FilterByConditions(cluster.Status.ProcessGroups, map[fdbv1beta2.ProcessGroupConditionType]bool{
-		fdbv1beta2.IncorrectCommandLine: true,
-		fdbv1beta2.IncorrectPodSpec:     false,
-		fdbv1beta2.SidecarUnreachable:   false, // ignore all Process groups that are not reachable and therefore will not get any ConfigMap updates.
-	}, true)
-
+	processesToBounce := fdbv1beta2.FilterByConditions(cluster.Status.ProcessGroups, restarts.GetFilterConditions(cluster), true)
```
sbodagala, line 154: Don't we need to log a warning or an event (in a way that is queryable on Splunk or visible on Hubble) indicating that bounce/upgrade is held back because of SidecarUnreachable processes? After this point if any process fails for any reason it will be restarted with the updated version, correct?

sbodagala (approved):

controllers/bounce_processes.go
```diff
@@ -205,6 +203,22 @@ func getProcessesReadyForRestart(ctx context.Context, logger logr.Logger, r *Fou
...
+			curError: err,
+		}
+	}
+
+	// If we upgrade the cluster wait until all processes are ready for the restart. In the future we can adjust this
+	// to only be a requirement for version incompatible upgrades. In addition we probably only want to block for a
+	// certain threshold either as a percentage or as a fixed number (which could also be 0).
+	if cluster.IsBeingUpgraded() && counts.Total() != len(addresses) {
+		return nil, &requeue{
+			message:        fmt.Sprintf("expected %d processes, got %d processes ready to restart", counts.Total(), len(addresses)),
```
johscheuer, line 218: @sbodagala this message will be available in the logs of the operator.

Issue #611, opened 2021-04-14
Distinguishing conditions for incorrect metadata, incorrect pod specs, and incorrect config maps


brownleej:
We currently use the IncorrectPodSpec condition for multiple different scenarios - incorrect pod metadata, incorrect pod specs, and incorrect config maps. This can be confusing when looking at the cluster status, as these scenarios have different causes and different implications for remediation. I think we should separate these conditions into different condition types.

johscheuer:
We may also want to start a document to define what each condition means (or how to resolve them).

PR #1454, opened 2022-12-12
Ensure that the configmap is synced before restarting processes

johscheuer:
# Description

This ensures that the ConfigMap is synced before restarting processes (this includes the generated fdbmonitor file). The current behaviour only works because of this bug: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/611

## Type of change

## Discussion

-

## Testing

Unit tests and e2e tests.

## Documentation

-

## Follow-up

-


brownleej (approved):
👍

PR #1455, opened 2022-12-12
Ensure that Pods that are stuck in terminating but still have a fdbserver process reporting are marked as failed

johscheuer:
# Description

Before that we would ignore Pods that are stuck in terminating (have a non-zero deletionTimestamp) and have a fdbserver process reporting to the cluster. Now we catch those cases and the operator will eventually replace them once they are marked as failed (assuming auto replacements are enabled). If we keep those Pods around they will/can become problematic in some cases.

## Type of change

## Discussion

-

## Testing

Unit tests I don't think we currently have the test setup to simulate such a failure outside of our unit test code.

## Documentation

-

## Follow-up

-


brownleej (approved):
👍

PR #1456, opened 2022-12-12
Add update hooks to the mock Kubernetes client

johscheuer:
# Description

This is a similar feature to the create hooks, were a user can define a custom method that will be executed during the creation of objects, just for the update method. This will be helpful to mock controllers that manages other resources (like a replicaSet).

## Type of change

## Discussion

-

## Testing

Unit tests.

## Documentation

It's probably worth creating a new PR + issue to better document our mock client implementation.

## Follow-up

-


brownleej (approved):
👍

PR #1463, opened 2022-12-27
Small clean up

johscheuer:
# Description

Small code clean up when I was reading some code.

## Type of change

- Other

## Discussion

-

## Testing

Locally with unit tests.

## Documentation

-

## Follow-up

-

### DIFF
--- a/controllers/update_labels.go
+++ b/controllers/update_labels.go
@@ -80,7 +80,7 @@ func (updateLabels) reconcile(ctx context.Context, r *FoundationDBClusterReconci
 		}
 
 		pvc, ok := pvcMap[processGroup.ProcessGroupID]
-		if !ok || pod == nil {
+		if !ok {
 			logger.V(1).Info("Could not find PVC for process group ID",
 				"processGroupID", processGroup.ProcessGroupID)
 			continue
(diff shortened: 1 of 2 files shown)


controllers/update_labels.go
```diff
@@ -80,7 +80,7 @@ func (updateLabels) reconcile(ctx context.Context, r *FoundationDBClusterReconci
 		}
 
 		pvc, ok := pvcMap[processGroup.ProcessGroupID]
-		if !ok || pod == nil {
```
johscheuer, line 83: That `pod == nil` statement was probably a relict of some refactoring.

controllers/update_status.go
```diff
@@ -70,7 +70,7 @@ func (updateStatus) reconcile(ctx context.Context, r *FoundationDBClusterReconci
 			},
 		}
 	} else {
-		connectionString, err := tryConnectionOptions(cluster, r)
+		connectionString, err := tryConnectionOptions(logger, cluster, r)
```
johscheuer, line 73: We should have an issue to revisit the logger handling.

controllers/update_labels.go
sbodagala: So "pod" can never be "nil" if "ok == true", right?

sbodagala (approved):

controllers/update_labels.go
johscheuer: Pod can be nil but that doesn't matter since the Pod struct is never used in the statements below. The `pod == nil` was a leftover from some previous refactoring.

Issue #384, opened 2020-11-17
Remove usage of maps


johscheuer:
Like stated in this comment: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/380#discussion_r525134130 we should remove `maps` where possible n the API.

brownleej:
I went through the CRD YAML files looking for occurrences of "additionalProperties", which seems to signal the use of a map type in the CRD. Most of the cases I found were either deprecated or coming from core Kubernetes types like the pod spec. Here are the cases I found that are active and coming from our types:

* cluster.spec.processes
* cluster.spec.sidecarVersions

Both of these could potentially be converted to lists of named sub-objects.

johscheuer:
The `sidecarVersions` is already deprecated and only `cluster.spec.processes` is left. I think the best way to represent this and being conformant with the Kubernetes API guidelines is something like this:

```
type FoundationDBClusterSpec struct {
    ...
    ProcessesConfigs []ProcessesConfig
    ...
}

type ProcessesConfig struct {
    Class ProcessClass
    settings ProcessSettings
}
```

I can try to squeeze that into my list of work items this week.

johscheuer:
The change to the `ProcessConfigs` might be a good choice for the conversion: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1047

PR #1464, opened 2022-12-27
Convert process setting map to a slice

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/384

## Type of change

- Other

## Discussion

We can remove the process setting in the next CRD release (v1beta3).

## Testing

Unit tests (e2e tests will be running).

## Documentation

-

## Follow-up

-


internal/deprecations.go
```diff
@@ -119,17 +141,39 @@ func ensureImageConfigPresent(imageConfigs *[]fdbv1beta2.ImageConfig, expected f
...
+		spec.ProcessesConfigs[settingIdx] = generalConfig
+	} else {
+		spec.ProcessesConfigs = append(spec.ProcessesConfigs, generalConfig)
+	}
+
+	//generalSettings := spec.Processes[fdbv1beta2.ProcessClassGeneral]
+	//if generalSettings.PodTemplate == nil {
+	//	generalSettings.PodTemplate = &corev1.PodTemplateSpec{}
+	//}
+	//spec.Processes[fdbv1beta2.ProcessClassGeneral] = generalSettings
```
johscheuer, line 176: This code must be removed.

api/v1beta2/foundationdbcluster_types.go
```diff
@@ -219,6 +223,15 @@ type FoundationDBClusterSpec struct {
 	UseUnifiedImage *bool `json:"useUnifiedImage,omitempty"`
 }
 
+// ProcessesConfig represents the configuration for a processes. This will be used as a template for all processes
+// of with the specified process class.
+type ProcessesConfig struct {
+	// Class the process class that is affected by this configuration.
+	Class ProcessClass `json:"class,omitempty"`
```
brownleej, line 230: I think this is going to introduce a new string field in the CRD with no max-length. Maybe we should take this opportunity to put a max-length annotation on the ProcessClass type.

internal/replacements/replacements_test.go
```diff
@@ -501,14 +527,20 @@ var _ = Describe("replace_misconfigured_pods", func() {
...
+						setting.PodTemplate.Spec.Containers = append(setting.PodTemplate.Spec.Containers,
 							corev1.Container{
 								Resources: corev1.ResourceRequirements{
 									Requests: corev1.ResourceList{
 										corev1.ResourceCPU: newCPU,
 									},
 								},
 							})
+
+						cluster.UpdateBareProcessSettings(fdbv1beta2.ProcessClassGeneral, setting)
```
brownleej, line 543: If we're repeating this fetch-and-update pattern multiple times, could we can encapsulate it in a function that takes a callback?

brownleej (changes requested):

api/v1beta2/foundationdbcluster_types.go
johscheuer: Good point 👍

johscheuer (commented):
Pausing this effort here since the CRD size is exploding.

Issue #1468, opened 2023-01-10
Upgrades from 7.1 to 7.2 are not working


johscheuer:
### What happened?

When trying to do an upgrade from a 7.1 cluster to a 7.2 cluster the operator will be stuck in the `checkClientCompatibility` step since all fdbserver processes are now reported as clients and those processes won't have the new protocol version available.

A workaround for this is to set `IgnoreUpgradabilityChecks` to `true` until the behaviour is fixed.

### What did you expect to happen?

I would expect that the operator is able to upgrade the cluster without any issue and ignores fdbserver processes (since those will be upgraded by the operator itself).

### How can we reproduce it (as minimally and precisely as possible)?

Create a cluster with 7.1 and upgrade it to 7.2.0 (to any other 7.2 version).

### Anything else we need to know?

We didn't observe this issue earlier since all patch upgrades for 7.1 are version compatible.

One heuristic that we could use is to filter all clients that use the same IP address as a fdbserver process. For a Kubernetes deployment this could be good enough with the risk that we are not catching clients with the wrong client library that are colocated in the same Pod.

### FDB Kubernetes operator

<details>

```console
$ kubectl fdb version
latest
```

</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Cloud provider

<details>

</details>

johscheuer:
This will be fixed by: https://github.com/apple/foundationdb/pull/9225 we should make sure to document that upgrades to 7.2 are only support from the according 7.1 release (probably 7.1.26+).

johscheuer:
Upgrades from 7.1.26 -> 7.2 should be working fine: https://github.com/apple/foundationdb/pull/9252

PR #1469, opened 2023-01-10
Ignore fdbserver processes in check compatibility step

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1468

## Type of change

## Discussion

Like discussed in the issue, the risk with this approach is that we are not catching incompatible processes that are running in the same Pod as a fdbserver process. I think the risk is minimal and in general we are not recommending to run any client workload colocated in the same Pod.

## Testing

Unit tests + e2e tests for upgrades from 7.1.25 -> 7.2.0

## Documentation

We could document the limitation and that we are not recommending to run any client workload in the same Pods.

## Follow-up

-


sbodagala:
> We could document the limitation and that we are not recommending to run any client workload in the same Pods.

Maybe we need to understand why fdbserver processes are now getting reported as clients before we modify the operator to work with it (considering the limitation that it will impose)? (I see that nobody replied to your question on Slack, probably we should post this in the other broader channel.)

johscheuer (commented):
Asked on the FDB Slack channel why fdbserver processes are added as clients and if there's a better way to filter them.

internal/foundationdb_status.go
```diff
@@ -79,3 +81,50 @@ func GetMinimumUptimeAndAddressMap(cluster *fdbv1beta2.FoundationDBCluster, stat
...
+	}
+
+	var unsupportedClients []string
+	for _, versionInfo := range status.Cluster.Clients.SupportedVersions {
+		if versionInfo.ProtocolVersion == "Unknown" {
+			continue
+		}
+
+		if versionInfo.ProtocolVersion != protocolVersion {
+			for _, client := range versionInfo.MaxProtocolClients {
```
brownleej, line 114: My understanding is that `max_protocol_clients` is a sample of the clients that have this version as their max supported protocol. Does that mean it's possible that all of the clients in the sample will be fdbserver processes, but we will have normal clients that are not included in the sample, leading to false negatives from this check?

brownleej (changes requested):

internal/foundationdb_status.go
brownleej: I checked a 7.1 cluster with a large number of clients, and confirmed that the number of items in `max_protocol_clients` is smaller than the `max_protocol_count`, indicating that some of the clients are omitted.

johscheuer: We discussed this offline and that's an expected behaviour. The list only contains a sample for the clients. This means we can't proceed with this PR as we might miss some real clients.

johscheuer (commented):
I'm moving this PR into draft to signalling that we need some more information before proceeding.

johscheuer:
Closed in favour of: https://github.com/apple/foundationdb/pull/9225

PR #1470, opened 2023-01-11
Allow to configure how many processes must be ready for the upgrade bounce

johscheuer:
# Description

Before this change we expected that 100% of the processes are ready for the bounce, in large environment this requirement can be a blocker for upgrades. This change allows to specify either an absolute value or a percentage

## Type of change

## Discussion

This will relax the requirement when upgrading a cluster. The assumption is that we can tolerate 10% of processes not being ready (and maybe being partitioned). The operator will recreate those Pods in the rolling bounce phase with the new image (which will ensure they run the new version even if they were missed during the bounce).

If a user doesn't want to use this feature they can disable it by setting the value to `100%`.

## Testing

Local testing + e2e.

## Documentation

-

## Follow-up

-


controllers/bounce_processes.go
```diff
@@ -192,22 +192,14 @@ func getProcessesReadyForRestart(logger logr.Logger, cluster *fdbv1beta2.Foundat
 		return nil, &requeue{message: "Waiting for config map to sync to all pods", delayedRequeue: true}
 	}
 
-	counts, err := cluster.GetProcessCountsWithDefaults()
-	if err != nil {
-		return nil, &requeue{
-			curError: err,
+	// If we upgrade the cluster wait until the minimum number of required processes are ready for the restart.
```
sbodagala, line 195: Can't this cause the cluster to become unavailable (what if the processes that get left out are in a state like SidecarUnreachable then those processes will keep running at the older version and if their count exceeds the replication factor then reads won't succeed)?

sbodagala:
> The operator will recreate those Pods in the rolling bounce phase with the new image 

Is this guaranteed in all cases?

johscheuer (commented):
I discussed this with @brownleej offline and we will take a different approach to ensure we are not blocking because of a partitioned process.

Issue #1438, opened 2022-11-21
Change crashLoop setup to allow to select container


johscheuer:
### What would you like to be added/changed?

The operator supports to add buggify to a running cluster e.g. by using `crashLoop`. We should adjust this setting (by creating a new setting and deprecate thee old one) to allow to select the container the struct could look like this:

```yaml
buggify:
  crashLoopContainer:
  - container: foundationdb
     targets:
     - storage-1
     - storage-2
  - container: foundationdb-kubernetes-sidecar
     targets:
     - storage-3
```

This allows to inject the crash loop behaviour into different containers.

PR #1472, opened 2023-01-13
Change crashLoop setup to allow to select container

vishesh:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1438


vishesh:
@johscheuer : Not sure what It happening, but I'm getting this error for build. Seems like there is a codegen to convert Go types to whatever kube yaml or whatever it understands., and it doesn't know how to convert this particular type? Which doesn't make sense as map is made of primitive values (string) here, its just nested.

```
/root/go/bin/controller-gen "crd:maxDescLen=0,crdVersions=v1,generateEmbeddedObjectMeta=true" rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
/root/src/fdb-kubernetes-operator/api/v1beta1/foundationdbcluster_types.go:1994:31: not a supported map value type: *ast.ParenExpr
/root/src/fdb-kubernetes-operator/api/v1beta2/foundationdbcluster_types.go:1691:31: not a supported map value type: *ast.ParenExpr
/root/src/fdb-kubernetes-operator/api/v1beta2/foundationdbcluster_types.go:1687:2: encountered struct field "NonTLS" without JSON tag in type "RequiredAddressSet"
Error: not all generators ran successfully
run `controller-gen crd:maxDescLen=0,crdVersions=v1,generateEmbeddedObjectMeta=true rbac:roleName=manager-role webhook paths=./... output:crd:artifacts:config=config/crd/bases -w` to see all available markers, or `controller-gen crd:maxDescLen=0,crdVersions=v1,generateEmbeddedObjectMeta=true rbac:roleName=manager-role webhook paths=./... output:crd:artifacts:config=config/crd/bases -h` for usage
```

api/v1beta1/foundationdbcluster_types.go
```diff
@@ -1991,14 +1991,16 @@ type RequiredAddressSet struct {
...
+
 // BuggifyConfig provides options for injecting faults into a cluster for testing.
 type BuggifyConfig struct {
 	// NoSchedule defines a list of process group IDs that should fail to schedule.
 	NoSchedule []string `json:"noSchedule,omitempty"`
 
 	// CrashLoops defines a list of process group IDs that should be put into a
 	// crash looping state.
-	CrashLoop []string `json:"crashLoop,omitempty"`
+	CrashLoop CrashLoopList `json:"crashLoop,omitempty"`
```
johscheuer, line 2003: All changes to v1beta1 can be revert. We don't change those files and they are only there until we remove the v1beta1 support from our CRD.

api/v1beta2/foundationdbcluster_types.go
```diff
@@ -1677,24 +1677,27 @@ type RoutingConfig struct {
...
-// RequiredAddressSet provides settings for which addresses we need to listen
+// Requiredaddressset provides settings for which addresses we need to listen
 // on.
 type RequiredAddressSet struct {
 	// TLS defines whether we need to listen on a TLS address.
 	TLS bool `json:"tls,omitempty"`
 
 	// NonTLS defines whether we need to listen on a non-TLS address.
-	NonTLS bool `json:"nonTLS,omitempty"`
+	NonTLS bool `jsogn:"nonTLS,omitempty"`
```
johscheuer, line 1687:
```suggestion
	NonTLS bool `json:"nonTLS,omitempty"`
```

api/v1beta2/foundationdbcluster_types.go
```diff
@@ -1677,24 +1677,27 @@ type RoutingConfig struct {
...
+
 // BuggifyConfig provides options for injecting faults into a cluster for testing.
 type BuggifyConfig struct {
 	// NoSchedule defines a list of process group IDs that should fail to schedule.
 	NoSchedule []string `json:"noSchedule,omitempty"`
 
 	// CrashLoops defines a list of process group IDs that should be put into a
 	// crash looping state.
-	CrashLoop []string `json:"crashLoop,omitempty"`
+	CrashLoop CrashLoopList `json:"crashLoop,omitempty"`
```
johscheuer, line 1700: We have to add a new CrashLoop list for adding the new feature otherwise we break our existing API. Internally we can convert the old format into the new format and define that only one of the two list are allowed to be set.

api/v1beta2/foundationdbcluster_types.go
```diff
@@ -1677,24 +1677,27 @@ type RoutingConfig struct {
...
 	// TLS defines whether we need to listen on a TLS address.
 	TLS bool `json:"tls,omitempty"`
 
 	// NonTLS defines whether we need to listen on a non-TLS address.
-	NonTLS bool `json:"nonTLS,omitempty"`
+	NonTLS bool `jsogn:"nonTLS,omitempty"`
 }
 
+// ProcessGroup -> ContainerName
+type CrashLoopList map[string](map[string]None)
```
johscheuer, line 1691: We shouldn't use maps here and the format should look like specified in the issue: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1438:

```yaml
buggify:
  crashLoopContainer:
  - container: foundationdb
     targets:
     - storage-1
     - storage-2
  - container: foundationdb-kubernetes-sidecar
     targets:
     - storage-3
```

Some background why maps shouldn't be used in Kubernetes structs: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps

api/v1beta2/foundationdbcluster_types.go
```diff
@@ -5,7 +5,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+	http://www.apache.org/licenses/LICENSE-2.0
```
johscheuer, line 8: Is there a reason for this change?

api/v1beta2/foundationdbcluster_types.go
```diff
@@ -1696,6 +1705,10 @@ type BuggifyConfig struct {
 	// crash looping state.
 	CrashLoop []string `json:"crashLoop,omitempty"`
 
+	// CrashLoopContainers defines a list of process group IDs and containers
+	// that should be put into a crash looping state.
+	CrashLoopContainers []CrashLoopContainerObject `json:"crashLoopContainers,omitempty"`
```
johscheuer, line 1710: Can we add a Deprecation notice on the `CrashLoop`?

api/v1beta2/foundationdbcluster_types_test.go
```diff
@@ -5,7 +5,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+	http://www.apache.org/licenses/LICENSE-2.0
```
johscheuer, line 8: Is that change required?

docs/cluster_spec.md
```diff
@@ -136,6 +138,17 @@ CoordinatorSelectionSetting defines the process class and the priority of it. A
 
 [Back to TOC](#table-of-contents)
 
+## CrashLoopContainerObject
+
+
```
johscheuer, line 143: That seems a little bit off. You can generate the docs with `make clean all`

johscheuer (commented):
You can use the [NormalizeClusterSpec](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/internal/deprecations.go#L45) method to convert the existing `CrashLoop` into the new format. That makes the internal handling easier and we only have to work with the CrashLoopContainers. You can put the conversion method here: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/internal/deprecations.go#L59 (between the custom parameter validation and the if statement).

api/v1beta2/foundationdbcluster_types_test.go
vishesh: It's done by go fmt.

api/v1beta2/foundationdbcluster_types.go
vishesh: It's done by go fmt or something automatic.

vishesh:
Are we using existing `CrashLoop` somewhere? I couldn't find any use of `CrashLoop` in our tests repository. If we are not using it, don't think we need to support conversions, and can just remove `CrashLoop` at some point.

api/v1beta2/foundationdbcluster_types.go
johscheuer: Could this be your emacs configuration, it seems like the spaces are replaced with one tab. If possible we should revert the change.

internal/pod_models.go
```diff
@@ -431,6 +445,7 @@ func GetPodSpec(cluster *fdbv1beta2.FoundationDBCluster, processClass fdbv1beta2
 			return nil, err
 		}
 	} else {
+		//TODO (Vishesh): Add crash-loop here too. What is unified image?
```
johscheuer, line 448: Is that todo still open? If so, can we create a radar to follow up on that?

internal/pod_models_test.go
```diff
@@ -940,6 +963,76 @@ var _ = Describe("pod_models", func() {
...
+				Expect(mainContainer.Args).To(Equal([]string{
+					"fdbmonitor --conffile /var/dynamic-conf/fdbmonitor.conf" +
+						" --lockfile /var/dynamic-conf/fdbmonitor.lockfile" +
+						" --loggroup operator-test-1" +
+						" >> /var/log/fdb-trace-logs/fdbmonitor-$(date '+%Y-%m-%d').log 2>&1",
+				}))
+			})
+		})
+
+		Context("with a different process group crash looping in container", func() {
```
johscheuer, line 990: This test is the same as the test above `with an instance that is crash looping container` correct?

internal/pod_models_test.go
```diff
@@ -940,6 +963,76 @@ var _ = Describe("pod_models", func() {
...
+				Expect(mainContainer.Name).To(Equal(fdbv1beta2.MainContainerName))
+				Expect(mainContainer.Args).To(Equal([]string{"crash-loop"}))
+
+				sidecarContainer := spec.Containers[1]
+				Expect(sidecarContainer.Name).To(Equal(fdbv1beta2.SidecarContainerName))
+				Expect(sidecarContainer.Args).ToNot(Equal([]string{"crash-loop"}))
+			})
+		})
+
+		Context("with a different process group crash looping in container", func() {
```
johscheuer, line 1013:
```suggestion
		Context("with a process group that defines crash looping for the sidecar container", func() {
```

internal/pod_models_test.go
```diff
@@ -940,6 +963,76 @@ var _ = Describe("pod_models", func() {
...
+					{
+						ContainerName: fdbv1beta2.SidecarContainerName,
+						Targets:       []string{"storage-1"},
+					},
+				}
+				spec, err = GetPodSpec(cluster, fdbv1beta2.ProcessClassStorage, 1)
+				Expect(err).NotTo(HaveOccurred())
+			})
+
+			It("should have the normal start command", func() {
```
johscheuer, line 1025:
```suggestion
			It("should set the correct start commands", func() {
```

Separating the It command into two statements would make the failure more obvious e.g. if the issue is in the sidecar setup or the main container setup, but I'm fine with leaving it as it is.

internal/replacements/replace_failed_process_groups.go
```diff
@@ -73,6 +75,12 @@ func ReplaceFailedProcessGroups(log logr.Logger, cluster *fdbv1beta2.FoundationD
 			continue
 		}
 
+		for _, targets := range cluster.GetCrashLoopContainerProcessGroups() {
```
johscheuer, line 78: Should we move the `cluster.GetCrashLoopContainerProcessGroups()` statement out of the for loop to calculate the map only once instead of multiple times? In the current implementation we calculate the same map for every iteration.

johscheuer (changes requested):

internal/pod_models_test.go
vishesh: More or less. One is checking Command and another is checking Args field. Just followed the pattern of existing crash loop test.

internal/replacements/replace_failed_process_groups.go
```diff
@@ -62,6 +62,9 @@ func ReplaceFailedProcessGroups(log logr.Logger, cluster *fdbv1beta2.FoundationD
 
 	maxReplacements := getMaxReplacements(cluster, cluster.GetMaxConcurrentAutomaticReplacements())
 	hasReplacement := false
+	containerProcessGroups := cluster.GetCrashLoopContainerProcessGroups()
```
johscheuer, line 65:
```suggestion
	crashLoopContainerProcessGroups := cluster.GetCrashLoopContainerProcessGroups()
```

johscheuer (approved):
The failure is unrelated to the changes and we have an issue to investigate that failure.I'm fine with merging this PR as it is and continue working on the feature in https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1484.

Thanks!

PR #1478, opened 2023-01-19
Disable local FDB client per default

johscheuer:
# Description

Disable the local FDB client which should not be used, since we make use of the multi version client.

## Type of change

## Discussion

We can put that setting behind a flag to let people enable/disable it.

## Testing

Unit tests + e2e tests will be run.

## Documentation

We can add a section about that in the manual.

## Follow-up

-

### DIFF
--- a/setup/setup.go
+++ b/setup/setup.go
@@ -29,6 +29,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/apple/foundationdb/bindings/go/src/fdb"
+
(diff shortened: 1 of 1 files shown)


johscheuer (commented):
Seems like the transaction is not working anymore.

```text
..."controllers.updateDatabaseConfiguration","requeueAfter":0,"error":"FoundationDB error code 2000 (Invalid API call)"...
```

PR #1484, opened 2023-01-23
Migrate old CrashLoop to new Spec

vishesh:
Follow up for https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1472

### DIFF
--- a/controllers/delete_pods_for_buggification.go
+++ b/controllers/delete_pods_for_buggification.go
@@ -77,6 +77,10 @@ func (d deletePodsForBuggification) reconcile(ctx context.Context, r *Foundation
 
 		shouldCrashLoop := false
 		for _, targets := range crashLoopContainerProcessGroups {
+			_, shouldCrashLoop = targets["*"]
+			if shouldCrashLoop {
+				break
+			}
 			_, shouldCrashLoop = targets[processGroup.ProcessGroupID]
 			if shouldCrashLoop {
 				break
--- a/controllers/delete_pods_for_buggification.go
+++ b/controllers/delete_pods_for_buggification.go
@@ -44,8 +44,7 @@ func (d deletePodsForBuggification) reconcile(ctx context.Context, r *Foundation
 	}
 
 	podMap := internal.CreatePodMap(cluster, pods)
-	crashLoopPods, crashLoopAll := cluster.GetCrashLoopProcessGroups()
-	crashLoopContainerPods := cluster.GetCrashLoopContainerProcessGroups()
+	crashLoopContainerProcessGroups := cluster.GetCrashLoopContainerProcessGroups()
 
 	noSchedulePods := make(map[string]fdbv1beta2.None, len(cluster.Spec.Buggify.NoSchedule))
 	for _, processGroupID := range cluster.Spec.Buggify.NoSchedule {
@@ -76,17 +75,11 @@ func (d deletePodsForBuggification) reconcile(ctx context.Context, r *Foundation
 			}
 		}
 
-		shouldCrashLoop := crashLoopAll
-		if !shouldCrashLoop {
-			_, shouldCrashLoop = crashLoopPods[processGroup.ProcessGroupID]
-		}
-
-		if !shouldCrashLoop {
-			for _, targets := range crashLoopContainerPods {
-				_, shouldCrashLoop = targets[processGroup.ProcessGroupID]
-				if shouldCrashLoop {
-					break
-				}
+		shouldCrashLoop := false
+		for _, targets := range crashLoopContainerProcessGroups {
+			_, shouldCrashLoop = targets[processGroup.ProcessGroupID]
+			if shouldCrashLoop {
+				break
 			}
 		}
 
--- a/internal/deprecations.go
+++ b/internal/deprecations.go
@@ -104,12 +104,6 @@ func NormalizeClusterSpec(cluster *fdbv1beta2.FoundationDBCluster, options Depre
 		}
 
 		for _, pid := range cluster.Spec.Buggify.CrashLoop {
-			if pid == "*" {
-				for _, proc := range cluster.Status.ProcessGroups {
-					crashLoopContainers[fdbv1beta2.MainContainerName][proc.ProcessGroupID] = fdbv1beta2.None{}
-				}
-			}
-
 			crashLoopContainers[fdbv1beta2.MainContainerName][pid] = fdbv1beta2.None{}
 		}
 
--- a/internal/deprecations.go
+++ b/internal/deprecations.go
@@ -97,6 +97,39 @@ func NormalizeClusterSpec(cluster *fdbv1beta2.FoundationDBCluster, options Depre
 		updateImageConfigs(&cluster.Spec, cluster.GetUseUnifiedImage())
 	}
 
+	if len(cluster.Spec.Buggify.CrashLoop) > 0 {
+		crashLoopContainers := cluster.GetCrashLoopContainerProcessGroups()
+		if _, ok := crashLoopContainers[fdbv1beta2.MainContainerName]; !ok {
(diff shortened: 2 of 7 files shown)


internal/deprecations.go
```diff
@@ -97,6 +97,42 @@ func NormalizeClusterSpec(cluster *fdbv1beta2.FoundationDBCluster, options Depre
 		updateImageConfigs(&cluster.Spec, cluster.GetUseUnifiedImage())
 	}
 
+	if cluster.Spec.Buggify.CrashLoop != nil && len(cluster.Spec.Buggify.CrashLoop) > 0 {
```
johscheuer, line 100:
```suggestion
	if len(cluster.Spec.Buggify.CrashLoop) > 0 {
```

The len check will automatically return 0 for a nil slice: https://go.dev/play/p/THQM6xELEet

internal/deprecations.go
```diff
@@ -97,6 +97,42 @@ func NormalizeClusterSpec(cluster *fdbv1beta2.FoundationDBCluster, options Depre
...
+		crashLoopContainers := cluster.GetCrashLoopContainerProcessGroups()
+		if _, ok := crashLoopContainers[fdbv1beta2.MainContainerName]; !ok {
+			crashLoopContainers[fdbv1beta2.MainContainerName] = make(map[string]fdbv1beta2.None)
+			crashLoopContainers[fdbv1beta2.SidecarContainerName] = make(map[string]fdbv1beta2.None)
```
johscheuer, line 104:
```suggestion
			crashLoopContainers[fdbv1beta2.MainContainerName] = make(map[string]fdbv1beta2.None)
```

We don't have to add the sidecar container here.

internal/deprecations.go
```diff
@@ -97,6 +97,42 @@ func NormalizeClusterSpec(cluster *fdbv1beta2.FoundationDBCluster, options Depre
...
+		}
+
+		for _, pid := range cluster.Spec.Buggify.CrashLoop {
+			if pid == "*" {
+				for _, proc := range cluster.Status.ProcessGroups {
+					crashLoopContainers[fdbv1beta2.MainContainerName][proc.ProcessGroupID] = fdbv1beta2.None{}
+					crashLoopContainers[fdbv1beta2.SidecarContainerName][proc.ProcessGroupID] = fdbv1beta2.None{}
```
johscheuer, line 111:
```suggestion
					crashLoopContainers[fdbv1beta2.MainContainerName][proc.ProcessGroupID] = fdbv1beta2.None{}
```

The old behaviour was to only put the main container into crash loop and we should keep it like that for compatibility.

internal/deprecations.go
```diff
@@ -97,6 +97,42 @@ func NormalizeClusterSpec(cluster *fdbv1beta2.FoundationDBCluster, options Depre
...
+				}
+			}
+
+			crashLoopContainers[fdbv1beta2.MainContainerName][pid] = fdbv1beta2.None{}
+			crashLoopContainers[fdbv1beta2.SidecarContainerName][pid] = fdbv1beta2.None{}
```
johscheuer, line 116:
```suggestion
			crashLoopContainers[fdbv1beta2.MainContainerName][pid] = fdbv1beta2.None{}
```

johscheuer (changes requested):
How did you test the behaviour? Could you please add some tests to ensure the conversion is working as expected.

controllers/delete_pods_for_buggification.go
```diff
@@ -45,6 +45,8 @@ func (d deletePodsForBuggification) reconcile(ctx context.Context, r *Foundation
 
 	podMap := internal.CreatePodMap(cluster, pods)
 	crashLoopPods, crashLoopAll := cluster.GetCrashLoopProcessGroups()
+	crashLoopContainerPods := cluster.GetCrashLoopContainerProcessGroups()
```
johscheuer, line 48:
```suggestion
	crashLoopContainerProcesGroups := cluster.GetCrashLoopContainerProcessGroups()
```

internal/replacements/replace_failed_process_groups.go
```diff
@@ -62,6 +62,9 @@ func ReplaceFailedProcessGroups(log logr.Logger, cluster *fdbv1beta2.FoundationD
 
 	maxReplacements := getMaxReplacements(cluster, cluster.GetMaxConcurrentAutomaticReplacements())
 	hasReplacement := false
+	containerProcessGroups := cluster.GetCrashLoopContainerProcessGroups()
```
johscheuer, line 65:
```suggestion
	crashLoopContainerProcesGroups := cluster.GetCrashLoopContainerProcessGroups()
```

internal/replacements/replace_failed_process_groups.go
```diff
@@ -73,6 +76,12 @@ func ReplaceFailedProcessGroups(log logr.Logger, cluster *fdbv1beta2.FoundationD
 			continue
 		}
 
+		for _, targets := range containerProcessGroups {
```
johscheuer, line 79: We can remove the if statement above since the normalise method will take care of the conversion.

internal/pod_models.go
```diff
@@ -443,8 +457,16 @@ func GetPodSpec(cluster *fdbv1beta2.FoundationDBCluster, processClass fdbv1beta2
 				args = "crash-loop"
 			}
 		}
-		mainContainer.Args = []string{args}
 
+		for _, crashObjs := range cluster.Spec.Buggify.CrashLoopContainers {
```
johscheuer, line 461: We can remove the for loop above that is iterating over the deprecated feature.

internal/pod_models.go
```diff
@@ -250,6 +243,27 @@ func configureContainersForUnifiedImages(cluster *fdbv1beta2.FoundationDBCluster
 		corev1.VolumeMount{Name: "fdb-trace-logs", MountPath: "/var/log/fdb-trace-logs"},
 	)
 
+	for _, crashLoopInstanceID := range cluster.Spec.Buggify.CrashLoop {
+		if processGroupID == crashLoopInstanceID || crashLoopInstanceID == "*" {
+			mainContainer.Command = []string{"crash-loop"}
+			mainContainer.Args = []string{"crash-loop"}
+		}
```
johscheuer, line 250: This can be removed.

internal/deprecations.go
```diff
@@ -97,6 +97,39 @@ func NormalizeClusterSpec(cluster *fdbv1beta2.FoundationDBCluster, options Depre
 		updateImageConfigs(&cluster.Spec, cluster.GetUseUnifiedImage())
 	}
 
+	if len(cluster.Spec.Buggify.CrashLoop) > 0 {
```
johscheuer, line 100: Can we make sure we have unit tests for this conversion to ensure it behaves as expected?

johscheuer (changes requested):

internal/deprecations.go
vishesh: I think, the existing CrashLoop tests should pass if this is implemented correctly which should verify if the behavior is preserved.

johscheuer: I'm okay with that personally I would prefer to have a dedicated that ensures the conversions works in all cases e.g.:

- CrashLoop is empty and CrashLoopContainers is empty
- CrashLoop is not empty and CrashLoopContainers is empty
- CrashLoop is not empty and CrashLoopContainers is not empty

I believe the existing tests only cover `CrashLoop is not empty and CrashLoopContainers is empty`.

johscheuer (approved):

internal/deprecations.go
vishesh: Existing CrashLoop tests will cover "CrashLoop empty/not-empty and CrashLoopContainer empty", as the behavior expected by those tests are fulfilled by the conversions. The only combination left is CrashLoopContainer empty.

PR #1491, opened 2023-01-27
Improve the way how the operator handles partitial upgrades

johscheuer:
# Description

This change makes sure we are not setting any FDB environment variables for the `fdbcli` command to prevent the use of the multi version client for 7.1+ fdbclis. In addition we added more fallback logic to the `GetStatus` method to transparently try the new (desired) version if the runningVersion is not working.

## Type of change

## Discussion

This adds some more complexity around the upgrade handling but the multi version client is not reliable in cases where not all coordinators are upgraded.

## Testing

I will see if I can add some additional unit tests and I will do a test with out new upgrade test.

## Documentation

Will be covered in the other upgrade doc issue.

## Follow-up

-

### DIFF
--- a/fdbclient/admin_client.go
+++ b/fdbclient/admin_client.go
@@ -238,7 +238,6 @@ func (client *cliAdminClient) runCommand(command cliCommand) (string, error) {
 
 		// If we hit a timeout report it as a timeout error
 		if strings.Contains(string(output), "Specified timeout reached") {
-			client.log.Info("runner has timeout issue")
 			// See: https://apple.github.io/foundationdb/api-error-codes.html
 			// 1031: Operation aborted because the transaction timed out
 			return "", &fdbv1beta2.TimeoutError{Err: err}
@@ -310,8 +309,6 @@ func (client *cliAdminClient) getStatusFromCli() (*fdbv1beta2.FoundationDBStatus
 		return nil, err
 	}
 
-	client.log.V(1).Info("Fetched status JSON with fdbcli", "version", client.Cluster.GetRunningVersion(), "status", status)
-
 	return status, nil
 }
 
@@ -325,7 +322,7 @@ func (client *cliAdminClient) getStatus() (*fdbv1beta2.FoundationDBStatus, error
 
 	// If the status doesn't contain any processes and we are doing an upgrade, we probably use the wrong fdbcli version
 	// and we have to fallback to the on specified in out spec.version.
-	if len(status.Cluster.Processes) == 0 && client.Cluster.IsBeingUpgraded() {
+	if (status == nil || len(status.Cluster.Processes) == 0) && client.Cluster.IsBeingUpgraded() {
 		// Create a copy of the cluster and make use of the desired version instead of the last observed running version.
 		clusterCopy := client.Cluster.DeepCopy()
 		clusterCopy.Status.RunningVersion = clusterCopy.Spec.Version
--- a/fdbclient/admin_client.go
+++ b/fdbclient/admin_client.go
@@ -291,24 +291,14 @@ func (client *cliAdminClient) runCommandWithBackoff(command string) (string, err
 	return rawResult, err
 }
 
-func (client *cliAdminClient) getStatus(useClientLibrary bool) (*fdbv1beta2.FoundationDBStatus, error) {
-	var contents []byte
-	var err error
-
-	if useClientLibrary {
-		// This will call directly the database and fetch the status information
-		// from the system key space.
-		contents, err = getStatusFromDB(client.Cluster, client.log)
-	} else {
-		var rawResult string
-		rawResult, err = client.runCommandWithBackoff("status json")
-		if err != nil {
-			return nil, err
-		}
-
-		contents, err = internal.RemoveWarningsInJSON(rawResult)
+// getStatusFromCli uses the fdbcli to connect to the FDB cluster
+func (client *cliAdminClient) getStatusFromCli() (*fdbv1beta2.FoundationDBStatus, error) {
+	output, err := client.runCommandWithBackoff("status json")
+	if err != nil {
+		return nil, err
 	}
 
+	contents, err := internal.RemoveWarningsInJSON(output)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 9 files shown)


johscheuer (commented):
I found some other areas that needs some improvement for all the edge cases.

johscheuer:
Ran the test suite with the new test that keeps one coordinator at the old version:

fdbclient/admin_client.go
```diff
@@ -318,7 +309,30 @@ func (client *cliAdminClient) getStatus(useClientLibrary bool) (*fdbv1beta2.Foun
...
+// the new version for fdbcli will be tried.
+func (client *cliAdminClient) getStatus() (*fdbv1beta2.FoundationDBStatus, error) {
+	status, err := client.getStatusFromCli()
+	if err != nil {
+		return nil, err
+	}
+
+	// If the status doesn't contain any processes and we are doing an upgrade, we probably use the wrong fdbcli version
+	// and we have to fallback to the on specified in out spec.version.
+	if len(status.Cluster.Processes) == 0 && client.Cluster.IsBeingUpgraded() {
```
brownleej, line 328: It looks like this has less logging in this scenario than the existing code, which worries me because of how that logging has been helpful in recent debugging. Can we add that logging back? We might even want to expand it to include the invalid status we got from the client library to give more context.

fdbclient/command_runner.go
```diff
@@ -37,28 +40,61 @@ type realCommandRunner struct {
 	log logr.Logger
 }
 
+// getEnvironmentVariablesWithoutBlacklisted returns the current environment variables for the new process with some
```
brownleej, line 43:
```suggestion
// getEnvironmentVariablesWithoutExcludedFdbEnv returns the current environment variables for the new process with some
```

fdbclient/command_runner.go
```diff
@@ -37,28 +40,61 @@ type realCommandRunner struct {
...
+// FDB specific variables filtered out to ensure we don't set any variables that could change the behaviour of fdbcli or
+// the other fdb tools.
+func getEnvironmentVariablesWithoutExcludedFdbEnv() []string {
+	blackListedEnvironmentVariables := map[string]fdbv1beta2.None{
```
brownleej, line 47: I think we generally try to avoid the term "black list" for these kind of use cases. Can we go with something like "blocked" or "excluded" instead?

brownleej (changes requested):

fdbclient/admin_client.go
johscheuer: I don't think that this scenario has less logging than before. We still report the `"Fetched status JSON with fdbcli"` log line including the used version. In addition we still log the command that is executed including the full binary path and the timeout parameter and I haven't removed any log statements. What logging statement exactly are you missing/looking for?

fdbclient/command_runner.go
johscheuer: Yes, I adjusted the method name but forgot about the variable name 🤦

fdbclient/admin_client.go
johscheuer: I added some additional logging.

fdbclient/admin_client.go
```diff
@@ -238,9 +238,10 @@ func (client *cliAdminClient) runCommand(command cliCommand) (string, error) {
 
 		// If we hit a timeout report it as a timeout error
 		if strings.Contains(string(output), "Specified timeout reached") {
+			client.log.Info("runner has timeout issue")
```
johscheuer, line 241: We can drop this log statement again, I only used it for debugging.

fdbclient/admin_client.go
```diff
@@ -318,7 +309,30 @@ func (client *cliAdminClient) getStatus(useClientLibrary bool) (*fdbv1beta2.Foun
 	if err != nil {
 		return nil, err
 	}
-	client.log.V(1).Info("Fetched status JSON", "status", status)
+
+	client.log.V(1).Info("Fetched status JSON with fdbcli", "version", client.Cluster.GetRunningVersion(), "status", status)
```
johscheuer, line 313:
```suggestion
	client.log.V(1).Info("Fetched status JSON with fdbcli", "version", client.Cluster.GetRunningVersion())
```

We already log the full status at the return of the GetStatus method.

fdbclient/admin_client.go
johscheuer, line 328:
```suggestion
	if (status == nil || len(status.Cluster.Processes) == 0) && client.Cluster.IsBeingUpgraded() {
```

fdbclient/common.go
```diff
@@ -142,8 +143,19 @@ func getConnectionStringFromDB(cluster *fdbv1beta2.FoundationDBCluster, log logr
...
+		return nil, err
+	}
+
+	status := &fdbv1beta2.FoundationDBStatus{}
+	err = json.Unmarshal(contents, status)
+	if err != nil {
+		return nil, err
+	}
+
+	return status, err
```
johscheuer, line 158:
```suggestion
	return status, nil
```

brownleej (approved):
👍

sbodagala:
> This change makes sure we are not setting any FDB environment variables for the fdbcli command to prevent the use of the multi version client for 7.1+ fdbclis.

Why are we doing this?

fdbclient/admin_client.go
```diff
@@ -318,7 +308,30 @@ func (client *cliAdminClient) getStatus(useClientLibrary bool) (*fdbv1beta2.Foun
...
+
+	return status, nil
+}
+
+// getStatus uses fdbcli to connect to the FDB cluster, if the cluster is upgraded and the initial version returns no processes
```
sbodagala, line 315: You don't want to handle the partial upgrade scenario (where a subset of processes are still/stuck at the older version for whatever reason) here?

It appears you are trying to implement (a version of) the multiversion client logic here, probably because of a limitation of that logic? Maybe you can explain that limitation and how you are addressing that here (so we can follow what you are trying to do)? Thanks!

johscheuer: > You don't want to handle the partial upgrade scenario (where a subset of processes are still/stuck at the older version for whatever reason) here?

What scenario do you have in mind? As far as I understand the logic in FDB the majority of coordinators will define the version of the cluster e.g. a cluster can only be at one version (and have a few other processes running at a different version but not reporting to the cluster). We observed a few cases where the kill was not restarting one coordinator (for different reasons) and that resulted in the multi version client behaving badly.

> It appears you are trying to implement (a version of) the multiversion client logic here, probably because of a limitation of that logic? Maybe you can explain that limitation and how you are addressing that here (so we can follow what you are trying to do)? Thanks!

We are implementing it because the multi version logic is currently broken (I'm not sure if that is limited to 6.3 and is fixed with 7.1), there are some comments about this in the FDB Slack. Basically by using the fdbcli binary without multi version configuration we force the operator to use a specific version to connect to the DB.

johscheuer (commented):
>> This change makes sure we are not setting any FDB environment variables for the fdbcli command to prevent the use of the multi version client for 7.1+ fdbclis.

> Why are we doing this?

Otherwise `fdbcli` will use the multi-version client with the same result as if we use the multi version client from the operator (go bindings). I think the multi version support for fdbcli was added in 7.1.

fdbclient/admin_client.go
```diff
@@ -291,24 +291,14 @@ func (client *cliAdminClient) runCommandWithBackoff(command string) (string, err
 	return rawResult, err
 }
 
-func (client *cliAdminClient) getStatus(useClientLibrary bool) (*fdbv1beta2.FoundationDBStatus, error) {
-	var contents []byte
-	var err error
-
-	if useClientLibrary {
```
sbodagala, line 298: So "useClientLibrary" is not used in the modified version, right?

fdbclient/command_runner.go
```diff
@@ -37,9 +40,32 @@ type realCommandRunner struct {
 	log logr.Logger
 }
 
+// getEnvironmentVariablesWithoutBlacklisted returns the current environment variables for the new process with some
+// FDB specific variables filtered out to ensure we don't set any variables that could change the behaviour of fdbcli or
+// the other fdb tools.
+func getEnvironmentVariablesWithoutExcludedFdbEnv() []string {
```
sbodagala, line 46: Do you want to exclude these variables only when running "status" command? You are doing it for all commands below.

PR #1493, opened 2023-01-31
extending the cordon command for all clusters

09harsh:
# Description

Currently the `cordon` command in plugin requires a cluster name to remove all the pods that are running of a specific node on a cluster. This PR is to extend that behaviour to remove all the pods that are running on that node irrespective of the cluster name. This requires us to use a new flag `use-custom-label` which enables this new feature and then we also need to provide a `custom-label`. The default value of `custom-label` is `fdb-cluster-name` is nothing was given. 

## Type of change

## Discussion

No

## Testing

Manual testing?*

Added unit test

## Documentation

Yes

## Follow-up

### DIFF
--- a/api/v1beta2/foundationdbcluster_types.go
+++ b/api/v1beta2/foundationdbcluster_types.go
@@ -2275,18 +2275,3 @@ func (cluster *FoundationDBCluster) Validate() error {
 
 	return fmt.Errorf(strings.Join(validations, ", "))
 }
-
-// ContainsPod checks if the given Pod is part of the cluster or not.
-func (cluster *FoundationDBCluster) ContainsPod(pod corev1.Pod) bool {
-	clusterMatchingLabels := cluster.GetMatchLabels()
-	podLabels := pod.GetLabels()
-	if len(clusterMatchingLabels) > len(podLabels) {
-		return false
-	}
-	for clusterLabelKey, clusterLabelValue := range clusterMatchingLabels {
-		if podLabelValue, found := podLabels[clusterLabelKey]; !found || clusterLabelValue != podLabelValue {
-			return false
-		}
-	}
-	return true
-}
--- a/api/v1beta2/foundationdbcluster_types.go
+++ b/api/v1beta2/foundationdbcluster_types.go
@@ -2275,3 +2275,17 @@ func (cluster *FoundationDBCluster) Validate() error {
 
 	return fmt.Errorf(strings.Join(validations, ", "))
 }
+
+func (cluster *FoundationDBCluster) ContainsPod(pod corev1.Pod) bool {
+	clusterMatchingLabels := cluster.GetMatchLabels()
+	podLabels := pod.GetLabels()
+	if len(clusterMatchingLabels) > len(podLabels) {
+		return false
+	}
+	for clusterLabelKey, clusterLabelValue := range clusterMatchingLabels {
+		if podLabelValue, found := podLabels[clusterLabelKey]; !found || clusterLabelValue != podLabelValue {
+			return false
+		}
+	}
+	return true
+}
--- a/api/v1beta2/foundationdbcluster_types.go
+++ b/api/v1beta2/foundationdbcluster_types.go
@@ -2276,6 +2276,7 @@ func (cluster *FoundationDBCluster) Validate() error {
 	return fmt.Errorf(strings.Join(validations, ", "))
 }
 
+// ContainsPod checks if the given Pod is part of the cluster or not.
 func (cluster *FoundationDBCluster) ContainsPod(pod corev1.Pod) bool {
 	clusterMatchingLabels := cluster.GetMatchLabels()
 	podLabels := pod.GetLabels()
--- a/internal/pod_models.go
+++ b/internal/pod_models.go
@@ -1039,3 +1039,18 @@ func GetObjectMetadata(cluster *fdbv1beta2.FoundationDBCluster, base *metav1.Obj
 func GetPodDNSName(cluster *fdbv1beta2.FoundationDBCluster, podName string) string {
 	return fmt.Sprintf("%s.%s.%s.svc.%s", podName, cluster.Name, cluster.Namespace, cluster.GetDNSDomain())
 }
+
+// ContainsPod checks if the given Pod is part of the cluster or not.
+func ContainsPod(cluster *fdbv1beta2.FoundationDBCluster, pod corev1.Pod) bool {
+	clusterMatchingLabels := cluster.GetMatchLabels()
+	podLabels := pod.GetLabels()
+	if len(clusterMatchingLabels) > len(podLabels) {
+		return false
+	}
+	for clusterLabelKey, clusterLabelValue := range clusterMatchingLabels {
+		if podLabelValue, found := podLabels[clusterLabelKey]; !found || clusterLabelValue != podLabelValue {
+			return false
+		}
+	}
+	return true
+}
--- a/kubectl-fdb/cmd/cordon.go
+++ b/kubectl-fdb/cmd/cordon.go
@@ -22,11 +22,9 @@ package cmd
 
 import (
 	"fmt"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
-	"strings"
-
-	corev1 "k8s.io/api/core/v1"
 
 	"github.com/spf13/cobra"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -63,7 +61,7 @@ func newCordonCmd(streams genericclioptions.IOStreams) *cobra.Command {
 			if err != nil {
 				return err
 			}
-			customLabel, err := cmd.Flags().GetString("custom-label")
+			customLabels, err := cmd.Flags().GetStringArray("custom-labels")
 			if err != nil {
 				return err
 			}
@@ -88,10 +86,10 @@ func newCordonCmd(streams genericclioptions.IOStreams) *cobra.Command {
 					return err
 				}
 
-				return cordonNode(kubeClient, clusterName, nodes, namespace, withExclusion, wait, sleep, customLabel)
+				return cordonNode(kubeClient, clusterName, nodes, namespace, withExclusion, wait, sleep, customLabels)
 			}
 
-			return cordonNode(kubeClient, clusterName, args, namespace, withExclusion, wait, sleep, customLabel)
+			return cordonNode(kubeClient, clusterName, args, namespace, withExclusion, wait, sleep, customLabels)
 		},
 		Example: `
 # Evacuate all process groups for a cluster in the current namespace that are hosted on node-1
@@ -123,13 +121,13 @@ kubectl fdb cordon --node-selector machine=a,disk=fast
 	cmd.Flags().StringP("fdb-cluster", "c", "", "evacuate process group(s) from the provided cluster.")
 	cmd.Flags().StringToStringVarP(&nodeSelectors, "node-selector", "", nil, "node-selector to select all nodes that should be cordoned. Can't be used with specific nodes.")
 	cmd.Flags().BoolP("exclusion", "e", true, "define if the process groups should be removed with exclusion.")
-	cmd.Flags().StringP("custom-label", "l", "fdb-cluster-name", "space separated custom label to extract appropriate pods")
+	cmd.Flags().StringArrayP("custom-labels", "l", []string{"fdb-cluster-name"}, "space separated custom label to extract appropriate pods")
 	o.configFlags.AddFlags(cmd.Flags())
 
 	return cmd
 }
 
-func getClusterNames(kubeClient client.Client, inputClusterName string, namespace string, node string, customLabel string) ([]string, error) {
+func getClusterNames(kubeClient client.Client, inputClusterName string, namespace string, node string, customLabels []string) ([]string, error) {
 	if len(inputClusterName) != 0 {
 		// Cluster name already given.
 		return []string{inputClusterName}, nil
@@ -137,12 +135,12 @@ func getClusterNames(kubeClient client.Client, inputClusterName string, namespac
 	var pods corev1.PodList
 	err := kubeClient.List(ctx.Background(), &pods,
 		client.InNamespace(namespace),
-		client.HasLabels(strings.Split(customLabel, " ")),
+		client.HasLabels(customLabels),
 		client.MatchingFieldsSelector{
 			Selector: fields.OneTermEqualSelector("spec.nodeName", node),
 		})
 	if err != nil {
-		return nil, fmt.Errorf("error fetching pods with custom labels %v", strings.Split(customLabel, " "))
+		return nil, fmt.Errorf("error fetching pods with custom labels %v", customLabels)
 	}
 	clusterNames := make([]string, 0, len(pods.Items))
 	for _, pod := range pods.Items {
@@ -157,7 +155,7 @@ func getClusterNames(kubeClient client.Client, inputClusterName string, namespac
 }
 
 // cordonNode gets all process groups of this cluster that run on the given nodes and add them to the remove list
-func cordonNode(kubeClient client.Client, inputClusterName string, nodes []string, namespace string, withExclusion bool, wait bool, sleep uint16, customLabel string) error {
+func cordonNode(kubeClient client.Client, inputClusterName string, nodes []string, namespace string, withExclusion bool, wait bool, sleep uint16, customLabels []string) error {
 	fmt.Printf("Start to cordon %d nodes\n", len(nodes))
 	if len(nodes) == 0 {
 		return nil
@@ -165,7 +163,7 @@ func cordonNode(kubeClient client.Client, inputClusterName string, nodes []strin
 
 	operationFailed := false
 	for _, node := range nodes {
-		clusterNames, err := getClusterNames(kubeClient, inputClusterName, namespace, node, customLabel)
+		clusterNames, err := getClusterNames(kubeClient, inputClusterName, namespace, node, customLabels)
 		if err != nil {
 			return fmt.Errorf("unable to fetch cluster names")
 		}
--- a/kubectl-fdb/cmd/cordon.go
+++ b/kubectl-fdb/cmd/cordon.go
@@ -146,7 +146,7 @@ func cordonNode(cmd *cobra.Command, kubeClient client.Client, inputClusterName s
 		}
 
 		clusterNames := getClusterNames(cmd, inputClusterName, pods, clusterLabel)
-		for _, clusterName := range clusterNames {
+		for clusterName := range clusterNames {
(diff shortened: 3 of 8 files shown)


kubectl-fdb/cmd/cordon.go
```diff
@@ -67,6 +66,14 @@ func newCordonCmd(streams genericclioptions.IOStreams) *cobra.Command {
 			if err != nil {
 				return err
 			}
+			useCustomLabel, err := cmd.Flags().GetBool("use-custom-label")
+			if err != nil {
+				return err
+			}
+			customLabel, err := cmd.Flags().GetString("custom-label")
```
johscheuer, line 73:
```suggestion
			customLabel, err := cmd.Flags().GetString("cluster-label")
```

Why should we name it `custom-label` and not `cluster-label`?

kubectl-fdb/cmd/cordon.go
```diff
@@ -67,6 +66,14 @@ func newCordonCmd(streams genericclioptions.IOStreams) *cobra.Command {
 			if err != nil {
 				return err
 			}
+			useCustomLabel, err := cmd.Flags().GetBool("use-custom-label")
```
johscheuer, line 69: Is this additional flag really required?

kubectl-fdb/cmd/cordon.go
```diff
@@ -116,18 +139,15 @@ kubectl fdb cordon -c cluster --node-selector machine=a,disk=fast
 	cmd.Flags().StringP("fdb-cluster", "c", "", "evacuate process group(s) from the provided cluster.")
 	cmd.Flags().StringToStringVarP(&nodeSelectors, "node-selector", "", nil, "node-selector to select all nodes that should be cordoned. Can't be used with specific nodes.")
 	cmd.Flags().BoolP("exclusion", "e", true, "define if the process groups should be removed with exclusion.")
-	err := cmd.MarkFlagRequired("fdb-cluster")
-	if err != nil {
-		log.Fatal(err)
-	}
-
+	cmd.Flags().Bool("use-custom-label", false, "define if the process groups should be removed using label instead of cluster name.")
+	cmd.Flags().StringP("custom-label", "l", "fdb-cluster-name", "evacuate process group(s) from the provided cluster.")
```
johscheuer, line 143:
```suggestion
	cmd.Flags().StringP("custom-label", "l", fdbv1beta2. FDBClusterLabel, "evacuate process group(s) from the provided cluster.")
```

We should use the default label here if we define a default value for the flag. If we want to allow multiple custom labels we should make use of `StringArray` instead of a String and doing some custom parsing.

kubectl-fdb/cmd/cordon.go
```diff
@@ -137,13 +157,22 @@ func cordonNode(kubeClient client.Client, cluster *fdbv1beta2.FoundationDBCluste
...
-			client.MatchingLabels(cluster.GetMatchLabels()),
-			client.MatchingFieldsSelector{
-				Selector: fields.OneTermEqualSelector("spec.nodeName", node),
-			})
-
+		var err error
+		if useCustomLabel {
+			err = kubeClient.List(ctx.Background(), &pods,
+				client.InNamespace(namespace),
+				client.HasLabels(strings.Split(customLabel, " ")),
```
johscheuer, line 164: I don't see how the cluster is dynamically fetched here.

johscheuer (changes requested):
Have you tried running this command with the new changes? I believe it won't work. Currently it's still required to pass the cluster otherwise the command will fail. In addition we never dynamically fetch the cluster to add the process groups fro replacement. The current implementation would add all process groups to the one cluster that is selected with `-c`.

kubectl-fdb/cmd/cordon_test.go
```diff
@@ -100,41 +101,80 @@ var _ = Describe("[plugin] cordon command", func() {
...
+					customLabel:    fdbv1beta2.FDBClusterLabel,
+				}),
+			Entry("Cordon no nodes with exclusion with custom label",
+				testCase{
+					nodes:                     []string{""},
+					WithExclusion:             true,
+					ExpectedInstancesToRemove: []string{},
+					ExpectedInstancesToRemoveWithoutExclusion: []string{},
+					useCustomLabel: true,
+					customLabel:    fdbv1beta2.FDBClusterLabel,
```
johscheuer, line 177: Can we add some test cases where we remove multiple Pods from different clusters? The current test case is not enough to test the behaviour with multiple Pods from different clusters running on the same node.

johscheuer (changes requested):

kubectl-fdb/cmd/cordon.go
09harsh: The idea is to allow other labels as well as per user requirement. 
Eg:
```
 kubectl fdb cordon -l "fdb-cluster-name some-other-label" node-1
 ```

File: kubectl-fdb/cmd/suite_test.go [hunk 6]
```diff
@@ -51,3 +41,17 @@ var _ = JustBeforeEach(func() {
 var _ = AfterEach(func() {
 	k8sClient.Clear()
 })
+
+func createCluster(givenName string, givenNamespace string) *fdbv1beta2.FoundationDBCluster {
```
johscheuer, line 45:
```suggestion
func createCluster(name string, namespace string) *fdbv1beta2.FoundationDBCluster {
```

Do you think the prefix "given" adds additional value to this parameters?

The name `createCluster` is a bit misleading as we are not really creating a cluster, we only generate the cluster struct (e.g. the `createPods` actually "creates" those Pods in the Kubernetes instance).

kubectl-fdb/cmd/cordon_test.go
```diff
@@ -32,110 +32,190 @@ import (
...
+					ExpectedInstancesToRemove: []string{},
+					ExpectedInstancesToRemoveWithoutExclusion: []string{},
+					clusterName:  "",
+					clusterLabel: fdbv1beta2.FDBClusterLabel,
 				}),
 		)
 	})
 })
+
+func createPods(inputClusterName string, inputNamespace string, id int) error {
```
johscheuer, line 182:
```suggestion
func createPods(clusterName string, namespace string, id int) error {
```

I think we can drop the prefix input here.

kubectl-fdb/cmd/cordon_test.go
```diff
@@ -32,110 +32,190 @@ import (
...
+	pods := []corev1.Pod{
+		{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      fmt.Sprintf("instance-%d", id),
```
johscheuer, line 186: Is there a reason we have to use the ID? It's probably better to prefix the Pod name with the cluster name, that way it's pretty obvious which Pods are associated with which cluster.

kubectl-fdb/cmd/cordon_test.go
```diff
@@ -32,110 +32,190 @@ import (
...
+				err := cordonNode(k8sClient, input.clusterName, input.nodes, namespace, input.WithExclusion, false, 0, input.clusterLabel)
 				Expect(err).NotTo(HaveOccurred())
 
-				var resCluster fdbv1beta2.FoundationDBCluster
-				err = k8sClient.Get(context.Background(), client.ObjectKey{
-					Namespace: namespace,
-					Name:      clusterName,
-				}, &resCluster)
+				var clusterNames []string
+				if len(input.clusterName) == 0 {
```
johscheuer, line 65:
```suggestion
				if input.clusterName == "" {
```

What is the reason for this additional check? Shouldn't we always check both clusters to ensure the code is not accidentally adding replacements to the second cluster even if it shouldn't?

kubectl-fdb/cmd/cordon_test.go
```diff
@@ -32,110 +32,190 @@ import (
...
+					clusterNames = []string{clusterName, secondClusterName}
+				} else {
+					clusterNames = []string{input.clusterName}
+				}
+
+				var currentInstancesToRemove []string
```
johscheuer, line 71:
```suggestion
				var instancesToRemove []string
```

kubectl-fdb/cmd/cordon_test.go
```diff
@@ -32,110 +32,190 @@ import (
...
+				var currentInstancesToRemoveWithoutExclusion []string
```
johscheuer, line 72:
```suggestion
				var instancesToRemoveWithoutExclusion []string
```

kubectl-fdb/cmd/cordon_test.go
```diff
@@ -32,110 +32,190 @@ import (
...
+				for _, currentClusterName := range clusterNames {
```
johscheuer, line 73:
```suggestion
				for _, clusterName := range clusterNames {
```

kubectl-fdb/cmd/cordon_test.go
```diff
@@ -32,110 +32,190 @@ import (
...
+					clusterLabel: "",
 				}),
 			Entry("Cordon all nodes with exclusion",
 				testCase{
 					nodes:                     []string{"node-1", "node-2"},
 					WithExclusion:             true,
 					ExpectedInstancesToRemove: []string{"instance-1", "instance-2"},
 					ExpectedInstancesToRemoveWithoutExclusion: []string{},
+					clusterLabel: fdbv1beta2.FDBClusterLabel,
+					clusterName:  clusterName,
```
johscheuer, line 131:
```suggestion
					clusterName:  clusterName,
					clusterLabel: fdbv1beta2.FDBClusterLabel,
```

johscheuer (commented):
Have you tested those changes against a real cluster?

kubectl-fdb/cmd/cordon.go
```diff
@@ -116,54 +122,111 @@ kubectl fdb cordon -c cluster --node-selector machine=a,disk=fast
...
+						cmd.Printf("could not fetch process group ID from Pod: %s\n", pod.Name)
+						continue
+					}
+					processGroups = append(processGroups, processGroup)
+				}
+			}
+			err = replaceProcessGroups(kubeClient, cluster.Name, processGroups, namespace, withExclusion, wait, false, true, sleep)
+			if err != nil {
+				operationFailed = true
+				errors = append(errors, fmt.Sprintf("unable to cordon all Pods for cluster %s\n", cluster.Name))
```
johscheuer, line 179: Any reason not to print the error here directly?

kubectl-fdb/cmd/cordon.go
```diff
@@ -116,54 +122,111 @@ kubectl fdb cordon -c cluster --node-selector machine=a,disk=fast
...
 			}
-			processGroups = append(processGroups, processGroup)
 		}
 	}
+	if operationFailed {
+		return fmt.Errorf("following operation failed, please check and retry: \n, %s", errors)
+	}
```
johscheuer, line 185: If we already keep a list of errors, what is the benefit of having `operationFailed`? A check for `len(errors) > 0` should provide the same information.

kubectl-fdb/cmd/cordon.go
```diff
@@ -116,54 +122,111 @@ kubectl fdb cordon -c cluster --node-selector machine=a,disk=fast
...
+	return nil
+}
 
-	return replaceProcessGroups(kubeClient, cluster.Name, processGroups, namespace, withExclusion, wait, false, true, sleep)
+func fetchPods(kubeClient client.Client, inputClusterName string, namespace string, node string, clusterLabel string) (corev1.PodList, error) {
```
johscheuer, line 189:
```suggestion
func fetchPods(kubeClient client.Client, clusterName string, namespace string, node string, clusterLabel string) (corev1.PodList, error) {
```

kubectl-fdb/cmd/cordon.go
```diff
@@ -116,54 +122,111 @@ kubectl fdb cordon -c cluster --node-selector machine=a,disk=fast
...
+		}
+		err = kubeClient.List(ctx.Background(), &pods,
+			client.InNamespace(namespace),
+			client.MatchingLabels(cluster.GetMatchLabels()),
+			client.MatchingFieldsSelector{
+				Selector: fields.OneTermEqualSelector("spec.nodeName", node),
+			})
+	}
+	if err != nil {
+		return pods, fmt.Errorf("unable to fetch pods")
```
johscheuer, line 212: Why are we not returning the actual error?

kubectl-fdb/cmd/cordon.go
```diff
@@ -116,54 +122,111 @@ kubectl fdb cordon -c cluster --node-selector machine=a,disk=fast
...
+	var err error
+	if inputClusterName == "" {
+		err = kubeClient.List(ctx.Background(), &pods,
+			client.InNamespace(namespace),
+			client.HasLabels([]string{clusterLabel}),
+			client.MatchingFieldsSelector{
+				Selector: fields.OneTermEqualSelector("spec.nodeName", node),
+			})
+	} else {
+		cluster, error := loadCluster(kubeClient, namespace, inputClusterName)
```
johscheuer, line 200: Now we are fetching the cluster multiple times (once for every node). This is probably okay for now, but we should create an issue to resolve this.

kubectl-fdb/cmd/cordon.go
```diff
@@ -116,54 +122,111 @@ kubectl fdb cordon -c cluster --node-selector machine=a,disk=fast
...
+func getClusterNames(cmd *cobra.Command, inputClusterName string, pods corev1.PodList, clusterLabel string) map[string]fdbv1beta2.None {
+	if inputClusterName != "" {
+		return map[string]fdbv1beta2.None{inputClusterName: {}}
+	}
+
+	clusterNames := make(map[string]fdbv1beta2.None)
+	for _, pod := range pods.Items {
+		clusterName, ok := pod.Labels[clusterLabel]
+		if !ok {
+			printStatement(cmd, fmt.Sprintf("could not fetch cluster name from Pod: %s\n", pod.Name), errorMessage)
```
johscheuer, line 226: I don't think we want to reuse this method here, this will always add a ❌ in front of the error message.

kubectl-fdb/cmd/cordon_test.go
```diff
@@ -21,121 +21,219 @@
...
+				}
 
 				Expect(err).NotTo(HaveOccurred())
 				// Use equality.Semantic.DeepEqual here since the Equal check of gomega is to strict
-				Expect(equality.Semantic.DeepEqual(input.ExpectedInstancesToRemove, resCluster.Spec.ProcessGroupsToRemove)).To(BeTrue())
-				Expect(equality.Semantic.DeepEqual(input.ExpectedInstancesToRemoveWithoutExclusion, resCluster.Spec.ProcessGroupsToRemoveWithoutExclusion)).To(BeTrue())
+				fmt.Printf("\ninstancesToRemove: %v\n", instancesToRemove)
+				Expect(equality.Semantic.DeepEqual(input.ExpectedInstancesToRemove, instancesToRemove)).To(BeTrue())
+				fmt.Printf("\ninstancesToRemoveWithoutExclusion: %v\n", instancesToRemoveWithoutExclusion)
+				Expect(equality.Semantic.DeepEqual(input.ExpectedInstancesToRemoveWithoutExclusion, instancesToRemoveWithoutExclusion)).To(BeTrue())
```
johscheuer, line 90:
```suggestion
				Expect(input.ExpectedInstancesToRemove).To(ConsistOf(instancesToRemove))
				Expect(input.ExpectedInstancesToRemoveWithoutExclusion).To(ConsistOf(instancesToRemoveWithoutExclusion))
```

See https://pkg.go.dev/github.com/onsi/gomega#ConsistOf, that check was "recently" added in gomega and provides more details on what element is missing.

kubectl-fdb/cmd/cordon_test.go
```diff
@@ -21,121 +21,219 @@
...
+				testCase{
+					nodes:         []string{"node-1", "node-2"},
+					WithExclusion: true,
+					ExpectedInstancesToRemove: []string{
+						fmt.Sprintf("%s-instance-1", clusterName),
+						fmt.Sprintf("%s-instance-2", clusterName),
+					},
+					ExpectedInstancesToRemoveWithoutExclusion: []string{},
+					clusterName:  clusterName,
+					clusterLabel: fdbv1beta2.FDBClusterLabel,
```
johscheuer, line 138: Is there a reason we add the cluster label here too but in other test cases we leave it empty?

kubectl-fdb/cmd/cordon_test.go
```diff
@@ -21,121 +21,219 @@
...
 				testCase{
 					nodes:                     []string{"node-1", "node-2"},
+					WithExclusion:             false,
+					ExpectedInstancesToRemove: []string{},
+					ExpectedInstancesToRemoveWithoutExclusion: []string{
+						fmt.Sprintf("%s-instance-1", clusterName),
+						fmt.Sprintf("%s-instance-2", clusterName),
+					},
+					clusterName:  clusterName,
+					clusterLabel: fdbv1beta2.FDBClusterLabel,
```
johscheuer, line 150: Same as above. We should either add it to all test cases or leave it empty and have a dedicated test case where we pass down the cluster name and the cluster label.

api/v1beta2/foundationdbcluster_types.go
```diff
@@ -2275,3 +2275,18 @@ func (cluster *FoundationDBCluster) Validate() error {
 
 	return fmt.Errorf(strings.Join(validations, ", "))
 }
+
+// ContainsPod checks if the given Pod is part of the cluster or not.
+func (cluster *FoundationDBCluster) ContainsPod(pod corev1.Pod) bool {
```
johscheuer, line 2280: Is that something we want to expose from the cluster struct of should this be rather part of the internal package? What do you think?

johscheuer (commented):
Changes look good so far, a few comments.

kubectl-fdb/cmd/cordon.go
09harsh: mainly just to print all the errors at the end else user have to search the error with other logs. Another thing we can do here is print the error here and all the errors at the end as well.

kubectl-fdb/cmd/cordon.go
09harsh: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1496

johscheuer (approved):
Changes LGTM, could you please rebase the PR. That should make the e2e tests pass.

internal/pod_models.go
```diff
@@ -1051,3 +1051,18 @@ func GetObjectMetadata(cluster *fdbv1beta2.FoundationDBCluster, base *metav1.Obj
 func GetPodDNSName(cluster *fdbv1beta2.FoundationDBCluster, podName string) string {
 	return fmt.Sprintf("%s.%s.%s.svc.%s", podName, cluster.Name, cluster.Namespace, cluster.GetDNSDomain())
 }
+
+// ContainsPod checks if the given Pod is part of the cluster or not.
+func ContainsPod(cluster *fdbv1beta2.FoundationDBCluster, pod corev1.Pod) bool {
```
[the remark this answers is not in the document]
09harsh: done

johscheuer (approved):
Thanks, LGTM 👍

Issue #1492, opened 2023-01-30
Better handling for node failures/taints


johscheuer:
### What would you like to be added/changed?

We should add optional support for detecting node failures and taints in the operator to proactively replace/move Pods away from such nodes.

xumengpanda:
I'm going to use this issue as the meta-issue for all PRs related to taint feature.

xumengpanda:
Feature PR: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1581

xumengpanda:
Documentation PR: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1632

PR #1502, opened 2023-02-14
Initial design for replacing Pods running on tainted nodes

johscheuer:
# Description

Initial work for: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1492

## Type of change

- Documentation

## Discussion

The focus of the doc is to handle nodes that are tainted. We could extend the design to also handle nodes that are in a `NotReady` state but still host Pods.

## Testing

-

## Documentation

Updated the design.

## Follow-up

Implementation.

### DIFF
--- a/docs/design/node_taint_eviction.md
+++ b/docs/design/node_taint_eviction.md
@@ -4,17 +4,20 @@
 
 * Authors: @johscheuer
 * Created: 2023-02-14
-* Updated: 2023-02-14
+* Updated: 2023-03-08
 
 ## Background
 
-The operator is currently able to detect failures of Pods and replace them if required, but there is still a gap between an issue that manifests in issue that are detected by the operator and issues that could potentially happen.
+The operator is currently able to detect failures of Pods and replace them if required, the replacement is happening based on the [conditions that need replacement](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/v1.14.0/api/v1beta2/foundationdbcluster_types.go#L65).
+One of the missing pieces in failure mitigation is to evict Pods from tainted nodes.
 In Kubernetes it's common to use [taints](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration) to indicate potential problems with a node or planned maintenance.
-This information could be consumed by the operator to proactively replace Pods that are running on node with a given set of taints.
+This information could be consumed by the operator to proactively replace Pods that are running on nodes with a given set of taints.
 
 ## General Design Goals
 
-The goal of this design is to implement a mechanism that allows the operator to replace Pods that are running on nodes with a taint.
+The goal of this design is to implement a mechanism that allows the operator to replace Pods that are running on nodes with a given set of taints.
+It's expected that the operator will detect the Pods running on tainted nodes in a timely manner, either by receiving an event triggering the reconciliation loop or when running the periodic reconciliation, which will happen every 10h per default.
+The actual replacement of all Pods running on tainted nodes, depends on the data that must be replicated and the number of affected Pods.
 
 ## Current Implementation
 
@@ -24,8 +27,8 @@ The closest implementation to this is the [Replace Failed ProcessGroups reconcil
 ## Proposed Design
 
 We have to make sure that this feature can be enabled and disabled, e.g. if someone wants to run the operator inside a cluster without node access.
-The `AutomaticReplacementOptions` will be extended to support configuration of replacements for Pods that run on tainted nodes.
-The new field will be called `taints` and will contain a list, default would be empty, of taint configurations.
+The `AutomaticReplacementOptions` will be extended to support the configuration of replacements for Pods that run on tainted nodes.
+The new field will be called `taints` and will contain a list, the default would be an empty list, of taint configurations.
 That configuration will include the taint key, this has to be an exact match or a wildcard `*`, and the duration this taint must be present on the node.
 The `durationInSeconds` allow to specify how sensitive the operator should be to taints.
 If a wildcard is specified as key this will be used as default value, except there is an exact match defined for a taint key.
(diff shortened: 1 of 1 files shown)


docs/design/node_taint_eviction.md
```diff
@@ -0,0 +1,78 @@
...
+  taints:
+  - key: "example.org/maintenance"
+    durationInSeconds: 7200 # Ensure the taint is present for at least 2 hours before replacing Pods on a node with this taint
+  - key: "*" # The wildcard would allow to define a catch all configuration
+    durationInSeconds: 3600 # Ensure the taint is present for at least 1 hour before replacing Pods on a node with this taint
+```
+
+The actual implementation will extend the [ReplaceFailedProcessGroups](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/internal/replacements/replace_failed_process_groups.go) to handle those cases.
+
+```go
```
johscheuer, line 49: The formatting seems to be of in this code block.

docs/design/node_taint_eviction.md
```diff
@@ -0,0 +1,78 @@
...
+	if maxReplacements <= 0 {
+		return hasReplacements
+    }
+
+	pods, err := r.PodLifecycleManager.GetPods(ctx, r, cluster, internal.GetPodListOptions(cluster, "", "")...)
+    if err != nil {
+        return &requeue{curError: err}
+    }
+
+	nodeMap := internal.GetNodeMap(pods) // This will generate a map containing the node name as key and the pods as value.
```
johscheuer, line 64: We either can make use of the `maxConcurrentReplacements` as a simple starting point or we could directly design it in a way that we always "drain" a node. Draining a node is probably more beneficial instead of only replacing a single Pod. We might want to add a `maxConcurrentNodeDrains` which would be a little bit more complex to track.

xumengpanda: A cluster only cares about how many concurrent pods can be put offline w/o losing availability. 
I think both replacement-on-failure and drain-on-taints are the same to cluster's availability.  
So a global `maxConcurrentReplacements` works better, imo.

docs/design/node_taint_eviction.md
```diff
@@ -0,0 +1,78 @@
...
+	for node, pods := range nodeMap {
+		// Check if node has taint for the expected duration
+		// in that case we will replace all Pods of that node.
+    }
+```
+
+The limitation of this approach would be that we only detect Pods running on tainted nodes when a reconciliation is happening (those are not coupled to node events).
```
xumengpanda, line 71: we can use the periodic reconciliation in operator as a starting point? 
Since we don't do anything until a taint has been there for a certain duration (say 2 hours), periodically check should be timely enough.

docs/design/node_taint_eviction.md
```diff
@@ -0,0 +1,78 @@
...
+  maxConcurrentReplacements: 1
+  enableNodeFailureDetection: true
+  taints:
+  - key: "example.org/maintenance"
+    durationInSeconds: 7200 # Ensure the taint is present for at least 2 hours before replacing Pods on a node with this taint
+  - key: "*" # The wildcard would allow to define a catch all configuration
+    durationInSeconds: 3600 # Ensure the taint is present for at least 1 hour before replacing Pods on a node with this taint
+```
+
+The actual implementation will extend the [ReplaceFailedProcessGroups](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/internal/replacements/replace_failed_process_groups.go) to handle those cases.
```
brownleej, line 47: Rather than modifying this reconciler, could we add a process group condition indicating that a pod is on a tainted node, and make use of the existing logic for replacing process groups? That might involve less changes to the code, and allow more re-use.

johscheuer: I was thinking about the same and the drawback of this approach would be that we are not able to replace all Pods on the tainted node "at once". I guess we can try that approach first before adding more complex logic.

docs/design/node_taint_eviction.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Metadata
+
+* Authors: @johscheuer
+* Created: 2023-02-14
+* Updated: 2023-02-14
+
+## Background
+
+The operator is currently able to detect failures of Pods and replace them if required, but there is still a gap between an issue that manifests in issue that are detected by the operator and issues that could potentially happen.
```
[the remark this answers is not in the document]
johscheuer: Could you extend what you mean by Pod is in unhealthy state? A Pod running on a tainted node doesn't necessarily imply the Pod is in an unhealthy state.

docs/design/node_taint_eviction.md
```diff
@@ -0,0 +1,78 @@
...
+In Kubernetes it's common to use [taints](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration) to indicate potential problems with a node or planned maintenance.
+This information could be consumed by the operator to proactively replace Pods that are running on node with a given set of taints.
```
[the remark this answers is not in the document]
johscheuer: That should be `nodes`

docs/design/node_taint_eviction.md
```diff
@@ -0,0 +1,78 @@
...
+
+## General Design Goals
+
+The goal of this design is to implement a mechanism that allows the operator to replace Pods that are running on nodes with a taint.
```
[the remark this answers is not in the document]
johscheuer: I'm not sure how valuable those "performance" goals will be since they depend on a lot of different variables.

docs/design/node_taint_eviction.md
johscheuer: Based on https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1502/files#r1120462131 it's probably the best to use our normal mechanism and not add another setting.

docs/design/node_taint_eviction.md
johscheuer: Yes, I thought I expressed that with this statement. Let me add some more context to make it clear we are targeting for the simple approach first.

docs/design/node_taint_eviction.md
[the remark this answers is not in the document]
xumengpanda: how about changing `in unhealthy state` to `in undesired state`

docs/design/node_taint_eviction.md
johscheuer: Updated the design to reflect, that we use the same replacement mechanism and just add a new condition for Pods running on a tainted node.

docs/design/node_taint_eviction.md
```diff
@@ -0,0 +1,94 @@
...
+  enableNodeFailureDetection: true
+  taints:
+  - key: "example.org/maintenance"
+    durationInSeconds: 7200 # Ensure the taint is present for at least 2 hours before replacing Pods on a node with this taint
+  - key: "*" # The wildcard would allow to define a catch all configuration
+    durationInSeconds: 3600 # Ensure the taint is present for at least 1 hour before replacing Pods on a node with this taint
+```
+
+We will add a new [ProcessGroupConditionType](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/api/v1beta2/foundationdbcluster_types.go#L673), called `RunningOnTaintedNode` to mark process groups that are running on a tainted node.
+In addition we will extend the  [conditions that need replacement](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/v1.14.0/api/v1beta2/foundationdbcluster_types.go#L65) by the newly created `RunningOnTaintedNode` condition.
```
johscheuer, line 51: We discussed this offline and we will actually add 2 status, one indicating that a process group has a Pod running on a tainted node and another one to indicate that a replacement can be triggered.

brownleej (approved):
👍

Issue #1497, opened 2023-02-08
Make use of version compatibility when upgrading patch versions


johscheuer:
### What would you like to be added/changed?

Patch versions in FDB are version compatible and we should make use of that to remove the requirement of staging the new binaries. This will add reliability to patch upgrades and removes some complexity when doing patch upgrades.

Issue #1473, opened 2023-01-13
Write in depth document around upgrades for FDB on Kubernetes


johscheuer:
### What would you like to be added/changed?

Upgrades of FDB on Kubernetes are complex with multiple different potential failure models. We should write down a in depth document covering the whole process, the failure modes and how to recover.

PR #1506, opened 2023-02-17
Remove staging phase for version compatible upgrades

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1497

Before this change we used the same mechanism for all kinds of upgrades. With this change we only use the staging approach (copying the new binary inside a running container and then run the kill command) for version incompatible upgrades.

The basic idea of this PR is to skip the staging phase for version compatible upgrades and directly recreate Pods with the new desired image (based on the deployment strategy). This reduces the work we have to do for version compatible upgrades.

## Type of change

## Discussion

## Testing

Unit tests and running e2e tests (I'm going to paste the results here once they completed).

## Documentation

This will be done as a follow up in https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1473

## Follow-up

https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1473

### DIFF
--- a/controllers/bounce_processes.go
+++ b/controllers/bounce_processes.go
@@ -62,24 +62,7 @@ func (bounceProcesses) reconcile(ctx context.Context, r *FoundationDBClusterReco
 		return &requeue{curError: err}
 	}
 
-	// In the case of version compatible upgrades we have to check if some processes are already running with the new
-	// desired version e.g. because they were restarted by an event outside of the control of the operator.
-	var upgradedProcesses int
-	if cluster.VersionCompatibleUpgradeInProgress() {
-		for _, process := range status.Cluster.Processes {
-			dcID := process.Locality[fdbv1beta2.FDBLocalityDCIDKey]
-			// Ignore processes that are not managed by this operator instance
-			if dcID != cluster.Spec.DataCenter {
-				continue
-			}
-
-			if process.Version == cluster.Spec.Version {
-				upgradedProcesses++
-			}
-		}
-	}
-
-	addresses, req := getProcessesReadyForRestart(logger, cluster, addressMap, upgradedProcesses)
+	addresses, req := getProcessesReadyForRestart(logger, cluster, addressMap)
 	if req != nil {
 		return req
 	}
@@ -117,7 +100,7 @@ func (bounceProcesses) reconcile(ctx context.Context, r *FoundationDBClusterReco
 		return &requeue{curError: err}
 	}
 
-	upgrading := cluster.Status.RunningVersion != cluster.Spec.Version
+	upgrading := cluster.IsBeingUpgradedWithVersionIncompatibleVersion()
 
 	if useLocks && upgrading {
 		processGroupIDs := make([]fdbv1beta2.ProcessGroupID, 0, len(cluster.Status.ProcessGroups))
@@ -175,7 +158,7 @@ func (bounceProcesses) reconcile(ctx context.Context, r *FoundationDBClusterReco
 
 // getProcessesReadyForRestart returns a slice of process addresses that can be restarted. If addresses are missing or not all processes
 // have the latest configuration this method will return a requeue struct with more details.
-func getProcessesReadyForRestart(logger logr.Logger, cluster *fdbv1beta2.FoundationDBCluster, addressMap map[fdbv1beta2.ProcessGroupID][]fdbv1beta2.ProcessAddress, upgradedProcesses int) ([]fdbv1beta2.ProcessAddress, *requeue) {
+func getProcessesReadyForRestart(logger logr.Logger, cluster *fdbv1beta2.FoundationDBCluster, addressMap map[fdbv1beta2.ProcessGroupID][]fdbv1beta2.ProcessAddress) ([]fdbv1beta2.ProcessAddress, *requeue) {
 	addresses := make([]fdbv1beta2.ProcessAddress, 0, len(cluster.Status.ProcessGroups))
 	allSynced := true
 	var missingAddress []fdbv1beta2.ProcessGroupID
@@ -237,7 +220,7 @@ func getProcessesReadyForRestart(logger logr.Logger, cluster *fdbv1beta2.Foundat
 	// block the restart command if a process is missing longer than the specified GetIgnoreMissingProcessesSeconds.
 	// Those checks should ensure we only run the restart command if all processes that have to be restarted and are connected
 	// to cluster are ready to be restarted.
-	if cluster.IsBeingUpgraded() && (counts.Total()-missingProcesses-upgradedProcesses) != len(addresses) {
+	if cluster.IsBeingUpgradedWithVersionIncompatibleVersion() && (counts.Total()-missingProcesses) != len(addresses) {
 		return nil, &requeue{
 			message:        fmt.Sprintf("expected %d processes, got %d processes ready to restart", counts.Total(), len(addresses)),
 			delayedRequeue: true,
--- a/controllers/bounce_processes_test.go
+++ b/controllers/bounce_processes_test.go
@@ -434,34 +434,6 @@ var _ = Describe("bounceProcesses", func() {
 				Expect(pendingUpgrades).NotTo(BeEmpty())
 			})
 		})
-
-		When("one process is already upgraded for a version compatible upgrade", func() {
-			var upgradedProcessGroup *fdbv1beta2.ProcessGroupStatus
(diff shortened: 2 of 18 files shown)


controllers/change_coordinators.go
```diff
@@ -46,13 +46,13 @@ func (c changeCoordinators) reconcile(ctx context.Context, r *FoundationDBCluste
...
 	if err != nil {
-		return &requeue{curError: err}
+		return &requeue{curError: err, delayedRequeue: true}
 	}
 	defer adminClient.Close()
 
 	status, err := adminClient.GetStatus()
 	if err != nil {
-		return &requeue{curError: err}
+		return &requeue{curError: err, delayedRequeue: true}
```
sbodagala, line 55: I don't follow why "delayedRequeue" is getting added here, and its connection to version compatible upgrades.

controllers/bounce_processes.go
```diff
@@ -237,7 +220,7 @@ func getProcessesReadyForRestart(logger logr.Logger, cluster *fdbv1beta2.Foundat
 	// block the restart command if a process is missing longer than the specified GetIgnoreMissingProcessesSeconds.
 	// Those checks should ensure we only run the restart command if all processes that have to be restarted and are connected
 	// to cluster are ready to be restarted.
-	if cluster.IsBeingUpgraded() && (counts.Total()-missingProcesses-upgradedProcesses) != len(addresses) {
+	if cluster.IsBeingUpgradedWithVersionIncompatibleVersion() && (counts.Total()-missingProcesses) != len(addresses) {
```
sbodagala, line 223: Why "upgradeProcesses" is not getting included here? Shouldn't this check/logic be same (as before) for version incompatible upgrades?

controllers/bounce_processes_test.go
```diff
@@ -434,34 +434,6 @@ var _ = Describe("bounceProcesses", func() {
 				Expect(pendingUpgrades).NotTo(BeEmpty())
 			})
 		})
-
-		When("one process is already upgraded for a version compatible upgrade", func() {
```
sbodagala, line 438: Why this test is getting dropped? Instead you should probably modify it to verify that upgrade completes if a process has already been upgraded?

controllers/check_client_compatibility.go
```diff
@@ -38,7 +38,7 @@ type checkClientCompatibility struct{}
 // reconcile runs the reconciler's work.
 func (c checkClientCompatibility) reconcile(_ context.Context, r *FoundationDBClusterReconciler, cluster *fdbv1beta2.FoundationDBCluster) *requeue {
 	logger := log.WithValues("namespace", cluster.Namespace, "cluster", cluster.Name, "reconciler", "checkClientCompatibility")
-	if !cluster.Status.Configured {
+	if !cluster.Status.Configured && !cluster.IsBeingUpgraded() {
```
sbodagala, line 41: Does it mean we should still do the client compatibility check for version compatible upgrades?

controllers/update_sidecar_versions.go
```diff
@@ -41,7 +41,7 @@ type updateSidecarVersions struct{}
 func (updateSidecarVersions) reconcile(ctx context.Context, r *FoundationDBClusterReconciler, cluster *fdbv1beta2.FoundationDBCluster) *requeue {
 	logger := log.WithValues("namespace", cluster.Namespace, "cluster", cluster.Name, "reconciler", "updateSidecarVersions")
 	// We don't need to upgrade the sidecar if no upgrade is in progress, we can skip any further work here.
-	if !cluster.IsBeingUpgraded() {
```
sbodagala, line 44: You might want to update the comment above to "no version incompatible upgrade"?

controllers/bounce_processes.go
johscheuer: For version incompatible upgrades we have no way to differentiate the case of a process being missing and already being upgraded since those processes will be missing from the cluster status (in theory they should appear in the incompatible processes list). So the `upgradedProcesses` was only interesting during a patch upgrade, with the removal of the staging phase the cluster wide restart was removed.

controllers/bounce_processes_test.go
johscheuer: The reason for dropping this test is that we are not restarting processes with the kill command during a patch upgrade. We will do a rolling recreation of all Pods (except the transaction system Pods, those will be replaced to reduce the number of recoveries).

controllers/change_coordinators.go
johscheuer: It's not really related to this change an I can revert it. In general we want to delay the requeue everywhere to make sure a single step is not blocking the operator from making progress and potentially dead lock itself.

controllers/check_client_compatibility.go
johscheuer: A few lines below we already have:

```go
	if version.IsProtocolCompatible(runningVersion) {
		return nil
	}
```

So we will skip the check for version compatible upgrades. The only modification here is to ignore this reconciler is we are not upgrading.

controllers/update_sidecar_versions.go
johscheuer: Good catch!

controllers/bounce_processes.go
sbodagala: So in the case of version incompatible upgrades if there are incompatible processes (processes at the older version) in the cluster then we will still go ahead with the upgrade, which appears to me as a change in the behavior/logic of version incompatible upgrades. Probably we should have a document that talks about the difference cases that the processes could be in (missing vs all processes present, if missing then incompatible or not, if incompatible then at an older version or at the upgrade version, version compatible/incompatible upgrade or not, what if those processes have a special role (like coordinator), etc., and have a description of how the operator is supposed to handle those cases - to have confidence that the operator can correctly handle those scenarios, and also that would make code reviews easier (and we won't need to have design discussions in PRs).

controllers/change_coordinators.go
sbodagala: You can keep the change, that's fine.

controllers/bounce_processes.go
johscheuer: I started the document here: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1511

> So in the case of version incompatible upgrades if there are incompatible processes (processes at the older version) in the cluster then we will still go ahead with the upgrade, which appears to me as a change in the behavior/logic of version incompatible upgrades. 

Can you add some more details on this? I'm not sure I understand that concern based on the code changes. The changes to the code only affected version compatible upgrades.

sbodagala: Thanks for starting the documentation!

> The changes to the code only affected version compatible upgrades.

Not totally correct, right? The change above affects version incompatible upgrades - the check used to be "(counts.Total()-missingProcesses-upgradedProcesses) != len(addresses)", now it is "(counts.Total()-missingProcesses) != len(addresses)", that is, you are not subtracting "upgradedProcesses" now?

> For version incompatible upgrades we have no way to differentiate the case of a process being missing and already being upgraded since those processes will be missing from the cluster status

Do we know for sure this is what "status" will show for upgraded processes? As I said in another place(s), I don't think we have server side testing for status output of a cluster that has processes in different versions. We need to write a kubernetes test for it, and verify what it outputs.

So the change you are making here in the context of version incompatible upgrades may be a good change - hopefully we will have an explanation of how missing/upgraded processes will be handled in the case of version incompatible upgrades.

controllers/check_client_compatibility.go
sbodagala: Ok.

fdbclient/admin_client.go
```diff
@@ -322,7 +322,7 @@ func (client *cliAdminClient) getStatus() (*fdbv1beta2.FoundationDBStatus, error
 
 	// If the status doesn't contain any processes and we are doing an upgrade, we probably use the wrong fdbcli version
 	// and we have to fallback to the on specified in out spec.version.
-	if (status == nil || len(status.Cluster.Processes) == 0) && client.Cluster.IsBeingUpgraded() {
+	if (status == nil || len(status.Cluster.Processes) == 0) && client.Cluster.IsBeingUpgradedWithVersionIncompatibleVersion() {
```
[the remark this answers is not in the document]
sbodagala: Ok.

sbodagala (approved):

johscheuer (commented):
The test failures are unrelated to those changes.

PR #1511, opened 2023-02-21
Initial round of documenting the upgrade process

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1473

## Type of change

- Documentation

## Discussion

-

## Testing

-

## Documentation

Updated.

## Follow-up

-

### DIFF
--- a/docs/manual/upgrades.md
+++ b/docs/manual/upgrades.md
@@ -44,7 +44,7 @@ Once all Pods have the new configuration and binaries present the operator can m
 
 A potentional blocker for moving towards the next Phase can be an issue with a subset of the Pods e.g. if the sidecar is unreachable.
 If the sidecar is unreachable the operator is not able to confirm that the new configuration and binary is in place.
-All Process Groups affected by such an issue can be discovered with the `SidecarUnreachable` condition.
+All Process Groups affected by such an issue can be discovered with the `SidecarUnreachable` condition, until this issue is resolved the upgrade will be blocked.
 The best way forward for this is to replace the affected Pods with the following [kubectl fdb plugin](../../kubectl-fdb/Readme.md) command:
 
 ```bash
@@ -98,12 +98,6 @@ In that case the best way forward is to restart all processes still running on t
 kubectl fdb restart -c cluster --process-condition=IncorrectCommandLine
 ```
 
-or
-
-```bash
-kubectl fdb restart -c cluster --process-condition=MissingProcesses
-```
-
 One special case here is when the majority of the coordinators are restarted and the rest of the processes are still running on the old version.
 In this case the operator is not able to handle the upgrade correctly since the returned cluster status has no information besides the coordinators that are reachable.
 During this time the FDB cluster will be unavailable.
@@ -144,6 +138,7 @@ kubectl fdb restart -c cluster --all-processes
 The operator tries per default to read the [cluster status json](https://apple.github.io/foundationdb/mr-status.html) with the [multi-version](https://apple.github.io/foundationdb/api-general.html#multi-version-client) from the system key-space.
 During an upgrade it can happen that the multi-version client tries to connect to the cluster with the wrong version especially if not all coordinators are upgraded.
 In this case the operator will fallback using `fdbcli` for the different versions (the one that is desired and the current running version).
+This requires that the majority of coordinators are upgraded to the new desired version.
 In addition to that we will retry the command if the `fdbcli` hits a timeout responding.
 The code in the [fdbclient package](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/v1.14.0/fdbclient/admin_client.go#L337-L360) has some additional details on how we handle those cases.
 
--- a/docs/manual/upgrades.md
+++ b/docs/manual/upgrades.md
@@ -32,6 +32,18 @@ In the main container the new binaries will be present in `/var/dynamic-conf/bin
 The `UpdatePodConfig` subreconciler will ensure that all Pods have the new configuration present, this configuration ensures that the `fdbmonitor` will restart processes with the new binary present in the shared volume.
 Once all Pods have the new configuration and binaries present the operator can move to the next phase.
 
+A potentional blocker for moving towards the next Phase can be an issue with a subset of the Pods e.g. if the sidecar is unreachable.
+If the sidecar is unreachable the operator is not able to confirm that the new configuration and binary is in place.
+All Process Groups affected by such an issue can be discovered with the `SidecarUnreachable` condition.
+The best way forward for this is to replace the affected Pods with the following [kubectl fdb plugin](../../kubectl-fdb/Readme.md) command:
+
+```bash
+kubectl fdb remove process-group -c cluster pod-1
+```
+
+If the whole Pod is affected by the networking issue the operator will replace the Pod automatically if automatic replacements are enabled, which is the default.
+In order to replace the Pod the operator will wait for the Process Group to have a [failure condition](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/v1.14.0/api/v1beta2/foundationdbcluster_types.go#L65) for at least the defined [FailureDetectionTimeSeconds](../cluster_spec.md#automaticreplacementoptions), by default those are 7200 seconds (2 hours).
+
 #### Restart Phase
 
 The `BounceProcesses` subreconciler will handle the restart of all `fdbserver` processes.
@@ -41,20 +53,20 @@ After the `kill` command the operator will initiate a new reconciliation loop to
 
 #### Recreation of Pods Phase
 
---> Add info incompatbile clients
-
-
 Technically the upgrade is now done and the cluster is running on the new version but the operator has to do some cleanup.
 The cleanup includes the recreation of all Pods to ensure they use the new FoundationDB image for the desired version and the configuration will be updated to point again to the binary present in the container.
 Depending on the defined [Pod update strategy](../cluster_spec.md#podupdatestrategy) the operator will handle the recreation of the Pods differently.
 In the default setting `ReplaceTransactionSystem` the operator will recreate the storage Pods zone by zone and replaces Pods of the transaction system (log and stateless processes).
 The replacement of the transaction system has the benefit that the recoveries of the transaction system is reduced to potentially one recovery.
+During this phase the operator will also run the `RemoveIncompatibleProcesses` subreconciler, this reconciler will recreate all Pods that host a `fdbserver` process that is in the list of the `incompatible_connections` and is missing the the process list of the [cluster status json](https://apple.github.io/foundationdb/mr-status.html).
+This reconciler should ensure that the cluster is moving faster towards the desired configuration if some processes where not restarted properly.
+
 Once all Pods are updated to the new image the upgrade is done and the cluster status of the FoundationDB cluster resource in Kubernetes should show that the reconciliation is done.
 
 ### Known issues
 
 There are a number of known issues that can occur during an upgrade of FoundationDB running on Kubernetes.
-Most of the issues are only a challenged for version incompatible upgrades and most issues are handled automatically by the operator.
+Most of the issues are only a challenge for version incompatible upgrades and most issues are handled automatically by the operator.
 If an issue cannot be handled by the operator we will outline a way to manually resolve the issue.
 
 #### Process restarted during staging Phase
@@ -123,6 +135,32 @@ During an upgrade it can happen that the multi-version client tries to connect t
 In this case the operator will fallback using `fdbcli` for the different versions (the one that is desired and the current running version).
 In addition to that we will retry the command if the `fdbcli` hits a timeout responding.
 The code in the [fdbclient package](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/v1.14.0/fdbclient/admin_client.go#L337-L360) has some additional details on how we handle those cases.
+
+#### Processes running on a different version
+
+We discussed in [Process restarted during staging Phase](#process-restarted-during-staging-phase) and [Processes are not restarted during the restart Phase](#processes-are-not-restarted-during-the-restart-phase) cases where a process could be running in a different version than expected.
+For version incompatible upgrades the operator has no way to differentiate between processes not reporting to the cluster and processes running in the incompatible version, both will be missing from the processes list.
+The only difference is that the incompatible processes might be reported in the `incompatible_connections` field.
+
+### How we test upgrades in the operator
+
+For every PR against the operator repository we run a set of tests in addition to that we are running those tests on a regular basis with different versions to ensure we test major, minor and patch upgrades.
+
+Those tests include:
+
+- Upgrades without any failure injection
+- Upgrading a cluster with a partitioned Pod
+- Upgrading a cluster with a partitioned Pod which eventually gets replaced
+- Upgrading a cluster with a random Pod deleted during rolling bounce phase
+- Upgrading a cluster where one coordinator gets restarted during the staging phase
+- Upgrading a cluster with a crash looping sidecar process
+- Upgrading a cluster and one coordinator is not restarted
+- Upgrading a cluster and multiple processes are not restarted
+- Upgrading a cluster with link that drops some packets
+- Upgrading a cluster and no coordinator is restarted
+
+In addition we have some more tests that will run against a cluster in multi-region configuration.
+Those tests are extended if some new edge cases are discovered.
 
 ## Next
 
--- a/docs/manual/upgrades.md
+++ b/docs/manual/upgrades.md
@@ -1,0 +1,129 @@
+# FoundationDB Upgrades
+
+## Overview
+
+The operator supports to upgrade FDB clusters by adjusting the `version` setting in the `FoundationDBCluster` spec.
+It also offers limited support for downgrades, the limitation is that the operator only supports downgrades to an older patch version of the running FDB version.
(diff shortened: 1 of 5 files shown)


johscheuer:
JFYI @xumengpanda

docs/manual/upgrades.md
```diff
@@ -0,0 +1,129 @@
...
+
+#### Staging Phase
+
+This phase ensures that all Pods are in a state that the operator can restart all `fdbserver` processes and they will be restarting with the new FDB version.
+The first step is the `UpdateSidecarVersions`, this subreconciler updates the image of the `foundationdb-kubernetes-sidecar` container to match the desired FDB version.
+Updating the image of a single container doesn't require the recreation of the Pod and keeps the main container that runs the `fdbserver` process running.
+The sidecar container will copy the new binary inside the main container over a shared volume.
+In the main container the new binaries will be present in `/var/dynamic-conf/bin/$fdb_version`, this version will be used once the `fdbserver` process is restarted.
+The `UpdatePodConfig` subreconciler will ensure that all Pods have the new configuration present, this configuration ensures that the `fdbmonitor` will restart processes with the new binary present in the shared volume.
+Once all Pods have the new configuration and binaries present the operator can move to the next phase.
```
johscheuer, line 33: Add some information if the sidecar is not reachable.

docs/manual/upgrades.md
```diff
@@ -0,0 +1,129 @@
+# FoundationDB Upgrades
+
+## Overview
+
+The operator supports to upgrade FDB clusters by adjusting the `version` setting in the `FoundationDBCluster` spec.
```
sbodagala, line 5: Typo? "to upgrade" -> "upgrade to"?

AnnigeriShambu (approved):

docs/manual/upgrades.md
```diff
@@ -0,0 +1,129 @@
...
+This check will ensure that a cluster is only upgraded if all clients support the new version, for patch upgrades this check is skipped.
+More information about this reconciler can be found in the [technical design](technical_design.md#checkclientcompatibility).
+Clients not supporting the new version will be reported in the logs of the operator with the message `Deferring reconciliation due to unsupported clients` and in addition the operator will emit a Kubernetes event.
+This prevents upgrading a database until all clients have been updated with a compatible client library.
+
+#### Staging Phase
+
+This phase ensures that all Pods are in a state that the operator can restart all `fdbserver` processes and they will be restarting with the new FDB version.
+The first step is the `UpdateSidecarVersions`, this subreconciler updates the image of the `foundationdb-kubernetes-sidecar` container to match the desired FDB version.
+Updating the image of a single container doesn't require the recreation of the Pod and keeps the main container that runs the `fdbserver` process running.
```
sbodagala, line 29: "Updating the image of a single container" - you mean "sidecar container" here?

docs/manual/upgrades.md
```diff
@@ -0,0 +1,129 @@
...
+The sidecar container will copy the new binary inside the main container over a shared volume.
+In the main container the new binaries will be present in `/var/dynamic-conf/bin/$fdb_version`, this version will be used once the `fdbserver` process is restarted.
+The `UpdatePodConfig` subreconciler will ensure that all Pods have the new configuration present, this configuration ensures that the `fdbmonitor` will restart processes with the new binary present in the shared volume.
```
sbodagala, line 32: This subreconciler essentially updates "monitor.conf" to point at "/var/dynamic-conf/bin/$fdb_version" binary? Or is this something deeper than that?

docs/manual/upgrades.md
```diff
@@ -0,0 +1,129 @@
...
+Once all Pods have the new configuration and binaries present the operator can move to the next phase.
+
+#### Restart Phase
+
+The `BounceProcesses` subreconciler will handle the restart of all `fdbserver` processes.
+For FoundationDB clusters that are spanned across multiple Kubernetes clusters the operator will follow a special process which is document in the [technical design](technical_design.md#bounceprocesses).
+The subreconciler will also ensure to wait until all `fdbserver` processes are ready to be restarted to prevent cases where only a subset of processes are restarted.
```
sbodagala, line 51: A process is "ready to be restarted" when it has completed the staging phase (configuration updated and the binary copied), correct?

JoshuaMcManus (changes requested):
couple small comments but otherwise looks good to me.

docs/manual/upgrades.md
johscheuer: Correct and it makes sure that the new binary is present in the shared directory.

docs/manual/upgrades.md
johscheuer: Correct, add one more sentence to this.

docs/manual/upgrades.md
```diff
@@ -0,0 +1,129 @@
...
+After the `kill` command the operator will initiate a new reconciliation loop to detect the new running version, this is handled in the `UpdateState` subreconciler and the version is detected based on the output of the [cluster status json](https://apple.github.io/foundationdb/mr-status.html).
+
+#### Recreation of Pods Phase
+
+--> Add info incompatbile clients
+
+
+Technically the upgrade is now done and the cluster is running on the new version but the operator has to do some cleanup.
+The cleanup includes the recreation of all Pods to ensure they use the new FoundationDB image for the desired version and the configuration will be updated to point again to the binary present in the container.
```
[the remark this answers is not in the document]
johscheuer: That is mostly a limitation of the reconciliation model and how we use it. After the upgrade the operator notices that the desired image is different from the current image and will start recreating those Pods. The interesting/important part is that we are not seeing issues during this phase, we only see issues in the cases prior that and even the `RemoveIncompatibleProcesses` subreconciler is technically only cleaning processes that were missed by the `kill` command.

docs/manual/upgrades.md
```diff
@@ -32,6 +32,18 @@ In the main container the new binaries will be present in `/var/dynamic-conf/bin
 The `UpdatePodConfig` subreconciler will ensure that all Pods have the new configuration present, this configuration ensures that the `fdbmonitor` will restart processes with the new binary present in the shared volume.
 Once all Pods have the new configuration and binaries present the operator can move to the next phase.
 
+A potentional blocker for moving towards the next Phase can be an issue with a subset of the Pods e.g. if the sidecar is unreachable.
+If the sidecar is unreachable the operator is not able to confirm that the new configuration and binary is in place.
+All Process Groups affected by such an issue can be discovered with the `SidecarUnreachable` condition.
```
sbodagala, line 47: If any of the process groups are in "SidecarUnreachable" condition do we still advance to the Restart phase?

docs/manual/upgrades.md
```diff
@@ -32,6 +32,18 @@ In the main container the new binaries will be present in `/var/dynamic-conf/bin
...
+The best way forward for this is to replace the affected Pods with the following [kubectl fdb plugin](../../kubectl-fdb/Readme.md) command:
```
sbodagala, line 48: The pod replacement will cause the containers/server process within the main container to get restarted, correct?

sbodagala: Is there a reason we don't want to do automatic replacement (the way described below) in this case?

docs/manual/upgrades.md
johscheuer: No, this will basically block the upgrade until the issue is resolved. I can add an additional paragraph to make that more precise.

docs/manual/upgrades.md
johscheuer: > The pod replacement will cause the containers/server process within the main container to get restarted, correct?

Correct.

> Is there a reason we don't want to do automatic replacement (the way described below) in this case?

I think it should be fine to replace Pods if they have that condition. I don't recall if we made any specific decision on skipping that decision for a specific reason.

docs/manual/upgrades.md
```diff
@@ -0,0 +1,129 @@
...
+#### Staging Phase
+
+This phase ensures that all Pods are in a state that the operator can restart all `fdbserver` processes and they will be restarting with the new FDB version.
+The first step is the `UpdateSidecarVersions`, this subreconciler updates the image of the `foundationdb-kubernetes-sidecar` container to match the desired FDB version.
+Updating the image of a single container doesn't require the recreation of the Pod and keeps the main container that runs the `fdbserver` process running.
+The sidecar container will copy the new binary inside the main container over a shared volume.
+In the main container the new binaries will be present in `/var/dynamic-conf/bin/$fdb_version`, this version will be used once the `fdbserver` process is restarted.
+The `UpdatePodConfig` subreconciler will ensure that all Pods have the new configuration present, this configuration ensures that the `fdbmonitor` will restart processes with the new binary present in the shared volume.
+Once all Pods have the new configuration and binaries present the operator can move to the next phase.
+
```
sbodagala, line 44: Same question here: we don't advance to the Restart phase if not all pods have the new configuration/binaries present, correct?

docs/manual/upgrades.md
```diff
@@ -0,0 +1,129 @@
...
+In that case the best way forward is to restart all processes still running on the old version, this can either be done manually or by using the [kubectl fdb plugin](../../kubectl-fdb/Readme.md) with the following command:
+
+```bash
+kubectl fdb restart -c cluster --process-condition=IncorrectCommandLine
+```
+
+or
+
+```bash
+kubectl fdb restart -c cluster --process-condition=MissingProcesses
```
sbodagala, line 104: This doesn't help (because the MissingProcesses have already restarted), right?

docs/manual/upgrades.md
sbodagala: Ok, that's what I remember.

docs/manual/upgrades.md
sbodagala: Ok, we should do automatic replacement in this case then (if we are not doing it already). Also, we probably should make FailureDetectionTimeSeconds to be < 7200 seconds in order for the replacement to be effective. A process(es) (that has been staged correctly) can restart and can get the cluster into a bad state in 7200 seconds. Any reason we picked such a high value for FailureDetectionTimeSeconds?

docs/manual/upgrades.md
```diff
@@ -0,0 +1,129 @@
...
+Most of the issues are only a challenged for version incompatible upgrades and most issues are handled automatically by the operator.
+If an issue cannot be handled by the operator we will outline a way to manually resolve the issue.
+
+#### Process restarted during staging Phase
+
+During the staging phase it can happen that a `fdbserver` process gets restarted by an external event like the OOM killer or any issue that results in the `fdbserver` crashing.
+In that case the `fdbmonitor` will restart the `fdbserver` process with the new version and for version incompatible upgrades, that means this process is not able to connect to the cluster.
+The operator will mark this Process Group with the `MissingProcesses` condition, since the process will be missing in the [cluster status json](https://apple.github.io/foundationdb/mr-status.html).
+If the process has this condition longer than the [IgnoreMissingProcessesSeconds](../cluster_spec.md#foundationdbclusterautomationoptions) setting it will be ignored in the [Restart Phase](#restart-phase) to ensure the operator can move forward with the upgrade.
```
sbodagala, line 88: It appears to me this step is not needed (that is, there is no need to wait for IgnoreMissingProcessesSeconds). Because those processes have already been restarted and there is no way to get them to restart at the older version (as they have been staged), and this will only delay restarting the rest of the processes/the upgrade process without giving any additional benefit.

docs/manual/upgrades.md
```diff
@@ -0,0 +1,129 @@
...
+### Known issues
+
+There are a number of known issues that can occur during an upgrade of FoundationDB running on Kubernetes.
+Most of the issues are only a challenged for version incompatible upgrades and most issues are handled automatically by the operator.
+If an issue cannot be handled by the operator we will outline a way to manually resolve the issue.
+
+#### Process restarted during staging Phase
+
+During the staging phase it can happen that a `fdbserver` process gets restarted by an external event like the OOM killer or any issue that results in the `fdbserver` crashing.
+In that case the `fdbmonitor` will restart the `fdbserver` process with the new version and for version incompatible upgrades, that means this process is not able to connect to the cluster.
```
sbodagala, line 86: If that process happens to be a stateless process then this will cause a recovery, correct? I wonder how recovery will work if the restarted process is at a later version? Especially, if this process happens to be a tlog process (as the sequencer has to read state from tlogs in order to proceed with recovery)? And, what if multiple tlogs (equal to the tlog replication factor) restart, can the sequencer even complete recovery?

docs/manual/upgrades.md
```diff
@@ -0,0 +1,129 @@
...
+The operator will mark this Process Group with the `MissingProcesses` condition, since the process will be missing in the [cluster status json](https://apple.github.io/foundationdb/mr-status.html).
+If those processes are reported in the `incompatible_connections` field of the cluster status json the operator will recreate those Pods directly with the new image in the `RemoveIncompatibleProcesses` subreconciler.
+Otherwise the Pod will recreate those Pods in the [recreation of Pods phase](#recreation-of-pods-phase) with the new image.
+During that time the cluster might run with reduced capacity since not all processes are reporting to the cluster.
+
+The operator can handle those cases automatically.
+If you want to speed up this process you can restart the processes with the [kubectl fdb plugin](../../kubectl-fdb/Readme.md):
+
+```bash
+kubectl fdb restart -c cluster --process-condition=MissingProcesses
```
sbodagala, line 125: So these processes are not going to be marked with "IncorrectCommandLine" condition? If not, it could be a problem as the operator/user cannot distinguish between these cases but they would need to specify a different "--process-condition" condition to resolve them.

docs/manual/upgrades.md
```diff
@@ -0,0 +1,129 @@
...
+
+```bash
+kubectl fdb restart -c cluster --all-processes
+```
+
+#### Detect correct version
+
+The operator tries per default to read the [cluster status json](https://apple.github.io/foundationdb/mr-status.html) with the [multi-version](https://apple.github.io/foundationdb/api-general.html#multi-version-client) from the system key-space.
+During an upgrade it can happen that the multi-version client tries to connect to the cluster with the wrong version especially if not all coordinators are upgraded.
+In this case the operator will fallback using `fdbcli` for the different versions (the one that is desired and the current running version).
```
sbodagala, line 146: My guess is that this works only when a majority of coordinators are on the "desired" version).

docs/manual/upgrades.md
sbodagala: Also, why don't we do automated replacements in this case too (so this can be resolved without requiring human intervention)?

docs/manual/upgrades.md
johscheuer: Good point. I add that 👍

docs/manual/upgrades.md
johscheuer: Created https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1518 for that. 2 hours is the default value and can be adjusted by users if wanted. The value was picked to ensure replacements are not triggered to quickly in cases where a host is unavailable for less than 2 hours.

docs/manual/upgrades.md
johscheuer: Correct.

docs/manual/upgrades.md
johscheuer: Correct, if the process is a log process this will issue also a recovery. 

> I wonder how recovery will work if the restarted process is at a later version? Especially, if this process happens to be a tlog process (as the sequencer has to read state from tlogs in order to proceed with recovery)? And, what if multiple tlogs (equal to the tlog replication factor) restart, can the sequencer even complete recovery?

Those are good questions I cannot answer directly and we have to dig deeper into the recovery process. In our test case where we keep half of the stateless and log processes at the old version, we see that the cluster will come up and recover.

docs/manual/upgrades.md
johscheuer: I believe we decided to set this value to 30s as default to prevent cases where a process is only missing for a short time. I agree with you that we should revisit the default value here to something shorter.

Created: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1519

docs/manual/upgrades.md
johscheuer: Correct, let me remove that.

docs/manual/upgrades.md
johscheuer: They are not marked with `IncorrectCommandLine` as those processes are running on the old FDB version and therefore are not part of the cluster status json output and as they are missing they will be marked as `MissingProcesses`.

> Also, why don't we do automated replacements (that is, replace these pods) in this case too (so this can be resolved without requiring human intervention)?

If the cluster is still available the operator will recreate those Pods with the new desired spec running the new FDB version, this is done in the `RemoveIncompatibleProcesses` subreconciler as mentioned above.

docs/manual/upgrades.md
sbodagala: The problem here could be that we are using the same condition to mark processes that are in different states. Thanks for filing that issue!

docs/manual/upgrades.md
sbodagala: > They are not marked with IncorrectCommandLine
So the operator is marking both the processes in this category and the processes in the previous category with the same condition, but the user would need to specify different "--process-condition" option to "kubectl fdb" to resolve these scenarios? That could be a problem. Again, I think the underlying problem is that the operator is using the same condition to mark processes that are in different states.

docs/manual/upgrades.md
sbodagala: Ok.

sbodagala:
Thanks a lot for preparing this document!

sbodagala (approved):

PR #1498, opened 2023-02-09
Creating a new type ProcessGroupID

09harsh:
# Description

*Please include a summary of the change and which issue is addressed. If this change resolves an issue, please include the issue number in the description.*

Created a new type `ProcessGroupID` to represent process group ID instead of using string, resolves #1476 

## Type of change

*Please select one of the options below.*

- New feature (non-breaking change which adds functionality)

## Discussion

*Are there any design details that you would like to discuss further?*
No

## Testing

*Please describe the tests that you ran to verify your changes. Unit tests?
Manual testing?*

*Do we need to perform additional testing once this is merged, or perform in a larger testing environment?*
Not needed, ran the existing test cases and checked fit lint errors
## Documentation

*Did you update relevant documentation within this repository?*

*If this change is adding new functionality, do we need to describe it in our user manual?*

*If this change is adding or removing subreconcilers, have we updated the core technical design doc to reflect that?*

*If this change is adding new safety checks or new potential failure modes, have we documented and how to debug potential issues?*

## Follow-up

*Are there any follow-up issues that we should pursue in the future?*

*Does this introduce new defaults that we should re-evaluate in the future?*

PR #1513, opened 2023-02-23
Fix the internal handling of the lock client for get pending upgrades

johscheuer:
# Description

In https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1498 we added a new typed string but we forgot to adjust the type in some places.

## Type of change

## Discussion

-

## Testing

e2e test. Currently this code path is not unit tested, I'm. going to create a new issue for doing that.

## Documentation

-

## Follow-up

-


09harsh (approved):

johscheuer:
Verified locally with an e2e test run that this solved the upgrade issue.

PR #1474, opened 2023-01-13
Improve operator restart logic for upgrades

johscheuer:
# Description

This is an attempt to relax the constraint for upgrades to ensure we account for scenarios where some processes are already upgraded (or are missing). Otherwise it could happen that a upgrade is blocked by a single process that was restarted and is already running on the new version.

## Type of change

*Please select one of the options below.*

- Bug fix (non-breaking change which fixes an issue)

## Discussion

-

## Testing

Unit tests + e2e tests.

## Documentation

I will write down a more in depth document about the upgrade process: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1473

## Follow-up

See above.

PR #1514, opened 2023-02-23
Revert fault tolerance check to only check availability

johscheuer:
# Description

In https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1474/files#diff-d0bee063a26a37c83244adbffd63b8a0daaf66032647b5e17bd78e0ca4cf4310R254 we made a change to check for fault tolerance and not only for availability. In theory this sounds like a good idea but we discovered in our e2e testings that this can block the operator from moving forward with the upgrade e.g. in the case where some processes are restarted before the operator dis the `kill` command.

## Type of change

## Discussion

-

## Testing

e2e tests.

## Documentation

Will be updated in https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1511

## Follow-up

-

### DIFF
--- a/controllers/bounce_processes.go
+++ b/controllers/bounce_processes.go
@@ -255,6 +255,10 @@ func getAddressesForUpgrade(logger logr.Logger, r *FoundationDBClusterReconciler
 		return nil, &requeue{curError: err}
 	}
 
+	// We don't want to check for fault tolerance here to make sure the operator is able to restart processes if some
+	// processes where restarted before the operator issued the cluster wide restart. For version incompatible upgrades
+	// that would mean that the processes restarted earlier are not part of the cluster anymore leading to a fault tolerance
+	// drop.
 	if !databaseStatus.Client.DatabaseStatus.Available {
 		r.Recorder.Event(cluster, corev1.EventTypeNormal, "UpgradeRequeued", "Database is unavailable")
 		return nil, &requeue{message: "Deferring upgrade until database is available"}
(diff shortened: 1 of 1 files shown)


johscheuer:
> * Commit ID: [dd38f75](https://github.com/FoundationDB/fdb-kubernetes-operator/commit/dd38f75d7b191fc510df81d4fe8fb92e2a314e03)
> * Duration 2:55:23
> * Result: ❌ **FAILED**
> * [Build Log](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator/fdb-kubernetes-operator-pr/77cccab1-afbb-4017-9ce6-004779ba38de.gz) terminal output (available for 30 days)
> * [Build Workspace](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator-pr/77cccab1-afbb-4017-9ce6-004779ba38de/workspace.zip) zip file of the working directory (available for 30 days)

One test failure is the empty monitor config issue (ignoring as @09harsh is looking into that). I look into the HA failure.

sbodagala:
> this can block the operator from moving forward with the upgrade

Is this because a storage server process already got restarted and that reduced fault tolerance?

controllers/bounce_processes.go
```diff
@@ -255,9 +255,9 @@ func getAddressesForUpgrade(logger logr.Logger, r *FoundationDBClusterReconciler
 		return nil, &requeue{curError: err}
 	}
```
sbodagala, line 257: A comment about why we don't want to do fault tolerance check here would be nice.

sbodagala (approved):

controllers/bounce_processes.go
johscheuer: Good point, I'm adding the comment.

johscheuer:
> > this can block the operator from moving forward with the upgrade
> 
> Is this because a storage server process already got restarted and that reduced fault tolerance?

Correct and it's better to move forward with restarting the rest of the processes.

PR #1521, opened 2023-02-27
Add Version Check to Redwood Experimental Storage Engine Type

mengzhbj:
# Description

FDB 7.0.0 is the first release of the Redwood Storage Engine. This PR complements version supporting check for Redwood Experimental storage engine type. 

## Type of change

## Discussion

No.

## Testing

Manual testing.
I think no additional testing needed once this is merge.

## Documentation

Didn't update any documentation. I just complemented the operator part, didn't summary any new doc.

## Follow-up

No.

### DIFF
--- a/api/v1beta2/foundationdb_database_configuration.go
+++ b/api/v1beta2/foundationdb_database_configuration.go
@@ -737,6 +737,8 @@ const (
 	StorageEngineRocksDbV1 StorageEngine = "ssd-rocksdb-v1"
 	// StorageEngineShardedRocksDB defines the storage engine ssd-sharded-rocksdb.
 	StorageEngineShardedRocksDB StorageEngine = "ssd-sharded-rocksdb"
+	// StorageEngineRedwood1Experimental defines the storage engine ssd-redwood-1-experimental.
+	StorageEngineRedwood1Experimental StorageEngine = "ssd-redwood-1-experimental"
 )
 
 // RoleCounts represents the roles whose counts can be customized.
--- a/api/v1beta2/foundationdbcluster_types_test.go
+++ b/api/v1beta2/foundationdbcluster_types_test.go
@@ -4548,6 +4548,17 @@ var _ = Describe("[api] FoundationDBCluster", func() {
 				},
 				fmt.Errorf("storage engine ssd-rocksdb-v1 is not supported on version 6.3.2"),
 			),
+			Entry("using invalid storage engine",
+				&FoundationDBCluster{
+					Spec: FoundationDBClusterSpec{
+						Version: "6.3.24",
+						DatabaseConfiguration: DatabaseConfiguration{
+							StorageEngine: StorageEngineRedwood1Experimental,
+						},
+					},
+				},
+				fmt.Errorf("storage engine ssd-redwood-1-experimental is not supported on version 6.3.24"),
+			),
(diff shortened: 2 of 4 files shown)


api/v1beta2/foundationdb_database_configuration.go
```diff
@@ -737,6 +737,8 @@ const (
 	StorageEngineRocksDbV1 StorageEngine = "ssd-rocksdb-v1"
 	// StorageEngineShardedRocksDB defines the storage engine ssd-sharded-rocksdb.
 	StorageEngineShardedRocksDB StorageEngine = "ssd-sharded-rocksdb"
+	// StorageEngineRedwood1Experimental defines the storage engine ssd-redwood-1-experimental.
+	StorageEngineRedwood1Experimental StorageEngine = "ssd-redwood-1-experimental"
```
johscheuer, line 741: Those changes are not required and the operator already supports the Redwood storage engine. The actual validation is done here: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/api/v1beta2/foundationdb_database_configuration.go#L42. If we add this constant we should also add it to the `IsStorageEngineSupported` check: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/api/v1beta2/foundationdb_version.go#L184.

api/v1beta2/foundationdbcluster_types_test.go
```diff
@@ -4548,6 +4548,17 @@ var _ = Describe("[api] FoundationDBCluster", func() {
 				},
 				fmt.Errorf("storage engine ssd-rocksdb-v1 is not supported on version 6.3.2"),
 			),
+			Entry("using invalid storage engine",
```
johscheuer, line 4551: This test needs some adjustments in https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/api/v1beta2/foundationdb_version.go#L184.

controllers/cluster_controller_test.go
```diff
@@ -2668,6 +2668,61 @@ var _ = Describe("cluster_controller", func() {
 					})
 				})
 			})
+			When("using ssd-redwood-1-experimental", func() {
```
johscheuer, line 2671: This test is nested in the wrong `When` statement, currently it's part of `When("creating a cluster with RocksDB as storage engine", func() {` which is wrong, that generates a test case like this: `when creating a cluster with RocksDB as storage engine when using ssd-redwood-1-experimental when using 6.3.24 generations are not matching`.

This test is currently failing since the `IsStorageEngineSupported` is not updated.

johscheuer (changes requested):
Thanks for the PR a few comments before we can merge it.

mengzhao-ibm:
@johscheuer  Thank a lot for your detailed review. Yes, kubebuilder:validation is enough and I didn't realize that. I see the kubebuilder:validation has covered several more storage engine types than StorageEngine. Do you think it's worth to add the version supporting check?
Anyway I'll add the check. If you and team think it's unnecessary, feel free to close the PR. Thanks.

controllers/cluster_controller_test.go
mengzhao-ibm: Sorry for this.

johscheuer (approved):
Thanks!

PR #1525, opened 2023-02-28
Add MaxUnavailablePods option for limiting number of unavailable pods…

### LINKED ISSUE #1444: Implement Max Unavailable

We should implement a `Max Unavailable` setting similar to what a Kubernetes Deployment provides: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#max-unavailable to prevent cases where the operator is deleting Pods and the new Pods are stuck in Pending. See https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1236#discussion_r1032580981 for some more context.

vishesh:
… during upgrade

FIXES #1444


manfontan:
Nice.

It would be nice to be able to set a limit depending on the process class. 

In our experience most of the problems happen when upgrading storage pods. So it would be nice to allow the operator to update process groups of other types for example logs while waiting for a storage pod to be available.

Issue #1089, opened 2022-03-07
Generate initial cluster file doesn't honour coordinator selection


johscheuer:
### What happened?

In the generate initial cluster file reconciler we only select storage Pods: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/controllers/generate_initial_cluster_file.go#L47. Instead we should make use of the https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/api/v1beta1/foundationdbcluster_types.go#L3055-L3058 and filter all Pods out that are not valid coordinators. This is a rather small bug without a huge impact since we hit this case only when we create a new cluster and the side effect will be that we have potentially 2 coordinator changes instead one. For consistency we should fix this bug.

### What did you expect to happen?

That the generate initial cluster file sub reconciler makes use of the coordinator selection mechanism.

### How can we reproduce it (as minimally and precisely as possible)?

Create a new cluster and change and watch the logs (you can also change the default coordinator selection).

### Anything else we need to know?

-

### FDB Kubernetes operator

main (latest)

### Kubernetes version

-

### Cloud provider

-

PR #1527, opened 2023-03-03
Generate initial cluster file based on "CoordinatorSelection"

sbodagala:
# Description

Modify the initial cluster file sub reconciler to make use of the coordinator selection mechanism. And, update the affected tests.

Addresses issue: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1089

## Type of change

## Discussion

- Does the current operator logic always select storage processes as coordinators? If so, this PR may cause log processes to also be selected as coordinators? And, if so, what is the impact of that during recovery? So, overall, is this a good change?

## Testing

Ran the operator unit tests successfully (by doing "make fmt lint test").

## Documentation

TBD: Do we need to update any documentation?

### DIFF
--- a/api/v1beta2/foundationdbcluster_types.go
+++ b/api/v1beta2/foundationdbcluster_types.go
@@ -1840,6 +1840,19 @@ func (cluster *FoundationDBCluster) IsEligibleAsCandidate(pClass ProcessClass) b
 	return false
 }
 
+// GetEligibleCandidateClasses returns process classes that are eligible to become coordinators.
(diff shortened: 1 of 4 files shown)


johscheuer (approved):
We are doing now more calls to the Kubernetes API (or the local cache), that could be prevented by using a label selector that selects multiple process classes but that would be an optimisation, that is only relevant in the initial cluster create phase so 👍 for the simplicity.

Thanks!

PR #1536, opened 2023-03-09
Adjust client check for cases where a newer client version is available

johscheuer:
# Description

There could be cases where a client already supports a newer version than the version the cluster will be upgraded to. Before the check was failing, with the new approach we ensure all connected clients support a compatible version.

## Type of change

## Discussion

## Testing

Unit tests added.

## Documentation

Updated documentation.

## Follow-up

None.

### DIFF
--- a/controllers/check_client_compatibility.go
+++ b/controllers/check_client_compatibility.go
@@ -79,17 +79,9 @@ func (c checkClientCompatibility) reconcile(_ context.Context, r *FoundationDBCl
 		return &requeue{curError: err}
 	}
 
-	var unsupportedClients []string
-	for _, versionInfo := range status.Cluster.Clients.SupportedVersions {
-		if versionInfo.ProtocolVersion == "Unknown" {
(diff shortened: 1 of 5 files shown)


controllers/check_client_compatibility.go
```diff
@@ -102,5 +94,47 @@ func (c checkClientCompatibility) reconcile(_ context.Context, r *FoundationDBCl
...
+			continue
+		}
+
+		parsedVersion, err := fdbv1beta2.ParseFdbVersion(versionInfo.ClientVersion)
+		if err != nil {
+			return nil, -1, err
+		}
+
+		if versionInfo.ProtocolVersion == protocolVersion {
+			compatibleClientsCount += versionInfo.Count
```
brownleej, line 127: What's the behavior here if a client has two multi-version libraries with the same protocol version? Do they get double-counted, or do they only get counted under one of the two libraries?

Issue #1535, opened 2023-03-07
Upgrade blocked when storageServersPerPod > 1


johscheuer:
### What happened?

See: [Upgrade never completes when storageServersPerPod > 1](https://forums.foundationdb.org/t/upgrade-never-completes-when-storageserversperpod-1/3844)

### What did you expect to happen?

Upgrade should complete.

### How can we reproduce it (as minimally and precisely as possible)?

See: [Upgrade never completes when storageServersPerPod > 1](https://forums.foundationdb.org/t/upgrade-never-completes-when-storageserversperpod-1/3844)

### FDB Kubernetes operator

<details>

```console
$ kubectl fdb version
# paste output here
```

</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Cloud provider

<details>

</details>

PR #1538, opened 2023-03-13
Fix upgrade logic for multiple storage server per Pod

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1535

## Type of change

## Discussion

-

## Testing

Unit test, we should also add a e2e test for this.

## Documentation

-

## Follow-up

-

### DIFF
--- a/controllers/bounce_processes.go
+++ b/controllers/bounce_processes.go
@@ -228,7 +228,7 @@ func getProcessesReadyForRestart(logger logr.Logger, cluster *fdbv1beta2.Foundat
 	// If more than one storage server per Pod is running we have to account for this. In this case we have to add the
 	// additional storage processes.
 	if cluster.Spec.StorageServersPerPod > 1 {
-		expectedProcesses = counts.Storage * (cluster.Spec.StorageServersPerPod - 1)
+		expectedProcesses += counts.Storage * (cluster.Spec.StorageServersPerPod - 1)
 	}
 
 	if cluster.IsBeingUpgradedWithVersionIncompatibleVersion() && expectedProcesses != len(addresses) {
--- a/controllers/bounce_processes.go
+++ b/controllers/bounce_processes.go
@@ -71,6 +71,8 @@ func (bounceProcesses) reconcile(ctx context.Context, r *FoundationDBClusterReco
 		return nil
 	}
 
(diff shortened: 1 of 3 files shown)


controllers/bounce_processes.go
```diff
@@ -219,8 +223,10 @@ func getProcessesReadyForRestart(logger logr.Logger, cluster *fdbv1beta2.Foundat
 	// if some processes are already upgraded e.g. in the case of version compatible upgrades and we also don't want to
 	// block the restart command if a process is missing longer than the specified GetIgnoreMissingProcessesSeconds.
 	// Those checks should ensure we only run the restart command if all processes that have to be restarted and are connected
-	// to cluster are ready to be restarted.
-	if cluster.IsBeingUpgradedWithVersionIncompatibleVersion() && (counts.Total()-missingProcesses) != len(addresses) {
+	// to cluster are ready to be restarted. If more than one storage server per Pod is running we have to account for this.
+	expectedProcesses := counts.Total() - missingProcesses + (counts.Storage * (cluster.Spec.StorageServersPerPod - 1))
```
sbodagala: Is it better to modify "cluster.GetProcessCountsWithDefaults()" to take "cluster.Spec.StorageServersPerPod" into account, rather than doing this adjustment on the invoker's side? And, have we checked the other invoker functions of "cluster.GetProcessCountsWithDefaults()" to see if they need to do a similar adjustment?

johscheuer: Not really, the difference here is that the process counts in most cases reflect the number of Pods and not the processes (the naming here is not really great) and in most cases the operator is only working on Pods as a whole without taking into account how many processes are running inside that Pod.

sbodagala: Ok. (A comment above "cluster.GetProcessCountsWithDefaults()" that it doesn't take "cluster.Spec.StorageServersPerPod" would be nice, so other readers would be aware of this detail.)

sbodagala (approved):

controllers/bounce_processes.go
johscheuer: Good point 👍 Let me add a comment on the method.

Issue #560, opened 2021-03-19
Document how to spin up an HA cluster


johscheuer:
Currently we have limited documentation how to spin up an HA FDB cluster with the operator. We should define what steps are required to spin up a cluster existing of multiple FDB cluster instances (e.g. multiple FDB Cluster resources).

PR #1548, opened 2023-03-21
Update examples and better document how to run a multi-region FDB cluster

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/560 and updates the example version for FDB clusters.

## Type of change

- Documentation

## Discussion

-

## Testing

-

## Documentation

Updated

## Follow-up

-

### DIFF
--- a/config/samples/backup.yaml
+++ b/config/samples/backup.yaml
@@ -53,7 +53,7 @@ spec:
         secret:
           secretName: fdb-kubernetes-operator-secrets
   snapshotPeriodSeconds: 3600
-  version: 6.3.22
+  version: 7.1.26
 ---
 apiVersion: apps.foundationdb.org/v1beta2
 kind: FoundationDBCluster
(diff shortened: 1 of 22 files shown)


config/samples/backup.yaml
```diff
@@ -53,7 +53,7 @@ spec:
         secret:
           secretName: fdb-kubernetes-operator-secrets
   snapshotPeriodSeconds: 3600
-  version: 6.3.22
+  version: 7.1.26
```
sbodagala, line 56: Why "7.1.26" here (instead of "7.1.27")?

config/samples/client.yaml
```diff
@@ -149,4 +149,4 @@ spec:
     enableLivenessProbe: true
     enableReadinessProbe: false
   useExplicitListenAddress: true
-  version: 6.3.22
+  version: 7.1.26
```
sbodagala, line 152: Same question here (and below)?

sbodagala (approved):

config/samples/backup.yaml
johscheuer: I chose 7.1.26 as this version has no AVX support (otherwise it could result in some errors that are not obvious if you don't know that AVX support is required).

PR #1559, opened 2023-03-29
Ensure we only mark the database as unavailable if we can talk to the majority of coordinators

johscheuer:
# Description

This change makes the status invariant checker more robust during upgrades. If the operator is not able to communicate to the quorum of the coordinator we cannot make any assumption about the cluster state.

## Type of change

## Discussion

-

## Testing

Running e2e tests locally (will report once they finished).

## Documentation

Added in code.

## Follow-up

-

### DIFF
--- a/e2e/fixtures/status.go
+++ b/e2e/fixtures/status.go
@@ -290,7 +290,9 @@ func (fdbCluster FdbCluster) InvariantClusterStatusAvailableWithThreshold(
 func (fdbCluster FdbCluster) InvariantClusterStatusAvailable() error {
 	return fdbCluster.StatusInvariantChecker(
 		"InvariantClusterStatusAvailable",
-		0,
+		// Per default we allow 5 seconds unavailability. Otherwise we could get a few test failures when we do operations
+		// like a replacement on a transaction system Pod and the recovery takes longer.
+		5*time.Second,
 		checkAvailability,
 	)
 }
--- a/e2e/fixtures/status.go
+++ b/e2e/fixtures/status.go
@@ -52,7 +52,6 @@ func (fdbCluster *FdbCluster) getStatusFromOperatorPod() *fdbv1beta2.FoundationD
 		return err
 	}).WithTimeout(2 * time.Minute).WithPolling(1 * time.Second).ShouldNot(gomega.HaveOccurred())
 
-	// TODO remove the gabs dependency in this part and make use of the struct
 	return status
 }
 
--- a/e2e/fixtures/status.go
+++ b/e2e/fixtures/status.go
@@ -260,14 +260,8 @@ func (fdbCluster FdbCluster) StatusInvariantChecker(
 	)
 }
 
-// checkAvailability returns nil if the cluster is reachable or if not a quorum of the coordinators are reachable. If
-// the cluster is unreachable an error will be returned.
+// checkAvailability returns nil if the cluster is reachable. If the cluster is unreachable an error will be returned.
 func checkAvailability(status *fdbv1beta2.FoundationDBStatus) error {
-	// If we are not able to reach the quorum of coordinators, we cannot make any assumption about the status of the Cluster.
-	if !status.Client.Coordinators.QuorumReachable {
-		return nil
-	}
-
 	if !status.Client.DatabaseStatus.Available {
 		return fmt.Errorf("cluster is not available")
 	}
--- a/e2e/fixtures/status.go
+++ b/e2e/fixtures/status.go
@@ -260,6 +260,21 @@ func (fdbCluster FdbCluster) StatusInvariantChecker(
 	)
 }
 
+// checkAvailability returns nil if the cluster is reachable or if not a quorum of the coordinators are reachable. If
+// the cluster is unreachable an error will be returned.
+func checkAvailability(status *fdbv1beta2.FoundationDBStatus) error {
+	// If we are not able to reach the quorum of coordinators, we cannot make any assumption about the status of the Cluster.
+	if !status.Client.Coordinators.QuorumReachable {
+		return nil
+	}
+
+	if !status.Client.DatabaseStatus.Available {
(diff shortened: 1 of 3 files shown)


e2e/fixtures/status.go
```diff
@@ -260,35 +260,40 @@ func (fdbCluster FdbCluster) StatusInvariantChecker(
...
-		func(status *fdbv1beta2.FoundationDBStatus) error {
-			if !status.Client.DatabaseStatus.Available {
-				return fmt.Errorf("cluster.database_available=false")
-			}
-
-			return nil
-		},
+		// Per default we allow 5 seconds unavailability. Otherwise we could get a few test failures when we do operations
+		// like a replacement on a transaction system Pod and the recovery takes longer.
+		5*time.Second,
```
johscheuer, line 295: Previously the check was too sensitive and we had some failures like this:

```text
invariant InvariantClusterStatusAvailable failed after: 343ns
```

johscheuer:
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1562

e2e/fixtures/status.go
```diff
@@ -260,20 +260,29 @@ func (fdbCluster FdbCluster) StatusInvariantChecker(
...
 
+// checkAvailability returns nil if the cluster is reachable or if not a quorum of the coordinators are reachable. If
+// the cluster is unreachable an error will be returned.
+func checkAvailability(status *fdbv1beta2.FoundationDBStatus) error {
+	// If we are not able to reach the quorum of coordinators, we cannot make any assumption about the status of the Cluster.
+	if !status.Client.Coordinators.QuorumReachable {
+		return nil
+	}
+
+	if !status.Client.DatabaseStatus.Available {
```
sbodagala, line 271: Sorry, I am missing the context here - if a quorum of coordinators is not reachable then shouldn't the status be empty and so this flag should have been cleared? Maybe I need to check the logic used for setting/clearing these flags.

e2e/fixtures/status.go
```diff
@@ -260,20 +260,29 @@ func (fdbCluster FdbCluster) StatusInvariantChecker(
 	)
 }
 
+// checkAvailability returns nil if the cluster is reachable or if not a quorum of the coordinators are reachable. If
```
sbodagala, line 263: Same/Similar question: how can the cluster be reachable if a quorum of coordinators is not reachable?

e2e/fixtures/status.go
johscheuer: The problem here is that this value is not a pointer. So if the value is not set and we convert it to the `FoundationDBStatus` it will be `false` (default) value. Since this is an exposed "API" or struct we cannot just change it as it would break things. I have to verify if we can have the json struct point to two fields.

e2e/test_operator_upgrades/operator_upgrades_test.go
```diff
@@ -334,7 +341,8 @@ var _ = Describe("Operator Upgrades", Label("e2e"), func() {
 				Skip("this test case only affects version incompatible upgrades")
 			}
 
-			clusterSetup(beforeVersion)
+			// We ignore the availability check here since this check is sometimes flaky if not all coordinators are running.
+			clusterSetup(beforeVersion, false)
```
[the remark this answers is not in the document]
johscheuer: Yeah, this specific test should be fine to enable it.

e2e/test_operator_upgrades/operator_upgrades_test.go
```diff
@@ -453,7 +461,8 @@ var _ = Describe("Operator Upgrades", Label("e2e"), func() {
 				Skip("operator doesn't support feature for test case")
 			}
 
-			clusterSetup(beforeVersion)
+			// We ignore the availability check here since this check is sometimes flaky if not all coordinators are running.
+			clusterSetup(beforeVersion, false)
```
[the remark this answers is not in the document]
johscheuer: Same here it should be fine to be enabled. I disabled the check in `upgrading a cluster and multiple processes are not restarted` as it could bring down the transaction system. Same in `upgrading a cluster and no coordinator is restarted`, this will cause the cluster to be unavailable for some time.

e2e/fixtures/status.go
sbodagala: > So if the value is not set and we convert it to the FoundationDBStatus it will be false (default) value.

The default value should work here, isn't it? Are you saying "Available" gets set to "true" if a quorum of coordinators is not reachable?

e2e/test_operator_upgrades/operator_upgrades_test.go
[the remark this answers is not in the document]
sbodagala: Ok.

johscheuer:
> * Commit ID: [a14dda1](https://github.com/FoundationDB/fdb-kubernetes-operator/commit/a14dda1a3c789a28aa7fd4839dfca21f2bb61b70)
> * Duration 3:08:56
> * Result: ❌ **FAILED**
> * [Build Log](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator/fdb-kubernetes-operator-pr/073f7c4e-c831-46c8-844d-e793f6943bf6.gz) terminal output (available for 30 days)
> * [Build Workspace](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator-pr/073f7c4e-c831-46c8-844d-e793f6943bf6/workspace.zip) zip file of the working directory (available for 30 days)

```text
ERROR: This exclude may cause the total free space in the cluster to drop below 10%.\nType `exclude FORCE <ADDRESS...>' to exclude without checking free space.\n
```

e2e/fixtures/status.go
johscheuer: Updated the code.

sbodagala (approved):

PR #1563, opened 2023-03-30
Correct the way how the provided images and registry are used

johscheuer:
# Description

Fix the way how the sidecar is used in the operator setup. I removed the singleton setup and created a method in the factory struct. This should simplify the usage of the images.

## Type of change

## Discussion

-

## Testing

I did a local run with a different registry.

## Documentation

-

## Follow-up

-


kakaiu (approved):
Thank you for the fix!

PR #1565, opened 2023-03-31
Refactor e2e test suite to make it easier to use the sidecar config

johscheuer:
# Description

Expose the sidecar config to make it easier to consume it from outside the test suite.

## Type of change

- Other

## Discussion

I changed the type from string to fdbv1beta2.version to remove some steps where we parsed the version multiple times.

## Testing

Ran a an e2e upgrade test.

## Documentation

Added in the code itself.

## Follow-up

-

### DIFF
--- a/e2e/fixtures/factory.go
+++ b/e2e/fixtures/factory.go
@@ -503,58 +503,46 @@ func (factory *Factory) GetBeforeVersion() string {
 	return factory.beforeVersion
 }
 
-func getUpgradeVersions(upgradeString string) []string {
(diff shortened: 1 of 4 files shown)


hfu94 (approved):

johscheuer (commented):
All tests passed except for one, that one failed with:

```
"ERROR: This exclude may cause the total free space in the cluster to drop below 10%.\nType `exclude FORCE <ADDRESS...>' to exclude without checking free space.\n"
```

We are already investigating that.

PR #1573, opened 2023-04-06
Add kind setup script for e2e tests

johscheuer:
# Description

Add a script to setup a kind cluster for e2e tests. Those can be used to spin up a kind cluster in CI to run our e2e tests against a Kind cluster.

## Type of change

## Discussion

-

## Testing

I ran those scripts locally.

## Documentation

Updated the e2e documentation.

## Follow-up

Change CI tests to make use of this new script.


e2e/scripts/setup-local-fdb-cluster.sh
```diff
@@ -0,0 +1,69 @@
...
+- role: worker
+  image: kindest/node:${KUBE_VERSION}
+EOF
+else
+    echo "===Kind cluster already exists"
+fi
+
+kubectl config use-context "kind-${CLUSTER}"
+
+echo "===Start building operator"
```
johscheuer, line 54: Once I'm back I'm going to remove some parts of this script here and move it to a document. This currently breaks some stuff in our GitHub actions (and was flaky before that).

e2e/Makefile
```diff
@@ -39,7 +39,20 @@ clean:
 # This target can be used to delete namespaces that were created during testing. If a namespace is provided the user and
 # the namespace (as a prefix) must match.
 destroy-my-namespaces:
-	./remove_namespaces ${TEST_USERNAME} ${NAMESPACE}
+	./scripts/remove_namespaces ${TEST_USERNAME} ${NAMESPACE}
+
+# This target can be used to create a kind cluster that can be used for e2e testing.
+kind-setup:
+	@CHAOS_NAMESPACE=$(CHAOS_NAMESPACE) FDB_VERSION=$(FDB_VERSION) REGISTRY=$(REGISTRY) $(shell pwd)/scripts/setup_e2e.sh
```
xumengpanda, line 46: setup_e2e.sh creates an HA cluster by default. 
Good to have a setup to create non-HA cluster, which uses less resource and allows us to cover non-HA scenario.

e2e/scripts/setup-local-fdb-cluster.sh
xumengpanda: which part is flaky?

e2e/Makefile
johscheuer: This calls the `./e2e/scrips/setup_e2e.sh` not the other script (`./scripts/setup_e2e.sh`) (I will rename that and move it into the `./e2e/scripts/` folder).

e2e/scripts/setup-local-fdb-cluster.sh
johscheuer: I left the script as it is but removed the usage from the CI. The setup had the following problems:

1.) We installed the latest CRD directly, so this check: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/.github/workflows/pull_request.yml#L87-L101 was broken.
2.) Based on all runs that I have seen the FDB cluster was never running: https://github.com/FoundationDB/fdb-kubernetes-operator/actions/runs/4669101303/jobs/8267089024#step:7:2729, creating the cluster without any further checks brings no value.

sbodagala (approved):

PR #1578, opened 2023-04-07
Update FDB version used for e2e tests

johscheuer:
# Description

Make use of the latest FDB 7.1 release (7.1.31) for our e2e tests.

## Type of change

- Other

## Discussion

This should resolved the last flakiness of our tests where we same the exclude command failing with `10% free space`.

## Testing

e2e pipeline will be running.

## Documentation

-

## Follow-up

-

### DIFF
--- a/e2e/Makefile
+++ b/e2e/Makefile
@@ -9,7 +9,7 @@ TIMEOUT?=168h
 
 NAMESPACE?=
 CONTEXT?=
-FDB_VERSION?=7.1.27
+FDB_VERSION?=7.1.31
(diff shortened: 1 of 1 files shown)


e2e/Makefile
```diff
@@ -9,7 +9,7 @@ TIMEOUT?=168h
 
 NAMESPACE?=
 CONTEXT?=
-FDB_VERSION?=7.1.27
+FDB_VERSION?=7.1.31
```
johscheuer, line 12: Seems like the new image is not yet available in out testing infrastructure.

sbodagala (approved):

Issue #1549, opened 2023-03-24
Ensure we document how the operator handles exclusions


johscheuer:
### What would you like to be added/changed?

We should make sure that we document the way how the operator is doing exclusions and how it detects that it's safe to delete a specific process.

PR #1579, opened 2023-04-10
Run the exclude command with the highest timeout

johscheuer:
# Description

If we exclude a stateful process, the exclusion can take longer than the timeout, resulting in a timeout error. If we run the exclude command with the backoff we potentially trigger multiple exclusions for the same set of addresses, since the operator cannot differentiate between the exclude command timing out (exclusion is not started) and the exclusion is started but the processes still have data and it's not safe to delete them. One workaround would be to check the cluster status after the exclusion commands but this would be a rather expensive check compared to running with the highest timeout.

## Type of change

## Discussion

-

## Testing

Unit tests + e2e tests that will run.

## Documentation

I'm planning to work on https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1549 next week.

## Follow-up

-

### DIFF
--- a/fdbclient/admin_client.go
+++ b/fdbclient/admin_client.go
@@ -110,6 +110,16 @@ func NewCliAdminClient(cluster *fdbv1beta2.FoundationDBCluster, _ client.Client,
 	}, nil
 }
 
+// getMaxTimeout returns the maximum timeout, this is either the default of 40 seconds or if the provided default timeout
+// is higher it will be the default cli timeout.
+func (client *cliAdminClient) getMaxTimeout() time.Duration {
+	if DefaultCLITimeout > 40*time.Second {
+		return DefaultCLITimeout
+	}
+
+	return 40 * time.Second
+}
+
 // cliCommand describes a command that we are running against FDB.
 type cliCommand struct {
 	// binary is the binary to run.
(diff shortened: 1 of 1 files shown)


johscheuer (commented):
Ping @sbodagala

fdbclient/admin_client.go
```diff
@@ -110,6 +110,16 @@ func NewCliAdminClient(cluster *fdbv1beta2.FoundationDBCluster, _ client.Client,
 	}, nil
 }
 
+// getMaxTimeout returns the maximum timeout, this is either the default of 40 seconds or if the provided default timeout
+// is higher it will be the default cli timeout.
+func (client *cliAdminClient) getMaxTimeout() time.Duration {
+	if DefaultCLITimeout > 40*time.Second {
+		return DefaultCLITimeout
```
sbodagala, line 117: It appears "DefaultCLITimeout" is getting set to "10 * time.Second" by default. Can the user change it?

sbodagala (approved):

fdbclient/admin_client.go
johscheuer: Yes, the user can use this flag: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/setup/setup.go#L90, which will be used here: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/setup/setup.go#L141

PR #1582, opened 2023-04-11
Fix the way we build the node list in the getNodes function.

brownleej:
# Description

This fixes a bug where we would have a bunch of blank entries in the node list before the real nodes.

## Type of change

## Discussion

No.

## Testing

Manual testing?*

I ran the command from my local build.

No.

## Documentation

No.

## Follow-up

No.

No.


JoshuaMcManus (approved):

johscheuer (commented):
Thanks!

PR #1587, opened 2023-04-19
Add MaxZonesWithUnavailablePods for the update process. Fixes #1444.

### LINKED ISSUE #1444: Implement Max Unavailable

manfontan:
# Description

`MaxZonesWithUnavailablePods` option limits the number of zones (fault domains) with unavailable pods during an update. By default it is disabled

fixes #1444 

duplicates #1525 

## Type of change

## Discussion

*Are there any design details that you would like to discuss further?
NO

## Testing

Manual testing?
Unit tests
Manual testing 

Replace FDB base image for all log,storage and stateless:

Fault Tolerance was always above 1 and replicas were always >=2.

Screenshot 2023-06-02 at 17 25 07

The feature prevented the operator from deleting pods too fast 36 times with messages like:

```
fdb-kubernetes-operator-controller-manager-66dbb578c-hzrmt manager {"level":"info","ts":1685722511.3212197,"logger":"controller","msg":"Skip process group for update, the number zones with unavailable Pods equals the limit but the process group does not belong to a zone with unavailable Pods","namespace":"timeseries","cluster":"foundationdb-cluster","reconciler":"updatePods","processGroupID":"storage-8","maxZonesWithUnavailablePods":1,"faultDomain":"fdb1-26081535-vmss0000z2","faultDomainsWithUnavailablePods":{"fdb1-26081535-vmss0000yv":{}}}
```

## Documentation

*Did you update relevant documentation within this repository? Yes

*If this change is adding new functionality, do we need to describe it in our user manual? Yes

*If this change is adding or removing subreconcilers, have we updated the core technical design doc to reflect that?N/A

## Follow-up

*Are there any follow-up issues that we should pursue in the future? No

*Does this introduce new defaults that we should re-evaluate in the future? Yes. By default the max number of unavailable pods during updates is not limited. In the future it could be set to a specific number.


controllers/update_pods.go
```diff
@@ -87,25 +87,34 @@ func (updatePods) reconcile(ctx context.Context, r *FoundationDBClusterReconcile
 func getPodsToUpdate(logger logr.Logger, reconciler *FoundationDBClusterReconciler, cluster *fdbv1beta2.FoundationDBCluster, podMap map[fdbv1beta2.ProcessGroupID]*corev1.Pod) (map[string][]*corev1.Pod, error) {
 	updates := make(map[string][]*corev1.Pod)
 
+	var unavailablePods int
+
 	for _, processGroup := range cluster.Status.ProcessGroups {
 		if processGroup.IsMarkedForRemoval() {
 			logger.V(1).Info("Ignore removed Pod",
 				"processGroupID", processGroup.ProcessGroupID)
+			unavailablePods++
```
johscheuer, line 96: Any reason why a process group marked for removal is counted as unavailable? I would have assumed that unavailable means the Pod is not existing or in a not running state?

controllers/update_pods.go
```diff
@@ -87,25 +87,34 @@ func (updatePods) reconcile(ctx context.Context, r *FoundationDBClusterReconcile
...
 			logger.V(1).Info("Ignore pending Pod",
 				"processGroupID", processGroup.ProcessGroupID)
+			unavailablePods++
 			continue
 		}
 
 		if cluster.NeedsReplacement(processGroup) {
 			logger.V(1).Info("Skip process group for deletion, requires a replacement",
 				"processGroupID", processGroup.ProcessGroupID)
+			unavailablePods++
```
johscheuer, line 110: I think increasing the number of `unavailablePods` is wrong here.

api/v1beta2/foundationdbcluster_types.go
```diff
@@ -217,6 +217,12 @@ type FoundationDBClusterSpec struct {
 	// UseUnifiedImage determines if we should use the unified image rather than
 	// separate images for the main container and the sidecar container.
 	UseUnifiedImage *bool `json:"useUnifiedImage,omitempty"`
+
+	// MaxUnavailablePods defines the maximum number of pods that can be unavailable during the update process.
+	// When set to 0 there is no limit on the number of pods that can be unavailable.
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:default:=0
+	MaxUnavailablePods int `json:"maxUnavailablePods,omitempty"`
```
johscheuer, line 225: There is an `intstr` type that we might want to use, to have the possibility to use a percentage instead of using an integer,

johscheuer (changes requested):
Could you please add tests that show that the newly added setting is actually behaving as expected and preventing the operator from taking down more than thee specified Pods?

config/rbac/role.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-kind: Role
+kind: ClusterRole
```
manfontan, line 3: not sure why this changed. I did not do it 🤔

controllers/update_pods.go
```diff
@@ -87,13 +87,24 @@ func (updatePods) reconcile(ctx context.Context, r *FoundationDBClusterReconcile
...
+	var unavailablePods int
+
 	for _, processGroup := range cluster.Status.ProcessGroups {
 		if processGroup.IsMarkedForRemoval() {
 			logger.V(1).Info("Ignore removed Pod",
 				"processGroupID", processGroup.ProcessGroupID)
 			continue
 		}
 
+		if cluster.Spec.MaxUnavailablePods.IntValue() > 0 {
```
manfontan, line 99: I have created a dedicated validation since SkipProcessGroups uses the ignore pending time that defaults to 5min. 
And I think it is simpler to check it separately than having to set two values to get the functionality. Specially because the default ignore pending time value results in Pending Pods not being counted unless they have been pending for more than 5 min.

config/samples/deployment.yaml
```diff
@@ -150,7 +150,7 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
-        image: foundationdb/fdb-kubernetes-operator:v1.16.0
+        image: foundationdb/max_unavailable_pods-2023-04-24-19-45-600346b1
```
manfontan, line 153: ignore this change I will revert it.

controllers/update_pods.go
```diff
@@ -182,6 +199,10 @@ func getPodsToUpdate(logger logr.Logger, reconciler *FoundationDBClusterReconcil
 		updates[zone] = append(updates[zone], pod)
 	}
 
+	if maxUnavailablePods > 0 && unavailablePods >= maxUnavailablePods {
+		return nil, fmt.Errorf("unavailable Pods reached cluster.Spec.MaxUnavailablePods limit: %d", maxUnavailablePods)
```
johscheuer, line 203: Should we print here the number of unavailable Pods?

controllers/update_pods_test.go
```diff
@@ -246,5 +249,70 @@ var _ = Describe("update_pods", func() {
...
+				expectedError = true
+				cluster.Spec.MaxUnavailablePods = intstr.FromString("1%")
+				Expect(k8sClient.Update(context.TODO(), cluster)).NotTo(HaveOccurred())
+
+				var numPendingPods int
+				for _, processGroup := range cluster.Status.ProcessGroups {
+					if processGroup.ProcessClass.IsStateful() {
+						processGroup.ProcessGroupConditions = append(processGroup.ProcessGroupConditions, fdbv1beta2.NewProcessGroupCondition(fdbv1beta2.PodPending))
+						numPendingPods++
+						if numPendingPods == 8 {
```
johscheuer, line 264: This case is never hit right? The initial value of `numPendingPods` will be 0 and before we do the check here we are increasing the value by 1, so it will always be > 0 and all process groups will get this condition? Was the idea of `numPendingPods` to be used for debugging?

controllers/update_pods_test.go
```diff
@@ -246,5 +249,70 @@ var _ = Describe("update_pods", func() {
...
+			})
+
+			It("should return an error and nil updates", func() {
+				Expect(updates).To(BeNil())
+				Expect(expectedError).To(BeTrue())
+				Expect(err.Error()).Should(ContainSubstring("unavailable Pods reached cluster.Spec.MaxUnavailablePods limit: 1"))
+			})
+		})
+
+		When("max unavailable pods is set with an int value and there are process groups with pods in pending status", func() {
```
johscheuer, line 278:
```suggestion
		When("max unavailable pods is set with an int value and there are two process groups with pods in pending status", func() {
```

controllers/update_pods_test.go
```diff
@@ -246,5 +249,70 @@ var _ = Describe("update_pods", func() {
 				Expect(updates).To(HaveLen(1))
 			})
 		})
+
+		When("max unavailable pods is set with a percent value and there are process groups with pods in pending status", func() {
```
johscheuer, line 253: Can we add a test case where we actually update some Pods even if some of them are unavailable?

controllers/update_pods.go
```diff
@@ -182,6 +199,10 @@ func getPodsToUpdate(logger logr.Logger, reconciler *FoundationDBClusterReconcil
...
+	}
+
```
johscheuer, line 205: I believe the current logic is not doing what it proposes: `MaxUnavailablePods defines the maximum number or percent of pods that can be unavailable during the update process.`

But we are actually only checking if `unavailablePods >= maxUnavailablePods` before doing the updates. So we could hit a case where `maxUnavailablePods` is `5` and `unavailablePods` is `4`, this would mean `unavailablePods >= maxUnavailablePods` is false and we have an update budget of 1 Pod, but in the current implementation we are not verifying if the `updates` will only update `1` Pod and not more (otherwise we have more unavailable Pods than defined in `maxUnavailablePods`)

johscheuer (changes requested):
Most of the logic looks fine, we just have to make sure we are not updating more Pods then we have a "budget" for.

controllers/update_pods_test.go
manfontan: The idea here was to set the PodPending condition for a certain number (8 in this case) of process groups. Once the desired number is reached `numPendingPods` we skip setting the PodsPending condition for the remaining process groups.

Does that make sense?

controllers/update_pods.go
```diff
@@ -182,6 +199,15 @@ func getPodsToUpdate(logger logr.Logger, reconciler *FoundationDBClusterReconcil
 		updates[zone] = append(updates[zone], pod)
 	}
 
+	var numPodsToUpdate int
+	for _, zone := range updates {
+		numPodsToUpdate += len(zone)
+	}
+
+	if maxUnavailablePods > 0 && (unavailablePods+numPodsToUpdate) >= maxUnavailablePods {
```
johscheuer, line 207: I believe this calculation is still wrong. Assume a user sets `maxUnavailablePods` to `1` to make sure only 1 Pod is unavailable at a time (at least from planned activities) and now the user makes a change to the cluster that requires the operator to update all Pods in the cluster. For simplicity assume the cluster has 3 Pods, in this case the operator will never be able to make progress as `(unavailablePods+numPodsToUpdate) >= maxUnavailablePods` will be `3 >= 1` which will be blocking forever.

I believe a better approach is to return the updates map and the number of Pods that can be updated, in the case above that would be 1 if all Pods are available. If the updates map has at lease one entry but the Pods that can be updated is 0 (or a negative number) the operator should do a (delayed) requeue. If the number of Pods that can be updated is greater than 0 the operator should prefer to update zones that have less or equal the number of Pods that can be updated (to make sure we update the zone at once) if not the operator should update as many Pods as possible from the zone.

That way we make sure the operator is able to make progress in the different cases.

controllers/update_pods_test.go
```diff
@@ -246,5 +249,98 @@ var _ = Describe("update_pods", func() {
...
+				Expect(expectedError).To(BeTrue())
+				Expect(err.Error()).Should(ContainSubstring("cluster has 2 Pods that are pending or missing and 0 Pods will be updated, exceeding cluster.Spec.MaxUnavailablePods: 1"))
+			})
+		})
+
+		When("max unavailable pods is greater than the number of pods requiring update plus the number of missing or pending pods", func() {
+			BeforeEach(func() {
+				expectedError = false
+				// Set max unavailable pods to 100 (arbitrary large number)
+				cluster.Spec.MaxUnavailablePods = intstr.FromInt(100)
```
johscheuer, line 307: Can we use a more reasonable value here to e.g. allow only to update 1 Pod. Basically the value 100 in this test case is the same as setting the limit to 0 (not having any limit) as the Pods created by this cluster is much smaller than 100.

controllers/update_pods.go
```diff
@@ -87,13 +88,26 @@ func (updatePods) reconcile(ctx context.Context, r *FoundationDBClusterReconcile
 func getPodsToUpdate(logger logr.Logger, reconciler *FoundationDBClusterReconciler, cluster *fdbv1beta2.FoundationDBCluster, podMap map[fdbv1beta2.ProcessGroupID]*corev1.Pod) (map[string][]*corev1.Pod, error) {
 	updates := make(map[string][]*corev1.Pod)
 
+	var unavailablePods int
+	// When maxUnavailablePods is set to 0 any number of unavailable Pods is allowed.
+	maxUnavailablePods, err := intstr.GetScaledValueFromIntOrPercent(&cluster.Spec.MaxUnavailablePods, len(cluster.Status.ProcessGroups), true)
```
johscheuer, line 93: I believe we should count Pods that have a deletionTimestamp !=0 as unavailable, what do you think?

controllers/update_pods_test.go
johscheuer: That makes sense, I just was confused what the benefit of the previous check was.

johscheuer (changes requested):

controllers/update_pods.go
manfontan: That makes total sense. I like the idea. I hope I got it right this time :D

johscheuer (commented):
Once we merge the new `TrimUpdatedToMaxPodsToUpdate` with the `getPodsToDelete` this PR is good. Do you mind to add a section about the new setting in https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/docs/manual/replacements_and_deletions.md? Could you also create an issue to build a dedicated e2e test for this new feature?

manfontan:
> Once we merge the new `TrimUpdatedToMaxPodsToUpdate` with the `getPodsToDelete` this PR is good. Do you mind to add a section about the new setting in https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/docs/manual/replacements_and_deletions.md? Could you also create an issue to build a dedicated e2e test for this new feature?

I will add a section in the doc. 

About merging `TrimUpdatedToMaxPodsToUpdate` I don't think it is necessary since the new `getPodsToDelete` logic already does the same thing. 

Depending on the deletion Mode:
- PodUpdateModeAll: maxPodsToUpdate is ignored.
- PodUpdateModeProcessGroup: we only delete pods if maxPodsToUpdate>=1
- PodUpdateModeZone:
  - If the zone process count < maxPodsToUpdate we pick the entire zone to delete.
  - If the zone is > maxPodsToUpdate we trim the slice to get only the first maxPodsToUpdate processes
- PodUpdateModeNone: deletes nothing

Does that make sense?

api/v1beta2/foundationdbcluster_types.go
```diff
@@ -218,6 +219,11 @@ type FoundationDBClusterSpec struct {
 	// UseUnifiedImage determines if we should use the unified image rather than
 	// separate images for the main container and the sidecar container.
 	UseUnifiedImage *bool `json:"useUnifiedImage,omitempty"`
+
+	// MaxZonesWithUnavailablePods defines the maximum number or percent of zones that can have unavailable pods during the update process.
+	// When set to 0 there is no limit on the number or percent of zones with unavailable pods.
+	// +kubebuilder:default:=0
+	MaxZonesWithUnavailablePods intstr.IntOrString `json:"maxZonesWithUnavailablePods,omitempty"`
```
johscheuer, line 226: I think in this case the percentage doesn't make sense as the zones can be quite dynamic, so converting this to an `*int` would be better.

controllers/update_pods.go
```diff
@@ -81,13 +82,35 @@ func (updatePods) reconcile(ctx context.Context, r *FoundationDBClusterReconcile
...
 
+	// Map of zones to the number of pods that are unavailable in that zone.
+	zoneToUnavailablePods := make(map[string]int)
+
+	// In order to the maxZonesWithUnavailablePods we need to know the number of zones.
+	// if the fault domain zone count is not set. The number of process groups is used as the zone count.
+	zoneCount := cluster.Spec.FaultDomain.ZoneCount
+	if zoneCount == 0 {
+		zoneCount = len(cluster.Status.ProcessGroups)
+	}
```
johscheuer, line 93: This calculation is not correct. As I mentioned it makes sense to use an integer here instead of trying to calculate the percentage.

controllers/update_pods.go
```diff
@@ -81,13 +82,35 @@ func (updatePods) reconcile(ctx context.Context, r *FoundationDBClusterReconcile
 func getPodsToUpdate(ctx context.Context, logger logr.Logger, reconciler *FoundationDBClusterReconciler, cluster *fdbv1beta2.FoundationDBCluster) (map[string][]*corev1.Pod, error) {
 	updates := make(map[string][]*corev1.Pod)
 
+	// Map of zones to the number of pods that are unavailable in that zone.
+	zoneToUnavailablePods := make(map[string]int)
```
johscheuer, line 86:
```suggestion
	faultDomainsWithUnavailablePods := make(map[string]fdbv1beta2{})
```

We don't need to count the Pods, as we don't care about the number of unavailable Pods but the number of fault domains containing at least one unavailable Pod.

johscheuer (commented):
A few comments around some thing that can be simplified otherwise the changes look good. Have you done any tests besides the unit tests?

manfontan:
I will deploy the changes in dev tomorrow and perform some testing with the latest updates.

api/v1beta2/foundationdbcluster_types.go
```diff
@@ -2528,3 +2533,8 @@ func (cluster *FoundationDBCluster) Validate() error {
 func (cluster *FoundationDBCluster) IsTaintFeatureDisabled() bool {
 	return len(cluster.Spec.AutomationOptions.Replacements.TaintReplacementOptions) == 0
 }
+
+// Get MaxZonesWithUnavailablePods returns the maximum number of zones that can have unavailable pods.
```
johscheuer, line 2537:
```suggestion
// GetMaxZonesWithUnavailablePods returns the maximum number of zones that can have unavailable pods.
```

api/v1beta2/foundationdbcluster_types.go
```diff
@@ -2528,3 +2533,8 @@ func (cluster *FoundationDBCluster) Validate() error {
...
+func (cluster *FoundationDBCluster) GetMaxZonesWithUnavailablePods() int {
+	return pointer.IntDeref(cluster.Spec.MaxZonesWithUnavailablePods, 0)
```
johscheuer, line 2539:
```suggestion
	return pointer.IntDeref(cluster.Spec.MaxZonesWithUnavailablePods, math.MaxInt)
```

If we just set this value to the maximum integer value we don't have to make any special cases around `0`.

controllers/update_pods.go
```diff
@@ -141,6 +156,30 @@ func getPodsToUpdate(ctx context.Context, logger logr.Logger, reconciler *Founda
 			continue
 		}
 
+		if maxZonesWithUnavailablePods > 0 {
```
johscheuer, line 159: When we change the default to `math.MaxInt` we can remove this check.

api/v1beta2/foundationdbcluster_types.go
```diff
@@ -218,6 +218,11 @@ type FoundationDBClusterSpec struct {
 	// UseUnifiedImage determines if we should use the unified image rather than
 	// separate images for the main container and the sidecar container.
 	UseUnifiedImage *bool `json:"useUnifiedImage,omitempty"`
+
+	// MaxZonesWithUnavailablePods defines the maximum number of zones that can have unavailable pods during the update process.
+	// When set to 0 there is no limit on the number of zones with unavailable pods.
+	// +kubebuilder:default:=0
```
johscheuer, line 224:
```suggestion
	// When unset, there is no limit on the number of zones with unavailable pods.
```

johscheuer (commented):
Once the lint issue is fixed we are good 👍 One last proposal on using `math.MaxInt` instead of `0` to represent no limit on zones with unavailable Pods, this will remove the need for any additional checks for `0`.

johscheuer (approved):
LGTM 👍 Once you've verified those changes in dev we can merge them. Thanks!

johscheuer (approved):

Issue #1564, opened 2023-03-30
Add information about fault domain into process group status


johscheuer:
### What would you like to be added/changed?

Add fault domain into process group status. We should also add an additional label like `foundationdb.org/fault-domain` with the fault domain.

PR #1608, opened 2023-05-04
Initial support for fault domains in CRD

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1564

## Type of change

## Discussion

This is currently only used as an additional information source that can be used in addition to the information from the cluster status.

## Testing

Unit tests.

## Documentation

Will be added before merging.

## Follow-up

We could add the fault domain as a label to make the selection of Pods easier.

### DIFF
--- a/api/v1beta2/foundationdbcluster_types.go
+++ b/api/v1beta2/foundationdbcluster_types.go
@@ -330,6 +330,9 @@ type ProcessGroupStatus struct {
 	ExclusionSkipped bool `json:"exclusionSkipped,omitempty"`
 	// ProcessGroupConditions represents a list of degraded conditions that the process group is in.
 	ProcessGroupConditions []*ProcessGroupCondition `json:"processGroupConditions,omitempty"`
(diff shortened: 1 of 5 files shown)


api/v1beta2/foundationdbcluster_types.go
```diff
@@ -329,6 +329,9 @@ type ProcessGroupStatus struct {
 	ExclusionSkipped bool `json:"exclusionSkipped,omitempty"`
 	// ProcessGroupConditions represents a list of degraded conditions that the process group is in.
 	ProcessGroupConditions []*ProcessGroupCondition `json:"processGroupConditions,omitempty"`
+	// FaultDomain represents either the logical grouping that is used to bin pack processes together or the last seen
+	// fault domain from the cluster status.
```
johscheuer, line 333:
```suggestion
	// FaultDomain represents the last seen fault domain from the cluster status.
```

PR #1628, opened 2023-05-09
Test upgrade when no remote processes are restarted

sbodagala:
# Description

Test upgrading an HA cluster when no remote processes are restarted.

Co-authored by: Johannes Scheuermann

## Type of change

- Other (testing related)

## Discussion

No.

## Testing

Manual testing?*

Ran the test locally. (Did four runs today, one of those runs didn't complete after about half-an-hour, I think (at which point I stopped the test), the other runs completed successfully.)

Yes (we will need to run this test as part of regression testing).

## Documentation

N/A

## Follow-up

We will need to investigate any failures of this test (in regression test runs).

No.

### DIFF
--- a/e2e/test_operator_ha_upgrades/operator_ha_upgrade_test.go
+++ b/e2e/test_operator_ha_upgrades/operator_ha_upgrade_test.go
@@ -434,4 +434,49 @@ var _ = Describe("Operator HA Upgrades", Label("e2e"), func() {
 		fixtures.GenerateUpgradeTableEntries(testOptions),
 	)
 
+	DescribeTable(
+		"upgrading a cluster when no remote processes are restarted",
+		func(beforeVersion string, targetVersion string) {
+			isAtLeast := factory.OperatorIsAtLeast(
+				"v1.14.0",
+			)
+
+			if !isAtLeast {
+				Skip("operator doesn't support feature for test case")
+			}
+
+			clusterSetup(beforeVersion, false)
+
+			// Select remote processes and use the buggify option to skip those
+			// processes during the restart command.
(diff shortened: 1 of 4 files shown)


johscheuer (commented):
The code looks good. Could you check the test failure?

sbodagala:
Not sure why the integration checks failed. Will try to run them again.

e2e/test_operator_ha_upgrades/operator_ha_upgrade_test.go
```diff
@@ -434,4 +434,49 @@ var _ = Describe("Operator HA Upgrades", Label("e2e"), func() {
...
+					fdbv1beta2.ProcessGroupID(pod.Labels[fdbCluster.GetRemote().GetCachedCluster().GetProcessGroupIDLabel()]),
+				)
+			}
+
+			log.Println(
+				"Selected Pods:",
+				ignoreDuringRestart,
+				"to be skipped during the restart",
+			)
+			fdbCluster.GetRemote().SetIgnoreDuringRestart(ignoreDuringRestart)
```
johscheuer, line 473: This is probably currently a limitation of the buggify setup for the ignore during restart, but I'm pretty sure this case will only work if the operator in the remote side is doing the restart, as the operator is currently iterating over the process groups from the cluster status (not the machine-readable status). We could change the logic here: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/internal/buggify/buggify.go#L64-L72 to actually iterate over the machine-readable status, to make this feature multi-cluster aware.

We should also call the `SetIgnoreDuringRestart` on all clusters not only the remote side.

Issue #1629, opened 2023-05-10
Add watch for nodes for the operator


johscheuer:
### What would you like to be added/changed?

We should add the watch (https://book.kubebuilder.io/reference/watching-resources/externally-managed.html) functionality for the operator to get updates (and maintain a cache) on nodes. This allows the operator to directly work on changes in the node resources that host at least one Pod of a FoundationDB cluster. The watch functionality should be enabled/disabled via a command line flag to allow user to disable it if they don't want to use the taint feature.

johscheuer:
Follow up for: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1581

PR #1632, opened 2023-05-11
Update taint feature design doc and user manual

xumengpanda:
# Description

Update design doc for the taint feature, because we changed the design during the implementation;
Update the user manual to add the taint feature. 

## Type of change

- Documentation

## Discussion

None

## Testing
I compiled it locally to ensure document is compiled with the right format.

### DIFF
--- a/docs/design/node_taint_eviction.md
+++ b/docs/design/node_taint_eviction.md
@@ -2,36 +2,41 @@
 
 ## Metadata
 
-* Authors: @johscheuer
+* Authors: @johscheuer, @xumengpanda
 * Created: 2023-02-14
-* Updated: 2023-03-08
+* Updated: 2023-05-11
 
 ## Background
 
 The operator is currently able to detect failures of Pods and replace them if required, the replacement is happening based on the [conditions that need replacement](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/v1.14.0/api/v1beta2/foundationdbcluster_types.go#L65).
-One of the missing pieces in failure mitigation is to evict Pods from tainted nodes.
+One of the missing pieces in failure mitigation is to evict Pods from tainted Nodes.
 In Kubernetes it's common to use [taints](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration) to indicate potential problems with a node or planned maintenance.
-This information could be consumed by the operator to proactively replace Pods that are running on nodes with a given set of taints.
+This information could be consumed by the operator to proactively replace Pods that are running on Nodes with a given set of taints.
 
 ## General Design Goals
 
-The goal of this design is to implement a mechanism that allows the operator to replace Pods that are running on nodes with a given set of taints.
-It's expected that the operator will detect the Pods running on tainted nodes in a timely manner, either by receiving an event triggering the reconciliation loop or when running the periodic reconciliation, which will happen every 10h per default.
-The actual replacement of all Pods running on tainted nodes, depends on the data that must be replicated and the number of affected Pods.
+The goal of this design is to implement a mechanism that allows the operator to replace Pods that are running on Nodes with a given set of taints.
+It's expected that the operator will detect the Pods running on tainted Nodes in a timely manner, either by receiving an event triggering the reconciliation loop or by periodic reconciliation events that happen every 10h by default. Once the operator marks Pods on tainted Nodes as `NodeTaintReplacing`, the operator will re-use the replacement logic to replace those Pods. The actual replacement speed depends on the speed of excluding processes, the number of Pods to replace, and the allowed maximum number of inflight replacements (defined by `automationOptions.replacements.maxConcurrentReplacements`).
 
 ## Current Implementation
+The operator does not have this feature. It does not have access to list Node information. It does not consider taint-related events in Pod's condition.
 
-We don't have a current implementation of this and the operator is currently not able to access any node information.
-The closest implementation to this is the [Replace Failed ProcessGroups reconciler](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/controllers/replace_failed_process_groups.go), which could be extended for this.
+The operator does replace failed ProcessGroups in the [Replace Failed ProcessGroups reconciler](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/controllers/replace_failed_process_groups.go). It can extend it to support this feature.
 
 ## Proposed Design
 
-We have to make sure that this feature can be enabled and disabled, e.g. if someone wants to run the operator inside a cluster without node access.
-The `AutomaticReplacementOptions` will be extended to support the configuration of replacements for Pods that run on tainted nodes.
-The new field will be called `taints` and will contain a list, the default would be an empty list, of taint configurations.
-That configuration will include the taint key, this has to be an exact match or a wildcard `*`, and the duration this taint must be present on the node.
-The `durationInSeconds` allow to specify how sensitive the operator should be to taints.
-If a wildcard is specified as key this will be used as default value, except there is an exact match defined for a taint key.
+### Design Requirements
+**Requirement 1.** The feature can be enabled and disabled by users, because (1) when someone wants to run the operator inside a k8s cluster without Node access, they should be able to bypass this feature; (2) when we notice problems during roll-out of this feature, we need to disable it online.
+
+The feature can be disabled by option `automationOptions.replacements.taintReplacementOptions = {}`.
+
+**Requirement 2.** The feature allows users to configure a set of specific taint keys to detect and replace.
+
+Users can configure the taint feature in the new field `taints` inside `automaticReplacementOptions`. The `taints` field is a list of `(key, durationInSeconds)` pairs, which defines operator's behavior on the taint key.
+
+Users can use `automaticReplacementOptions.taintReplacementTimeSeconds` to control how fast operator should replace Pods on tainted Nodes. Users can configure operator to replace Pods on tainted Nodes much faster than the normal replacement by setting a smaller value of `taintReplacementTimeSeconds`.
+
+Example configuration:
 
 ```yaml
 ...
@@ -42,51 +47,37 @@ automaticReplacementOptions:
   enableNodeFailureDetection: true
   taints:
   - key: "example.org/maintenance"
-    durationInSeconds: 7200 # Ensure the taint is present for at least 2 hours before replacing Pods on a node with this taint
-  - key: "*" # The wildcard would allow to define a catch all configuration
-    durationInSeconds: 3600 # Ensure the taint is present for at least 1 hour before replacing Pods on a node with this taint
+    durationInSeconds: 7200 # Mark Pods to be replaced after this taint key has been on the Node for at least 7200 seconds (i.e., 2 hours).
+  - key: "example.org/disconnected"
+    durationInSeconds: 1200 # Mark Pods to be replaced after this taint key has been on the Node for at least 1200 seconds (i.e., 20 mins).
+  - key: "*" # The wildcard will match all taint keys
+    durationInSeconds: 3600 # Mark Pods to be replaced after a taint key, which does not have exact match in the taints list, has been on the Node for at least 3600 seconds.
+  taintReplacementTimeSeconds: 1800 # A Pod marked to be replaced due to taints must stay in the to-be-replaced state for 1800 seconds before it is automatically replaced.
 ```
 
-We will add a new [ProcessGroupConditionType](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/api/v1beta2/foundationdbcluster_types.go#L673), called `RunningOnTaintedNode` to mark process groups that are running on a tainted node.
-In addition we will extend the  [conditions that need replacement](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/v1.14.0/api/v1beta2/foundationdbcluster_types.go#L65) by the newly created `RunningOnTaintedNode` condition.
-Those two steps will make sure that the marked process groups will be eventually replaced by the operator.
-The minimum time until a process group that is running on a newly tainted node will be replaced is `failureDetectionTimeSeconds` (default 2h) + `durationInSeconds` from the taint configuration.
-In the example above this will be 4 hours for a node that gets tainted with `example.org/maintenance`.
+In the above configuration, it takes at least 3000 seconds (1200 + 1800) for the operator to replace a Pod on a Node with taint key `example.com/disconnected`, because it takes `durationInSeconds`, i.e., 1200 seconds, to mark the Pod to be replaced and it takes another `taintReplacementTimeSeconds`, i.e., 1800 seconds, to start replacing the Pod.
 
-The [update_status](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/controllers/update_status.go) subreconciler will be extended to mark process groups running on a tainted node with the new condition.
-This can be achieved by iterating over all Pods and make requests to the node object, one example implementation could look like this:
+**Requirement 3.** The feature lists the taint status for process groups. This allows users to have a holistic view of which Pods in a FDB cluster are running on tainted Nodes and being replaced.
 
-```go
-func checkProcessGroupsRunningOnTaintedNodes(pods []*corev1.Pod, status *fdbv1beta2.FoundationDBClusterStatus) {
-    if !cluster.NodeFailureDetectionEnabled() {
-        return
-    }
+### Implementation
 
-    nodeTaintedMap := internal.GetNodes(pods) // This will generate a map containing the node name as key and a boolean value indicating if the node is "tainted"
-    for node, _ := range nodeMap {
-        // Check if node has a taint for the minimum duration in that case we will set the value to true.
-    }
+We need to make the following four types of changes:
 
-    podMap := internal.CreatePodMap(cluster, pods)
-    for _, processGroup := range status.ProcessGroups {
-        pod, podExists := podMap[processGroup.ProcessGroupID]
-        if !podExists {
-            continue
-        }
+**CRD change.** We need to extend the `automaticReplacementOptions` with the new fields `taints` and `taintReplacementTimeSeconds`
 
-        tainted, nodeFound := nodeTaintedMap[pod.Spec.NodeName]
-        if !nodeFound {
-            continue
-        }
+**Detect Node taint and change ProcessGroup condition.**  We can extend the existing `validateProcessGroup()` function in `update_status` reconciler to detect if a Pod is running on a tainted Node and mark the ProcessGroup's taint-related condition accordingly.
 
-        processGroup.UpdateCondition(fdbv1beta2.RunningOnTaintedNode, tainted, status.ProcessGroups, processGroup.ProcessGroupID)
-    }
-}
-```
+**Status of ProcessGroup taint-related condition.** We will add two [ProcessGroupConditionType](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/api/v1beta2/foundationdbcluster_types.go#L673):
+- `NodeTaintDetected` represents a ProcessGroup Node is tainted, but not long enough to replace it.
+- `NodeTaintReplacing` represents a ProcessGroup whose Node has been tainted for longer-than-configured time and the ProcessGroup should be replaced.
 
-The limitation of this approach would be that we only detect Pods running on tainted nodes when a reconciliation is happening (those are not coupled to node events).
-To work around this we could extend the operator and let the operator watch node events, that could be potentially a high volume of events, so we have to make sure we get the right signals.
-The controller-runtime already supports to let the operator watch resources that are not managed by itself, see [Watching Externally Managed Resources](https://book.kubebuilder.io/reference/watching-resources/externally-managed.html).
+**Replace ProcessGroups with `NodeTaintReplacing` conditioin.** We can extend the existing replacement logic to replace ProcessGroups with `NodeTaintReplacing` condition. We can do this by adding the `NodeTaintReplacing` condition to [`conditionsThatNeedReplacement` list](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/d38d6bc7abf7764215976b75d581c933280389f8/api/v1beta2/foundationdbcluster_types.go#L65-L67)
+
+
+## Limitations
+The operator will only detect tainted Nodes when its reconciliation is triggered. This may increase the delay of detecting and replacing Pods on tainted Nodes.
+
+The limitation can be resolved by letting the operator watch node events. The controller-runtime already supports to let the operator watch resources that are not managed by itself, see [Watching Externally Managed Resources](https://book.kubebuilder.io/reference/watching-resources/externally-managed.html). [Tracked issue#1629](https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1629)
 
 ## Related Links
 
--- a/docs/design/node_taint_eviction.md
+++ b/docs/design/node_taint_eviction.md
@@ -26,9 +26,18 @@ The operator does replace failed ProcessGroups in the [Replace Failed ProcessGro
 ## Proposed Design
 
 ### Design Requirements
-**Requirement 1.** The feature can be enabled and disabled by users, because (1) when someone wants to run the operator inside a k8s cluster without Node access, they should be able to bypass this feature; (2) when we notice problems during roll-out of this feature, we need to disable it online.
+**Requirement 1.** The feature can be enabled and disabled by users, because (1) when someone wants to run the operator inside a k8s cluster without Node access, they should be able to bypass this feature; (2) when we notice problems during roll-out of this feature, we need to disable it.
 
-The feature can be disabled by option `automationOptions.replacements.taintReplacementOptions = {}`.
+The feature can be disabled by option `automationOptions.replacements.taintReplacementOptions = {}`. YAML example is as follows. Please note that `taintReplacementOptions` section is deleted.
+
+```yaml
+...
+automaticReplacementOptions:
+  enabled: true
+  failureDetectionTimeSeconds: 7200
+  maxConcurrentReplacements: 1
+  enableNodeFailureDetection: true
+```
 
 **Requirement 2.** The feature allows users to configure a set of specific taint keys to detect and replace.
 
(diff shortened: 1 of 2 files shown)


docs/changelog/v1.17.0.md
```diff
@@ -31,3 +31,4 @@ Add an upgrade test case where one sidecar is not ready for the upgrade [#1556](
 Update documentation for e2e tests [#1554](https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1554)
 Add e2e test setup for operator [#1534](https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1534)
 adding support for container name in plugin's buggify crashloop command [#1515](https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1515)
+Add taint feature to auto replace tainted node [#1581](https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1581)
```
johscheuer, line 34: Remove this line, the next release will be 1.18. We already release 1.17.

docs/design/node_taint_eviction.md
```diff
@@ -2,36 +2,42 @@
 
 ## Metadata
 
-* Authors: @johscheuer
+* Authors: @johscheuer, @xumengpanda
 * Created: 2023-02-14
 * Updated: 2023-03-08
+* Updated: 2023-05-11
```
johscheuer, line 8: Remove the line above. We only want to keep 1 updated statement.

docs/design/node_taint_eviction.md
```diff
@@ -2,36 +2,42 @@
...
 ## Proposed Design
 
-We have to make sure that this feature can be enabled and disabled, e.g. if someone wants to run the operator inside a cluster without node access.
-The `AutomaticReplacementOptions` will be extended to support the configuration of replacements for Pods that run on tainted nodes.
-The new field will be called `taints` and will contain a list, the default would be an empty list, of taint configurations.
-That configuration will include the taint key, this has to be an exact match or a wildcard `*`, and the duration this taint must be present on the node.
-The `durationInSeconds` allow to specify how sensitive the operator should be to taints.
-If a wildcard is specified as key this will be used as default value, except there is an exact match defined for a taint key.
+### Design Requirements
+**Requirement 1.** The feature can be enabled and disabled by users, because (1) when someone wants to run the operator inside a k8s cluster without Node access, they should be able to bypass this feature; (2) when we notice problems during roll-out of this feature, we need to disable it online.
```
johscheuer, line 30: What is the benefit of changing the layout here? What do you mean by "we need to disable it online"? Why is this a requirement?

docs/design/node_taint_eviction.md
```diff
@@ -2,36 +2,42 @@
...
+
+The feature can be disabled by option `automationOptions.replacements.taintReplacementOptions = {}`.
```
johscheuer, line 32: This is not a requirement but an implementation detail, this should be in the user manual.

docs/design/node_taint_eviction.md
```diff
@@ -2,36 +2,42 @@
...
+
+**Requirement 2.** The feature allows users to configure a set of specific taint keys to detect and replace. This is useful in a multi-tenant k8s environment where FDB operator only wants to react to a subset of taints used in the k8s cluster.
+
+Users can configure the taint feature in the new field `taints` inside `automaticReplacementOptions`. The `taints` field is a list of `(key, durationInSeconds)` pairs, which defines operator's behavior on the taint key.
+
+Users can use `automaticReplacementOptions.taintReplacementTimeSeconds` to control how fast operator should replace Pods on tainted Nodes. Users can ask operator to replace Pods on tainted Nodes much faster than the normal replacement by setting a smaller value of `taintReplacementTimeSeconds`.
```
johscheuer, line 38: `Users can ask operator to replace Pods on tainted Nodes`, how does a user "asks" the operator to replace Pods on a tainted node? Can we stick to a more technical language? I would prefer the word "configure" and not "ask".

docs/design/node_taint_eviction.md
```diff
@@ -42,51 +48,42 @@ automaticReplacementOptions:
   enableNodeFailureDetection: true
   taints:
   - key: "example.org/maintenance"
-    durationInSeconds: 7200 # Ensure the taint is present for at least 2 hours before replacing Pods on a node with this taint
-  - key: "*" # The wildcard would allow to define a catch all configuration
-    durationInSeconds: 3600 # Ensure the taint is present for at least 1 hour before replacing Pods on a node with this taint
+    durationInSeconds: 7200 # Mark Pods to be replaced after this taint key has been on the Node for at least 7200 seconds (i.e., 2 hours).
```
johscheuer, line 51: What is the benefit here with those changes? Adding `taintReplacementTimeSeconds` makes sense, but I'm not sure about the value of adding yet another example?

docs/design/node_taint_eviction.md
```diff
@@ -42,51 +48,42 @@ automaticReplacementOptions:
...
-
-        processGroup.UpdateCondition(fdbv1beta2.RunningOnTaintedNode, tainted, status.ProcessGroups, processGroup.ProcessGroupID)
-    }
-}
+In the above configuration, it takes at least 3000 seconds (1200 + 1800) for the operator to replace a Pod on a Node with taint key `example.com/disconnected`, because it takes `durationInSeconds`, i.e., 1200 seconds, to mark the Pod to be replaced and it takes another `taintReplacementTimeSeconds`, i.e., 1800 seconds, to start replacing the Pod.
+
+**Requirement 3.** The feature lists the taint status for process groups. This allows users to have a holistic view of which Pods in a FDB cluster are running on tainted Nodes and being replaced.
+
+### Implementation
+We need to make the following four type of changes:
```
johscheuer, line 64:
```suggestion
### Implementation

We need to make the following four type of changes:
```

docs/design/node_taint_eviction.md
```diff
@@ -42,51 +48,42 @@ automaticReplacementOptions:
...
+
+**CRD change.** We need to extend the `automaticReplacementOptions` with the new fields `taints` and `taintReplacementTimeSeconds`
```
johscheuer, line 66: This information is already implicitly given as we stated above the we are extending the `AutomaticReplacementOptions`.

docs/design/node_taint_eviction.md
```diff
@@ -42,51 +48,42 @@ automaticReplacementOptions:
...
+
+**Detect Node taint and change Pod condition.**  We can piggy-back the existing `validateProcessGroup()` function in `update_status` reconciler to detect if a Pod is running on a tainted Node and mark the Pod's taint-related condition accordingly.
```
johscheuer, line 68:
```suggestion
**Detect Node taint and change ProcessGroup condition.**  We can extend the existing `validateProcessGroup()` function in `update_status` reconciler to detect if a Pod is running on a tainted Node and mark the ProcessGroup taint-related condition accordingly.
```

We are not changing any Pod condition. We change the `ProcessGroup` condition, those are different things.

I'm not sure what "type" of change this is but I would have expected something like "code changes", "CRD changes", "documentation changes" etc. as "type" of change.

docs/design/node_taint_eviction.md
```diff
@@ -42,51 +48,42 @@ automaticReplacementOptions:
...
+
+A Pod's Node can be listed by
 ```
+node = &corev1.Node{}
+err := r.Get(ctx, client.ObjectKey{Name: pod.Spec.NodeName}, node)
```
johscheuer, line 73: What is the value with this statement? This is a basic example for how to use the controller runtime, can we remove this? What is the value of adding this example here?

docs/design/node_taint_eviction.md
```diff
@@ -42,51 +48,42 @@ automaticReplacementOptions:
...
+```
+
+**Status of Pod taint-related condition.** We will add two  [ProcessGroupConditionType](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/api/v1beta2/foundationdbcluster_types.go#L673): `NodeTaintDetected`, which represents a Pod's Node is tainted but not long enough for operator to replace it, and `NodeTaintReplacing`, which represents a Pod whose Node has been tainted for longer-than-configured time and the operator should replace it.
```
johscheuer, line 76:
```suggestion
**Status of ProcessGroup taint-related condition.** We will add two [ProcessGroupConditionType](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/api/v1beta2/foundationdbcluster_types.go#L673):

- `NodeTaintDetected` represents a ProcessGroup Node is tainted, but not long enough to replace it.
- `NodeTaintReplacing` represents a ProcessGroup whose Node has been tainted for longer-than-configured time and the ProcessGroup should be replaced.
```

Same as above in this context we are talking about `ProcessGroup`s and not `Pod`s. I personally would prefer a listing instead of a long sentence.

docs/design/node_taint_eviction.md
```diff
@@ -42,51 +48,42 @@ automaticReplacementOptions:
...
+
+**Replace Pods with `NodeTaintReplacing` conditioin.** We can piggy-back existing replacement logic to replace Pods with `NodeTaintReplacing` condition. We can do this by adding `NodeTaintReplacing` to [`conditionsThatNeedReplacement` list](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/d38d6bc7abf7764215976b75d581c933280389f8/api/v1beta2/foundationdbcluster_types.go#L65-L67)
```
johscheuer, line 78:
```suggestion
**Replace ProcessGroup with `NodeTaintReplacing` condition.** We can extend the existing replacement logic to replace ProcessGroups with `NodeTaintReplacing` condition. We can do this by adding the `NodeTaintReplacing` condition to [`conditionsThatNeedReplacement` list](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/d38d6bc7abf7764215976b75d581c933280389f8/api/v1beta2/foundationdbcluster_types.go#L65-L67)
```

Same as above in this context we are talking about ProcessGroups and not Pods.

docs/design/node_taint_eviction.md
```diff
@@ -42,51 +48,42 @@ automaticReplacementOptions:
...
+
+
+## Limitations
+The operator will only detect tainted Nodes when its reconciliation is triggered periodically. This may increase the delay of detecting and replacing Pods on tainted Nodes.
```
johscheuer, line 82:
```suggestion
## Limitations

The operator will only detect tainted Nodes when its reconciliation is triggered. This may increase the delay of detecting and replacing Pods on tainted Nodes.
```

The periodically statement is wrong here. The operator will also (potentially) detected this condition if a reconciliation was triggered by an unrelated event, saying that the tainted nodes are only detected by the periodic reconciliation is wrong. We can add a statement here that node changes are not triggering any reconciliation.

docs/design/node_taint_eviction.md
```diff
@@ -42,51 +48,42 @@ automaticReplacementOptions:
...
+
+The limitation can be resolved by letting the operator watch node events. The controller-runtime already supports to let the operator watch resources that are not managed by itself, see [Watching Externally Managed Resources](https://book.kubebuilder.io/reference/watching-resources/externally-managed.html).
```
johscheuer, line 84:
```suggestion
The limitation can be resolved by letting the operator watch node events. The controller-runtime already supports to let the operator watch resources that are not managed by itself, see [Watching Externally Managed Resources](https://book.kubebuilder.io/reference/watching-resources/externally-managed.html).
```

You could also reference: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1629, so users can see if the feature is implemented.

docs/design/node_taint_eviction.md
```diff
@@ -42,51 +48,42 @@ automaticReplacementOptions:
...
 
-The limitation of this approach would be that we only detect Pods running on tainted nodes when a reconciliation is happening (those are not coupled to node events).
-To work around this we could extend the operator and let the operator watch node events, that could be potentially a high volume of events, so we have to make sure we get the right signals.
-The controller-runtime already supports to let the operator watch resources that are not managed by itself, see [Watching Externally Managed Resources](https://book.kubebuilder.io/reference/watching-resources/externally-managed.html).
+Because a k8s cluster may have a high volume of node events, we must make sure we get the right signals and not overload the operator by excessive events.
```
johscheuer, line 86: This is a limitation? I would rather see this as a requirement. Is this statement part of the sentence above?

docs/manual/replacements_and_deletions.md
```diff
@@ -27,7 +27,7 @@ The following changes can only be rolled out through replacement:
 The number of inflight replacements can be configured by setting `maxConcurrentReplacements`, per default the operator will replace all misconfigured process groups.
 Depending on the cluster size this can require a quota that is has double the capacity of the actual required resources.
 
-## Automatic Replacements
+## Automatic Replacements for Pods in Bad State
```
johscheuer, line 30:
```suggestion
## Automatic Replacements for ProcessGroups in Bad State
```

Please use the correct term in the documentation, the operator acts on ProcessGroups and the Pods are only one part of it. What is the benefit of adding `for ProcessGroups in Bad State`? What else should the operator automatically be replacing?

docs/manual/replacements_and_deletions.md
```diff
@@ -42,11 +42,45 @@ The following conditions are currently eligible for replacement:
 * `MissingPod`: This indicates a process group that doesn't have a Pod assigned.
 * `MissingPVC`: This indicates that a process group that doesn't have a PVC assigned.
 * `MissingService`: This indicates that a process group that doesn't have a Service assigned.
-* `PodPending`: This indicates that a process group where the pod is in a pending state.
+* `PodPending`: This indicates that a process group where the Pod is in a pending state.
+* `NodeTaintReplacing`: This indicates that a process group where the pod is running on a tainted Node
```
johscheuer, line 46:
```suggestion
* `NodeTaintReplacing`: This indicates that a process group where the pod is running on a tainted Node.
```

This is only half true, if the node is not tainted anymore the ProcessGroup will be replaced anyway (at least in the current implementation) if the `NodeTaintReplacing` condition is set. We should add this details here. The implication there is that if a ProcessGroup has the `NodeTaintReplacing` condition the replacement cannot be stopped, even if the node was fixed.

docs/manual/replacements_and_deletions.md
```diff
@@ -42,11 +42,45 @@ The following conditions are currently eligible for replacement:
...
 
 Process groups that are set into the crash loop state with the `Buggify` setting won't be replaced by the operator.
 If the `cluster.Spec.Buggify.EmptyMonitorConf` setting is active the operator won't replace any process groups.
 
+## Automatic Replacements for Pods on Tainted Nodes
+The operator has an option to automatically replace Pods that are running on tainted Nodes. This feature is disabled by default but can be enabled by setting `automationOptions.replacements.taintReplacementOptions` and `automationOptions.replacements.taintReplacementTimeSeconds`.
```
johscheuer, line 52:
```suggestion
## Automatic Replacements for Pods on Tainted Nodes

The operator has an option to automatically replace ProcessGroups where the associated Pod is running on a tainted Node. This feature is disabled by default, but can be enabled by setting `automationOptions.replacements.taintReplacementOptions`.
```

`automationOptions.replacements.taintReplacementTimeSeconds` is actually an optional setting. We should remove this here.

docs/manual/replacements_and_deletions.md
```diff
@@ -42,11 +42,45 @@ The following conditions are currently eligible for replacement:
...
+
+**Example 1.** We can use the following setup to let operator detect Pods running on Nodes with taint key `example.com/maintenance`, set the Pods' condition to `NodeTaintReplacing` if their Nodes have been tainted for 3600 seconds, and replace the Pods after 1800 seconds.
```
johscheuer, line 54:
```suggestion
### Example 1

The following setup lets the operator detect Pods running on Nodes with taint key `example.com/maintenance`, set the ProcessGroup' condition to `NodeTaintReplacing`, if their Nodes have been tainted for 3600 seconds, and replace the Pods after 1800 seconds.
```

Can we use proper markdown syntax and maybe a better/more precise heading than `example 1`? I personally find it a little bit hard to read this sentence.

johscheuer (changes requested):

docs/design/node_taint_eviction.md
```diff
@@ -2,36 +2,42 @@
...
+* Authors: @johscheuer, @xumengpanda
 * Created: 2023-02-14
 * Updated: 2023-03-08
+* Updated: 2023-05-11
 
 ## Background
 
 The operator is currently able to detect failures of Pods and replace them if required, the replacement is happening based on the [conditions that need replacement](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/v1.14.0/api/v1beta2/foundationdbcluster_types.go#L65).
-One of the missing pieces in failure mitigation is to evict Pods from tainted nodes.
+One of the missing pieces in failure mitigation is to evict Pods from tainted Nodes.
```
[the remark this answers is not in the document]
xumengpanda: It's the same reason of using uppercase for Pods.

docs/design/node_taint_eviction.md
```diff
@@ -2,36 +2,42 @@
...
-The actual replacement of all Pods running on tainted nodes, depends on the data that must be replicated and the number of affected Pods.
+The goal of this design is to implement a mechanism that allows the operator to replace Pods that are running on Nodes with a given set of taints.
+It's expected that the operator will detect the Pods running on tainted Nodes in a timely manner, either by receiving an event triggering the reconciliation loop or by periodic reconciliation events that happen every 10h by default. Once the operator marks Pods on tainted Nodes as to-be-replaced, the operator will re-use the replacement logic to replace those Pods. The actual replacement speed depends on the speed of excluding a Pods, the number of Pods to replace, and the allowed maximum number of inflight replacements (defined by `automationOptions.replacements.maxConcurrentReplacements`).
 
 ## Current Implementation
+The operator does not have this feature. It does not have access to list Node information. It does not consider taint-related events in Pod's condition.
```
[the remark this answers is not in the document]
xumengpanda: I think this is clearer and simpler.

docs/design/node_taint_eviction.md
```diff
@@ -2,36 +2,42 @@
...
 
-We don't have a current implementation of this and the operator is currently not able to access any node information.
-The closest implementation to this is the [Replace Failed ProcessGroups reconciler](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/controllers/replace_failed_process_groups.go), which could be extended for this.
+The operator does replace failed ProcessGroups in the [Replace Failed ProcessGroups reconciler](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/controllers/replace_failed_process_groups.go). It can extend it to support this feature.
```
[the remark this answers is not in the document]
xumengpanda: I think the changed version is clearer and simpler.

docs/design/node_taint_eviction.md
xumengpanda: we need to be able to disable it online (ie., on-the-fly)

docs/design/node_taint_eviction.md
xumengpanda: This is the API definition/design for users.
During our discussion in implementation PR, we debated on "should we support using {key=*, duration=-1}" to disable the entire taint feature.  If the design defines it clearly,  the implementation will have less ambiguity.

docs/design/node_taint_eviction.md
```diff
@@ -2,36 +2,42 @@
...
-The new field will be called `taints` and will contain a list, the default would be an empty list, of taint configurations.
-That configuration will include the taint key, this has to be an exact match or a wildcard `*`, and the duration this taint must be present on the node.
-The `durationInSeconds` allow to specify how sensitive the operator should be to taints.
-If a wildcard is specified as key this will be used as default value, except there is an exact match defined for a taint key.
+### Design Requirements
+**Requirement 1.** The feature can be enabled and disabled by users, because (1) when someone wants to run the operator inside a k8s cluster without Node access, they should be able to bypass this feature; (2) when we notice problems during roll-out of this feature, we need to disable it online.
+
+The feature can be disabled by option `automationOptions.replacements.taintReplacementOptions = {}`.
+
+**Requirement 2.** The feature allows users to configure a set of specific taint keys to detect and replace. This is useful in a multi-tenant k8s environment where FDB operator only wants to react to a subset of taints used in the k8s cluster.
```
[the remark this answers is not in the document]
xumengpanda: :) Sure.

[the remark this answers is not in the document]
xumengpanda: I was thinking about k8s or C* team may use taint for their own purpose and we don't want to replace Pods for that particular taint.

docs/design/node_taint_eviction.md
xumengpanda: it's to show we can add a list of different taint keys and wildcard taint key.

docs/design/node_taint_eviction.md
xumengpanda: this is to echo the design requirement.

docs/design/node_taint_eviction.md
xumengpanda: This item is a procedure, which is like algorithm. 
The CRD changes are like data-structure change.

docs/design/node_taint_eviction.md
xumengpanda: yes

xumengpanda: removed this sentecne

docs/manual/replacements_and_deletions.md
xumengpanda: It's to distinguish from Replacements for Pods on Tainted Nodes.

docs/manual/replacements_and_deletions.md
```diff
@@ -42,11 +42,45 @@ The following conditions are currently eligible for replacement:
...
+```
+automationOptions.replacements.taintReplacementOptions = {
+    {
+        Key: example.com/maintenance
+        DurationInSeconds: 3600
+    },
+}
+automationOptions.replacements.taintReplacementTimeSeconds = 1800
+```
```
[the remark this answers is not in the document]
xumengpanda: changed it to yaml format

docs/manual/replacements_and_deletions.md
```diff
@@ -42,11 +42,45 @@ The following conditions are currently eligible for replacement:
...
+
+If there are multiple Pods on tainted Nodes, the operator will simultaneously replace at most `automationOptions.replacements.maxConcurrentReplacements` Pods.
```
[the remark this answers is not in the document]
xumengpanda: the sentence above is in a parallel section "Automatic Replacements for ProcessGroups in Bad State". 
Having this sentence here is to explicitly state the option is also in effect for "Automatic Replacements for ProcessGroups on Tainted Nodes".
It helps to avoid ambiguity to readers.

docs/manual/replacements_and_deletions.md
```diff
@@ -42,11 +42,45 @@ The following conditions are currently eligible for replacement:
...
+
+**Example 2.** We can enable taint feature on all taint keys except one taint key with the following configuration:
```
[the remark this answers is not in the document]
xumengpanda: Changed it to 
```

We use three examples below to illustrate how to set up the feature.

### Example Setup 1
...
### Example Setup 2
...
### Example Setup 3
...
```

docs/manual/replacements_and_deletions.md
```diff
@@ -42,11 +42,45 @@ The following conditions are currently eligible for replacement:
...
+		Key:               *,
+		DurationInSeconds: 3600,
+	},
+	{
+		Key:               example.com/taint-key-to-ignore,
+		DurationInSeconds: 9223372036854775807,
+	},
+}
+```
+The operator will detect and mark all Pods on tainted Nodes with `NodeTaintDetected` condition. But the operator will not mark Pods on Nodes with taint key `example.com/taint-key-to-ignore` with `NodeTaintReplacing` condition because the key's `DurationInSeconds` is set to max of int64.
```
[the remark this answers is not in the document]
xumengpanda: `
The operator will detect and mark all Pods on tainted Nodes with `NodeTaintDetected` condition. But the operator will ignore the taint key `example.com/taint-key-to-ignore` when it adds `NodeTaintReplacing` condition to Pods, because the key's `DurationInSeconds` is set to max of int64. For example, if a Node has only the taint key `example.com/taint-key-to-ignore`, its Pods will only be marked with  `NodeTaintDetected` condition. When the Node has another taint key, say `example.com/any-other-key`, its Pods will be added `NodeTaintReplacing` condition when the other taint key has been on the Node for 3600 seconds.
`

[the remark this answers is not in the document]
xumengpanda: changed to this ^^^

docs/design/node_taint_eviction.md
johscheuer: Not sure why this has to be "online" and a redeployment of the operator with the according settings to disable wouldn't be good enough.

docs/design/node_taint_eviction.md
johscheuer: If this is intended for users we should make this example in `yaml` in the context of the `FoundationDBClusterSpec` and not a one line go code example as most (if not all) users will be creating and managing clusters using yaml files with the `FoundationDBCluster` definition.

docs/design/node_taint_eviction.md
johscheuer: ok.

johscheuer (approved):
👍

docs/design/node_taint_eviction.md
xumengpanda: I see. redeployment also works. let me remove `online`.

docs/design/node_taint_eviction.md
xumengpanda: add a yaml example

docs/manual/replacements_and_deletions.md
```diff
@@ -27,7 +27,7 @@ The following changes can only be rolled out through replacement:
 The number of inflight replacements can be configured by setting `maxConcurrentReplacements`, per default the operator will replace all misconfigured process groups.
 Depending on the cluster size this can require a quota that is has double the capacity of the actual required resources.
 
-## Automatic Replacements
+## Automatic Replacements for ProcessGroups in Bad State
```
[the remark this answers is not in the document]
xumengpanda: I changed it as well.

xumengpanda:
This is a documentation change. The CI PR builder failure is running tests against operator code, which is irrelevant to the PR change.

PR #1633, opened 2023-05-16
Make sure we only set the ResourceTerminating condition if a process group is marked for removal and is fully excluded

johscheuer:
# Description

Ensure we set the `ResourcesTerminating` condition only if the process group is marked for removal and is fully excluded.
I also removed some of the old code to make things easier to understand and remove some of the old complexity that was not needed.

## Type of change

## Discussion

-

## Testing

Added unit tests and e2e test will be running.

## Documentation

-

## Follow-up

-


johscheuer:
> * Commit ID: [644f11f](https://github.com/FoundationDB/fdb-kubernetes-operator/commit/644f11fcae82e3450d2c692029f9336fde2a517e)
> * Duration 3:48:09
> * Result: ❌ **FAILED**
> * [Build Log](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator/fdb-kubernetes-operator-pr/1912f67b-0f13-4ab1-92ee-bdb8113ca475.gz) terminal output (available for 30 days)
> * [Build Workspace](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator-pr/1912f67b-0f13-4ab1-92ee-bdb8113ca475/workspace.zip) zip file of the working directory (available for 30 days)

```text

Summarizing 1 Failure:
  [FAIL] Operator Upgrades upgrading a cluster with a crash looping sidecar process [It] Upgrade from 6.3.25 to 7.1.31 with crash-looping sidecar [e2e]
  /codebuild/output/src392374028/src/github.com/FoundationDB/fdb-kubernetes-operator/e2e/test_operator_upgrades/operator_upgrades_test.go:88
```

api/v1beta2/foundationdbcluster_types.go
```diff
@@ -576,15 +568,14 @@ func (processGroupStatus *ProcessGroupStatus) addCondition(oldProcessGroups []*P
...
-		for _, condition := range oldProcessGroupStatus.ProcessGroupConditions {
-			if condition.ProcessGroupConditionType == conditionType {
-				// Assumption: processGroupStatus doesn't have the conditionType
-				processGroupStatus.ProcessGroupConditions = append(processGroupStatus.ProcessGroupConditions, condition)
-				return
-			}
+	// If a process group is marked for removal and is fully excluded we only keep the ResourcesTerminating condition.
+	if processGroupStatus.IsMarkedForRemoval() && processGroupStatus.IsExcluded() {
+		if conditionType != ResourcesTerminating {
+			return
```
sbodagala, line 574: Add a log statement here explaining why the condition is not being set?

api/v1beta2/foundationdbcluster_types.go
```diff
@@ -576,15 +568,14 @@ func (processGroupStatus *ProcessGroupStatus) addCondition(oldProcessGroups []*P
...
 		}
+
+		// Reset all previous conditions
```
sbodagala, line 577: This is a bit odd. The user is asking a condition to be set, instead this function is clear all existing conditions. Maybe an argument needs to be added so user can specify whether to clear these conditions or not? And, if clearing these conditions is important why is the user not explicitly clearing them, instead of doing that as a side-effect of calling this function?

api/v1beta2/foundationdbcluster_types.go
johscheuer: That sounds good.

api/v1beta2/foundationdbcluster_types.go
johscheuer: Good point and I agree with the statement that this side-effect might be surprising. I will create a dedicated method to clean conditions if a process group is marked for removal 👍

api/v1beta2/foundationdbcluster_types.go
johscheuer: I haven't added a log statement here as this would probably noisy and we would only want to log this at the debug level and we would need to pass down the logger from all calls.

sbodagala (approved):

Issue #1612, opened 2023-05-04
Rethink the current usage of the cache and map mapping for Pods and other resources


johscheuer:
### What would you like to be added/changed?

Currently we have `internal.CreatePodMap` to create a map of proceessGroupIDs to Pods:

```go
// CreatePodMap creates a map with the process group ID as a key and the according Pod as a value
func CreatePodMap(cluster *fdbv1beta2.FoundationDBCluster, pods []*corev1.Pod) map[fdbv1beta2.ProcessGroupID]*corev1.Pod {
	podProcessGroupMap := make(map[fdbv1beta2.ProcessGroupID]*corev1.Pod, len(pods))
	for _, pod := range pods {
		processGroupID := GetProcessGroupIDFromMeta(cluster, pod.ObjectMeta)
		if processGroupID == "" {
			continue
		}
		podProcessGroupMap[processGroupID] = pod
	}

	return podProcessGroupMap
}
```

We use this method in multiple places and subreconcilers and we already have a few method to get the Pod name from the processGroupID and vice versa, or from the processClass and the ID number of the process group.

The controller runtime brings per default a cache and we added an indexer for fetching the resources efficiently from the cache by name: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/controllers/cluster_controller.go#L197-L216.

We should try to remove the internal map creation and directly get the resource name (as they all follow a specific pattern), this would possibly improve the efficiency of the operator and could reduce the resource requirements for the operator. We have to verify which metrics we should use to compare the improvements.

The idea is to change calls like this:

```go
		pod, podExists := podMap[processGroup.ProcessGroupID]
```

to something like this:

```go
		pod := &corev1.Pod{}
		err := r.Get(ctx, client.ObjectKey{Namespace: cluster.Namespace, Name: processGroup.GetPodName()},pod)
		if err != nil {
			if k8serrors.IsNotFound(err) {
				podExists = false
			} else { 
				// error handling
			}
		}
		
```

Issue #1619, opened 2023-05-05
Revisit the usage of ParseProcessGroupID


johscheuer:
### What would you like to be added/changed?

We have multiple places where we parse the process group ID only to get the id number (the last digit of the ID), the code looks like this:

```go

			_, idNum, err := podmanager.ParseProcessGroupID(processGroup.ProcessGroupID)
			if err != nil {
				return &requeue{curError: err}
			}
```

I don't see any place where we actually use the returned prefix, so it might be more efficient for us to build a new method on the `type ProcessGroupID string` that directly returns the ID number. We should also consider doing some string parsing instead of using a regex as this will be probably more efficient too.

johscheuer:
I think as a general guideline it makes sense to move things into the process group struct and away from the pod manager pkg.

PR #1637, opened 2023-05-17
Remove the internal Pod map and use the controller runtime cache

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1612

For large clusters it will be more efficient to use the controller-runtime cache directly instead of doing a list and then building our own cache on top of the controller-runtime cache.

## Type of change

## Discussion

This is a breaking change as we extend the `PodLifecycleManager` interface.

## Testing

Adjusted unit tests and e2e tests will be running.

## Documentation

-

## Follow-up

Could be: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1619 to make sure we reduce the number of conversions and parsing in our operator.

### DIFF
--- a/controllers/replace_failed_process_groups_test.go
+++ b/controllers/replace_failed_process_groups_test.go
@@ -74,7 +74,7 @@ var _ = Describe("replace_failed_process_groups", func() {
 		taintKeyStarDuration := int64(20)
 		taintKeyMaintenance := "foundationdb.org/maintenance"
 		taintKeyMaintenanceDuration := int64(10)
-		var allPods []*corev1.Pod
+
 		var allPvcs *corev1.PersistentVolumeClaimList
 		var podOnTaintedNode *corev1.Pod
 		var targetPodProcessGroupID fdbv1beta2.ProcessGroupID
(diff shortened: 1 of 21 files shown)


controllers/replace_failed_process_groups_test.go
```diff
@@ -139,7 +136,7 @@ var _ = Describe("replace_failed_process_groups", func() {
 			targetPodProcessGroupID = internal.GetProcessGroupIDFromMeta(cluster, podOnTaintedNode.ObjectMeta)
 
 			// Call validateProcessGroups to set processGroupStatus to tainted condition
-			processGroupsStatus, err := validateProcessGroups(ctx.TODO(), clusterReconciler, cluster, &cluster.Status, processMap, configMap, allPods, allPvcs, logger)
+			processGroupsStatus, err := validateProcessGroups(ctx.TODO(), clusterReconciler, cluster, &cluster.Status, processMap, configMap, allPvcs, logger)
```
sbodagala, line 139: Why using "allPvcs" instead of "allPods"?

sbodagala: Same question for the changes below.

internal/pod_helper.go
```diff
@@ -226,20 +226,6 @@ func GetSidecarImage(cluster *fdbv1beta2.FoundationDBCluster, pClass fdbv1beta2.
 	return GetImage(image, imageConfigs, cluster.Spec.Version, false)
 }
 
-// CreatePodMap creates a map with the process group ID as a key and the according Pod as a value
-func CreatePodMap(cluster *fdbv1beta2.FoundationDBCluster, pods []*corev1.Pod) map[fdbv1beta2.ProcessGroupID]*corev1.Pod {
```
sbodagala, line 230: This map needed to be built by all controllers/reconcilers (that is, all controllers that needed to use this map)? Why not build it once (and update the map where needed)?

sbodagala:
If there is already a controller run-time cache it makes sense to use that rather than building another index with the same functionality.

sbodagala (approved):

controllers/replace_failed_process_groups_test.go
johscheuer: This is only needed until we fetch the pvcs directly from the cache similar to what we did in this PR for Pods. We still validate all Pods but the Pod is now fetched from the controller-runtime cache, so we don't have to pass down the Pod map. Created: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1638 as follow up.

PR #1643, opened 2023-05-23
Add the additional processes only once and not multiple times

johscheuer:
# Description

This is a small bug fix to add the additional processes only once. The logic was working before, because we use a map, so we added the same process multiple times into the map with the same key.

## Type of change

## Discussion

-

## Testing

Unit test.

## Documentation

-

## Follow-up

-

### DIFF
--- a/pkg/fdbadminclient/mock/admin_client_mock.go
+++ b/pkg/fdbadminclient/mock/admin_client_mock.go
@@ -230,7 +230,7 @@ func (client *AdminClient) GetStatus() (*fdbv1beta2.FoundationDBStatus, error) {
 			}
 
 			var uptimeSeconds float64 = 60000
-			if client.MaintenanceZone == pod.Name || client.MaintenanceZone == "simulation" {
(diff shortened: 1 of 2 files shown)


pkg/fdbadminclient/mock/admin_client_mock.go
```diff
@@ -274,34 +274,34 @@ func (client *AdminClient) GetStatus() (*fdbv1beta2.FoundationDBStatus, error) {
...
 			}
 		}
+	}
 
-		for _, processGroup := range client.additionalProcesses {
-			locality := map[string]string{
-				fdbv1beta2.FDBLocalityInstanceIDKey: string(processGroup.ProcessGroupID),
-				fdbv1beta2.FDBLocalityZoneIDKey:     string(processGroup.ProcessGroupID),
-			}
+	for _, processGroup := range client.additionalProcesses {
```
johscheuer, line 279: The diff seems a little bit off, I basically moved the for loop `for _, processGroup := range client.additionalProcesses` out of the for loop over the `Pods`.

sbodagala (approved):

PR #1647, opened 2023-05-24
Update maintenance mode information in the update_status controller

sbodagala:
# Description

Make "update_status" reconciler to also update maintenance mode information in cluster status.

## Type of change

- Other (re-organizes code; likely we will need to revisit the logic in the maintenance_mode reconciler)

## Discussion

As mentioned above, it is likely we will need to revisit the logic in the maintenance_mode reconciler.

## Testing

Manual testing?*

CI will run the operator unit tests.

Yes, nightly operator regression tests.

## Documentation

No.

Not sure (this PR is providing the functionality that is needed in order to implement the CanI sub-system).

No (maybe at a later point).

No (maybe at a later point).

## Follow-up

As mentioned above, we will likely need to revisit the logic in the maintenance_mode reconciler.

Not sure.

### DIFF
--- a/controllers/update_status.go
+++ b/controllers/update_status.go
@@ -264,12 +264,6 @@ func (updateStatus) reconcile(ctx context.Context, r *FoundationDBClusterReconci
 		return status.ProcessGroups[i].ProcessGroupID < status.ProcessGroups[j].ProcessGroupID
 	})
 
-	// Update maintenance mode information in "status".
-	err = updateMaintenanceModeInfo(r, cluster, &status)
-	if err != nil {
-		return &requeue{curError: err}
-	}
-
 	cluster.Status = status
 
 	_, err = cluster.CheckReconciliation(log)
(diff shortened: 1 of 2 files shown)


controllers/update_status.go
```diff
@@ -869,3 +875,34 @@ func updateFaultDomains(logger logr.Logger, processes map[fdbv1beta2.ProcessGrou
 		status.ProcessGroups[idx].FaultDomain = faultDomain
 	}
 }
+
+func updateMaintenanceModeInfo(r *FoundationDBClusterReconciler, cluster *fdbv1beta2.FoundationDBCluster, status *fdbv1beta2.FoundationDBClusterStatus) error {
```
johscheuer, line 879: I don't think we need an additional method here and we don't need the differentiation between those two cases (set by this operator or another operator). All we need is adding a single line in the update_status reconciler (here: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/controllers/update_status.go#L135-L153) to add the information:


```go
status.MaintenanceModeInfo.ZoneID = databaseStatus.Cluster.MaintenanceZone
```

Could you add a unit tests to make sure this field is set? The status mock client has a method to simulate the maintenance mode: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/pkg/fdbadminclient/mock/admin_client_mock.go#L783-L787. We should also add an e2e test for this to make sure the operator picks up the value in a real cluster too.

johscheuer (changes requested):

controllers/update_status.go
sbodagala: Ok, that change makes sense. (As I mentioned to you yesterday, I also wasn't sure about how the code I copied would help/didn't fully understand the overall plan.)

Regarding the unit test: the method you mentioned only sets the maintenance information on the client, how would that help with writing a unit test? (I am still trying to understand the operator unit test framework, how the state/status gets set without setting up a foundationdb cluster.)

johscheuer (approved):
LGTM 👍 Thanks!

Issue #1648, opened 2023-05-25
Add e2e test case for maintenance mode


johscheuer:
### What would you like to be added/changed?

The operator supports to use the maintenance mode before doing any operation on Pods, like doing an update. This feature is currently not e2e tested, so we have to add test cases to prove the feature behaves as expected.

PR #1649, opened 2023-05-25
Simplify maintenance mode checker logic

johscheuer:
# Description

With https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1647 we can remove some complexity in the maintenance mode checker that is not required anymore.

## Type of change

- Other (refactoring)

## Discussion

-

## Testing

Unit tests. We have to add an e2e test for this mode. I'll create a GitHub issue to add a new test for this mode.

## Documentation

-

## Follow-up

https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1648

### DIFF
--- a/controllers/maintenance_mode_checker.go
+++ b/controllers/maintenance_mode_checker.go
@@ -23,8 +23,6 @@ package controllers
 import (
 	"context"
 	"fmt"
-	"time"
-
 	fdbv1beta2 "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta2"
 )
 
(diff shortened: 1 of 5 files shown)


controllers/update_pods.go
```diff
@@ -249,22 +249,10 @@ func deletePodsForUpdates(ctx context.Context, r *FoundationDBClusterReconciler,
 		}
 	}
 
+	// TODO(jscheuermann): If a maintenance zone is already set, we should allow to delete Pods in this zone.
```
johscheuer, line 252: I can create issues as a follow up for those todos.

controllers/update_pods.go
```diff
@@ -249,22 +249,10 @@ func deletePodsForUpdates(ctx context.Context, r *FoundationDBClusterReconciler,
...
-		cluster.Status.MaintenanceModeInfo = fdbv1beta2.MaintenanceModeInfo{
-			StartTimestamp: &metav1.Time{Time: time.Now()},
-			ZoneID:         zone,
-			ProcessGroups:  processGroups,
-		}
-		err = r.updateOrApply(ctx, cluster)
-		if err != nil {
-			return &requeue{curError: err}
-		}
+
```
johscheuer, line 255: We should add a check here to only set the maintenance mode if at least 1 storage process is included in the list. Otherwise the maintenance mode will be useless, as it only works on storage processes.

controllers/update_pods.go
```diff
@@ -209,6 +208,7 @@ func getPodsToDelete(deletionMode fdbv1beta2.PodUpdateMode, updates map[string][
 		}
 	}
 
+	// TODO(jscheuermann): If a maintenance zone is already set, we should favour Pods in this zone to be updated.
```
johscheuer, line 211: Same here, I can create an issue to track the todo.

controllers/maintenance_mode_checker.go
```diff
@@ -39,77 +37,59 @@ func (maintenanceModeChecker) reconcile(ctx context.Context, r *FoundationDBClus
...
-	}
-	for _, process := range status.Cluster.Processes {
-		if _, ok := processGroupsToCheck[process.Locality[fdbv1beta2.FDBLocalityInstanceIDKey]]; !ok {
+
+	logger.Info("Cluster in maintenance mode", "zone", cluster.Status.MaintenanceModeInfo.ZoneID)
+
+	var processGroupsToUpdate int
+	var hasProcessGroupsInZone bool
+	for _, processGroup := range cluster.Status.ProcessGroups {
+		if processGroup.FaultDomain != cluster.Status.MaintenanceModeInfo.ZoneID {
```
johscheuer, line 49: We can ignore all process groups that are not using the `storage` class.

controllers/maintenance_mode_checker_test.go
```diff
@@ -37,18 +33,17 @@ import (
 	fdbv1beta2 "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta2"
 )
 
-var _ = Describe("maintenance_mode_checker", func() {
+var _ = FDescribe("maintenance_mode_checker", func() {
```
sbodagala, line 36: Focus needs to be removed here.

sbodagala (approved):

Issue #1615, opened 2023-05-05
Cache machine-readable status from FDB


johscheuer:
### What would you like to be added/changed?

Currently the operator is always fetching the machine-readable status when required. Fetching the machine-readable status from a FDB cluster can be costly in terms of time and additional load on a cluster. For large cluster the machine-readable status can be a few double digit MBs that have to be sent to the operator. The fetched machine-readable status has to be parsed multiple times in this case. In most cases we don't need the latest status and it should be fine to work on a cached version of the machine-readable status, especially that FoundationDB itself also has some caching or detection times for the machine-readable status. The caching will be especially beneficial for large clusters and for cases where the result of the status takes a longer time to return.

There are (at least) two ways how we can implement such a cache:

1. Time-based cache e.g. cache the status for x seconds.
2. Reconcile based cache e.g. fetch the status once in the (Reconcile)[https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/controllers/cluster_controller.go#L82] method and pass it down to the sub-reconcilers.

I believe we should start with the 2. approach as the risk should be minimal and the required code changes shouldn't be huge. The caching mechanism should be disabled per default and a command line flag should allow a user to enable/disable it.

Currently the following sub-reconcilers are querying the status directly:

```bash
$ rg -i  -g '!*_test.go' -g '!*e2e*' -g '!*admin*' 'GetStatus'
controllers/bounce_processes.go
56:     status, err := adminClient.GetStatus()

controllers/check_client_compatibility.go
72:     status, err := adminClient.GetStatus()

controllers/change_coordinators.go
53:     status, err := adminClient.GetStatus()

controllers/choose_removals.go
61:     status, err := adminClient.GetStatus()

controllers/update_status.go
84:             databaseStatus, err = adminClient.GetStatus()

controllers/remove_process_groups.go
82:     status, err := adminClient.GetStatus()

controllers/update_database_configuration.go
50:     status, err := adminClient.GetStatus()

controllers/maintenance_mode_checker.go
76:     status, err := adminClient.GetStatus()

internal/fault_tolerance.go
37:     status, err := adminClient.GetStatus()

controllers/remove_incompatible_processes.go
73:     status, err := adminClient.GetStatus()
```

PR #1661, opened 2023-06-01
Initial support for cache status per reconciliation loop

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1615. The caching mechanism can be enabled or disabled per `FoundationDBCluster` config, and the default value (if unset) can be configured per command line flag. The idea is to fetch the machine-readable status only once per reconciliation loop instead of multiple times. This should improve the performance of the operator for large clusters (and many), as the operator has to spend less time fetching the machine-readable status and parsing it.

A reconciliation loop should end pretty fast (a few seconds), therefore the risk should be minimal, but a user can decide if they want to enable this feature or not.

## Type of change

## Discussion

For the testing I set the default value to true, I will change that to false before merging.

## Testing

Unit tests and e2e tests will be running.

## Documentation

Will be added before merging.

## Follow-up

-

### DIFF
--- a/pkg/podmanager/pod_lifecycle_manager.go
+++ b/pkg/podmanager/pod_lifecycle_manager.go
@@ -123,6 +123,7 @@ func (manager StandardPodLifecycleManager) DeletePod(ctx context.Context, r clie
 
 // CanDeletePods checks whether it is safe to delete Pods.
 func (manager StandardPodLifecycleManager) CanDeletePods(ctx context.Context, adminClient fdbadminclient.AdminClient, cluster *fdbv1beta2.FoundationDBCluster) (bool, error) {
+	// TODO(j-scheuermann): Can we pass down a cached status here?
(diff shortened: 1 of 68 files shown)


pkg/podmanager/pod_lifecycle_manager.go
```diff
@@ -123,6 +123,7 @@ func (manager StandardPodLifecycleManager) DeletePod(ctx context.Context, r clie
 
 // CanDeletePods checks whether it is safe to delete Pods.
 func (manager StandardPodLifecycleManager) CanDeletePods(ctx context.Context, adminClient fdbadminclient.AdminClient, cluster *fdbv1beta2.FoundationDBCluster) (bool, error) {
+	// TODO(j-scheuermann): Can we pass down a cached status here?
```
johscheuer, line 126: This is currently still one of the places where we fetch the status. we have to change this method.

AnnigeriShambu (approved):

PR #1666, opened 2023-06-12
BugFix: init faultDomainsWithUnavailablePods before checking ProcessGroups updates

manfontan:
# Description

`maxZonesWithUnavailablePods` feature #1587 is updating the `faultDomainsWithUnavailable` pods map in the same loop as the updates map. This may result in `updates` map with more pods than the `maxZonesWithUnavailablePods` pods limit depending on the order of `cluster.Status.ProcessGroups`.

In this PR the `faultDomainsWithUnavailable` is calculated before the for loop to process the updates. This way we make sure that the `maxZonesWithUnavailablePods` is respected.  

This PR also extends the number of cases when a ProcessGroup is considered unavailable. Including Pods in `PodInitializing` state and Process Groups with `PodFailing` condition (has containers with none ready status).

## Type of change

## Discussion

## Testing

*Please describe the tests that you ran to verify your changes. Unit tests? Yes
Manual testing?* 

## Documentation

## Follow-up


controllers/update_pods.go
```diff
@@ -300,6 +314,10 @@ func deletePodsForUpdates(ctx context.Context, r *FoundationDBClusterReconciler,
 		}
 	}
 
+	// TODO(manfontan): if deletions size is larger than the fault tolerance of the cluster. The number of zones from
```
manfontan, line 317: This is unrelated to maxZonesWithUnavailablePods. I will work on this in a new PR. If that makes sense

manfontan: 🤔 actually if we are using the default mode the number of zones will be one. So a check here won't add any additional security.

api/v1beta2/foundationdbcluster_types.go
```diff
@@ -779,6 +779,8 @@ const (
 	SidecarUnreachable ProcessGroupConditionType = "SidecarUnreachable"
 	// PodPending represents a process group where the pod is in a pending state.
 	PodPending ProcessGroupConditionType = "PodPending"
+	// PodInitializing represents a process group where the pod is in an initializing state.
+	PodInitializing ProcessGroupConditionType = "PodInitializing"
```
johscheuer, line 783: What is the benefit of this new condition?

controllers/update_pods.go
```diff
@@ -72,11 +72,47 @@ func (updatePods) reconcile(ctx context.Context, r *FoundationDBClusterReconcile
...
+	faultDomainsWithUnavailablePods := make(map[fdbv1beta2.FaultDomain]fdbv1beta2.None)
+
+	for _, processGroup := range cluster.Status.ProcessGroups {
+		unavailable := false
+		// If the Pod is pending, we count it as unavailable.
+		if processGroup.GetConditionTime(fdbv1beta2.PodPending) != nil {
+			unavailable = true
+		}
+		// If the Pods is initializing, we count it as unavailable.
+		if processGroup.GetConditionTime(fdbv1beta2.PodInitializing) != nil {
```
johscheuer, line 86: I would expect that `MissingProcesses` is a better condition to check for.

controllers/update_pods.go
```diff
@@ -72,14 +72,38 @@ func (updatePods) reconcile(ctx context.Context, r *FoundationDBClusterReconcile
 	return deletePodsForUpdates(ctx, r, cluster, updates, logger, status)
 }
 
+// GetFaultDomainsWithUnavailablePods returns a map of fault domains with unavailable Pods. The map has the fault domain as key and the value is not used.
+func GetFaultDomainsWithUnavailablePods(ctx context.Context, logger logr.Logger, reconciler *FoundationDBClusterReconciler, cluster *fdbv1beta2.FoundationDBCluster) map[fdbv1beta2.FaultDomain]fdbv1beta2.None {
+	faultDomainsWithUnavailablePods := make(map[fdbv1beta2.FaultDomain]fdbv1beta2.None)
+
+	for _, processGroup := range cluster.Status.ProcessGroups {
```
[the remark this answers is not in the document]
johscheuer: Getting Pods is cheap, all those Pods are cached internally by the controller-runtime. I would prefer to move the unavailability check into a dedicated method e.g. `processGroupIsUnavailable`, that allows to directly return the result and prevents doing all the checks, even if we already discovered that the process group is unavailable.

controllers/update_pods.go
```diff
@@ -72,11 +72,47 @@ func (updatePods) reconcile(ctx context.Context, r *FoundationDBClusterReconcile
...
+		// If the Pods is initializing, we count it as unavailable.
+		if processGroup.GetConditionTime(fdbv1beta2.PodInitializing) != nil {
+			unavailable = true
+		}
+		// If the Pod is running with failed containers, we count it as unavailable.
+		if processGroup.GetConditionTime(fdbv1beta2.PodFailing) != nil {
+			unavailable = true
+		}
+		pod, err := reconciler.PodLifecycleManager.GetPod(ctx, reconciler, cluster, processGroup.GetPodName(cluster))
+		// If a Pod is not found consider it as unavailable.
```
johscheuer, line 94: We probably should log the error if the error is a different one than `..not found`. I would also prefer to return any errors that are not `.. not found` otherwise we might be acting on the wrong signal.

controllers/update_pods_test.go
```diff
@@ -208,6 +208,92 @@ var _ = Describe("update_pods", func() {
 		})
 	})
 
+	Context("Validating getFaultDomainsWithUnavailablePods", func() {
```
johscheuer, line 211:
```suggestion
	When("getting the fault domains with unavailable Pods", func() {
```

controllers/update_pods_test.go
```diff
@@ -208,6 +208,92 @@ var _ = Describe("update_pods", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+			Expect(result.Requeue).To(BeFalse())
+			Expect(k8sClient.Get(context.TODO(), ctrlClient.ObjectKeyFromObject(cluster), cluster)).NotTo(HaveOccurred())
+		})
+
+		JustBeforeEach(func() {
+			processGroupsWithFaultDomains = getFaultDomainsWithUnavailablePods(context.Background(), clusterReconciler, cluster)
+		})
+
+		When("A Process Group has a Pod with pending condition", func() {
```
johscheuer, line 228:
```suggestion
		When("a Process Group has a Pod with pending condition", func() {
```

Those `When` and `Context` statements are put together, having a capital letter in the mid of a sentence looks weird :)

controllers/update_pods_test.go
```diff
@@ -208,6 +208,92 @@ var _ = Describe("update_pods", func() {
...
+					}
+				}
+			})
+
+			It("should be marked as unavailable", func() {
+				Expect(processGroupsWithFaultDomains).To(HaveLen(1))
+			})
+		})
+
+		When("A Process Group has a Pod with initializing condition", func() {
```
johscheuer, line 242:
```suggestion
		When("a Process Group has a Pod with initializing condition", func() {
```

controllers/update_pods_test.go
```diff
@@ -208,6 +208,92 @@ var _ = Describe("update_pods", func() {
...
+					}
+				}
+			})
+
+			It("should be marked as unavailable", func() {
+				Expect(processGroupsWithFaultDomains).To(HaveLen(1))
+			})
+		})
+
+		When("A Process Group has a Pod running with failed containers", func() {
```
johscheuer, line 256:
```suggestion
		When("a Process Group has a Pod running with failed containers", func() {
```

controllers/update_pods_test.go
```diff
@@ -208,6 +208,92 @@ var _ = Describe("update_pods", func() {
...
+					}
+				}
+			})
+
+			It("should be marked as unavailable", func() {
+				Expect(processGroupsWithFaultDomains).To(HaveLen(1))
+			})
+		})
+
+		When("A Process Group has a pod marked for deletion", func() {
```
johscheuer, line 270:
```suggestion
		When("a Process Group has a Pod marked for deletion", func() {
```

controllers/update_pods_test.go
```diff
@@ -208,6 +208,92 @@ var _ = Describe("update_pods", func() {
...
+				pods[0].DeletionTimestamp = &metav1.Time{Time: time.Now()}
+				err = clusterReconciler.PodLifecycleManager.UpdatePods(context.TODO(), clusterReconciler, cluster, []*corev1.Pod{pods[0]}, true)
+				Expect(err).NotTo(HaveOccurred())
+			})
+			It("should be marked as unavailable", func() {
+				Expect(processGroupsWithFaultDomains).To(HaveLen(1))
+			})
+		})
+
+		When("A Process Group has no matching pod", func() {
```
johscheuer, line 283:
```suggestion
		When("a Process Group has no matching Pod", func() {
```

controllers/update_status_test.go
```diff
@@ -502,6 +502,31 @@ var _ = Describe("update_status", func() {
 			})
 		})
 
+		When("the pod is initializing", func() {
```
johscheuer, line 505:
```suggestion
		When("the Pod is initializing", func() {
```

johscheuer (changes requested):

api/v1beta2/foundationdbcluster_types.go
manfontan: After some more testing looks like this is not necessary. So I will remove it

manfontan:
🤔 not sure what happened. Looks like I have accidentally deleted some files. Are there any changes to `make clean all` I should be aware of?

controllers/update_pods.go
```diff
@@ -72,14 +72,84 @@ func (updatePods) reconcile(ctx context.Context, r *FoundationDBClusterReconcile
...
+			faultDomainsWithUnavailablePods[processGroup.FaultDomain] = fdbv1beta2.None{}
+			continue
+		}
+		// If the Pod is marked for deletion, we count it as unavailable.
+		if pod.DeletionTimestamp != nil {
+			faultDomainsWithUnavailablePods[processGroup.FaultDomain] = fdbv1beta2.None{}
+			continue
+		}
+		// If the Pod is pending, we count it as unavailable.
+		if pod.Status.Phase == corev1.PodPending {
```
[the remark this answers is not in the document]
manfontan: Not really. My understanding is that the set of pods is the same during the status update than here. So it should be the same check. We added it for extra peace of mind.

controllers/update_pods.go
```diff
@@ -72,14 +72,84 @@ func (updatePods) reconcile(ctx context.Context, r *FoundationDBClusterReconcile
...
+	}
+	return false
+}
+
+// getFaultDomainsWithUnavailablePods returns a map of fault domains with unavailable Pods. The map has the fault domain as key and the value is not used.
+func getFaultDomainsWithUnavailablePods(ctx context.Context, logger logr.Logger, reconciler *FoundationDBClusterReconciler, cluster *fdbv1beta2.FoundationDBCluster) map[fdbv1beta2.FaultDomain]fdbv1beta2.None {
+	faultDomainsWithUnavailablePods := make(map[fdbv1beta2.FaultDomain]fdbv1beta2.None)
+
+	for _, processGroup := range cluster.Status.ProcessGroups {
+		// TODO(manfontan) only consider stateful processes. Because stateless processes do not change the fault tolerance of the cluster.
```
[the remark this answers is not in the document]
manfontan: Ok I will add it. We were wondering if it was better to do it in a separate PR since the filtering was not part of the original functionality. But I can add the filter now if that is fine.

PR #1670, opened 2023-06-14
Add better logging for spec hash changes

johscheuer:
# Description

This provides better insights into the spec changes for a replacement.

## Type of change

## Discussion

I decided to log the spec as base64 to prevent any json parsing issues.

## Testing

Locally.

## Documentation

-

## Follow-up

-

### DIFF
--- a/internal/replacements/replacements.go
+++ b/internal/replacements/replacements.go
@@ -23,6 +23,7 @@ package replacements
 import (
 	"context"
 	"encoding/base64"
+	"encoding/json"
 	"fmt"
 
 	"github.com/FoundationDB/fdb-kubernetes-operator/pkg/podmanager"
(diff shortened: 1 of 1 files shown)


internal/replacements/replacements.go
```diff
@@ -200,14 +201,23 @@ func processGroupNeedsRemoval(cluster *fdbv1beta2.FoundationDBCluster, pod *core
...
 			return false, err
 		}
 
 		if pod.ObjectMeta.Annotations[fdbv1beta2.LastSpecKey] != specHash {
 			logger.Info("Replace process group",
-				"reason", fmt.Sprintf("specHash has changed from %s to %s", specHash, pod.ObjectMeta.Annotations[fdbv1beta2.LastSpecKey]))
+				"reason", "specHash has changed",
+				"desiredSpecHash", specHash,
+				"currentSpecHash", pod.ObjectMeta.Annotations[fdbv1beta2.LastSpecKey],
+				"desiredSpec", base64.StdEncoding.EncodeToString([]byte(spec.String())),
```
johscheuer, line 219: This should be rather the json value for better readability.

internal/replacements/replacements.go
```diff
@@ -200,14 +201,23 @@ func processGroupNeedsRemoval(cluster *fdbv1beta2.FoundationDBCluster, pod *core
 	}
 
 	if cluster.NeedsReplacement(processGroupStatus) {
-		specHash, err := internal.GetPodSpecHash(cluster, processGroupStatus.ProcessClass, idNum, nil)
+		spec, err := internal.GetPodSpec(cluster, processGroupStatus.ProcessClass, idNum)
```
sbodagala, line 205: Oh, you are invoking this because you wanted to log "spec.String()" below.

sbodagala (approved):

internal/replacements/replacements.go
johscheuer: Yes, we had a few cases in the past where it would have been useful to see what the new spec looks like. Using the spec in json format and encoding it to base64 seems like the best solution to log the new desired spec (to prevent any parsing issues with the log forwarders).

PR #1676, opened 2023-06-15
Run HA upgrade tests on PRs

johscheuer:
# Description

Enable HA upgrade tests on PRs.

## Type of change

*Please select one of the options below.*

- Other

## Discussion

HA upgrades have some other timing characteristic, so we should make sure new changes are not breaking any assumptions in a HA setup.

## Testing

-

## Documentation

-

## Follow-up

-

Issue #1264, opened 2022-06-26
Limit range over processes to DC


johscheuer:
### What would you like to be added/changed?

We have multiple places where we iterate over all processes from the cluster status and convert that information in a map or another structure that will be used later. I think it makes sense to filter processes and only include processes that have a matching DC ID:

```go
// Only include processes that are managed by this instance of the FoundationDBCluster
if cluster.Spec.DataCenter != p.Locality[fdbv1beta2.FDBLocalityDCIDKey] {
    continue
}
```

If `cluster.Spec.DataCenter` is unset it will compare two empty strings which will also match. This would have that for HA clusters we don' have to allocate additional memory for unused information.

A simple grep revealed those places where we use. the cluster status information (I removed all test files):

```bash
$ rg -g '!*_test.go' -i 'status.Cluster.Processes'
controllers/choose_removals.go
63:     for _, process := range status.Cluster.Processes {

controllers/admin_client_mock.go
207:                    status.Cluster.Processes[fmt.Sprintf("%s-%d", pod.Name, processIndex)] = fdbv1beta2.FoundationDBStatusProcessInfo{
230:                    status.Cluster.Processes[processGroup.ProcessGroupID] = fdbv1beta2.FoundationDBStatusProcessInfo{

controllers/change_coordinators.go
125:    candidates := make([]localityInfo, 0, len(status.Cluster.Processes))
126:    for _, process := range status.Cluster.Processes {

controllers/update_status.go
98:     for _, process := range databaseStatus.Cluster.Processes {

controllers/bounce_processes.go
59:     addressMap := make(map[string][]fdbv1beta2.ProcessAddress, len(status.Cluster.Processes))
60:     for _, process := range status.Cluster.Processes {
224:    addresses := make([]fdbv1beta2.ProcessAddress, 0, len(databaseStatus.Cluster.Processes))
225:    for _, process := range databaseStatus.Cluster.Processes {

controllers/cluster_controller.go
617:    for _, process := range status.Cluster.Processes {

internal/foundationdb_status.go
32:     for _, pInfo := range status.Cluster.Processes {

internal/removals/remove.go
89:     for _, p := range status.Cluster.Processes {

fdbclient/admin_client.go
356:    for _, process := range status.Cluster.Processes {

kubectl-fdb/cmd/exclusion_status.go
166:            for _, process := range status.Cluster.Processes {
```

PR #1681, opened 2023-06-16
Reduce the size of the process map by selecting only processes in the right DC

johscheuer:
# Description

There is no reason to put all processes into the map.

## Type of change

## Discussion

-

## Testing

We should wait for https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1676 to be merged.

## Documentation

-

## Follow-up

We should verify all other places where we iterate over all processes.


AnnigeriShambu (approved):

sbodagala:
Any other functions where we could do such an optimization?

johscheuer (commented):
> Any other functions where we could do such an optimization?

Yes, we have some examples here: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1264

PR #1687, opened 2023-06-20
Add a small buffer for the expected process counts for automatic replacements

johscheuer:
# Description

Allow a small buffer for automatic replacements during an upgrade.

## Type of change

## Discussion

-

## Testing

Manual run.

## Documentation

-

## Follow-up

-

### DIFF
--- a/e2e/test_operator_upgrades/operator_upgrades_test.go
+++ b/e2e/test_operator_upgrades/operator_upgrades_test.go
@@ -29,10 +29,11 @@ Since FoundationDB is version incompatible for major and minor versions and the 
 
 import (
 	"fmt"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"log"
 	"strings"
 	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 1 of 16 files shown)


AnnigeriShambu (approved):

e2e/test_operator_upgrades/operator_upgrades_test.go
```diff
@@ -159,8 +160,9 @@ var _ = Describe("Operator Upgrades", Label("e2e", "pr"), func() {
 			Expect(err).NotTo(HaveOccurred())
 
 			// During an upgrade we expect that the transaction system processes are replaced, so we expect to have seen
-			// 2 times the process counts for transaction system processes.
-			expectedProcessCounts := (processCounts.Total() - processCounts.Storage) * 2
+			// 2 times the process counts for transaction system processes. Add a small buffer of 5 to allow automatic
+			// replacements during an upgrade.
+			expectedProcessCounts := (processCounts.Total()-processCounts.Storage)*2 + 5
```
sbodagala, line 165: Same buffer count in both HA and non-HA cases?

johscheuer: We can reduce the buffer for the non-HA case. The buffer is still small enough for the non-HA case to catch issues where we are doing 2 replacements of the whole transaction system Pods.

PR #1688, opened 2023-06-20
Make sure we set the connection string if the status is cached

johscheuer:
# Description

Make sure we propagate the fetched connection string only in one place and we check that the fetched connection string is not empty before assigning it.

## Type of change

## Discussion

-

## Testing

Manuel e2e test run.

## Documentation

-

## Follow-up

-

### DIFF
--- a/controllers/change_coordinators.go
+++ b/controllers/change_coordinators.go
@@ -58,17 +58,6 @@ func (c changeCoordinators) reconcile(ctx context.Context, r *FoundationDBCluste
 		}
 	}
 
-	if status.Cluster.ConnectionString != cluster.Status.ConnectionString {
-		logger.Info("Updating out-of-date connection string")
-		r.Recorder.Event(cluster, corev1.EventTypeNormal, "UpdatingConnectionString", fmt.Sprintf("Setting connection string to %s", status.Cluster.ConnectionString))
-		cluster.Status.ConnectionString = status.Cluster.ConnectionString
-		err = r.updateOrApply(ctx, cluster)
-
-		if err != nil {
-			return &requeue{curError: err, delayedRequeue: true}
-		}
-	}
-
 	coordinatorStatus := make(map[string]bool, len(status.Client.Coordinators.Coordinators))
 	for _, coordinator := range status.Client.Coordinators.Coordinators {
 		coordinatorStatus[coordinator.Address.String()] = false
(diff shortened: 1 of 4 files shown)


AnnigeriShambu (approved):

controllers/change_coordinators.go
```diff
@@ -58,17 +58,6 @@ func (c changeCoordinators) reconcile(ctx context.Context, r *FoundationDBCluste
 		}
 	}
 
-	if status.Cluster.ConnectionString != cluster.Status.ConnectionString {
```
sbodagala, line 61: I don't follow this fully. Are you saying updating connection string here is not needed? This is because you are now updating the connection string in "getStatusFromClusterOrDummyStatus()" function but that function is not getting called here.

controllers/update_status.go
```diff
@@ -56,7 +56,6 @@ func (updateStatus) reconcile(ctx context.Context, r *FoundationDBClusterReconci
 	// Pass through Maintenance Mode Info as the maintenance_mode_checker reconciler takes care of updating it
 	originalStatus.MaintenanceModeInfo.DeepCopyInto(&clusterStatus.MaintenanceModeInfo)
 	clusterStatus.Generations.Reconciled = cluster.Status.Generations.Reconciled
-	clusterStatus.ConnectionString = originalStatus.ConnectionString
```
sbodagala: I am not sure even this change is entirely correct. This line is essentially copying the connection string from "originalStatus" to "clusterStatus". Instead you are going to update "clusterStatus.connectionString" in "getStatusFromClusterOrDummyStatus()" now. But what if "clusterStatus.connectionString" won't get updated in that function (because "tryConnectionOptions()" returns an empty connection string)?

sbodagala: Actually, it appears you had added this line in your previous commit?

sbodagala: But, overall, I like the idea of updating connectionString in a single place, rather than spreading it across multiple functions.

controllers/change_coordinators.go
johscheuer: Having the update logic in one place is better than spring it across multiple places. There was an also a bug where the connection string was set to en empty string if the status had an empty connection string but the `GetStatus` was not returning an error. Updating the connection string here brings not a huge value and was probably a left over from some previous refactoring.

controllers/update_status.go
johscheuer: Correct, the was from an old commit that got removed.

PR #1689, opened 2023-06-20
Do fault tolerance checks on status object

sbodagala:
# Description

Do fault tolerance checks on FoundationDB status object.

## Type of change

## Discussion

No.

## Testing

Manual testing?*

Ran operator unit tests locally.

Yes, nightly regression tests will need to be run. (But I think nightly tests only run upgrade related tests and we don't test this functionality in those tests, so not sure if running nightly regression tests will help here.)

## Documentation

No (maybe at a later stage).

No (maybe at a later stage).

N/A

Not sure (maybe we need to document this at a later stage).

## Follow-up

No.

N/A


johscheuer (approved):
LGTM 👍 We probably want to expose the newly added methods (moving them out of the internal package), but we can do that in a follow up.

PR #1701, opened 2023-06-23
Update chaos-mesh and the controller runtime version

johscheuer:
# Description

Updates the controller runtime to a more recent version and updates the chaos-mesh API definitions.

## Type of change

- Other

## Discussion

-

## Testing

Manual e2e test run.

## Documentation

-

## Follow-up

-


controllers/backup_controller.go
```diff
@@ -197,7 +197,7 @@ func (r *FoundationDBBackupReconciler) updateOrApply(ctx context.Context, backup
 			Status: backup.Status,
 		}
 
-		return r.Status().Patch(ctx, patch, client.Apply, client.FieldOwner("fdb-operator"), client.ForceOwnership)
+		return r.Status().Patch(ctx, patch, client.Apply, client.FieldOwner("fdb-operator")) //, client.ForceOwnership)
```
sbodagala, line 200: I am not sure how this change is related to upgrading chaos-mesh.

controllers/cluster_controller.go
```diff
@@ -220,6 +220,13 @@ func (r *FoundationDBClusterReconciler) SetupWithManager(mgr ctrl.Manager, maxCo
 		return err
 	}
 
+	err = mgr.GetFieldIndexer().IndexField(context.Background(), &corev1.Node{}, "metadata.name", func(o client.Object) []string {
+		return []string{o.(*corev1.Node).Name}
```
sbodagala, line 224: Why are we making this change? Overall, I think I am not following the controller related changes (a summary of why these changes are being made would be nice).

sbodagala (approved):

controllers/backup_controller.go
johscheuer: Those changes are related to the new controller runtime, there were some changes that made the `client.ForceOwnership` not a correct argument for sub resources.

controllers/cluster_controller.go
johscheuer: Those where required by the the controller-runtime as they added some checks if all the requested fields have and indexer and when the node taint support was added, this was missed and the new controller-runtime cached this issue.

PR #1732, opened 2023-07-11
Make sure we update the status of the FoundationDBCluster only in one place

johscheuer:
# Description

The goal of those changes is to reduce the number of conflicts when the operator is updating the `FoundationDBCluster` resource in Kubernetes.

## Type of change

## Discussion

This will make it easier to debug issues, as the noise for the conflicts will be reduced.

## Testing

Unit tests and e2e tests.

## Documentation

N/A

## Follow-up


AnnigeriShambu (approved):

johscheuer:
Closing for now. we will focus on this in another PR.

Issue #1679, opened 2023-06-16
Document default ports for FDB


johscheuer:
### What would you like to be added/changed?

We should document the default ports used for FDB.

PR #1733, opened 2023-07-11
Document used FDB ports

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1679

## Type of change

- Documentation

## Discussion

-

## Testing

N/A

## Documentation

Added.

## Follow-up

-

### DIFF
--- a/docs/manual/warnings.md
+++ b/docs/manual/warnings.md
@@ -41,6 +41,12 @@ You should be careful when changing images, environment variables, commands, or 
 
 Other than the above, you can make any modifications to the pod definition you need to suit your requirements.
 
+## Limitations on FDB Port Customization
+
+The operator doesn't support to use custom ports for FDB.
+Per default a FDB cluster without TLS will use the ports `4501` and a FDB cluster using TLS will be using `4500`.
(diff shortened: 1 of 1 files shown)


docs/manual/warnings.md
```diff
@@ -41,6 +41,12 @@ You should be careful when changing images, environment variables, commands, or
 
 Other than the above, you can make any modifications to the pod definition you need to suit your requirements.
 
+## Limitations on FDB Port Customization
```
sbodagala, line 44: How is "processNumber" going to be decided? Is the port going to be assigned to a process (or the additional processes) in a deterministic way?

johscheuer: The process number is deterministic and is only used if more than one process is running per Pod e.g. when `storageServersPerPod` is set. The first process has the process number 1 and then the next processes will have the incremented process number e.g. 2, 3, 4 .., depending on how many processes should be running per Pod.

AnnigeriShambu (approved):

docs/manual/warnings.md
sbodagala: So the process number gets assigned/decided based on the order that the processes start?

sbodagala (approved):

docs/manual/warnings.md
johscheuer: Not really, the operator configures this in the fdbmonitor config.

PR #1752, opened 2023-07-20
Verify the processes that are excluded 

johscheuer:
# Description

Verify the processes that are assumed to be excluded by the machine-readable status are also excluded based on the exclude command

## Type of change

## Discussion

This change will make sure we have a second safety net for processes that might be missing the roles sporadic in the amchine-readable status.

## Testing

Unit tests. I will see if there is a good way to reproduce a process without any roles but not excluded.

## Documentation

-

## Follow-up

-

### DIFF
--- a/fdbclient/admin_client_test.go
+++ b/fdbclient/admin_client_test.go
@@ -897,7 +897,7 @@ protocol fdb00b071010000`,
 			}
 
 			mockRunner = &mockCommandRunner{
-				mockedError:  fdbv1beta2.TimeoutError{Err: fmt.Errorf("timed out")},
+				mockedError:  nil,
 				mockedOutput: "",
 			}
 		})
@@ -921,8 +921,39 @@ protocol fdb00b071010000`,
 				Expect(err).NotTo(HaveOccurred())
 			})
 
-			It("should not issue an exclude command", func() {
-				Expect(mockRunner.receivedBinary).To(BeEmpty())
+			It("should issue an exclude command to verify the exclusion", func() {
+				Expect(mockRunner.receivedBinary).To(HaveSuffix(fdbcliStr))
+				Expect(mockRunner.receivedArgs).To(ContainElements("exclude 192.168.0.1:4500 192.168.0.2:4500"))
+			})
+		})
+
+		When("all provided processes are fully excluded in the status but the exclude command returns an error", func() {
+			BeforeEach(func() {
+				addressesToCheck = []fdbv1beta2.ProcessAddress{
+					{
+						IPAddress: net.ParseIP("192.168.0.1"),
+						Port:      4500,
+					},
+					{
+						IPAddress: net.ParseIP("192.168.0.2"),
+						Port:      4500,
+					},
+				}
+
+				mockRunner = &mockCommandRunner{
+					mockedError:  fdbv1beta2.TimeoutError{Err: fmt.Errorf("timed out")},
+					mockedOutput: "",
+				}
+			})
+
+			It("should return an empty list and an error", func() {
+				Expect(result).To(HaveLen(0))
+				Expect(err).To(HaveOccurred())
+			})
+
+			It("should issue an exclude command to verify the exclusion", func() {
+				Expect(mockRunner.receivedBinary).To(HaveSuffix(fdbcliStr))
+				Expect(mockRunner.receivedArgs).To(ContainElements("exclude 192.168.0.1:4500 192.168.0.2:4500"))
 			})
 		})
 
@@ -942,6 +973,11 @@ protocol fdb00b071010000`,
 						Port:      4500,
 					},
 				}
+
+				mockRunner = &mockCommandRunner{
+					mockedError:  nil,
+					mockedOutput: "",
+				}
 			})
 
 			It("should return the one process that is not excluded", func() {
@@ -952,8 +988,8 @@ protocol fdb00b071010000`,
 				Expect(err).NotTo(HaveOccurred())
 			})
 
-			It("should not issue an exclude command", func() {
-				Expect(mockRunner.receivedBinary).To(BeEmpty())
+			It("should issue an exclude command to verify the exclusion", func() {
+				Expect(mockRunner.receivedBinary).To(HaveSuffix(fdbcliStr))
 			})
 		})
 
@@ -983,8 +1019,8 @@ protocol fdb00b071010000`,
 				Expect(err).NotTo(HaveOccurred())
 			})
 
-			It("should not issue an exclude command", func() {
-				Expect(mockRunner.receivedBinary).To(BeEmpty())
+			It("should issue an exclude command to verify the exclusion", func() {
+				Expect(mockRunner.receivedBinary).To(HaveSuffix(fdbcliStr))
 			})
 		})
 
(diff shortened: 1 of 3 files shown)


internal/removals/remove.go
```diff
@@ -164,6 +174,28 @@ func GetRemainingMap(logger logr.Logger, adminClient fdbadminclient.AdminClient,
...
+	for _, addr := range addresses {
+		_, ok := remainingMap[addr.MachineAddress()]
+		if ok {
+			continue
+		}
+
+		checkExclusions = append(checkExclusions, addr)
+	}
+
+	return adminClient.ExcludeProcesses(checkExclusions)
```
johscheuer, line 196: We probably should add a check here and if no addresses to check are present skip the exclude command.

internal/removals/remove.go
```diff
@@ -153,6 +153,16 @@ func GetRemainingMap(logger logr.Logger, adminClient fdbadminclient.AdminClient,
 		logger.Info("Exclusions to complete", "remainingServers", remaining)
 	}
 
+	if status.Cluster.RecoveryState.ActiveGenerations > 1 {
+		return nil, fmt.Errorf("too many active generations to run the exclude command to verify a process is fully excluded, current active generations: %d", status.Cluster.RecoveryState.ActiveGenerations)
```
sbodagala, line 157: Since we are only checking for a non-zero value is that still considered to be too many?

johscheuer: We are checking if there is more than one active generation, if there are more than one active generation, the cluster could be in a state where we don't want to run the `exclude` command to verify if processes are excluded. In theory this shouldn't be required, as if the cluster has more than one active generation the previous command will return all processes as not excluded: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/pkg/fdbstatus/status_checks.go#L60-L68

johscheuer: But I noticed, it's better to move this check inside the `verifyExcludedProcesses` and only do it if the exclude command would be used.

sbodagala: > the cluster could be in a state where we don't want to run the exclude command
Is that because exclude command can be run only if the cluster is fully available (and not going through recovery)? In turn, is that because exclude itself would force a recovery?

internal/removals/remove.go
```diff
@@ -153,6 +153,16 @@ func GetRemainingMap(logger logr.Logger, adminClient fdbadminclient.AdminClient,
...
+	}
+
+	// Make sure we use the exclude command again to verify that the processes are excluded based on the command.
```
sbodagala: It appears there are a bunch of issues, as seen from the operator side, around exclude. Maybe we should summarize those issues and come up with a comprehensive solution(s) (from both server and operator sides) to address those issues?

internal/removals/remove.go
```diff
@@ -164,6 +174,28 @@ func GetRemainingMap(logger logr.Logger, adminClient fdbadminclient.AdminClient,
 	return remainingMap, nil
 }
 
+// verifyExcludedProcesses will verify that the processes that should be excluded based on the FoundationDB machine-readble status are also excluded when issuing
```
sbodagala: > verifyExcludedProcesses will verify that the processes that should be excluded based on the FoundationDB machine-readble status

We are verifying because we don't trust the exclude information provided by machine-readable status? Or the exclude information provided by machine-readable status is not meant to be precise?

internal/removals/remove.go
sbodagala: This will force a cluster recovery?

sbodagala:
> Verify the processes that are assumed to be excluded by the machine-readable status are also excluded based on the exclude command

The PR says this is a bug fix. It is not clear to me where the bug is? Is it in the exclude related information (in machine readable status) provided by the server? Or is it in the way the operator is interpreting that output? Does the operator need to run "exclude" command again in order to be confident about this information? How does that address the bug? Thanks!

internal/removals/remove.go
johscheuer: We have a plan to do this as a follow up.

internal/removals/remove.go
johscheuer: We want to have an additional safeguard in cases where the machine-readable status might be missing some information like the role information for processes.

internal/removals/remove.go
johscheuer: Only if the cluster is not fully recovered: https://github.com/apple/foundationdb/blame/03b102d86aecbe700aa8402ae31d0431bfb0b2b9/fdbserver/ClusterRecovery.actor.cpp#L589

pkg/fdbstatus/status_checks.go
```diff
@@ -162,12 +163,21 @@ func CanSafelyRemoveFromStatus(logger logr.Logger, client fdbadminclient.AdminCl
...
 	}
 
+	// Verify that all processes that are assumed to be fully excluded based on the machine-readable status are actually
+	// not serving any roles by running the exclude command again. If those processes are actually fully excluded and are not
+	// serving any roles, the exclude command should terminate quickly, otherwise we will hit a timeout, and we know that
+	// not all processes are fully excluded. This is meant to be an additional safeguard if the machine-readable status
+	// returns the wrong signals.
+	if len(exclusions.fullyExcluded) > 0 {
+		// When we hit a timeout error here we know that at least one of the fullyExcluded is still not fully excluded.
+		return notSafeToDelete, client.ExcludeProcesses(exclusions.fullyExcluded)
```
johscheuer, line 173: CC @jzhou77 any concerns if we add the additional `exclude` command? The idea is to confirm that the processes that are assumed to be fully excluded based on the machine-readable, will be used in an additional `exclude` command. If the exclude command returns before it hits the timeout, those processes are actually excluded. If the command times out, we know that the machine-readable status was not correct and at least one process is still part of the key or log servers.

JoshuaMcManus: Do we have any safeguards in place to prevent the death spirals we saw in the past with trying to check for exclusions while a cluster was in a bad state? I expect it's better to have this than not do the check and potentially remove processes before they are fully excluded, but I think we should think about finding a middle ground where we're not doing unnecessary exclude commands when the cluster is already in a bad shape.

fdbclient/admin_client_test.go
```diff
@@ -921,8 +921,39 @@ protocol fdb00b071010000`,
...
 
-			It("should not issue an exclude command", func() {
-				Expect(mockRunner.receivedBinary).To(BeEmpty())
+			It("should issue an exclude command to verify the exclusion", func() {
+				Expect(mockRunner.receivedBinary).To(HaveSuffix(fdbcliStr))
+				Expect(mockRunner.receivedArgs).To(ContainElements("exclude 192.168.0.1:4500 192.168.0.2:4500"))
+			})
+		})
+
+		When("all provided processes are fully excluded in the status but the exclude command returns an error", func() {
```
JoshuaMcManus, line 930: Dumb question, but does this check the situation when there is no process with this IP address serving a role, but we still get a timeout from the exclude command? It's what I think this does, but want to confirm.

johscheuer: Correct. This "simulates" a case where a process is in the machine-readable status and has no roles but the exclude command times out because the process is not fully excluded.

JoshuaMcManus (approved):

PR #1754, opened 2023-07-24
updating ipv6 support in sidecar and service

09harsh:
# Description

Updated the service spec to include IPFamilies for IPv6. Also updated the sidecar args. 

## Type of change

## Discussion

No

## Testing

Manual testing?*
Added unit tests

Added / updated unit tests.

## Documentation

## Follow-up

No

### DIFF
--- a/controllers/add_services.go
+++ b/controllers/add_services.go
@@ -43,16 +43,7 @@ func (a addServices) reconcile(ctx context.Context, r *FoundationDBClusterReconc
 		existingService := &corev1.Service{}
 		err := r.Get(ctx, client.ObjectKey{Namespace: cluster.Namespace, Name: cluster.Name}, existingService)
 		if err == nil {
-			if cluster.IsPodIPFamily6() && (existingService.Spec.IPFamilies == nil || existingService.Spec.IPFamilies[0] != corev1.IPv6Protocol) {
-				logger.V(1).Info("Recreating service", "name", service.Name)
-				err = r.Delete(ctx, existingService)
-				if err != nil {
-					return &requeue{curError: err}
-				}
-				err = r.Create(ctx, service)
-			} else {
-				err = updateService(ctx, logger, r, existingService, service)
-			}
+			err = updateService(ctx, logger, cluster, r, existingService, service)
 			if err != nil {
 				return &requeue{curError: err}
 			}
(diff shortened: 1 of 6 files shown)


internal/pod_models.go
```diff
@@ -550,6 +550,9 @@ func configureSidecarContainer(container *corev1.Container, initMode bool, proce
 		if cluster.Spec.Routing.PodIPFamily != nil {
 			sidecarArgs = append(sidecarArgs, "--public-ip-family")
 			sidecarArgs = append(sidecarArgs, fmt.Sprint(*cluster.Spec.Routing.PodIPFamily))
+			if *cluster.Spec.Routing.PodIPFamily == 6 {
```
johscheuer, line 553: Can we use the `pointer` package here and to deference the setting if set? Otherwise we might get an nil pointer exception.

internal/pod_models.go
```diff
@@ -550,6 +550,9 @@ func configureSidecarContainer(container *corev1.Container, initMode bool, proce
...
+				sidecarArgs = append(sidecarArgs, "--bind-address", "[::]:8080")
```
johscheuer, line 554: Can we add a comment above this arg why we need it?

internal/service_helper.go
```diff
@@ -38,5 +39,9 @@ func GetHeadlessService(cluster *fdbv1beta2.FoundationDBCluster) *corev1.Service
 	service.Spec.ClusterIP = "None"
 	service.Spec.Selector = cluster.GetMatchLabels()
 
+	if cluster.Spec.Routing.PodIPFamily != nil && *cluster.Spec.Routing.PodIPFamily == 6 {
```
johscheuer, line 42: This will not work, see: https://kubernetes.io/docs/concepts/services-networking/dual-stack/#services

```text
Note: The .spec.ipFamilies field is immutable because the .spec.ClusterIP cannot be reallocated on a Service that already exists. If you want to change .spec.ipFamilies, delete and recreate the Service.
```

johscheuer (changes requested):
Please update the PR with information on how you tested this. Could you also verify that the documentation is updated for the IPv6 case?

internal/service_helper.go
09harsh: Here we are just creating a service type not the the actual service, the actual call for service creation is done in addService reconciler.

internal/pod_models.go
09harsh: done

internal/service_helper.go
09harsh: done

09harsh:
I checked the failing test and it is not failing due to the changes done in this PR. The test is also failing in other PRs as well.

brownleej (approved):
👍

JoshuaMcManus (changes requested):

controllers/add_services.go
```diff
@@ -45,8 +43,16 @@ func (a addServices) reconcile(ctx context.Context, r *FoundationDBClusterReconc
...
 		if err == nil {
-			// Update the existing service
-			err = updateService(ctx, logger, r, existingService, service)
+			if cluster.IsPodIPFamily6() && (existingService.Spec.IPFamilies == nil || existingService.Spec.IPFamilies[0] != corev1.IPv6Protocol) {
+				logger.V(1).Info("Recreating service", "name", service.Name)
+				err = r.Delete(ctx, existingService)
+				if err != nil {
+					return &requeue{curError: err}
+				}
+				err = r.Create(ctx, service)
```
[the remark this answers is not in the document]
09harsh: Yes, if the creation failed after the service deletion the operator will requeue  and the reconciliation loop will attempt to create the service again.

JoshuaMcManus (approved):

PR #1757, opened 2023-07-25
When remote storage is in place allow the operator to recreate pods instead or replace pods

manfontan:
# Description

With the current logic the operator only allows to replace PODs in some cases for example when the NodeSelector changes. This PR implements a new spec option `RemoteStorage` to allow updating pods by deletion when some conditions are met. Currently only node selector changes would be considered.

This configuration is useful for scenarios where remote storage is available and it is possible to recreate pods and wait for them to re-attach to the PVs. This will reduce the fault tolerance of the cluster until the Pod is reattached to the PV (normally this should be a very quick operation) but will allow the operator to avoid costly migrations.

## Type of change

## Discussion

*Are there any design details that you would like to discuss further?
The idea here is to skip replacements when the flag is enabled and perform an update instead. If I am missing something in the replacement logic it would need to be updated as well.

## Testing

Manual testing?*

Unit testing.
Manual Testing.

It would be good to write an e2e test for this feature.

## Documentation

*Did you update relevant documentation within this repository? Yes

*If this change is adding new functionality, do we need to describe it in our user manual? Maybe

## Follow-up

*Are there any follow-up issues that we should pursue in the future? Ideally it would be good to detect the disk configuration and prevent the operator or warn the user from using deletion strategy with local disks.

*Does this introduce new defaults that we should re-evaluate in the future? Yes. By default this feature is disabled because this way the operator works for both local and remote disks. If this changes in the future the default may need to be re-evaluated.


manfontan:
I will mark this PR as ready for review after some testing in our dev environment.
Any feedback in the meantime would be very appreciated :D

api/v1beta2/foundationdbcluster_types.go
```diff
@@ -202,6 +202,11 @@ type FoundationDBClusterSpec struct {
 	// +kubebuilder:default:=false
 	ReplaceInstancesWhenResourcesChange *bool `json:"replaceInstancesWhenResourcesChange,omitempty"`
 
+	// RemoteStorage defines if the cluster is using remote storage. This is used to determine
+	// how the operator should handle pod replacements.
```
johscheuer, line 206: Can we actually document the current limitations? e.g. this will only affect node selectors. We should also note here that some cases might require to set this value to false, even if remote storage is used. One example would be if a user wants to migrate the cluster from one AZ to another AZ in AWS (as EBS volumes are limited to an AZ).

docs/manual/replacements_and_deletions.md
```diff
@@ -20,7 +20,7 @@ The following changes can only be rolled out through replacement:
 * Changing the process group ID prefix
 * Changing the public IP source
 * Changing the number of storage servers per pod
-* Changing the node selector
+* Changing the node selector, when the `deletePodsWhenNodeSelectorChanges` flag is not set.
```
johscheuer, line 23: This must be updated to reflect the new setting?

internal/replacements/replacements_test.go
```diff
@@ -277,6 +277,28 @@ var _ = Describe("replace_misconfigured_pods", func() {
...
 
+		Context("when nodeSelector changes and RemoteStorage is true", func() {
+			BeforeEach(func() {
+				pClass = fdbv1beta2.ProcessClassStorage
+				remove = false
+			})
+
+			It("should not need a removal", func() {
+				remoteStorage := true
+				cluster.Spec.RemoteStorage = &remoteStorage
```
johscheuer, line 288:
```suggestion
				cluster.Spec.RemoteStorage = pointer.Bool(true)
```

johscheuer (changes requested):
Could you make sure to write GitHub issues for the todos, including adding e2e tests?

johscheuer:
Closing this one as it's a bit out of sync. We probably want to write down a document on how to better handle the Pod replacements when remote storage is used.

Thanks for your work. Hopefully we can build on that in the future.

PR #1759, opened 2023-07-25
Remove the additional get status call for getting the coordinator set

johscheuer:
# Description

This change removes the need for fetching the FoundationDB machine-readable status again to get the set of coordinators.

## Type of change

- Other

## Discussion

There is no need to issue the get status command again only to fetch the coordinator set.

## Testing

Changed the unit test and e2e tests will be running.

## Documentation

-

## Follow-up

-

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -411,16 +411,6 @@ func (r *FoundationDBClusterReconciler) newFdbPodClient(cluster *fdbv1beta2.Foun
 	return internal.NewFdbPodClient(cluster, pod, globalControllerLogger.WithValues("namespace", cluster.Namespace, "cluster", cluster.Name, "pod", pod.Name), r.GetTimeout, r.PostTimeout)
 }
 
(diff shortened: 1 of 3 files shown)


controllers/cluster_controller.go
```diff
@@ -411,16 +411,6 @@ func (r *FoundationDBClusterReconciler) newFdbPodClient(cluster *fdbv1beta2.Foun
 	return internal.NewFdbPodClient(cluster, pod, globalControllerLogger.WithValues("namespace", cluster.Namespace, "cluster", cluster.Name, "pod", pod.Name), r.GetTimeout, r.PostTimeout)
 }
 
-func (r *FoundationDBClusterReconciler) getCoordinatorSet(cluster *fdbv1beta2.FoundationDBCluster) (map[string]fdbv1beta2.None, error) {
```
sbodagala, line 414: You might want to keep this function as you might need it in a different context later.

sbodagala (approved):

controllers/cluster_controller.go
johscheuer: I rather delete the method and add it later again if we should need it. Otherwise we have a method that is not used and tested and might not be working anymore once we need it.

Issue #1760, opened 2023-07-26
e2e test should be running against a cluster with some minimal data


johscheuer:
### What would you like to be added/changed?

Currently all our tests are running against an empty FDB cluster, we should change this and load the cluster with a predefined amount of data to make sure we are testing all operations on a cluster with data.

PR #1762, opened 2023-07-27
Add DataLoader to make sure cluster has minimal data

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1760

## Type of change

## Discussion

-

## Testing

e2e tests.

## Documentation

Will be updated.

## Follow-up

-

### DIFF
--- a/e2e/fixtures/fdb_data_loader.go
+++ b/e2e/fixtures/fdb_data_loader.go
@@ -225,4 +225,7 @@ func (factory *Factory) WaitUntilDataLoaderIsRunning(cluster *FdbCluster) {
 
 		return runningPods
 	}).WithTimeout(5 * time.Minute).WithPolling(5 * time.Second).Should(gomega.BeNumerically(">", 0))
+
+	// Wait 1 minute to load data.
+	time.Sleep(1 * time.Minute)
 }
--- a/e2e/fixtures/fdb_data_loader.go
+++ b/e2e/fixtures/fdb_data_loader.go
@@ -1,0 +1,197 @@
+/*
+ * fdb_data_loader.go
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2023 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package fixtures
+
+import (
+	"bytes"
+	"errors"
+	"github.com/onsi/gomega"
+	"io"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer/yaml"
+	yamlutil "k8s.io/apimachinery/pkg/util/yaml"
+	"text/template"
+)
+
+const (
+	dataLoaderJob = `apiVersion: batch/v1
+kind: Job
+metadata:
+  name: fdb-data-loader
+  namespace: {{ .Namespace }}
+  labels:
+    app: fdb-data-loader
+spec:
+  backoffLimit: 2
+  completions: 2
+  parallelism: 2
+  template:
+    spec:
+      containers:
+      - image: {{ .Image }}
+        imagePullPolicy: Always
+        name: fdb-data-loader
+        # This configuration will load ~1GB per data loader.
+        args:
+        - --keys=1000000
+        - --batch-size=50
+        - --value-size=1000
+        env:
+          - name: FDB_CLUSTER_FILE
+            value: /var/dynamic-conf/fdb.cluster
+          - name: FDB_TLS_CERTIFICATE_FILE
+            value: /tmp/fdb-certs/tls.crt
+          - name: FDB_TLS_CA_FILE
+            value: /tmp/fdb-certs/ca.pem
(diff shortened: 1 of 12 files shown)


e2e/fixtures/fdb_data_loader.go
```diff
@@ -0,0 +1,197 @@
...
+		Image:           factory.GetDataLoaderImage(),
+		Namespace:       cluster.Namespace(),
+		SidecarVersions: factory.GetSidecarConfigs(),
+		ClusterName:     cluster.Name(),
+		SecretName:      factory.GetSecretName(),
+	}
+}
+
+// CreateDataLoaderIfAbsent will create the data loader for the provided cluster and load some random data into the cluster.
+func (factory *Factory) CreateDataLoaderIfAbsent(cluster *FdbCluster) {
```
johscheuer, line 166: We have to wait until the Job is running with at least one Pod.

johscheuer:
> * Commit ID: [ab7dc95](https://github.com/FoundationDB/fdb-kubernetes-operator/commit/ab7dc950534b16163cba996a467255f72ac275d8)
> * Duration 3:02:17
> * Result: ❌ **FAILED**
> * [Build Log](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator/fdb-kubernetes-operator-pr/1e68f8d9-400a-43eb-9b2f-2af08b52115f.gz) terminal output (available for 30 days)
> * [Build Workspace](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator-pr/1e68f8d9-400a-43eb-9b2f-2af08b52115f/workspace.zip) zip file of the working directory (available for 30 days)

The single cluster upgrades have failed with the availability check failing. My first guess would be that the cluster might be overloaded in some situations.

e2e/test_operator/operator_test.go
```diff
@@ -113,6 +113,9 @@ var _ = BeforeSuite(func() {
 		factory.GetClusterOptions()...,
 	)
 
+	// Load some data async into the cluster. We will only block as long as the Job is created.
+	factory.CreateDataLoaderIfAbsent(fdbCluster)
```
sbodagala, line 117: I assume the loading of data will happen in parallel with the tests running. Depending on how long it will take to load data, a subset of tests could run with data already loaded? Do you think this will introduce non-determinism, and we might not see the issues (that we see in nightly runs) when we run individual tests standalone?

e2e/fixtures/fdb_data_loader.go
```diff
@@ -0,0 +1,197 @@
...
+			}
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+		}
+
+		obj, _, err := yaml.NewDecodingSerializer(unstructured.UnstructuredJSONScheme).
+			Decode(rawObj.Raw, nil, nil)
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+		unstructuredMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj)
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+		unstructuredObj := &unstructured.Unstructured{Object: unstructuredMap}
```
sbodagala, line 196: Where is the client actually loading the data into the key-value store?

e2e/test_operator/operator_test.go
```diff
@@ -144,6 +144,9 @@ var _ = Describe("Operator", Label("e2e", "pr"), func() {
 		}
 		Expect(fdbCluster.WaitForReconciliation()).ToNot(HaveOccurred())
 		factory.StopInvariantCheck()
+		// Make sure all data is present in the cluster
```
sbodagala, line 147: Are these checks enough to ensure that all data is present in the cluster?

e2e/test_operator_ha_upgrades/operator_ha_upgrade_test.go
```diff
@@ -270,6 +270,9 @@ var _ = Describe("Operator HA Upgrades", Label("e2e", "pr"), func() {
 			// may get bounced at different times and also because of cluster controller
 			// change, which happens a number of times, during an upgrade).
 			Expect(finalGeneration).To(BeNumerically("<=", initialGeneration+80))
+			// Make sure the cluster has no data loss
+			fdbCluster.GetPrimary().EnsureTeamTrackersAreHealthy()
```
sbodagala, line 274: Probably you are using "data loss" here with a different meaning?

e2e/fixtures/fdb_data_loader.go
johscheuer: Here: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/sample-apps/data-loader/app.py#L23 we use the sample data loader app.

e2e/test_operator/operator_test.go
johscheuer: I can change this to wait that the Job has completed, this obviously will slow down the tests and we can add another issue to add some load during the tests.

johscheuer (commented):
I updated the tests to wait until data loading has finished. In my tests data loading takes roughly 3-5 min which should be fine.

e2e/test_operator_ha_upgrades/operator_ha_upgrade_test.go
johscheuer: What do you mean? This will check if all team trackers contain all data, if we lost some data because we removed a set of processes that still held data, I would expect that the team trackers report an issue. But like I said we can improve this check if we want, but I'd a first start and better than having no checks at all.

sbodagala:
> I updated the tests to wait until data loading has finished. In my tests data loading takes roughly 3-5 min which should be fine.

If the tests are stable/become stable then maybe we can modify it to load data in parallel with the tests.

e2e/test_operator_ha_upgrades/operator_ha_upgrade_test.go
sbodagala: I take "data loss" to mean the data loaded by data loader is all present in the key-value store/all shards (by taking replication into account). So checking for data loss will involve checking shard sizes etc, but it is fine to extend this check later (if needed).

sbodagala (approved):

sbodagala (approved):

Issue #1725, opened 2023-07-04
Test failure (flaky) - Operator HA Upgrades upgrading a cluster when no remote processes are restarted


johscheuer:
### What happened?

Test failure: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1721#issuecomment-1620011188

```text
2023/07/04 10:45:28 invariant InvariantClusterStatusAvailableWithThreshold failed after: 1h12m48.096324199s
2023/07/04 10:45:28 invariant InvariantClusterStatusAvailableWithThreshold true again
  [FAILED] in [AfterEach] - /codebuild/output/src2747221078/src/github.com/FoundationDB/fdb-kubernetes-operator/e2e/fixtures/fixtures.go:62 @ 07/04/23 10:45:35.742
• [FAILED] [5128.885 seconds]
Operator HA Upgrades upgrading a cluster when no remote processes are restarted [It] Upgrade from 6.3.25 to 7.1.31 when no remote processes are restarted [e2e, pr]
/codebuild/output/src2747221078/src/github.com/FoundationDB/fdb-kubernetes-operator/e2e/fixtures/upgrade_test_configuration.go:105

  [FAILED] Unexpected error:
      <*errors.errorString | 0xc0003bf970>: {
          s: "timed out waiting for the condition",
      }
      timed out waiting for the condition
  occurred
  In [It] at: /codebuild/output/src2747221078/src/github.com/FoundationDB/fdb-kubernetes-operator/e2e/test_operator_ha_upgrades/operator_ha_upgrade_test.go:568 @ 07/04/23 10:45:20.195

  There were additional failures detected.  To view them in detail run ginkgo -vv
```

We see this test failure sometimes and we have to investigate the reason for the failure.

### What did you expect to happen?

Test should be passing.

### How can we reproduce it (as minimally and precisely as possible)?

Running the tests multiple times.

### FDB Kubernetes operator

<details>

```console
$ kubectl fdb version
# paste output here
```

</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Cloud provider

<details>

</details>

PR #1769, opened 2023-08-01
For HA e2e test with remote processes not being restarted, ignore availability check for version incompatible upgrades

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1725

It's expected, that the FDB cluster will have availability issues if the remote region is not restarted.

## Type of change

## Discussion

-

## Testing

e2e test will be running.

## Documentation

Added a comment in the code.

## Follow-up

-

### DIFF
--- a/e2e/test_operator_ha_upgrades/operator_ha_upgrade_test.go
+++ b/e2e/test_operator_ha_upgrades/operator_ha_upgrade_test.go
@@ -536,7 +536,11 @@ var _ = Describe("Operator HA Upgrades", Label("e2e", "pr"), func() {
 				Skip("operator doesn't support feature for test case")
 			}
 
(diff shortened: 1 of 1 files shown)


sbodagala:
The referenced issues shows that "InvariantClusterStatusAvailableWithThreshold failed" after an hour. Should it be unavailable for that long?

johscheuer:
> The referenced issues shows that "InvariantClusterStatusAvailableWithThreshold failed" after an hour. Should it be unavailable for that long?

The referenced issue, was a timeout (the cluster was not able to get upgraded in time) but we also see a few test failures in our nightlies where the DB is unavailable for a longer time period (15 seconds +).

PR #1771, opened 2023-08-01
Document the maintenance mode feature

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1648
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1648

## Type of change

- Documentation
- Other

## Discussion

-

## Testing

Added a minimal upgrade test with the maintenance mode enabled.

## Documentation

Added documentation on how to use it and the current limitations.

## Follow-up

We have a few radars around the maintenance mode.

### DIFF
--- a/docs/manual/replacements_and_deletions.md
+++ b/docs/manual/replacements_and_deletions.md
@@ -27,6 +27,23 @@ The following changes can only be rolled out through replacement:
 The number of inflight replacements can be configured by setting `maxConcurrentReplacements`, per default the operator will replace all misconfigured process groups.
 Depending on the cluster size this can require a quota that is has double the capacity of the actual required resources.
 
+## Using The Maintenance Mode
+
+The FoundationDB Kubernetes operator supports to make use of the [maintenance mode](https://github.com/apple/foundationdb/wiki/Maintenance-mode) in FoundationDB.
+Using the maintenance mode in FoundationDB will reduce the data distribution and disruption when Storage Pods must be updated.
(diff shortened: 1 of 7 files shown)


docs/manual/replacements_and_deletions.md
```diff
@@ -27,6 +27,23 @@ The following changes can only be rolled out through replacement:
...
+
+The FoundationDB Kubernetes operator supports to make use of the [maintenance mode](https://github.com/apple/foundationdb/wiki/Maintenance-mode) in FoundationDB.
+Using the maintenance mode in FoundationDB will reduce the data distribution and disruption when Storage Pods must be updated.
+The following addition to the `FoundationDBCluster` resource will enable the maintenance mode for this cluster:
+
+```yaml
+spec:
+    automationOptions:
+      maintenanceModeOptions:
+        UseMaintenanceModeChecker: true
```
sbodagala, line 40: Setting maintenance mode is done by running a command from fdbcli, so I am not sure why this needs to be added to the FoundationDBCluster spec in order to enable maintenance mode.

sbodagala (approved):

docs/manual/replacements_and_deletions.md
johscheuer: That's right. This option is about the operator making use of the maintenance mode. Per default the operator will not use this option and just take down Pods if the Pod spec must be changed. If this setting is enabled the operator will make use of the maintenance mode before taking down the Pods.

Issue #1777, opened 2023-08-04
Buggify crashloop no longer accepts '*' argument in the targets


JoshuaMcManus:
### What happened?

PR https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1667 changed the acceptable pattern for the crashloop container targets to only match formats of `storage-1`, but the field is also supposed to accept `*` for crash looping all processes. This is useful in the event of needing to force a failover by crashing all running pods in one of the regions in an HA configuration.

The CRD is defined in apps.foundationdb.org_foundationdbclusters.yaml:

```
buggify:
                properties:
                  blockRemoval:
                    items:
                      maxLength: 63
                      pattern: ^([\w-]+)-(\d+)$
                      type: string
                    maxItems: 1000
                    type: array
                  crashLoop:
                    items:
                      maxLength: 63
                      pattern: ^([\w-]+)-(\d+)$
                      type: string
                    type: array
                  crashLoopContainers:
                    items:
                      properties:
                        containerName:
                          maxLength: 253
                          minLength: 1
                          type: string
                        targets:
                          items:
                            maxLength: 63
                            pattern: ^([\w-]+)-(\d+)$
                            type: string
                          maxItems: 10000
                          minItems: 0
```

We should modify the regex to allow `*` to be accepted while also enforcing all other previous matchers.

### What did you expect to happen?

To be able to add the following to the cluster spec:

```
  spec:
    buggify:
      crashLoopContainers:
        - containerName: foundationdb
          targets:
            - '*'
```

And the cluster will trigger all pods into a crashloop.

### How can we reproduce it (as minimally and precisely as possible)?

Modify a cluster spec using the above changes.

### FDB Kubernetes operator

latest version

### Kubernetes version

n/a

### Cloud provider

n/a

PR #1779, opened 2023-08-07
Fix CRD validation for Process Group ID and allow wildcard

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1777

## Type of change

## Discussion

-

## Testing

I added an e2e test and ran it manually. We have to add a better test, as the FDB cluster will be broken after the test.

## Documentation

-

## Follow-up

-

### DIFF
--- a/e2e/test_operator/operator_test.go
+++ b/e2e/test_operator/operator_test.go
@@ -1309,4 +1309,54 @@ var _ = Describe("Operator", Label("e2e", "pr"), func() {
 			})
 		})
 	})
+
+	// This test is pending, as all Pods will be restarted at the same time, which will lead to unavailability without
+	// using DNS.
+	PWhen("crash looping the sidecar for all Pods", func() {
+		BeforeEach(func() {
+			availabilityCheck = false
+			fdbCluster.SetCrashLoopContainers([]fdbv1beta2.CrashLoopContainerObject{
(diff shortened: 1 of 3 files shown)


e2e/test_operator/operator_test.go
```diff
@@ -1309,4 +1309,54 @@ var _ = Describe("Operator", Label("e2e", "pr"), func() {
 			})
 		})
 	})
+
+	// This test is pending, as all Pods will be restarted at the same time, which will lead to unavailability without
+	// using DNS.
+	PWhen("crash looping the sidecar for all Pods", func() {
```
JoshuaMcManus, line 1315: The test looks good, but is it actually testing the CRD (or is that even possible)? There are plenty of tests from what I was seeing that check the behviour of this field, but none of them check that the CRD will reject the value `*`.

JoshuaMcManus (approved):

e2e/test_operator/operator_test.go
johscheuer: It does test the CRD, as we are setting the `*`, if the CRD doesn't allow this the test would fail. The problem with this test is, that the cluster will be broken after this. I will create a follow up issue to fix this.

PR #1783, opened 2023-08-22
Update the default e2e FDB versions

johscheuer:
# Description

Update the default e2e test versions to 7.1 and 7.3.

## Type of change

- Other

## Discussion

-

## Testing

e2e test with the new versions will be triggered.

## Documentation

-

## Follow-up

-

### DIFF
--- a/e2e/Makefile
+++ b/e2e/Makefile
@@ -10,8 +10,9 @@ TIMEOUT?=168h
 NAMESPACE?=
 PREFIX?=
 CONTEXT?=
-FDB_VERSION?=7.1.31
-PREVIOUS_FDB_VERSION?=6.3.25
+FDB_VERSION?=7.1.37
+# This will be the version used for upgrade tests.
+NEXT_FDB_VERSION?=7.3.15
 ## Expectation is that you are running standard build image which generates both regular and debug (Symbols) images.
 FDB_IMAGE?=foundationdb/foundationdb:$(FDB_VERSION)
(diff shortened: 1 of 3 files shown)


e2e/Makefile
```diff
@@ -51,7 +52,7 @@ destroy-my-namespaces:
 
 # This target can be used to create a kind cluster that can be used for e2e testing.
 kind-setup:
-	@CHAOS_NAMESPACE=$(CHAOS_NAMESPACE) FDB_VERSION=$(FDB_VERSION) UPGRADE_VERSIONS=$(UPGRADE_VERSIONS) PREVIOUS_FDB_VERSION=$(PREVIOUS_FDB_VERSION) REGISTRY=$(REGISTRY) $(shell pwd)/scripts/setup_e2e.sh
+	@CHAOS_NAMESPACE=$(CHAOS_NAMESPACE) FDB_VERSION=$(FDB_VERSION) UPGRADE_VERSIONS=$(UPGRADE_VERSIONS) PREVIOUS_FDB_VERSION=$(NEXT_FDB_VERSION) REGISTRY=$(REGISTRY) $(shell pwd)/scripts/setup_e2e.sh
```
sbodagala, line 55: Do we still need to set "PREVIOUS_FDB_VERSION" (is it still being used)? If it needs to be set, shouldn't it be set to $(FDB_VERSION) (as the upgrade is now happening from $FDB_VERSION)?

johscheuer: Removed.

sbodagala (approved):

Issue #1756, opened 2023-07-25
Design: Suspend Pods before deleting them


johscheuer:
### What would you like to be added/changed?

It could be useful to keep the old PVCs before fully removing all resources associated with a process group, e.g. in cases where the operator makes a bad decision or where the FDB cluster returns the wrong signal. The idea is to remove the Pods, but keep the PVC around for some configurable time. This will speed up the recovery time.

PR #1785, opened 2023-08-22
Initial design doc for suspending Process Groups

johscheuer:
# Description

This design will allow the operator to suspend a Process Group (only delete the Pod) before deleting all resources.
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1756

## Type of change

- Other

## Discussion

-

## Testing

-

## Documentation

-

## Follow-up

Implementation

### DIFF
--- a/docs/design/suspending_pods.md
+++ b/docs/design/suspending_pods.md
@@ -29,7 +29,7 @@ The operator does not support suspending Pods.
 ## Proposed Design
 
 We will add a new subreconciler that will perform the same exclusion checks as the `removeProcessGroups` reconciler.
-If the suspension logic is disable the operator will skip all for in the new `suspendProcessGroups` subreconciler.
+If the suspension logic is disabled the operator will skip all for in the new `suspendProcessGroups` subreconciler.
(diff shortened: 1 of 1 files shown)


johscheuer (commented):
> Are there any negative side effects of not releasing the resources right away?

One negative side effect will be that the resources are kept a little bit longer. This shouldn't be a problem if not a lot of migrations are triggered and the wait time before actually removing is not too long, e.g. multiple weeks/days.

>> There could be a bug in the operator or FoundationDB that leads the operator to mistakenly removing a Pod and PVC that still has data on it.

> It appears to me the real problem is that we don't have a precise way of knowing (on operator's side) when exclusion is fully done? If so, probably we should extend the server to provide this information (rather than adding more logic on the operator's side to cover up for it)?

It's more about adding more safety nets. As the operator is managing a data service, we want to make sure that the risk of losing data because of a code bug in FDB or the operator is as low as possible and that the recovery time is small. Those changes will potentially help to protect against future bugs in the operator or FoundationDB, where the returned signals might be wrong.

sbodagala:
Adding a safety net is fine, but in this case it appears we already know that the details about full exclusion are not very clear to the operator and so we should probably put in effort to address that too.

sbodagala (approved):

brownleej (approved):
I left a minor writing suggestion, but otherwise this looks good to me.

PR #1792, opened 2023-09-06
Increase timeout for default single cluster upgrade

johscheuer:
# Description

Increase the timeout for the default single cluster upgrade. This should make sure that the operator has enough time to replace any process that might not be coming up correctly.

## Type of change

- Other

## Discussion

-

## Testing

I ran those test manually.

## Documentation

-

## Follow-up

-


sbodagala (approved):

johscheuer (commented):
Those test failures seem unrelated to the actual change in this PR.

Issue #1793, opened 2023-09-06
Update status reconciler should take the machine-readable messages into account


johscheuer:
### What would you like to be added/changed?

Currently the update status reconciler is performing multiple checks based on the created resources in Kubernetes and the machine-readable status. If the machine-readable status is missing some information e.g. with messages like `status_incomplete_timeout` the machine-readable status could reflect wrong information and the operator will use this information to add conditions to the processes. In order to prevent this we should check the machine-readable status to make sure the data is valid.

PR #1794, opened 2023-09-06
Only set the MissingProcesses condition if the machine-readable status contains at least one process

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1793 otherwise the process groups might get the `MissingProcesses` condition assigned if the machine-readable status is missing all processes.

## Type of change

## Discussion

## Testing

Unit tests and e2e tests will be running by CI.

## Documentation

-

## Follow-up

We can add more dedicated checks for specific messages in the machine-readable status.

### DIFF
--- a/controllers/update_status.go
+++ b/controllers/update_status.go
@@ -337,8 +337,9 @@ func checkAndSetProcessStatus(logger logr.Logger, r *FoundationDBClusterReconcil
 	}
 
 	processStatus := processMap[processID]
-
-	processGroupStatus.UpdateCondition(fdbv1beta2.MissingProcesses, len(processStatus) == 0)
+	// Only set the MissingProcesses condition if the machine-readable status has at least one process. We can improve this check
+	// later by validating additional messages in the machine-readable status.
+	processGroupStatus.UpdateCondition(fdbv1beta2.MissingProcesses, len(processStatus) == 0 && len(processMap) > 0)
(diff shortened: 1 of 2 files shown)


johscheuer (commented):
The failed test is:


```text
• [FAILED] [5227.144 seconds]
Operator HA Upgrades upgrading a cluster when no remote storage processes are restarted [It] Upgrade from 7.1.37 to 7.3.15 when no remote storage processes are restarted [e2e, pr]
/codebuild/output/src3343907186/src/github.com/FoundationDB/fdb-kubernetes-operator/e2e/fixtures/upgrade_test_configuration.go:105
```

controllers/update_status.go
```diff
@@ -337,8 +337,9 @@ func checkAndSetProcessStatus(logger logr.Logger, r *FoundationDBClusterReconcil
 	}
 
 	processStatus := processMap[processID]
-
-	processGroupStatus.UpdateCondition(fdbv1beta2.MissingProcesses, len(processStatus) == 0)
+	// Only set the MissingProcesses condition if the machine-readable status has at least one process. We can improve this check
```
sbodagala, line 340: You might want to file a radar/issue about the todo item/improvement so we won't forget about it. Thanks!

sbodagala (approved):

PR #1800, opened 2023-09-13
Cluster unable to reconcile and come back up If all pods get deleted in succession without waiting for them to come back up.

adarshdwivedi-arkin:
# Description

* Cluster unable to reconcile and come back up If whole cluster gets deleted multiple times without waiting for previously deleted pods to come up.

* Changes:

    * Check if pods are not in running phase and cluster status shows NeedsConfigurationChange or DatabaseUnavailable or HasUnhealthyProcess. In that case return reconciliation status with missing configuration error.

     * This allows the reconciler to redeploy the required pods/processes
     * Once all pods are up it will do a reconciliation to make fdb.cluster file and monitor.conf changes on pods and bring the cluster back up.

## Type of change

## Discussion

## Testing

* Ran the prexisting Unit tests.
* Manual testing involved deleting all pods in the cluster in quick succession without waiting for pods to come up and AWS EKS node reboots

*Do we need to perform additional testing once this is merged, or perform in a larger testing environment?* No

## Documentation

*Did you update relevant documentation within this repository?* Not required

*If this change is adding new functionality, do we need to describe it in our user manual?* Not required

*If this change is adding or removing subreconcilers, have we updated the core technical design doc to reflect that?* Not required

*If this change is adding new safety checks or new potential failure modes, have we documented and how to debug potential issues?* Not required

## Follow-up

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -452,7 +452,6 @@ func (r *FoundationDBClusterReconciler) getAllPodStatus(ctx context.Context, log
 		}
 
 		pod, err := r.PodLifecycleManager.GetPod(ctx, r, cluster, processGroup.GetPodName(cluster))
-		// If a Pod is not found ignore it for now.
 		if err != nil {
 			logger.Info("Could not find Pod for process group ID",
 				"processGroupID", processGroup.ProcessGroupID, "error", err)
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -444,7 +444,6 @@ func (r *FoundationDBClusterReconciler) getAllPodStatus(ctx context.Context, log
 		return false
 	}
 
-	var pods = make([]*corev1.Pod, 0, processCounts.Total())
 	for _, processGroup := range cluster.Status.ProcessGroups {
 		if processGroup.IsMarkedForRemoval() {
 			logger.V(1).Info("Ignore process group marked for removal",
@@ -466,8 +465,6 @@ func (r *FoundationDBClusterReconciler) getAllPodStatus(ctx context.Context, log
 				"phase", pod.Status.Phase)
 			return false
 		}
-
-		pods = append(pods, pod)
 	}
 
 	return true
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -440,7 +440,7 @@ func (r *FoundationDBClusterReconciler) updateOrApply(ctx context.Context, clust
 // connecting to fdb cluster and get information on coordinator servers, etc. Once all pods are up and running it will return true else false.
 func (r *FoundationDBClusterReconciler) getAllPodStatus(ctx context.Context, logger logr.Logger, cluster *fdbv1beta2.FoundationDBCluster) bool {
 	processCounts, err := cluster.GetProcessCountsWithDefaults()
-	logger.Info("Cluster Info", cluster, "Process Counts", processCounts)
+	logger.Info("Cluster Info", "Cluster", cluster, "Process Counts", processCounts)
 	if err != nil {
 		return false
 	}
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -24,7 +24,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/FoundationDB/fdb-kubernetes-operator/internal/locality"
 	"regexp"
 	"time"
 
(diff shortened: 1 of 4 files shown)


controllers/cluster_controller.go
```diff
@@ -448,6 +483,24 @@ func (r *FoundationDBClusterReconciler) getStatusFromClusterOrDummyStatus(logger
 		}, nil
 	}
 
+	/*Get the status of all the pods in the cluster.
+	 *Until all the pods are up it will keep cluster in config missing state and ask reconciler to spawn required pods/pvc, etc.
+	 *Once all the pods are up and running it will force the cluster to reconcile
+	 */
+	podsStatus := r.getAllPodStatus(ctx, logger, cluster)
```
johscheuer, line 490: What is the benefit of doing this check always if the if statement below is only executed during the initial reconciliation?

controllers/cluster_controller.go
```diff
@@ -435,9 +435,44 @@ func (r *FoundationDBClusterReconciler) updateOrApply(ctx context.Context, clust
...
+
+	for _, processGroup := range cluster.Status.ProcessGroups {
+		if processGroup.IsMarkedForRemoval() {
+			logger.V(1).Info("Ignore process group marked for removal",
+				"processGroupID", processGroup.ProcessGroupID)
+			continue
+		}
+
+		pod, err := r.PodLifecycleManager.GetPod(ctx, r, cluster, processGroup.GetPodName(cluster))
+		// If a Pod is not found ignore it for now.
```
johscheuer, line 455: The comment is actually wrong, if a Pod is not found the whole method will return false. We probably can remove this comment.

controllers/cluster_controller.go
```diff
@@ -435,9 +435,44 @@ func (r *FoundationDBClusterReconciler) updateOrApply(ctx context.Context, clust
 	return r.Status().Update(ctx, cluster)
 }
 
+// getAllPodStatus will get the current status and phase of all the pods in the fdb cluster and check if all of them are up before proceeding with
+// connecting to fdb cluster and get information on coordinator servers, etc. Once all pods are up and running it will return true else false.
+func (r *FoundationDBClusterReconciler) getAllPodStatus(ctx context.Context, logger logr.Logger, cluster *fdbv1beta2.FoundationDBCluster) bool {
```
johscheuer, line 440: What is the benefit of waiting for all Pods to be up and running? For a huge cluster this could take a long time until all Pods are scheduled and running.

johscheuer (commented):
Hi 👋 Thanks for your PR!

I'm not sure if I understand the idea/need for this PR, could you add some more context? The current implementation will only work in the initial reconciliation, is that the intention? If so a better place would be the [generate_initial_cluster_file](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/controllers/generate_initial_cluster_file.go#L43) reconciler. What is the requirement that Pods are deleted during the initial reconciliation phase? 

Could you add an e2e test that shows what currently is broken and that it will be fixed with your changes? Also can we add unit tests to show that this behaviour improves the operator? 

Thanks!

controllers/cluster_controller.go
adarshdwivedi-arkin: @johscheuer Won't the cluster still be in an unhealthy state without all the log and storage servers pods up and running?
For both small or large cluster if the Storage Server pods are not up and healthy, cluster might throw unhealthy state with no data replica error.

controllers/cluster_controller.go
adarshdwivedi-arkin: @johscheuer From logs I've observed that this same flow is also used when a majority of pods are deleted and in the scenario I've described in PR description. From what I've seen if all the pods in the cluster are deleted multiple times without waiting for them to come up and running the reconciliation state and status has stale data. This stale state of cluster results in pods not coming up and cluster stuck in problem state.
If the state updates then cluster status shows the fields in if condition set to 1. I can try again to see if the other if conditions can be removed.

controllers/cluster_controller.go
johscheuer: What do you mean by an "unhealthy" state? The cluster won't show the reconciled state until all Pods are up and running.

By recovered one you mean a cluster where you deleted all Pods and waited until the operator recreated them?

controllers/cluster_controller.go
johscheuer: Could you please create a reproducing example in our e2e test suite? Could you add some more information about the test you did and how the cluster was configured?

controllers/cluster_controller.go
```diff
@@ -448,6 +482,24 @@ func (r *FoundationDBClusterReconciler) getStatusFromClusterOrDummyStatus(logger
...
 
+	/*Get the status of all the pods in the cluster.
+	 *Until all the pods are up it will keep cluster in config missing state and ask reconciler to spawn required pods/pvc, etc.
+	 *Once all the pods are up and running it will force the cluster to reconcile
+	 */
+	podsStatus := r.getAllPodStatus(ctx, logger, cluster)
+	if !podsStatus &&
+		(cluster.Status.Generations.NeedsConfigurationChange == 1 ||
+			cluster.Status.Generations.DatabaseUnavailable == 1 ||
+			cluster.Status.Generations.HasUnhealthyProcess == 1) {
```
johscheuer, line 493:
```suggestion
		(cluster.Status.Generations.NeedsConfigurationChange > 0 ||
			cluster.Status.Generations.DatabaseUnavailable > 0 ||
			cluster.Status.Generations.HasUnhealthyProcess > 0) {
```

Otherwise this check will only work for the first generation (initial state).

johscheuer (commented):
Sorry for the late response.

Issue #1819, opened 2023-09-29
Specifying triple redundancy mode elect 5 coordinators


iyuroch:
### What happened?

When I specify redundancy mode triple for coordinator fdb operator will elect 5 (cluster.MinimumFaultDomains() + cluster.DesiredFaultTolerance()).
It also force us to have more pods for coordinator (we have separate pods for coordinator to support physical backup).

### What did you expect to happen?

To have 3 coordinators

### How can we reproduce it (as minimally and precisely as possible)?

Deploy any cluster with redundancy mode triple

### Anything else we need to know?

I will be happy to contribute to fix the issue.

### FDB Kubernetes operator

<details>

```console
$ kubectl fdb version
# paste output here
```

</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Cloud provider

<details>

</details>

johscheuer:
I changed the label to enhancement as this is not a bug. The triple configuration is used for the transaction system and storage subsystem and will allow the failure of processes in 2 zones (as the data is replicated across 3 zones). Since the coordinators use Paxos we need 5 coordinators to tolerate the failure of 2 zones.

We can allow to modify the number of coordinators to choose if we think this makes sense. In your new desired configuration you could only lose 1 coordinator. In general it's a good practice to have at least 5 coordinators (maybe you can reduce the resources for the Pod if that's your concern as the coordinators are not in the commit/read critical path). But if you loose a majority of coordinators and you are not able to recover them, you have to restore the cluster from a backup.

iyuroch:
Makes sense since it's paxos, thanks for explanation. I'm not sure if it makes sense to have additional configuration (since triple redundancy - means losing 2 coordinators).

johscheuer:
Do you think this section in the docs explain that well enough: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/docs/manual/fault_domains.md#coordinators or should we polish it?

iyuroch:
I would include information about it being paxos thus reason for having that many

PR #1826, opened 2023-10-06
Update docs on fault domain and coordinators

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1819

## Type of change

- Documentation

## Discussion

-

## Testing

-

## Documentation

Updated.

## Follow-up

-

### DIFF
--- a/docs/manual/fault_domains.md
+++ b/docs/manual/fault_domains.md
@@ -267,6 +267,7 @@ func (cluster *FoundationDBCluster) DesiredCoordinatorCount() int {
 For all clusters that use more than one region the operator will recruit 9 coordinators.
 If the number of regions is `1` the number of recruited coordinators depends on the redundancy mode.
 The number of coordinators is chosen based on the fact that the coordinators use a consensus protocol (Paxos) that needs a majority of processes to be up.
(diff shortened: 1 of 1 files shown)


docs/manual/fault_domains.md
```diff
@@ -265,7 +265,9 @@ func (cluster *FoundationDBCluster) DesiredCoordinatorCount() int {
 ```
 
 For all clusters that use more than one region the operator will recruit 9 coordinators.
-If the number of regions is `1` the number of recruited coordinators depends on the redundancy mode:
+If the number of regions is `1` the number of recruited coordinators depends on the redundancy mode.
+The number of coordinators is chosen based on the fact that the coordinators use a consensus protocol (Paxos) that needs a majority of processes to be up.
```
iyuroch, line 269: Where cluster size = failure tolerance * 2 + 1. In case of triple redundancy we expect cluster to be available losing 2 nodes, so for coordinator we have n  =  2 * 2 + 1

iyuroch (approved):

docs/manual/fault_domains.md
iyuroch: can we add this line to documentation?

AnnigeriShambu (approved):

PR #1827, opened 2023-10-09
Add coordinator fault tolerance check

johscheuer:
# Description

Adding a fault tolerance check for coordinators and only allow operations if all coordinators are up and running. In a fallow up we could use those fault tolerances and only validate them for affected Pods, e.g. only validate the coordinator fault tolerance for coordinator changes.

## Type of change

## Discussion

Currently the requirement is that all coordinators are up and running. This is the most strict policy.

## Testing

I added new unit tests. Some of the existing tests are broken and they need some modification (or the mock setup).

## Documentation

-

## Follow-up

-

### DIFF
--- a/pkg/fdbstatus/status_checks.go
+++ b/pkg/fdbstatus/status_checks.go
@@ -390,9 +390,26 @@ func DoLogServerFaultDomainCheckOnStatus(status *fdbv1beta2.FoundationDBStatus) 
 }
 
 // DoCoordinatorFaultDomainCheckOnStatus does a coordinator related fault domain check over the given status object.
-// @note an empty function for now. We will revisit this later.
-func DoCoordinatorFaultDomainCheckOnStatus(_ *fdbv1beta2.FoundationDBStatus) error {
-	// TODO: decide if we need to do coordinator related check.
+func DoCoordinatorFaultDomainCheckOnStatus(status *fdbv1beta2.FoundationDBStatus) error {
+	notReachable := make([]string, 0, len(status.Client.Coordinators.Coordinators))
+	for _, coordinator := range status.Client.Coordinators.Coordinators {
+		if coordinator.Reachable {
+			continue
+		}
+
+		notReachable = append(notReachable, coordinator.Address.String())
+	}
+
+	if len(notReachable) > 0 {
+		return fmt.Errorf("not all coordinators are reachable, unreachable coordinators: %s", strings.Join(notReachable, ","))
+	}
+
+	// If this is the case the statements above should already catch the unreachable coordinators and printout a more
+	// detailed message.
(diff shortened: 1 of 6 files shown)


pkg/fdbstatus/status_checks.go
```diff
@@ -390,9 +390,26 @@ func DoLogServerFaultDomainCheckOnStatus(status *fdbv1beta2.FoundationDBStatus)
 }
 
 // DoCoordinatorFaultDomainCheckOnStatus does a coordinator related fault domain check over the given status object.
-// @note an empty function for now. We will revisit this later.
-func DoCoordinatorFaultDomainCheckOnStatus(_ *fdbv1beta2.FoundationDBStatus) error {
-	// TODO: decide if we need to do coordinator related check.
+func DoCoordinatorFaultDomainCheckOnStatus(status *fdbv1beta2.FoundationDBStatus) error {
```
iyuroch, line 393: I don't see how this or parent call is used in the code - but since status update is eventual and we read it from object we can run in situation where current coordinator is not reachable anymore and we assume it's based on this check

pkg/fdbstatus/status_checks_test.go
```diff
@@ -1121,18 +1121,147 @@ var _ = Describe("status_checks", func() {
...
 		})
 
-		Context("coordinator fault domain checks", func() {
-			It("no coordinator related information in status", func() {
+		When("doing the coordinator fault domain checks", func() {
+			BeforeEach(func() {
 				status = &fdbv1beta2.FoundationDBStatus{
+					Client: fdbv1beta2.FoundationDBStatusLocalClientInfo{
+						Coordinators: fdbv1beta2.FoundationDBStatusCoordinatorInfo{
+							QuorumReachable: true,
```
09harsh, line 1129: Can we export this to a variable and use it in all the test cases as that code is repeated multiple times in this file ?

09harsh (approved):
LGTM. Just  a minor comment.

pkg/fdbstatus/status_checks_test.go
johscheuer: We change this value exactly once, so I'm not sure about the value of moving it into a variable.

pkg/fdbstatus/status_checks.go
johscheuer: What do you mean by `status update is eventual`?

> we can run in situation where current coordinator is not reachable anymore and we assume it's based on this check

This can happen even if we fetch the status directly before doing this check e.g. because at the moment we fetched the status the coordinator was still reachable. The question is only how long the interval between the failure and the detection is. In general those checks intention are to reduce the risk of failures, we are not able to prevent any outages, only to reduce the risk of them.

PR #1834, opened 2023-10-12
Add failure condition to logging in automatic replacements

johscheuer:
# Description

This change adds additional information to the replacement logic for failed process groups. Historically we only replaced processes based on one condition, but now in the current implementation we have multiple reasons that could lead to an automatic replacement. Adding this information to the logging output, makes it easier to understand why the operator started the replacement.

## Type of change

- Other (improve logging)

## Discussion

-

## Testing

Updated unit tests + CI will run e2e tests.

## Documentation

-

## Follow-up

-

### DIFF
--- a/api/v1beta2/foundationdbcluster_types.go
+++ b/api/v1beta2/foundationdbcluster_types.go
@@ -480,15 +480,19 @@ func (processGroupStatus *ProcessGroupStatus) GetPodName(cluster *FoundationDBCl
 	return sb.String()
 }
 
-// NeedsReplacement checks if the ProcessGroupStatus has conditions so that it should be removed
-func (processGroupStatus *ProcessGroupStatus) NeedsReplacement(failureTime int, taintReplacementTime int) (bool, int64) {
+// NeedsReplacement checks if the ProcessGroupStatus has conditions that require a replacement of the failed Process Group.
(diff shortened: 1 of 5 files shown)


AnnigeriShambu (approved):

johscheuer (commented):
Failure for the test was:

```text
• [FAILED] [1141.278 seconds]
Operator Migrations when a migration is triggered and the namespace quota is limited [BeforeEach] should add the prefix to all instances [e2e, pr]
  [BeforeEach] /codebuild/output/src4239453049/src/github.com/FoundationDB/fdb-kubernetes-operator/e2e/test_operator_migrations/operator_migration_test.go:77
  [It] /codebuild/output/src4239453049/src/github.com/FoundationDB/fdb-kubernetes-operator/e2e/test_operator_migrations/operator_migration_test.go:104

  [FAILED] Timed out after 600.013s.
  Expected
      <int64>: 2
  to be zero-valued
  In [BeforeEach] at: /codebuild/output/src4239453049/src/github.com/FoundationDB/fdb-kubernetes-operator/e2e/test_operator_migrations/operator_migration_test.go:100 @ 10/12/23 12:24:47.535
```

We have to monitor this and if that failures appears more often we have to investigate it.

Issue #1638, opened 2023-05-21
Remove the usage of the PVC map in the operator


johscheuer:
### What would you like to be added/changed?

We should use the cache from the controller-runtime instead of building our own map on to of this. The get requests are also better/cheaper than doing a list.

PR #1838, opened 2023-10-13
Remove the list call in the remove process groups reconciler and use the get method instead.

johscheuer:
# Description

Using the get method is more efficient than doing a List, especially if the operator is managing a large amount of resources.

## Type of change

## Discussion

Once https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1638 is fixed we are rid of all List command except the once in `update_status`, but those are intended to catch resources that are missing in the status.

## Testing

Ran unit tests, CI will run e2e test.

## Documentation

-

## Follow-up

- https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1638

### DIFF
--- a/controllers/remove_process_groups.go
+++ b/controllers/remove_process_groups.go
@@ -22,6 +22,7 @@ package controllers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"time"
@@ -146,6 +147,10 @@ func removeProcessGroup(ctx context.Context, r *FoundationDBClusterReconciler, c
 	var deletionError error
 
 	pod, err := r.PodLifecycleManager.GetPod(ctx, r, cluster, podName)
+	if err != nil && !k8serrors.IsNotFound(err) {
+		return err
+	}
+
 	if err == nil && pod.DeletionTimestamp.IsZero() {
 		deletionError = r.PodLifecycleManager.DeletePod(ctx, r, pod)
 		if err != nil {
@@ -163,12 +168,7 @@ func removeProcessGroup(ctx context.Context, r *FoundationDBClusterReconciler, c
 		logr.FromContextOrDiscard(ctx).V(1).Info("Deleting pvc", "name", pvcs.Items[0].Name)
 		err = r.Delete(ctx, &pvcs.Items[0])
 		if err != nil {
-			pvcError := fmt.Errorf("could not delete PVC: %w", err)
-			if deletionError == nil {
-				deletionError = pvcError
-			} else {
-				deletionError = fmt.Errorf("previous error: %w, %w", deletionError, pvcError)
-			}
+			deletionError = errors.Join(deletionError, fmt.Errorf("could not delete PVC: %w", err))
 		}
 	} else if len(pvcs.Items) > 1 {
 		return fmt.Errorf("multiple PVCs found for cluster %s, processGroupID %s", cluster.Name, processGroup.ProcessGroupID)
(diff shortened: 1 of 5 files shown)


controllers/remove_process_groups.go
```diff
@@ -138,108 +141,112 @@ func (u removeProcessGroups) reconcile(ctx context.Context, r *FoundationDBClust
...
+func removeProcessGroup(ctx context.Context, r *FoundationDBClusterReconciler, cluster *fdbv1beta2.FoundationDBCluster, processGroup *fdbv1beta2.ProcessGroupStatus) error {
+	podName := processGroup.GetPodName(cluster)
+	var deletionError error
 
-	if len(pods) == 1 && pods[0].DeletionTimestamp.IsZero() {
-		err = r.PodLifecycleManager.DeletePod(ctx, r, pods[0])
+	pod, err := r.PodLifecycleManager.GetPod(ctx, r, cluster, podName)
+	if err == nil && pod.DeletionTimestamp.IsZero() {
+		deletionError = r.PodLifecycleManager.DeletePod(ctx, r, pod)
 		if err != nil {
```
09harsh, line 151: We are checking `err==nil` in the outer if statement so this check will be useless.

09harsh: Should be `deletionError`

controllers/remove_process_groups.go
```diff
@@ -138,108 +141,112 @@ func (u removeProcessGroups) reconcile(ctx context.Context, r *FoundationDBClust
...
-	if err != nil {
-		return err
-	}
+func removeProcessGroup(ctx context.Context, r *FoundationDBClusterReconciler, cluster *fdbv1beta2.FoundationDBCluster, processGroup *fdbv1beta2.ProcessGroupStatus) error {
+	podName := processGroup.GetPodName(cluster)
+	var deletionError error
 
-	if len(pods) == 1 && pods[0].DeletionTimestamp.IsZero() {
-		err = r.PodLifecycleManager.DeletePod(ctx, r, pods[0])
+	pod, err := r.PodLifecycleManager.GetPod(ctx, r, cluster, podName)
```
09harsh, line 148: Should we also check whether this err is nil or not instead of just ignoring it ?

controllers/remove_process_groups.go
```diff
@@ -138,108 +141,112 @@ func (u removeProcessGroups) reconcile(ctx context.Context, r *FoundationDBClust
...
+			pvcError := fmt.Errorf("could not delete PVC: %w", err)
+			if deletionError == nil {
+				deletionError = pvcError
+			} else {
+				deletionError = fmt.Errorf("previous error: %w, %w", deletionError, pvcError)
+			}
 		}
 	} else if len(pvcs.Items) > 1 {
-		return fmt.Errorf("multiple PVCs found for cluster %s, processGroupID %s", cluster.Name, processGroupID)
+		return fmt.Errorf("multiple PVCs found for cluster %s, processGroupID %s", cluster.Name, processGroup.ProcessGroupID)
```
09harsh, line 174: do you think we should also include `deletionError` here ? Otherwise if all the previous error conditions are true those errors will be ignored completely.

controllers/remove_process_groups.go
```diff
@@ -138,108 +141,112 @@ func (u removeProcessGroups) reconcile(ctx context.Context, r *FoundationDBClust
...
-	services := &corev1.ServiceList{}
-	err = r.List(ctx, services, listOptions...)
-	if err != nil {
-		return err
-	}
-	if len(services.Items) == 1 && services.Items[0].DeletionTimestamp.IsZero() {
-		logr.FromContextOrDiscard(ctx).V(1).Info("Deleting service", "name", services.Items[0].Name)
-		err = r.Delete(ctx, &services.Items[0])
+	service := &corev1.Service{}
+	err = r.Get(ctx, client.ObjectKey{Name: podName, Namespace: cluster.Namespace}, service)
```
09harsh, line 178: Should we check if the `err` here is not nil ? Currently we are not checking that and if `deletionError` is nil, our function will return nil instead of err.

controllers/remove_process_groups.go
johscheuer: We probably can ignore that here, this code will be changed in a few weeks anyway to make use of the same get mechanism as for service and Pod.

AnnigeriShambu (approved):

johscheuer:
> * Commit ID: [6fd1232](https://github.com/FoundationDB/fdb-kubernetes-operator/commit/6fd1232959285c431a69f748f0e850c0bec91f88)
> * Duration 2:49:02
> * Result: ❌ **FAILED**
> * [Build Log](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator/fdb-kubernetes-operator-pr/248c1df9-e301-4e16-ae8e-4cdbf94b35e9.gz) terminal output (available for 30 days)
> * [Build Workspace](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator-pr/248c1df9-e301-4e16-ae8e-4cdbf94b35e9/workspace.zip) zip file of the working directory (available for 30 days)

The reason of those tests failures seems to be multiple concurrent running e2e tests.

PR #1844, opened 2023-10-13
If locality-based exclusions are enabled, only make use of the locality

johscheuer:
# Description

Otherwise we might issue an exclusions with the IP address.

## Type of change

## Discussion

-

## Testing

I will add some additional unit tests for this behaviour.

## Documentation

-

## Follow-up

-

### DIFF
--- a/internal/removals/remove.go
+++ b/internal/removals/remove.go
@@ -131,15 +131,18 @@ func GetRemainingMap(logger logr.Logger, adminClient fdbadminclient.AdminClient,
 			continue
 		}
 
+		// If we use localities for exclusions we don't have to care about the addresses.
+		if cluster.UseLocalitiesForExclusion() {
+			addresses = append(addresses, fdbv1beta2.ProcessAddress{StringAddress: processGroup.GetExclusionString()})
+			continue
+		}
+
 		if len(processGroup.Addresses) == 0 {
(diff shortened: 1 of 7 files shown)


johscheuer (commented):
This change is blocked on a new release containing: https://github.com/apple/foundationdb/pull/10976. We have to add additional version checks to make sure locality based exclusions are only used for this version that has the fix.

internal/removals/remove.go
```diff
@@ -126,15 +126,18 @@ func GetRemainingMap(logger logr.Logger, adminClient fdbadminclient.AdminClient,
 			continue
 		}
 
-		if len(processGroup.Addresses) == 0 {
-			logger.Info("Getting remaining removals to check for exclusion", "processGroupID", processGroup.ProcessGroupID, "reason", "missing address")
+		// If we use localities for exclusions we don't have to care about the addresses.
+		if cluster.UseLocalitiesForExclusion() {
```
johscheuer, line 130: The use localities for exclusion must have an additional check for the released version.

johscheuer: We have to add the IP addresses for log processes, in case they might be missing in the status.

johscheuer:
This PR: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1868 add the version check.

AnnigeriShambu (approved):

PR #1852, opened 2023-10-18
Make sure we test the status conditions before running the configure command

johscheuer:
# Description

The intention of this PR is to make sure the operator only trigger a configuration change if it is safe to do so.

## Type of change

## Discussion

-
## Testing

## Documentation

Will be updated.

## Follow-up

-

### DIFF
--- a/controllers/update_database_configuration.go
+++ b/controllers/update_database_configuration.go
@@ -23,12 +23,15 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"time"
+
+	"github.com/FoundationDB/fdb-kubernetes-operator/pkg/fdbstatus"
 
 	"github.com/go-logr/logr"
 
 	"k8s.io/utils/pointer"
 
-	fdbtypes "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta2"
+	fdbv1beta2 "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta2"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 )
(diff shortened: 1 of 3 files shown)


pkg/fdbstatus/status_checks.go
```diff
@@ -469,3 +478,32 @@ func HasDesiredFaultToleranceFromStatus(log logr.Logger, status *fdbv1beta2.Foun
 
 	return true
 }
+
+// ConfigurationChangeAllowed will return an error if the configuration change is assumed to be unsafe. If no error
+// is returned the configuration change can be applied.
+func ConfigurationChangeAllowed(status *fdbv1beta2.FoundationDBStatus, useRecoveryState bool) error {
```
johscheuer, line 484: @sbodagala and @jzhou77 are there any other checks that the operator should be performing before trying to change the cluster configuration?

sbodagala: I am still getting familiar with this area so I can't comment on what other checks the operator should perform. Checking that the status hasn't returned any error is reasonable. But the changeConfig API itself appears to be doing a number of checks and returns an error (https://github.com/apple/foundationdb/blob/967a546e15e81714e1dd763108b8afc3775416c5/fdbclient/include/fdbclient/GenericManagementAPI.actor.h#L47) if it can't change the configuration. Why are you checking the storage server lag below (I am trying to understand how it is related to/could affect changing the cluster configuration)?

controllers/update_database_configuration.go
```diff
@@ -71,21 +74,28 @@ func (u updateDatabaseConfiguration) reconcile(ctx context.Context, r *Foundatio
...
 		}
 		configurationString, _ := nextConfiguration.GetConfigurationString(cluster.Spec.Version)
 
-		dataState := status.Cluster.Data.State
-		if !(initialConfig || dataState.Healthy) {
-			logger.Info("Waiting for data distribution to be healthy", "stateName", dataState.Name, "stateDescription", dataState.Description)
-			r.Recorder.Event(cluster, corev1.EventTypeNormal, "NeedsConfigurationChange",
-				fmt.Sprintf("Spec require configuration change to `%s`, but data distribution is not fully healthy: %s (%s)", configurationString, dataState.Name, dataState.Description))
-			return nil
+		if !initialConfig {
```
sbodagala, line 91: If the cluster was not already configured then we wouldn't want to do these checks?

sbodagala (approved):

AnnigeriShambu (approved):

Issue #1854, opened 2023-10-19
Include processes should make use of Excluded server list instead of iterating over all the ProcessGroups


09harsh:
### What would you like to be added/changed?

In remove_process_groups reconciler we are iterating over all the ProcessGroups to get all the Processes we want to include (https://github.com/FoundationDB/fdb-kubernetes-operator/blob/be14c239ba5ccb84abdbb1aab5f743485a7ea62c/controllers/remove_process_groups.go#L273C6-L273C27). As an alternative a better approach would be to get the list of ExcludedServers and include them instead.

PR #1857, opened 2023-10-20
including processes using excluded server list

09harsh:
# Description

Updated the existing code to include processes using excluded server list.

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1854

## Type of change

## Discussion

N/A

## Testing

Added unit tests

## Documentation

N/A

## Follow-up

N/A

### DIFF
--- a/controllers/remove_process_groups.go
+++ b/controllers/remove_process_groups.go
@@ -301,8 +301,9 @@ func getProcessesToInclude(logger logr.Logger, cluster *fdbv1beta2.FoundationDBC
 	for _, processGroup := range cluster.Status.ProcessGroups {
 		if processGroup.IsMarkedForRemoval() && removedProcessGroups[processGroup.ProcessGroupID] {
 			foundInExcludedServerList := false
-			if _, ok := excludedServersMap[processGroup.GetExclusionString()]; ok {
-				fdbProcessesToInclude = append(fdbProcessesToInclude, fdbv1beta2.ProcessAddress{StringAddress: processGroup.GetExclusionString()})
+			exclusionString := processGroup.GetExclusionString()
+			if _, ok := excludedServersMap[exclusionString]; ok {
+				fdbProcessesToInclude = append(fdbProcessesToInclude, fdbv1beta2.ProcessAddress{StringAddress: exclusionString})
 				foundInExcludedServerList = true
 			}
 			for _, pAddr := range processGroup.Addresses {
--- a/controllers/remove_process_groups.go
+++ b/controllers/remove_process_groups.go
@@ -24,6 +24,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/FoundationDB/fdb-kubernetes-operator/pkg/fdbadminclient"
 	"net"
 	"time"
 
@@ -134,7 +135,7 @@ func (u removeProcessGroups) reconcile(ctx context.Context, r *FoundationDBClust
 	// This will return a map of the newly removed ProcessGroups and the ProcessGroups with the ResourcesTerminating condition
 	removedProcessGroups := r.removeProcessGroups(ctx, logger, cluster, zoneRemovals, zonedRemovals[removals.TerminatingZone])
 
-	err = includeProcessGroup(ctx, r, cluster, removedProcessGroups)
+	err = includeProcessGroup(ctx, r, cluster, removedProcessGroups, logger)
 	if err != nil {
 		return &requeue{curError: err}
 	}
@@ -251,14 +252,18 @@ func confirmRemoval(ctx context.Context, logger logr.Logger, r *FoundationDBClus
 	return true, canBeIncluded, nil
 }
 
-func includeProcessGroup(ctx context.Context, r *FoundationDBClusterReconciler, cluster *fdbv1beta2.FoundationDBCluster, removedProcessGroups map[fdbv1beta2.ProcessGroupID]bool) error {
+func includeProcessGroup(ctx context.Context, r *FoundationDBClusterReconciler, cluster *fdbv1beta2.FoundationDBCluster, removedProcessGroups map[fdbv1beta2.ProcessGroupID]bool, logger logr.Logger) error {
 	adminClient, err := r.getDatabaseClientProvider().GetAdminClient(cluster, r)
 	if err != nil {
 		return err
 	}
 	defer adminClient.Close()
 
-	fdbProcessesToInclude := getProcessesToInclude(cluster, removedProcessGroups)
+	fdbProcessesToInclude, err := getProcessesToInclude(cluster, removedProcessGroups, adminClient, logger)
+	if err != nil {
+		return err
+	}
+
 	if len(fdbProcessesToInclude) > 0 {
 		r.Recorder.Event(cluster, corev1.EventTypeNormal, "IncludingProcesses", fmt.Sprintf("Including removed processes: %v", fdbProcessesToInclude))
 
@@ -276,23 +281,44 @@ func includeProcessGroup(ctx context.Context, r *FoundationDBClusterReconciler, 
 	return nil
 }
 
-func getProcessesToInclude(cluster *fdbv1beta2.FoundationDBCluster, removedProcessGroups map[fdbv1beta2.ProcessGroupID]bool) []fdbv1beta2.ProcessAddress {
+func getProcessesToInclude(cluster *fdbv1beta2.FoundationDBCluster, removedProcessGroups map[fdbv1beta2.ProcessGroupID]bool, adminClient fdbadminclient.AdminClient, logger logr.Logger) ([]fdbv1beta2.ProcessAddress, error) {
 	fdbProcessesToInclude := make([]fdbv1beta2.ProcessAddress, 0)
 
 	if len(removedProcessGroups) == 0 {
-		return fdbProcessesToInclude
+		return fdbProcessesToInclude, nil
+	}
+
+	excludedServers, err := adminClient.GetExclusions()
+	if err != nil {
+		return fdbProcessesToInclude, fmt.Errorf("unable to get excluded servers from status, %s", err)
+	}
+	excludedServersMap := make(map[string]fdbv1beta2.None, len(excludedServers))
+	for _, excludedServer := range excludedServers {
+		excludedServersMap[excludedServer.String()] = fdbv1beta2.None{}
 	}
 
 	idx := 0
 	for _, processGroup := range cluster.Status.ProcessGroups {
 		if processGroup.IsMarkedForRemoval() && removedProcessGroups[processGroup.ProcessGroupID] {
-			if cluster.UseLocalitiesForExclusion() {
+			isIncluded := false
+			if _, ok := excludedServersMap[processGroup.GetExclusionString()]; ok {
 				fdbProcessesToInclude = append(fdbProcessesToInclude, fdbv1beta2.ProcessAddress{StringAddress: processGroup.GetExclusionString()})
+				isIncluded = true
 			}
 			for _, pAddr := range processGroup.Addresses {
-				fdbProcessesToInclude = append(fdbProcessesToInclude, fdbv1beta2.ProcessAddress{IPAddress: net.ParseIP(pAddr)})
+				if _, ok := excludedServersMap[pAddr]; ok {
+					fdbProcessesToInclude = append(fdbProcessesToInclude, fdbv1beta2.ProcessAddress{IPAddress: net.ParseIP(pAddr)})
+					isIncluded = true
+				}
 			}
-			continue
+			if !isIncluded {
+				// This means that the process is marked for exclusion and is also removed in the previous step but is missing
+				// its entry in the excluded servers in the status. This should not throw an error as this will block the
+				// inclusion for other processes, but we should have a record of this event happening in the logs.
+				logger.Info("processGroup ", processGroup, "is not included back as this is missing from excluded server list.")
+			} else {
+				continue
+			}
 		}
 		cluster.Status.ProcessGroups[idx] = processGroup
 		idx++
@@ -301,7 +327,7 @@ func getProcessesToInclude(cluster *fdbv1beta2.FoundationDBCluster, removedProce
 	// Remove the trailing duplicates.
 	cluster.Status.ProcessGroups = cluster.Status.ProcessGroups[:idx]
 
-	return fdbProcessesToInclude
+	return fdbProcessesToInclude, nil
 }
 
 func (r *FoundationDBClusterReconciler) getProcessGroupsToRemove(logger logr.Logger, cluster *fdbv1beta2.FoundationDBCluster, remainingMap map[string]bool, cordSet map[string]fdbv1beta2.None) (bool, bool, []*fdbv1beta2.ProcessGroupStatus) {
--- a/controllers/remove_process_groups.go
+++ b/controllers/remove_process_groups.go
@@ -24,7 +24,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/FoundationDB/fdb-kubernetes-operator/pkg/fdbadminclient"
 	"net"
 	"time"
 
@@ -259,7 +258,7 @@ func includeProcessGroup(ctx context.Context, logger logr.Logger, r *FoundationD
 	}
 	defer adminClient.Close()
 
-	fdbProcessesToInclude, err := getProcessesToInclude(logger, cluster, removedProcessGroups, adminClient, status)
+	fdbProcessesToInclude, err := getProcessesToInclude(logger, cluster, removedProcessGroups, status)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


controllers/remove_process_groups.go
```diff
@@ -251,14 +252,18 @@ func confirmRemoval(ctx context.Context, logger logr.Logger, r *FoundationDBClus
 	return true, canBeIncluded, nil
 }
 
-func includeProcessGroup(ctx context.Context, r *FoundationDBClusterReconciler, cluster *fdbv1beta2.FoundationDBCluster, removedProcessGroups map[fdbv1beta2.ProcessGroupID]bool) error {
+func includeProcessGroup(ctx context.Context, r *FoundationDBClusterReconciler, cluster *fdbv1beta2.FoundationDBCluster, removedProcessGroups map[fdbv1beta2.ProcessGroupID]bool, logger logr.Logger) error {
```
johscheuer, line 255:
```suggestion
func includeProcessGroup(ctx context.Context, logger logr.Logger, r *FoundationDBClusterReconciler, cluster *fdbv1beta2.FoundationDBCluster, removedProcessGroups map[fdbv1beta2.ProcessGroupID]bool) error {
```

I would prefer if we keep in out methods the `ctx` as first argument and the `logger` as second argument.

controllers/remove_process_groups.go
```diff
@@ -276,23 +281,44 @@ func includeProcessGroup(ctx context.Context, r *FoundationDBClusterReconciler,
...
-func getProcessesToInclude(cluster *fdbv1beta2.FoundationDBCluster, removedProcessGroups map[fdbv1beta2.ProcessGroupID]bool) []fdbv1beta2.ProcessAddress {
+func getProcessesToInclude(cluster *fdbv1beta2.FoundationDBCluster, removedProcessGroups map[fdbv1beta2.ProcessGroupID]bool, adminClient fdbadminclient.AdminClient, logger logr.Logger) ([]fdbv1beta2.ProcessAddress, error) {
 	fdbProcessesToInclude := make([]fdbv1beta2.ProcessAddress, 0)
 
 	if len(removedProcessGroups) == 0 {
-		return fdbProcessesToInclude
+		return fdbProcessesToInclude, nil
+	}
+
+	excludedServers, err := adminClient.GetExclusions()
```
johscheuer, line 291: We already have the status in the calling method: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/controllers/remove_process_groups.go#L58-L64, is there any reason to fetch the exclusion list again instead of taking it from the status?

controllers/remove_process_groups.go
```diff
@@ -276,23 +281,44 @@ func includeProcessGroup(ctx context.Context, r *FoundationDBClusterReconciler,
...
+	excludedServersMap := make(map[string]fdbv1beta2.None, len(excludedServers))
+	for _, excludedServer := range excludedServers {
+		excludedServersMap[excludedServer.String()] = fdbv1beta2.None{}
 	}
 
 	idx := 0
 	for _, processGroup := range cluster.Status.ProcessGroups {
 		if processGroup.IsMarkedForRemoval() && removedProcessGroups[processGroup.ProcessGroupID] {
-			if cluster.UseLocalitiesForExclusion() {
+			isIncluded := false
```
johscheuer, line 303:
```suggestion
			foundAddressToInclude := false
```

controllers/remove_process_groups.go
```diff
@@ -276,23 +281,44 @@ func includeProcessGroup(ctx context.Context, r *FoundationDBClusterReconciler,
...
+					fdbProcessesToInclude = append(fdbProcessesToInclude, fdbv1beta2.ProcessAddress{IPAddress: net.ParseIP(pAddr)})
+					isIncluded = true
+				}
+			}
+			if !isIncluded {
+				// This means that the process is marked for exclusion and is also removed in the previous step but is missing
+				// its entry in the excluded servers in the status. This should not throw an error as this will block the
+				// inclusion for other processes, but we should have a record of this event happening in the logs.
+				logger.Info("processGroup ", processGroup, "is not included back as this is missing from excluded server list.")
+			} else {
```
johscheuer, line 319: What happens if a process group was never scheduled and there fore never got an address?

pkg/fdbadminclient/mock/admin_client_mock.go
```diff
@@ -595,15 +595,19 @@ func (client *AdminClient) GetExclusions() ([]fdbv1beta2.ProcessAddress, error)
 		return nil, client.mockError
 	}
 
-	pAddrs := make([]fdbv1beta2.ProcessAddress, len(client.ExcludedAddresses))
+	pAddrs := make([]fdbv1beta2.ProcessAddress, 0)
```
johscheuer, line 598:
```suggestion
	pAddrs := make([]fdbv1beta2.ProcessAddress, 0,  len(client.ExcludedAddresses))
```

Any reason to remove the init method with capacity?

09harsh: During debugging I noticed its first parameter is always nil since we are appending new values in the slice.

controllers/remove_process_groups.go
```diff
@@ -276,21 +280,48 @@ func includeProcessGroup(ctx context.Context, r *FoundationDBClusterReconciler,
...
+	for _, excludedServer := range excludedServers {
+		excludedServersMap[excludedServer.String()] = fdbv1beta2.None{}
 	}
 
 	idx := 0
 	for _, processGroup := range cluster.Status.ProcessGroups {
 		if processGroup.IsMarkedForRemoval() && removedProcessGroups[processGroup.ProcessGroupID] {
-			if cluster.UseLocalitiesForExclusion() {
+			foundInExcludedServerList := false
+			if _, ok := excludedServersMap[processGroup.GetExclusionString()]; ok {
```
johscheuer, line 303:
```suggestion
			exclusionString := processGroup.GetExclusionString()
			if _, ok := excludedServersMap[exclusionString]; ok {
```

And adjust the `append` call below, otherwise we create the same string multiple times.

controllers/remove_process_groups_test.go
```diff
@@ -436,6 +439,8 @@ var _ = Describe("remove_process_groups", func() {
 				},
 			}
 			removedProcessGroups = make(map[fdbv1beta2.ProcessGroupID]bool)
+			Expect(err).NotTo(HaveOccurred())
```
johscheuer, line 442: What error are we checking here?

controllers/remove_process_groups_test.go
```diff
@@ -406,6 +407,8 @@ var _ = Describe("remove_process_groups", func() {
 
 	Context("validating getProcessesToInclude", func() {
 		var removedProcessGroups map[fdbv1beta2.ProcessGroupID]bool
+		var status *fdbv1beta2.FoundationDBStatus
+		var err error
```
johscheuer, line 411: Is there any need for this variable?

controllers/remove_process_groups_test.go
```diff
@@ -436,6 +439,8 @@ var _ = Describe("remove_process_groups", func() {
 				},
 			}
 			removedProcessGroups = make(map[fdbv1beta2.ProcessGroupID]bool)
+			Expect(err).NotTo(HaveOccurred())
+			status = &fdbv1beta2.FoundationDBStatus{}
```
johscheuer, line 443: Since we are not updating the status most of the test cases test the case where no exclusion is present, is that intended? Shouldn't this be rather the exception case instead of the default case we want to test?

controllers/remove_process_groups_test.go
```diff
@@ -489,17 +501,70 @@ var _ = Describe("remove_process_groups", func() {
...
+
+			When("one excluded process is missing from excluded servers", func() {
+				var removedProcessGroup *fdbv1beta2.ProcessGroupStatus
+				var removedProcessGroup2 *fdbv1beta2.ProcessGroupStatus
+
+				BeforeEach(func() {
+					removedProcessGroup = cluster.Status.ProcessGroups[0]
+					Expect(removedProcessGroup.ProcessGroupID).To(Equal(fdbv1beta2.ProcessGroupID("storage-1")))
+					removedProcessGroup.MarkForRemoval()
+					cluster.Status.ProcessGroups[0] = removedProcessGroup
```
johscheuer, line 549: Is this step required? Since we are working with a pointer, that step shouldn't be required.

controllers/remove_process_groups_test.go
```diff
@@ -489,17 +501,70 @@ var _ = Describe("remove_process_groups", func() {
...
+					removedProcessGroups[removedProcessGroup.ProcessGroupID] = true
+
+					removedProcessGroup2 = cluster.Status.ProcessGroups[1]
+					Expect(removedProcessGroup2.ProcessGroupID).To(Equal(fdbv1beta2.ProcessGroupID("storage-2")))
+					removedProcessGroup2.MarkForRemoval()
+					cluster.Status.ProcessGroups[1] = removedProcessGroup2
+					status.Cluster.DatabaseConfiguration.ExcludedServers = append(status.Cluster.DatabaseConfiguration.ExcludedServers, fdbv1beta2.ExcludedServers{Locality: removedProcessGroup2.GetExclusionString()})
```
johscheuer, line 556: Why are we doing an append here? Do we expect that the `ExcludedServers` contain any elements?

controllers/remove_process_groups_test.go
```diff
@@ -489,17 +501,70 @@ var _ = Describe("remove_process_groups", func() {
...
+			When("including a process which is excluded both by IP and locality", func() {
+				var removedProcessGroup *fdbv1beta2.ProcessGroupStatus
+
+				BeforeEach(func() {
+					removedProcessGroup = cluster.Status.ProcessGroups[0]
+					Expect(removedProcessGroup.ProcessGroupID).To(Equal(fdbv1beta2.ProcessGroupID("storage-1")))
+					removedProcessGroup.MarkForRemoval()
+					cluster.Status.ProcessGroups[0] = removedProcessGroup
+
+					status.Cluster.DatabaseConfiguration.ExcludedServers = append(status.Cluster.DatabaseConfiguration.ExcludedServers, fdbv1beta2.ExcludedServers{Locality: removedProcessGroup.GetExclusionString()})
```
johscheuer, line 526: Why are we doing an append here? Do we expect that the `ExcludedServers` contain any elements?

controllers/remove_process_groups_test.go
```diff
@@ -489,17 +501,70 @@ var _ = Describe("remove_process_groups", func() {
 					Expect(removedProcessGroup.ProcessGroupID).To(Equal(fdbv1beta2.ProcessGroupID("storage-1")))
 					removedProcessGroup.MarkForRemoval()
 					cluster.Status.ProcessGroups[0] = removedProcessGroup
+					status.Cluster.DatabaseConfiguration.ExcludedServers = append(status.Cluster.DatabaseConfiguration.ExcludedServers, fdbv1beta2.ExcludedServers{Locality: removedProcessGroup.GetExclusionString()})
```
johscheuer, line 504: Why are we doing an append here? Do we expect that the `ExcludedServers` contain any elements?

controllers/remove_process_groups_test.go
```diff
@@ -489,17 +501,70 @@ var _ = Describe("remove_process_groups", func() {
...
+					removedProcessGroup2.MarkForRemoval()
+					cluster.Status.ProcessGroups[1] = removedProcessGroup2
+					status.Cluster.DatabaseConfiguration.ExcludedServers = append(status.Cluster.DatabaseConfiguration.ExcludedServers, fdbv1beta2.ExcludedServers{Locality: removedProcessGroup2.GetExclusionString()})
+					removedProcessGroups[removedProcessGroup2.ProcessGroupID] = true
+				})
+
+				It("should include two process", func() {
+					fdbProcessesToInclude, err := getProcessesToInclude(logr.Logger{}, cluster, removedProcessGroups, status)
+					Expect(err).NotTo(HaveOccurred())
+					Expect(len(fdbProcessesToInclude)).To(Equal(2))
```
johscheuer, line 563: Since only `removedProcessGroup2` is present in the `ExcludedServer` with the locality based exclusion string, shouldn't `fdbProcessesToInclude` also include the IP address from `removedProcessGroup`?

pkg/fdbadminclient/mock/admin_client_mock.go
johscheuer: Right, the difference is `make([]fdbv1beta2.ProcessAddress, 0)` vs. (my proposal): `make([]fdbv1beta2.ProcessAddress, 0, len(client.ExcludedAddresses))` the latter creates an empty slice with the capacity of `len(client.ExcludedAddresses)`. Another option would have been to use the index instead of appending.

In general we want to preallocate the capacity if possible.

johscheuer (commented):
Do you think there is a way to add some additional checks in one of the e2e tests for this behaviour?

controllers/remove_process_groups_test.go
09harsh: It is needed now at line # 445

controllers/remove_process_groups_test.go
09harsh: Excluded servers are managed through `ExcludedAddresses` in our test suit. Currently this is acting as a place holder for the function as the excluded servers are being extracted from `ExcludedAddresses` in the `adminClient.GetExclusionsFromStatus(status)` function.

controllers/remove_process_groups_test.go
09harsh: Since we are not using the locality based exclusion setting and we are using the excluded server list we are re including all the servers the way they are excluded. So for e.g if a process is excluded using locality it will be reincluded using locality and not by both. Similarly if a process is excluded using IP, it will be reincluded using IP.

controllers/remove_process_groups.go
```diff
@@ -276,21 +281,42 @@ func includeProcessGroup(ctx context.Context, r *FoundationDBClusterReconciler,
...
+				if _, ok := excludedServersMap[pAddr]; ok {
+					fdbProcessesToInclude = append(fdbProcessesToInclude, fdbv1beta2.ProcessAddress{IPAddress: net.ParseIP(pAddr)})
+					foundInExcludedServerList = true
+				}
+			}
+			if !foundInExcludedServerList {
+				// This means that the process is marked for exclusion and is also removed in the previous step but is missing
+				// its entry in the excluded servers in the status. This should not throw an error as this will block the
+				// inclusion for other processes, but we should have a record of this event happening in the logs.
+				logger.Error(fmt.Errorf(""), "processGroup is included but is missing from excluded server list", "processGroup", processGroup)
```
johscheuer, line 319: It's probably good enough to print this out at the `info` level, as this is not an error.

pkg/fdbadminclient/admin_client.go
```diff
@@ -45,6 +45,10 @@ type AdminClient interface {
 	// database.
 	GetExclusions() ([]fdbv1beta2.ProcessAddress, error)
 
+	// GetExclusionsFromStatus gets a list of the addresses currently excluded from the
+	// database using status.
+	GetExclusionsFromStatus(status *fdbv1beta2.FoundationDBStatus) ([]fdbv1beta2.ProcessAddress, error)
```
johscheuer, line 50: Is there a reason to add a new method here and not just use his method: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/pkg/fdbstatus/status_checks.go#L251?

johscheuer (changes requested):

pkg/fdbadminclient/admin_client.go
09harsh: I needed this method to perform mocking. The real implementation of this function is calling the method you mentioned while the mocked version is used in unit tests that returns the ExcludedAdresses. Directly calling the `fdbstatus.GetExclusion()` method will break many unit tests as and various places of the mocked code we are adding excluded servers / ips in the `ExcludedAddresses` field of mocked admin client.

controllers/remove_process_groups.go
```diff
@@ -276,21 +281,42 @@ func includeProcessGroup(ctx context.Context, r *FoundationDBClusterReconciler,
...
+				if _, ok := excludedServersMap[pAddr]; ok {
+					fdbProcessesToInclude = append(fdbProcessesToInclude, fdbv1beta2.ProcessAddress{IPAddress: net.ParseIP(pAddr)})
+					foundInExcludedServerList = true
+				}
+			}
+			if !foundInExcludedServerList {
+				// This means that the process is marked for exclusion and is also removed in the previous step but is missing
+				// its entry in the excluded servers in the status. This should not throw an error as this will block the
+				// inclusion for other processes, but we should have a record of this event happening in the logs.
+				logger.V(1).Info("processGroup is included but is missing from excluded server list", "processGroup", processGroup)
```
johscheuer, line 319: You changed it now from `error` level to `debug` level, is that intended?

pkg/fdbadminclient/admin_client.go
johscheuer: I would prefer not to add another public exposed method to the admin client only to get some unit tests passed. From my perspective the correct approach would be to adjust the `GetStatus` method in the mock client to include the excluded servers in the generated status. That should solve the issue with the failing unit tests and will not require to expose another method with the benefit that the mocked output is more realistic.

johscheuer (commented):
The code changes LGTM. Can you make sure to rebase on the current main branch to pull in all new changes and e2e tests.

Do you thing we should add any new/additional e2e tests for this change?

johscheuer (approved):
LGTM 👍

PR #1875, opened 2023-11-08
Make sure the operator checks all provided addresses for exclusions

johscheuer:
# Description

In the previous implementation the operator only checked the localities in `getRemainingAndExcludedFromStatus`, this could lead to cases where the operator provided the locality and the IP address to be validated for exclusion. In that case the IP address was reported as `missing` (what was actually wrong), leading to triggering an exclusions, which could break the exclusion budget. The solution is to check for the IP address and the locality and if any of those should be verified.

## Type of change

## Discussion

-

## Testing

Updated the unit tests for this scenario.

## Documentation

-

## Follow-up

https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1844 will make sure that only the localities are used.


johscheuer (commented):
I spend some time on this, to think about a reliable test scenario.

AnnigeriShambu (approved):

johscheuer (commented):
Waiting for the e2e test to pass before merging.

PR #1879, opened 2023-11-09
Make sure the e2e test suite only runs locality based e2e tests if the provided versions support it

johscheuer:
# Description

If a FDB versions doesn't support the locality based exclusions, we should just skip this check.

## Type of change

## Discussion

-

## Testing

CI will run e2e tests.

## Documentation

-

## Follow-up

-

### DIFF
--- a/e2e/test_operator/operator_test.go
+++ b/e2e/test_operator/operator_test.go
@@ -374,7 +374,7 @@ var _ = Describe("Operator", Label("e2e", "pr"), func() {
 		})
 
 		BeforeEach(func() {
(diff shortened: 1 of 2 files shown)


AnnigeriShambu (approved):

johscheuer (commented):
Test failure is unrelated to the change:

```text
Summarizing 1 Failure:
  [FAIL] Operator when replacing a Pod stuck in Terminating state [It] should replace the Pod stuck in Terminating state [e2e, pr]
  /codebuild/output/src2005301706/src/github.com/FoundationDB/fdb-kubernetes-operator/e2e/test_operator/operator_test.go:954
```

PR #1881, opened 2023-11-09
Add check for plugin version upgrades

### LINKED ISSUE #1106: Implement a check-for-upgrade flag in the kubectl plugin

We should implement the functionality that our plugin can check if a newer version exists by simply checking the latest release (see: https://github.com/FoundationDB/fdb-kubernetes-operator/tree/main/kubectl-fdb#installation). This should be either a new sub command or a flag in the version sub command.

bktsh:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1106

## Type of change

## Discussion

An alternative approach would be saving minimum-supported-plugin-version as a label in operator deployment(instead of GitHub api) and use that to check if client's plugin is outdated. Would like to hear your idea @johscheuer .  

## Testing

Manual testing?*
Added unit test and ran manual test. 

No 

## Documentation

No

No

No

No

## Follow-up

No
No

### DIFF
--- a/kubectl-fdb/Readme.md
+++ b/kubectl-fdb/Readme.md
@@ -16,6 +16,9 @@ Install from release:
 
 ```bash
 pushd $TMPDIR
+if [ -f "$TMPDIR/latest.plugin" ]; then
+   rm "$TMPDIR/latest.plugin"
+fi
 OS=$(uname)
 ARCH=$(uname -m)
 VERSION="$(curl -s "https://api.github.com/repos/FoundationDB/fdb-kubernetes-operator/releases/latest" | jq -r '.tag_name')"
@@ -53,4 +56,4 @@ that we want to implement.
 Raise an issue if you miss a specific command to operate FDB on Kubernetes.
 
 ## Plugin Expiration
-We have added a commandline flag to the plugin that by default makes plugin to check its version against latest release(using GitHub API). If plugin is not using latest release version, it will print a warning message and skip the command. You may skip this version check by overriding default value of the flag to be false `--version-check=false`. The plugin will read the version info from GitHub API and store in a file in temp directory(`$TMPDIR/latest.plugin`), it'll use this temporary file to check version as long as the file is not older than 24 hours otherwise it'll repeat this process. If you get the warning message and want to update to the latest version, please remove `$TMPDIR/latest.plugin` file first.    
+We have added a commandline flag to the plugin that by default makes plugin to check its version against latest release(using GitHub API). If plugin is not using latest release version, it will print a warning message and skip the command. You may skip this version check by setting default value of the flag when issuing a command: `--version-check=false`. The plugin will read the version info from GitHub API and store in a file in temp directory(`$TMPDIR/latest.plugin`), it'll use this file to check the version and will refresh it daily. If you still get the warning message right after upgrading, please remove `$TMPDIR/latest.plugin` file and retry.    
--- a/kubectl-fdb/Readme.md
+++ b/kubectl-fdb/Readme.md
@@ -55,5 +55,10 @@ We have a list of [planned operations](https://github.com/FoundationDB/fdb-kuber
 that we want to implement.
 Raise an issue if you miss a specific command to operate FDB on Kubernetes.
 
-## Plugin Expiration
-We have added a commandline flag to the plugin that by default makes plugin to check its version against latest release(using GitHub API). If plugin is not using latest release version, it will print a warning message and skip the command. You may skip this version check by setting default value of the flag when issuing a command: `--version-check=false`. The plugin will read the version info from GitHub API and store in a file in temp directory(`$TMPDIR/latest.plugin`), it'll use this file to check the version and will refresh it daily. If you still get the warning message right after upgrading, please remove `$TMPDIR/latest.plugin` file and retry.    
+## Plugin Version Check
+
+Per default the plugin will check if there is a newer version present.
+For this the plugin will contact the GitHub API and check for the latest release.
+If a newer version was found the command will exit with the request to update the plugin.
+The version check can be disabled by setting `--version-check=false` when issuing a command.
+The plugin will create a local cache in the temp directory to reduce the calls to the GitHub API, the file is located under `$TMPDIR/kubectl-fdb-version.txt`.
--- a/kubectl-fdb/Readme.md
+++ b/kubectl-fdb/Readme.md
@@ -51,3 +51,6 @@ Run `kubectl fdb help` to get the latest help.
 We have a list of [planned operations](https://github.com/FoundationDB/fdb-kubernetes-operator/issues?q=is%3Aissue+is%3Aopen+label%3Aplugin)
 that we want to implement.
 Raise an issue if you miss a specific command to operate FDB on Kubernetes.
+
+## Plugin Expiration
+We have added a commandline flag to the plugin that by default makes plugin to check its version against latest release(using GitHub API). If plugin is not using latest release version, it will print a warning message and skip the command. You may skip this version check by overriding default value of the flag to be false `--version-check=false`. The plugin will read the version info from GitHub API and store in a file in temp directory(`$TMPDIR/latest.plugin`), it'll use this temporary file to check version as long as the file is not older than 24 hours otherwise it'll repeat this process. If you get the warning message and want to update to the latest version, please remove `$TMPDIR/latest.plugin` file first.    
--- a/kubectl-fdb/Readme.md
+++ b/kubectl-fdb/Readme.md
@@ -16,8 +16,8 @@ Install from release:
 
 ```bash
 pushd $TMPDIR
-if [ -f "$TMPDIR/latest.plugin" ]; then
-   rm "$TMPDIR/latest.plugin"
+if [ -f "$TMPDIR/kubectl-fdb-version.txt" ]; then
+   rm "$TMPDIR/kubectl-fdb-version.txt"
 fi
 OS=$(uname)
 ARCH=$(uname -m)
--- a/kubectl-fdb/cmd/plugin_version_checker.go
+++ b/kubectl-fdb/cmd/plugin_version_checker.go
@@ -12,8 +12,11 @@ import (
 	"time"
 )
 
+// KubectlFbReleaseURL is the public GirHub url we read the latest version
 const KubectlFbReleaseURL = "https://api.github.com/repos/FoundationDB/fdb-kubernetes-operator/releases/latest"
-const localTempVersionFileName = "latest"
+
+// LocalTempVersionFileName is where we cache plugin version for 24 hours, so we don't call GitHub for every single command(also rate-limit issue on GitHub api calls)
+const LocalTempVersionFileName = "latest"
 
 // PluginVersionDetails Contains [partial] plugin version details from GitHub
 type PluginVersionDetails struct {
@@ -25,7 +28,7 @@ type PluginVersionDetails struct {
 // It reads the latest fdb plugin version from local temp file,
 // if not exists, gets it from GitHub and store it locally and then returns it
 func getLatestPluginVersion() string {
-	fileName := filepath.Join(os.TempDir(), localTempVersionFileName)
+	fileName := filepath.Join(os.TempDir(), LocalTempVersionFileName)
 	_, err := os.Stat(fileName)
 	if os.IsNotExist(err) {
 		return updateLocalVersion(fileName)
@@ -33,10 +36,9 @@ func getLatestPluginVersion() string {
 	versionFileIsRecent := isVersionFileCreatedToday(fileName)
 	if versionFileIsRecent {
 		return readVersionFromLocalFile(fileName)
-	} else {
-		os.Remove(fileName)
-		return updateLocalVersion(fileName)
 	}
+	_ = os.Remove(fileName)
+	return updateLocalVersion(fileName)
 }
 
 // read version value from local temp file
@@ -71,10 +73,7 @@ func writeVersionToLocalTempFile(fileName string, version string) bool {
 	}
 	defer file.Close()
 	_, err = file.WriteString(version)
-	if err != nil {
-		return false
-	}
-	return true
+	return err == nil
 }
 
 // read version from HitHub and update local version file
@@ -102,10 +101,11 @@ func readVersionFromGitHub() (string, error) {
 		fmt.Println("Failed to fetch kubectl-fdb version from GitHub")
 		return "", err
 	}
-	body, err := io.ReadAll(resp.Body)
-	if resp != nil {
-		defer resp.Body.Close()
+	if resp == nil {
+		return "", nil
 	}
+	defer resp.Body.Close()
+	body, err := io.ReadAll(resp.Body)
 
 	if err != nil {
 		fmt.Println("Error in reading version from GitHub")
--- a/kubectl-fdb/cmd/plugin_version_checker.go
+++ b/kubectl-fdb/cmd/plugin_version_checker.go
@@ -12,7 +12,7 @@ import (
 	"time"
 )
 
-// KubectlFbReleaseURL is the public GirHub url we read the latest version
+// KubectlFbReleaseURL is the public GitHub URL we read the latest version from
 const KubectlFbReleaseURL = "https://api.github.com/repos/FoundationDB/fdb-kubernetes-operator/releases/latest"
 
 // LocalTempVersionFileName is where we cache plugin version for 24 hours, so we don't call GitHub for every single command(also rate-limit issue on GitHub api calls)
--- a/kubectl-fdb/cmd/plugin_version_checker.go
+++ b/kubectl-fdb/cmd/plugin_version_checker.go
@@ -15,7 +15,7 @@ import (
 const KubectlFbReleaseURL = "https://api.github.com/repos/FoundationDB/fdb-kubernetes-operator/releases/latest"
 const localTempVersionFileName = "latest"
 
-// PluginVersionDetails Contains [partial] plugin version details from github
+// PluginVersionDetails Contains [partial] plugin version details from GitHub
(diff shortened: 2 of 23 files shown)


kubectl-fdb/cmd/plugin_version_checker.go
```diff
@@ -0,0 +1,126 @@
...
+type PluginVersionDetails struct {
+	ID      int64  `json:"id"`
+	Version string `json:"tag_name"`
+	Name    string `json:"name"`
+}
+
+// It reads the latest fdb plugin version from local temp file,
+// if not exists, gets it from GitHub and store it locally and then returns it
+func getLatestPluginVersion() string {
+	fileName := filepath.Join(os.TempDir(), LocalTempVersionFileName)
```
johscheuer, line 31: Can we document this behaviour? We should make sure that people understand that the plugin tries to write to the temp dir. We should also add an additional argument to the root command to overwrite the default directory used for storing this information (could be useful for unit testing the behaviour of these methods).

kubectl-fdb/cmd/plugin_version_checker.go
```diff
@@ -0,0 +1,126 @@
...
+	if os.IsNotExist(err) {
+		return false
+	} else if err != nil {
+		return false
+	}
+	now := time.Now().UnixNano()
+	creationTime := fileInfo.ModTime().UnixNano()
+	diff := now - creationTime
+	hours := diff / int64(time.Hour)
+	return hours <= 24
```
johscheuer, line 65:
```suggestion
	return time.Since(fileInfo.ModTime()) <= 24 * time.hour
```

This should work too and would be a bit easier to understand.

kubectl-fdb/cmd/plugin_version_checker.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+// write version value in local temp file
+func writeVersionToLocalTempFile(fileName string, version string) bool {
```
johscheuer, line 69:
```suggestion
func writeVersionToLocalTempFile(fileName string, version string) error {
```

kubectl-fdb/cmd/plugin_version_checker.go
```diff
@@ -0,0 +1,126 @@
...
+	file, err := os.OpenFile(fileName, os.O_WRONLY|os.O_CREATE, 0644)
+	if err != nil {
+		return false
+	}
+	defer file.Close()
+	_, err = file.WriteString(version)
+	return err == nil
```
johscheuer, line 76:
```suggestion
	return err
```

kubectl-fdb/cmd/plugin_version_checker.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+// read version from HitHub and update local version file
+func updateLocalVersion(fileName string) string {
+	latestVersion, err := readVersionFromGitHub()
+	if err != nil {
+		return ""
```
johscheuer, line 83: Any reason not to return an error here?

kubectl-fdb/cmd/root.go
```diff
@@ -38,6 +38,9 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var skipCommand = false
```
johscheuer, line 41: Can we get rid of this global variable and make use of `PreRunE` instead of `PreRun`. Based on the documentation the whole chain should stop executing if `PreRuneE` is used and an error is returned.

kubectl-fdb/cmd/root.go
```diff
@@ -68,6 +71,7 @@ func NewRootCmd(streams genericclioptions.IOStreams) *cobra.Command {
 			return cmd.Help()
 		},
 	}
+	skipCommand = !usingLatestPluginVersion()
```
johscheuer, line 74: Can we add this as a `PreRunE`? Is there any benefit in the current setup compared to using `PreRunE`?

kubectl-fdb/cmd/root.go
```diff
@@ -145,3 +149,19 @@ func printStatement(cmd *cobra.Command, line string, mesType messageType) {
 	cmd.Printf("✔ %s\n", line)
 	color.Unset()
 }
+
+// will check plugin version and won't let any interaction happen with cluster, if it's not latest release version
+func usingLatestPluginVersion() bool {
+	if latestPluginVersion == "" {
```
johscheuer, line 155: Can we add an additional flag on the root to skip this check? Just to make sure people can skip this check if they want (default should be that the check is enabled).

kubectl-fdb/cmd/root.go
```diff
@@ -145,3 +149,19 @@ func printStatement(cmd *cobra.Command, line string, mesType messageType) {
...
+		return false
+	}
+	isLatest := strings.Compare(strings.ToLower(pluginVersion), "latest") == 0 || strings.Compare(pluginVersion, latestPluginVersion) >= 0
+	if !isLatest {
+		versionMessage = "Your kubectl-fdb plugin is not up-to-date, please install latest version and try again!\n" +
+			"Your version:[" + pluginVersion + "] vs. latest release version:[" + latestPluginVersion + "]\n" +
+			"Installation instructions can be found here: https://github.com/bktsh/fdb-kubernetes-operator/blob/main/kubectl-fdb/Readme.md"
+		fmt.Println(versionMessage)
```
johscheuer, line 163:
```suggestion
		cmd.Println(versionMessage)
```

When we make use of the `PreRunE` method we can make use of the cmd printer. That will print out the message to the according out writer (will be easier to test in unit tests).

kubectl-fdb/cmd/version.go
```diff
@@ -32,6 +32,7 @@ import (
 )
 
 var pluginVersion = "latest"
+var versionMessage = ""
```
johscheuer, line 35: Can we remove this global variable and instead return an error in `usingLatestPluginVersion` with the according message?

kubectl-fdb/cmd/root.go
```diff
@@ -38,6 +38,9 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var skipCommand = false
+var latestPluginVersion = getLatestPluginVersion()
```
johscheuer, line 42: Adding it here means we will run this command during our unit tests.

johscheuer (changes requested):

johscheuer (commented):
We need `PersistentPreRunE` instead of `PreRunE`. The `PersistentPreRunE` will be inherited by all subcommands.

kubectl-fdb/cmd/root.go
johscheuer: We have to make use of `PersistentPreRun` and return an error, then we can remove this global variable, see: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1881#pullrequestreview-1729917480.

kubectl-fdb/cmd/root.go
```diff
@@ -67,13 +72,20 @@ func NewRootCmd(streams genericclioptions.IOStreams) *cobra.Command {
...
+			}
+		},
 	}
 	cmd.SetOut(o.Out)
 	cmd.SetErr(o.ErrOut)
 	cmd.SetIn(o.In)
 
 	viper.SetDefault("license", "apache 2")
 	cmd.PersistentFlags().StringP("operator-name", "o", "fdb-kubernetes-operator-controller-manager", "Name of the Deployment for the operator.")
+	cmd.PersistentFlags().BoolP("version-check", "v", true, "If the plugin should compare its version with latest release to see if it is the latest.")
```
johscheuer, line 88:
```suggestion
	cmd.PersistentFlags().Bool("version-check", true, "If the plugin should check if a newer release of the plugin exists. If so the command will not be executed.")
```

We can drop the shorthand version of this flag.

kubectl-fdb/cmd/root.go
```diff
@@ -145,3 +157,20 @@ func printStatement(cmd *cobra.Command, line string, mesType messageType) {
...
+	}
+	isLatest := strings.Compare(strings.ToLower(pluginVersion), "latest") == 0 || strings.Compare(pluginVersion, latestPluginVersion) >= 0
+	if !isLatest {
+		versionMessage := "Your kubectl-fdb plugin is not up-to-date, please install latest version and try again!\n" +
+			"Your version:[" + pluginVersion + "] vs. latest release version:[" + latestPluginVersion + "]\n" +
+			"Installation instructions can be found here: https://github.com/bktsh/fdb-kubernetes-operator/blob/main/kubectl-fdb/Readme.md"
+		cmd.Println(versionMessage)
+		return false
+	}
+	return true
```
johscheuer, line 175:
```suggestion
	// If the user has a self build plugin we are not performing any checks.
	if strings.ToLower(pluginVersion) == "latest" {
		return true
	}
	latestPluginVersion, err := PluginVersionChecker.getLatestPluginVersion()
	if err != nil {
		return false
	}

	if  pluginVersion != latestPluginVersion {
		versionMessage := "kubectl-fdb plugin is not up-to-date, please install the latest version and try again!\n" +
			"Your version:[" + pluginVersion + "], latest release version:[" + latestPluginVersion + "].\n" +
			"Installation instructions can be found here: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/kubectl-fdb/Readme.md"
		cmd.Println(versionMessage)
		return false
	}

	return true
```

kubectl-fdb/main.go
```diff
@@ -31,7 +31,7 @@ import (
 func main() {
 	flags := pflag.NewFlagSet("kubectl-fdb", pflag.ExitOnError)
 	pflag.CommandLine = flags
-
+	cmd.PluginVersionChecker = &cmd.RealVersionChecker{}
```
johscheuer, line 34: See my comment above, we should pass this down to the `NewRootCmd` and not define a global variable.

kubectl-fdb/cmd/suite_test.go
```diff
@@ -19,6 +19,15 @@ var cluster *fdbv1beta2.FoundationDBCluster
 var clusterName = "test"
 var namespace = "test"
 
+// MockVersionChecker will help to mock the calls to GitHub API to get 'latest' version in tests
+type MockVersionChecker struct {
+	VersionChecker
+}
```
johscheuer, line 25:
```suggestion
type MockVersionChecker struct {
	MockedVersion string
}
```

It would be useful to allow a user to define a specific mocked version for different test cases.

kubectl-fdb/cmd/suite_test.go
```diff
@@ -19,6 +19,15 @@ var cluster *fdbv1beta2.FoundationDBCluster
...
+
+func (p *MockVersionChecker) getLatestPluginVersion() (string, error) {
+	return "latest", nil
+}
```
johscheuer, line 29:
```suggestion
func (versionChecker *MockVersionChecker) getLatestPluginVersion() (string, error) {
	if versionChecker. MockedVersion == "" {
		return "latest", nil
	}

	return versionChecker. MockedVersion, nil
}
```

kubectl-fdb/cmd/suite_test.go
```diff
@@ -32,6 +41,7 @@ var _ = BeforeSuite(func() {
 })
 
 var _ = BeforeEach(func() {
+	PluginVersionChecker = &MockVersionChecker{}
```
johscheuer, line 44: Same here, we should add this as a parameter to `NewRootCmd`.

kubectl-fdb/cmd/version_test.go
```diff
@@ -162,4 +162,33 @@ var _ = Describe("[plugin] version command", func() {
...
+
+			args := []string{"version", "--client-only"}
+			rootCmd.SetArgs(args)
+
+			err := rootCmd.Execute()
+			Expect(err).NotTo(HaveOccurred())
+		})
+		AfterEach(func() {
+			PluginVersionChecker = &MockVersionChecker{}
+		})
```
johscheuer, line 187: Why is this needed if the mocked version checker returns a hardcoded string?

kubectl-fdb/cmd/plugin_version_checker.go
```diff
@@ -0,0 +1,133 @@
...
+// PluginVersionDetails Contains [partial] plugin version details from GitHub
+type PluginVersionDetails struct {
+	ID      int64  `json:"id"`
+	Version string `json:"tag_name"`
+	Name    string `json:"name"`
+}
+
+// It reads the latest fdb plugin version from local temp file,
+// if not exists, gets it from GitHub and store it locally and then returns it
+func (p *RealVersionChecker) getLatestPluginVersion() (string, error) {
```
johscheuer, line 40:
```suggestion
func (versionChecker *RealVersionChecker) getLatestPluginVersion() (string, error) {
```

kubectl-fdb/cmd/plugin_version_checker.go
johscheuer: Looks like we decided against the flag to define the cache directory?

kubectl-fdb/cmd/plugin_version_checker.go
```diff
@@ -0,0 +1,133 @@
...
+
+// VersionChecker interface to help us mock test the version checker
+type VersionChecker interface {
+	getLatestPluginVersion() (string, error)
+}
+
+// RealVersionChecker will do actual call to GitHub API to get version
+type RealVersionChecker struct {
+	VersionChecker
+}
```
johscheuer, line 23:
```suggestion
type RealVersionChecker struct {}
```

It's not needed to add the interface to the struct. The Go compiler will detect that this struct implements all methods of the interface, e.g. https://go.dev/tour/methods/10.

kubectl-fdb/cmd/plugin_version_checker.go
```diff
@@ -0,0 +1,133 @@
...
+func (p *RealVersionChecker) getLatestPluginVersion() (string, error) {
+	fmt.Println("IN REAL")
+	fmt.Println("$$$$$					BBBBBBBBBBBBBBBBBBBBBBBBB					$$$$$$")
+	fileName := filepath.Join(os.TempDir(), LocalTempVersionFileName)
+	_, err := os.Stat(fileName)
+	if os.IsNotExist(err) {
+		return updateLocalVersion(fileName)
+	}
+	versionFileIsRecent := isVersionFileCreatedToday(fileName)
+	if versionFileIsRecent {
```
johscheuer, line 49:
```suggestion
	if isVersionFileCreatedToday(fileName) {
```

kubectl-fdb/cmd/plugin_version_checker.go
```diff
@@ -0,0 +1,133 @@
...
+	ID      int64  `json:"id"`
+	Version string `json:"tag_name"`
+	Name    string `json:"name"`
+}
+
+// It reads the latest fdb plugin version from local temp file,
+// if not exists, gets it from GitHub and store it locally and then returns it
+func (p *RealVersionChecker) getLatestPluginVersion() (string, error) {
+	fmt.Println("IN REAL")
+	fmt.Println("$$$$$					BBBBBBBBBBBBBBBBBBBBBBBBB					$$$$$$")
```
johscheuer, line 42: I guess those were used for debugging purposes?

johscheuer (changes requested):

bktsh:
> We need `PersistentPreRunE` instead of `PreRunE`. The `PersistentPreRunE` will be inherited by all subcommands.

Oh!  I totally missed the E at the end `PersistentPreRunE`!

kubectl-fdb/Readme.md
```diff
@@ -51,3 +54,6 @@ Run `kubectl fdb help` to get the latest help.
 We have a list of [planned operations](https://github.com/FoundationDB/fdb-kubernetes-operator/issues?q=is%3Aissue+is%3Aopen+label%3Aplugin)
 that we want to implement.
 Raise an issue if you miss a specific command to operate FDB on Kubernetes.
+
+## Plugin Expiration
+We have added a commandline flag to the plugin that by default makes plugin to check its version against latest release(using GitHub API). If plugin is not using latest release version, it will print a warning message and skip the command. You may skip this version check by setting default value of the flag when issuing a command: `--version-check=false`. The plugin will read the version info from GitHub API and store in a file in temp directory(`$TMPDIR/latest.plugin`), it'll use this file to check the version and will refresh it daily. If you still get the warning message right after upgrading, please remove `$TMPDIR/latest.plugin` file and retry.
```
johscheuer, line 59:
```suggestion
## Plugin Version Check

Per default the plugin will check if there is a newer version present.
For this the plugin will contact the GitHub API and check for the latest release.
If a newer version was found the command will exit with the request to update the plugin.
The version check can be disabled by setting `--version-check=false` when issuing a command.
The plugin will create a local cache in the temp directory to reduce the calls to the GitHub API, the file is located under `$TMPDIR/kubectl-fdb-version.txt`.
```

kubectl-fdb/cmd/plugin_version_checker.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+// RealVersionChecker will do actual call to GitHub API to get version
+type RealVersionChecker struct{}
+
+// KubectlFbReleaseURL is the public GitHub URL we read the latest version from
+const KubectlFbReleaseURL = "https://api.github.com/repos/FoundationDB/fdb-kubernetes-operator/releases/latest"
+
+// LocalTempVersionFileName is where we cache plugin version for 24 hours, so we don't call GitHub for every single command(also rate-limit issue on GitHub api calls)
+const LocalTempVersionFileName = "latest.plugin"
```
johscheuer, line 27:
```suggestion
const LocalTempVersionFileName = "kubectl-fdb-version.txt"
```

`latest.plugin` is a bit generic and I would prefer to have a more specific name.

johscheuer (commented):
One comment on rephrasing the Readme update and to use amore specific name for the cache file.

johscheuer (approved):

Issue #1675, opened 2023-06-15
Consider using controller client with caching in plugin


johscheuer:
### What would you like to be added/changed?

Currently the plugin uses the controller client with out any caching. We should consider setting up the controller client with caching and remove all the "manual" caching we build in the plugin.

Issue #1496, opened 2023-02-03
Avoid fetching cluster multiple time for each node in cordon command of fdb-plugin. 


09harsh:
### What would you like to be added/changed?

This came as a requirement for the PR: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1493, where we are fetching the clusters multiple time for each node which can be avoided.

PR #1888, opened 2023-11-17
Add cache to plugin

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1675
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1496 (fetching the same cluster multiple times is not an issue as they will be read from the cache)

## Type of change

## Discussion

-

## Testing

Ran some manual tests.

## Documentation

-

## Follow-up

We should remove some of the logic in the plugin to make sure we make use of the get calls and remove any caches we put on top of the controller client.


bktsh (approved):
:+1:

PR #1895, opened 2023-11-27
Update the way the operator deployment is created and verified

johscheuer:
# Description

We have seen cases where it took longer than the expected 60 seconds for the Pods to get scheduled and into a running state, so this PR increases the threshold to 120 seconds and the overall wait time. I removed some duplication code in `RecreateOperatorPods`. 

## Type of change

- Other

## Discussion

-

## Testing

CI will run tests.

## Documentation

-

## Follow-up

-

### DIFF
--- a/e2e/fixtures/fdb_operator_client.go
+++ b/e2e/fixtures/fdb_operator_client.go
@@ -40,8 +40,6 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer/yaml"
-	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/util/wait"
 	yamlutil "k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


e2e/fixtures/fdb_operator_client.go
```diff
@@ -507,80 +505,38 @@ func (factory *Factory) WaitUntilOperatorPodsRunning(namespace string) {
...
+		var runningReplicas int
 		for _, pod := range pods.Items {
-			if pod.Status.Phase == corev1.PodRunning {
+			if pod.Status.Phase == corev1.PodRunning && pod.DeletionTimestamp.IsZero() {
+				runningReplicas++
 				continue
 			}
 
-			allRunning = false
 			// If the Pod is not running after 60 seconds we delete it and let the Deployment controller create a new Pod.
```
09harsh, line 517:
```suggestion
			// If the Pod is not running after 120 seconds we delete it and let the Deployment controller create a new Pod.
```

e2e/fixtures/fdb_operator_client.go
```diff
@@ -507,80 +505,38 @@ func (factory *Factory) WaitUntilOperatorPodsRunning(namespace string) {
...
-			if time.Since(pod.CreationTimestamp.Time).Seconds() > 60.0 {
+			if time.Since(pod.CreationTimestamp.Time).Seconds() > 120.0 {
 				log.Println("operator Pod", pod.Name, "not running after 60 seconds, going to delete this Pod, status:", pod.Status)
```
09harsh, line 519:
```suggestion
				log.Println("operator Pod", pod.Name, "not running after 120 seconds, going to delete this Pod, status:", pod.Status)
```

09harsh (approved):
LGTM, just one change needed

Issue #985, opened 2021-11-15
Implement cache client label-selector


johscheuer:
We should add support for https://github.com/kubernetes-sigs/controller-runtime/pull/1435. This allows the client cache to be more efficient in clusters where multiple resources are running in the same namespace but only a small subset is managed by the operator.

PR #1922, opened 2024-01-22
Add cache client selector

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/985. This will reduce the cached resources to the once that matches the label selector if one is specified.

## Type of change

## Discussion

-

## Testing

Unit tests + CI will run e2e tests. I will run some additional tests.

## Documentation

-

## Follow-up

-

### DIFF
--- a/setup/setup.go
+++ b/setup/setup.go
@@ -25,16 +25,20 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"log"
 	"os"
 	"path"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


johscheuer (commented):
The tests are failing because some of the older versions are removed from the CI env. https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1923 will fix this.

AnnigeriShambu (approved):

09harsh (approved):

gm42:
@johscheuer I am seeing an odd issue with v1.33.0, which does not reproduce when reverting this change: when creating an empty cluster it fails to get the configmap for the cluster, and then fails to create it (because it actually already exists). Shall I open an issue with more details?

johscheuer:
> @johscheuer I am seeing an odd issue with v1.33.0, which does not reproduce when reverting this change: when creating an empty cluster it fails to get the configmap for the cluster, and then fails to create it (because it actually already exists). Shall I open an issue with more details?

That would be great. Thanks for reporting!

PR #1934, opened 2024-02-05
add ProcessGroupIdPrefix to testing generateClusterStruct and correct test cases

nicmorales9:
# Description

This PR sets `ProcessGroupIdPrefix` to the testing helper function `generateClusterStruct`, which makes it more correct for tests to refer to process group IDs with a <clustername> prefix, as they are doing right now.  This in turn makes test cases more consistent and easier to understand.

## Type of change

- Other: test bug fix?

## Discussion

Nope!

## Testing

unit-test fix

## Documentation

PR only includes unit test changes, I tried to clarify comments and names as needed

## Follow-up

N/A

### DIFF
--- a/kubectl-fdb/cmd/remove_process_group_test.go
+++ b/kubectl-fdb/cmd/remove_process_group_test.go
@@ -33,6 +33,7 @@ var _ = Describe("[plugin] remove process groups command", func() {
 	When("running remove process groups command", func() {
 		When("removing process groups from a cluster", func() {
 			BeforeEach(func() {
+				cluster.Spec.ProcessGroupIDPrefix = ""
 				cluster.Status = fdbv1beta2.FoundationDBClusterStatus{
 					ProcessGroups: []*fdbv1beta2.ProcessGroupStatus{
 						{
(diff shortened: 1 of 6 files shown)


kubectl-fdb/cmd/remove_process_group_test.go
```diff
@@ -33,6 +33,7 @@ var _ = Describe("[plugin] remove process groups command", func() {
 	When("running remove process groups command", func() {
 		When("removing process groups from a cluster", func() {
 			BeforeEach(func() {
+				cluster.Spec.ProcessGroupIDPrefix = ""
```
nicmorales9, line 36: This one is questionable whether or not I should be fixing all of the test cases or setting prefix to nothing.  It felt fine to be since the tests are setup this way and are consistent with no prefix, and it's not bad to test the empty prefix case too.  
I'll have some coverage on the non-empty prefix case in https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1930/files but let me know if it'd be more comfortable to have more test with the prefix set.

johscheuer: It's fine to keep the tests as they are, we can change them later if we want.

johscheuer (approved):
LGTM 👍

Issue #1702, opened 2023-06-24
Allow the remove sub command to remove by process class


johscheuer:
### What would you like to be added/changed?

In some cases it would be helpful to allow the remove process groups command to remove by process class.

PR #1935, opened 2024-02-07
Allow the remove sub command to remove by process class

nicmorales9:
# Description

This addresses https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1702

## Type of change

## Discussion

Let me know if the error raising is undesirable, also if a struct of common options for these commands might be helpful instead of the long list of parameters/options that includes many naked bools.

## Testing

Unit tests are in the PR and I tested it on some actual pods

I do not believe so, no.

## Documentation

I don't think this is needed beyond the docstrings

I do not think so, no

N/A

## Follow-up

No

No

### DIFF
--- a/kubectl-fdb/cmd/remove_process_group.go
+++ b/kubectl-fdb/cmd/remove_process_group.go
@@ -50,6 +50,10 @@ func newRemoveProcessGroupCmd(streams genericclioptions.IOStreams) *cobra.Comman
 			if err != nil {
 				return err
 			}
+			processClass, err := cmd.Flags().GetString("process-class")
+			if err != nil {
+				return err
+			}
 			withExclusion, err := cmd.Flags().GetBool("exclusion")
 			if err != nil {
 				return err
@@ -73,7 +77,7 @@ func newRemoveProcessGroupCmd(streams genericclioptions.IOStreams) *cobra.Comman
 				return err
 			}
 
-			return replaceProcessGroups(kubeClient, cluster, args, namespace, clusterLabel, withExclusion, wait, removeAllFailed, useProcessGroupID)
+			return replaceProcessGroups(kubeClient, cluster, args, namespace, clusterLabel, processClass, withExclusion, wait, removeAllFailed, useProcessGroupID)
 		},
 		Example: `
 # Remove process groups for a cluster in the current namespace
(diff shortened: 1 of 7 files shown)


kubectl-fdb/cmd/remove_process_group.go
```diff
@@ -109,10 +117,12 @@ kubectl fdb -n default remove process-group -c cluster --remove-all-failed
...
 
-// replaceProcessGroups adds process groups to the removal list of their respective clusters
-// if a clusterName is specified, it will ONLY do so for the specified cluster
-func replaceProcessGroups(kubeClient client.Client, clusterName string, ids []string, namespace string, clusterLabel string, withExclusion bool, wait bool, removeAllFailed bool, useProcessGroupID bool) error {
-	if len(ids) == 0 && !removeAllFailed {
+// replaceProcessGroups adds process groups to the removal list of their respective clusters.
+// If clusterName is specified, it will ONLY do so for the specified cluster.
+// If processClass is specified, it will ignore the given ids and remove all processes in the given cluster whose pods
+// have a processClassLabel matching the processClass.
+func replaceProcessGroups(kubeClient client.Client, clusterName string, ids []string, namespace string, clusterLabel string, processClass string, withExclusion bool, wait bool, removeAllFailed bool, useProcessGroupID bool) error {
```
nicmorales9, line 124: would a struct of common option fields be helpful or annoying to others? The list of parameters / options is getting lengthly

kubectl-fdb/cmd/remove_process_group.go
```diff
@@ -139,14 +149,22 @@ func replaceProcessGroups(kubeClient client.Client, clusterName string, ids []st
 		}
 		return err
 	}
-	// In this case the user has Pod name specified
+
 	var processGroupIDs []fdbv1beta2.ProcessGroupID
-	if !useProcessGroupID {
+	if processClass != "" { // match against a whole process class, ignore provided ids
+		if len(ids) != 0 {
+			return fmt.Errorf("process identifiers were provided along with a processClass and would be ignored, please only provide one or the other")
```
nicmorales9, line 156: the error handling is currently rather passive, but I felt like being explicit could be helpful here (and below on line 160). Let me know if this is not preferred, or if I should make other cases error more clearly.

kubectl-fdb/cmd/remove_process_group.go
johscheuer: I'm fine with creating a struct that keeps all parameters.

kubectl-fdb/cmd/remove_process_group.go
johscheuer: Sounds reasonable and makes sense to me 👍

kubectl-fdb/cmd/remove_process_group_test.go
```diff
@@ -178,15 +178,15 @@ var _ = Describe("[plugin] remove process groups command", func() {
 					podNames          []string
 					clusterNameFilter string // if used, then cross-cluster will not work
 					clusterLabel      string
-					RemoveAllFailed   bool
```
johscheuer, line 181: Was that option always false?

johscheuer (approved):
LGTM 👍

kubectl-fdb/cmd/remove_process_group_test.go
nicmorales9: yeah it wasn't looked at or used in those tests

AnnigeriShambu (approved):

Issue #1602, opened 2023-04-27
cordon command should print a summary


johscheuer:
### What would you like to be added/changed?

It would be nice to have a summary of the Pods remove from the cluster after a user ran the cordon command with `--wait=false`. Right know a user only sees the `Starting operation on ...` statements.

An idea how this could look like:

```text
Summary:

total replaced Pods: 12

Summary cluster X:

total replaced Pods: 5
Replaced Pods on node xxx: 4 [storage-1, stateless-3, ...]
Replaced Pods on node yyy: 1 [storage-9]

....
```

PR #1936, opened 2024-02-08
Improve verbosity of cordon and remove process commands

nicmorales9:
# Description

This improves the verbosity of the cordon and remove process commands. It should address https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1602 (if not exactly as described)

Example output for cordon (no wait):
```
Starting to cordon 2 nodes

Cordoning node: node-1
Cluster test/test2:
removed [test2-instance-1] (exclude: false)
Cluster test/test:
removed [test-instance-1] (exclude: false)

Cordoning node: node-2
Cluster test/test2:
removed [test2-instance-2] (exclude: false)
Cluster test/test:
removed [test-instance-2] (exclude: false)

Completed removal of 4 Pods
```

Example output for remove process-groups by cluster-label (with wait)
```
Remove [test-instance-1] from cluster test with exclude: true [y/n]: y
Cluster test:
removed [test-instance-1] (exclude: true)

Completed removal of 1 processGroups
```

## Type of change

- Cleanup

## Discussion

Error immediately or collate?

## Testing

Manual testing?*
No new tests, should be covered by old unit tests.  Verified with test output and on kubernetes.

No

## Documentation

N/A

## Follow-up

N/A

### DIFF
--- a/kubectl-fdb/cmd/cordon.go
+++ b/kubectl-fdb/cmd/cordon.go
@@ -137,7 +137,7 @@ func cordonNode(cmd *cobra.Command, kubeClient client.Client, inputClusterName s
 	for _, node := range nodes {
 		pods, err := fetchPodsOnNode(kubeClient, inputClusterName, namespace, node, clusterLabel)
 		if err != nil {
-			return fmt.Errorf("Issue fetching Pods running on node %s. Error: %s\n", node, err)
+			return fmt.Errorf("issue fetching Pods running on node %s. Error: %s", node, err)
 		}
 		var podNames []string
 		for _, pod := range pods.Items {
@@ -154,7 +154,7 @@ func cordonNode(cmd *cobra.Command, kubeClient client.Client, inputClusterName s
 			useProcessGroupID: false,
 		})
 		if err != nil {
-			return fmt.Errorf("Unable to cordon all Pods running on node %s. Error: %s\n", node, err.Error())
+			return fmt.Errorf("unable to cordon all Pods running on node %s. Error: %s", node, err.Error())
 		}
 		totalRemoved += removedFromNode
 	}
--- a/kubectl-fdb/cmd/cordon.go
+++ b/kubectl-fdb/cmd/cordon.go
@@ -132,18 +132,20 @@ func cordonNode(cmd *cobra.Command, kubeClient client.Client, inputClusterName s
 		return errors.New("no nodes were provided for cordoning")
 	}
 
+	var totalRemoved int
+
 	for _, node := range nodes {
 		pods, err := fetchPodsOnNode(kubeClient, inputClusterName, namespace, node, clusterLabel)
 		if err != nil {
-			return fmt.Errorf("Issue fetching Pods running on node: %s. Error: %s\n", node, err)
+			return fmt.Errorf("Issue fetching Pods running on node %s. Error: %s\n", node, err)
 		}
 		var podNames []string
 		for _, pod := range pods.Items {
 			podNames = append(podNames, pod.Name)
 		}
 
-		cmd.Printf("Cordoning node: %s\n", node)
-		err = replaceProcessGroups(cmd, kubeClient, inputClusterName, podNames, namespace, replaceProcessGroupsOptions{
+		cmd.Printf("\nCordoning node: %s\n", node)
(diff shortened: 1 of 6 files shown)


kubectl-fdb/cmd/cordon.go
```diff
@@ -141,42 +140,24 @@ func cordonNode(cmd *cobra.Command, kubeClient client.Client, inputClusterName s
...
-				}
-			}
-			err = replaceProcessGroups(kubeClient, cluster.Name, processGroups, namespace, "", "", withExclusion, wait, false, true)
-			if err != nil {
-				internalErr := fmt.Sprintf("unable to cordon all Pods for cluster %s\n", cluster.Name)
-				errors = append(errors, internalErr)
-				cmd.PrintErr(internalErr)
-			}
+		cmd.Printf("Cordoning node: %s\n", node)
+		err = replaceProcessGroups(cmd, kubeClient, inputClusterName, podNames, namespace, replaceProcessGroupsOptions{
```
nicmorales9, line 149: note: this swap would result in a behavioral difference: a failure to patch any of the clusters (or if the user says no) would have it error, rather than the current behavior which is to collate all the errors and have the user retry.   I think we should pick one behavior and have the node logic follow it (as-is, the node logic collates removal errors of non-collated cluster removal errors)

johscheuer: That change makes sense and simplifies the code 👍 Exiting early is also fine as the user can rerun the command again with the same result, e.g. the process group IDs will be added to the removal list.

kubectl-fdb/cmd/remove_process_group.go
```diff
@@ -117,62 +126,83 @@ kubectl fdb -n default remove process-group -c cluster --process-class="stateles
...
+
+// replaceProcessGroups adds process groups to the removal list of their respective clusters, and returns a count of
+// the number of removed processGroups (pods) and any encountered error.
 // If clusterName is specified, it will ONLY do so for the specified cluster.
 // If processClass is specified, it will ignore the given ids and remove all processes in the given cluster whose pods
 // have a processClassLabel matching the processClass.
-func replaceProcessGroups(kubeClient client.Client, clusterName string, ids []string, namespace string, clusterLabel string, processClass string, withExclusion bool, wait bool, removeAllFailed bool, useProcessGroupID bool) error {
-	if len(ids) == 0 && !removeAllFailed && processClass == "" {
-		return nil
+func replaceProcessGroups(cmd *cobra.Command, kubeClient client.Client, clusterName string, ids []string, namespace string, opts replaceProcessGroupsOptions) (int, error) {
```
nicmorales9, line 143: I don't love returning the total here but without it we can't get a removal total for cordon.  I think named return parameters could be helpful here, but I know some people don't like them.

kubectl-fdb/cmd/cordon.go
```diff
@@ -127,62 +127,38 @@ kubectl fdb cordon --node-selector machine=a,disk=fast -l fdb-cluster-label
...
+	var totalRemoved int
+
 	for _, node := range nodes {
 		pods, err := fetchPodsOnNode(kubeClient, inputClusterName, namespace, node, clusterLabel)
 		if err != nil {
-			internalErr := fmt.Sprintf("Issue fetching Pods running on node: %s. Error: %s\n", node, err)
-			cmd.PrintErr(internalErr)
-			errors = append(errors, internalErr)
-			continue
+			return fmt.Errorf("issue fetching Pods running on node %s. Error: %s", node, err)
```
09harsh, line 140: Is this intended to end the program here if we see error in any of the nodes ? 
Lets say hypothetically we are running it for 10 knodes and we see error in the first one, the previous implementation will work on the remaining 9 but the current one will stop the execution immediately.

nicmorales9: yes, discussion here: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1936#discussion_r1484164809 
Basically, to me, it seemed like we needed to converge on one strategy (collate errors or terminate on first).  Thread discusses it more but I lean towards making it obvious that the error occurred on the first node in your scenario, and having the user retry since the action is idempotent

09harsh: The new approach sounds good to me, just wanted to make sure we are leaning towards the new approach 👍

09harsh (approved):
👍 LGTM

johscheuer (approved):
LGTM 👍

Issue #615, opened 2021-04-15
Unify selection mechanism in the plugin


johscheuer:
To avoid duplication and different behaviour in the different commands we should ensure we are using the same "filter" mechanism for our plugin commands that interact with process groups (see: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/609#discussion_r613696114)

Issue #614, opened 2021-04-15
Add condition selector to remove process groups 


johscheuer:
I think it would bee useful to have a way to replace all process groups with a specific condition.

PR #1937, opened 2024-02-09
Create a unified process selection function and use it in removeProcessGroups

nicmorales9:
# Description

work towards https://github.com/FoundationDB/fdb-kubernetes-operator/issues/615 by creating a function which returns a map of processGroupIDs by FDB cluster based on process selection options.  
This PR only refactors removeProcessGroup's logic, but subsequent PRs will unify `buggify no-schedule`, `buggify crash-loop` and `restart` commands with the same selection function as well.  In order to meet feature-parity with `restart`, the selection function must have the ability to select by processGroup Condition, and thus will address https://github.com/FoundationDB/fdb-kubernetes-operator/issues/614 in a separate followup PR.

## Type of change

- Cleanup / Tech debt / New Feature

## Discussion

No

## Testing

Manual testing?*
Minor unit test changes, but this is mostly refactoring and is passively covered by existing tests

No

## Documentation

Not needed beyond docstring updates

## Follow-up

This does not solve https://github.com/FoundationDB/fdb-kubernetes-operator/issues/615 quite yet, but more PRs will follow.

No

### DIFF
--- a/kubectl-fdb/cmd/remove_process_group.go
+++ b/kubectl-fdb/cmd/remove_process_group.go
@@ -79,7 +79,7 @@ func newRemoveProcessGroupCmd(streams genericclioptions.IOStreams) *cobra.Comman
 			}
 
 			totalRemoved, err := replaceProcessGroups(cmd, kubeClient,
-				processSelectionOptions{
+				processGroupSelectionOptions{
 					ids:               args,
 					namespace:         namespace,
 					clusterName:       cluster,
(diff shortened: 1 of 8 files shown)


kubectl-fdb/cmd/remove_process_group.go
```diff
@@ -127,118 +134,78 @@ kubectl fdb -n default remove process-groups -c cluster --process-class="statele
...
-			err := replaceProcessGroupsFromCluster(kubeClient, cluster, pgs, namespace, opts.withExclusion, opts.wait, opts.removeAllFailed)
-			if err != nil {
-				return 0, err
-			}
-			totalRemoved += len(pgs)
-			cmd.Printf("Cluster %v/%v:\n", namespace, cluster.Name)
-			cmd.Printf("removed %v (exclude: %t)\n", pgs, opts.withExclusion)
-		}
-		return totalRemoved, nil
+func replaceProcessGroups(cmd *cobra.Command, kubeClient client.Client, pgSelectOpts processSelectionOptions, opts replaceProcessGroupsOptions) (int, error) {
```
johscheuer, line 147: I believe in most cases we tended to rather have longer names instead of abbreviations like `pg`.

kubectl-fdb/cmd/root.go
```diff
@@ -175,3 +176,58 @@ func usingLatestPluginVersion(cmd *cobra.Command, pluginVersionChecker VersionCh
...
+	namespace         string
+	clusterName       string
+	clusterLabel      string
+	processClass      string
+	useProcessGroupID bool
+}
+
+// getProcessGroupsByCluster returns a map of processGroupIDs by FDB cluster that match the criteria in the provided
+// processSelectionOptions.
+func getProcessGroupsByCluster(kubeClient client.Client, opts processSelectionOptions) (map[*fdbv1beta2.FoundationDBCluster][]fdbv1beta2.ProcessGroupID, error) {
```
johscheuer, line 191: Most of those selection and "util" methods are in the `k8s_client` file: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/kubectl-fdb/cmd/k8s_client.go.

Should we add dedicated tests for this method?

kubectl-fdb/cmd/remove_process_group.go
nicmorales9: Ack! wasn't sure here since it gets lengthly, I'll start defaulting to longer then 😄

kubectl-fdb/cmd/root.go
nicmorales9: not a bad idea, I think it's decently tested via the remove process group one right now but a basic set for it sounds good

johscheuer (approved):
LGTM 👍

Issue #1890, opened 2023-11-21
Replacement of failed processes without and address with locality-based exclusions


johscheuer:
### What would you like to be added/changed?

In our automatic replacement logic for failed processes we have a special check to see if a process group has an address assigned: 
https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/internal/replacements/replace_failed_process_groups.go#L95. If no addresses are assigned the exclusion will be skipped. This brings the potential risk that a Pod might get scheduled with a huge delay, but get's schedule shortly after the operator detected it as failed without an address. To mitigate this, we should always require an exclusion with locality-based exclusions, since the locality is known upfront we can make use of that. The use of the IP address is only possible once the Pod was actually scheduled and received an IP address.

PR #1938, opened 2024-02-14
Don't skip the exclusion for process groups without an address when using localities

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1890

## Type of change

## Discussion

-

## Testing

Added an additional test and CI will run e2e tests.

## Documentation

-

## Follow-up

-

### DIFF
--- a/internal/replacements/replace_failed_process_groups.go
+++ b/internal/replacements/replace_failed_process_groups.go
@@ -54,6 +54,7 @@ func ReplaceFailedProcessGroups(log logr.Logger, cluster *fdbv1beta2.FoundationD
 	maxReplacements := getMaxReplacements(cluster, cluster.GetMaxConcurrentAutomaticReplacements())
 	hasReplacement := false
 	crashLoopContainerProcessGroups := cluster.GetCrashLoopContainerProcessGroups()
+	localitiesUsedForExclusion := cluster.UseLocalitiesForExclusion()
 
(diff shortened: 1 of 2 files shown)


internal/replacements/replace_failed_process_groups.go
```diff
@@ -92,7 +93,9 @@ func ReplaceFailedProcessGroups(log logr.Logger, cluster *fdbv1beta2.FoundationD
 		}
 
 		skipExclusion := false
-		if len(processGroupStatus.Addresses) == 0 {
+		// Only if localities are not used for exclusions we want to skip the exclusion. Skipping the exclusion could
+		// lead to a race condition, see: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1890
```
nicmorales9, line 97: If I understand correctly, the race condition can still occur here, it's just that it can be prevented when we are able to exclude by locality, right?  If true, I think it would be slightly clearer worded like this:
```suggestion
		// Only if localities are not used for exclusions do we want to skip the exclusion. Skipping the exclusion could
		// lead to a race condition, which can be prevented if we are able to exclude by locality.
		// see: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1890
```

nicmorales9 (approved):
Code looks good to me! I'll leave the comment suggestion up to you

PR #1940, opened 2024-02-14
Add condition selection option to common processGroup selection function

nicmorales9:
The common function is currently only used by remove-process-group.  This solves https://github.com/FoundationDB/fdb-kubernetes-operator/issues/614.
It also includes some refactoring of the error logic in the unified function -- I tried to make it as clear as possible for end users and it definitely assumes that this is on CLI use.  Definitely open to complaints!
This is work towards https://github.com/FoundationDB/fdb-kubernetes-operator/issues/615; the next PR will involve having all the process-group selection commands (`buggify no-schedule`, `buggify crash-loop`, and `restart`) use the common function now that it supports all the selection options with this PR.

# Description

## Type of change

- Cleanup / Tech debt / New Feature

## Discussion

## Testing

Manual testing?*
Added unit tests, manually (via `buggify crash-loop`)

No

## Documentation

I updated the docstrings, error, and help messages to the best of my ability

## Follow-up

No

No

### DIFF
--- a/api/v1beta2/foundationdbcluster_types.go
+++ b/api/v1beta2/foundationdbcluster_types.go
@@ -964,10 +964,14 @@ func GetProcessGroupConditionType(processGroupConditionType string) (ProcessGrou
 		return MissingService, nil
 	case "MissingProcesses":
 		return MissingProcesses, nil
+	case "ResourcesTerminating":
+		return ResourcesTerminating, nil
 	case "SidecarUnreachable":
 		return SidecarUnreachable, nil
 	case "PodPending":
 		return PodPending, nil
+	case "Ready":
+		return ReadyCondition, nil
 	case "NodeTaintDetected":
 		return NodeTaintDetected, nil
 	case "NodeTaintReplacing":
--- a/kubectl-fdb/cmd/k8s_client.go
+++ b/kubectl-fdb/cmd/k8s_client.go
@@ -400,7 +400,7 @@ func getProcessGroupsByCluster(cmd *cobra.Command, kubeClient client.Client, opt
 	if opts.clusterName == "" && opts.clusterLabel == "" {
 		return nil, errors.New("processGroups will not be selected without cluster specification")
 	}
-	if opts.clusterName == "" { // cli has a valid default for clusterLabel
+	if opts.clusterName == "" { // cli has a default for clusterLabel
 		if opts.useProcessGroupID || opts.processClass != "" || len(opts.conditions) > 0 {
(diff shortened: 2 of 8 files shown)


api/v1beta2/foundationdbcluster_types.go
```diff
@@ -964,10 +964,14 @@ func GetProcessGroupConditionType(processGroupConditionType string) (ProcessGrou
 		return MissingService, nil
 	case "MissingProcesses":
 		return MissingProcesses, nil
+	case "ResourcesTerminating":
+		return ResourcesTerminating, nil
 	case "SidecarUnreachable":
 		return SidecarUnreachable, nil
 	case "PodPending":
 		return PodPending, nil
+	case "Ready":
```
johscheuer, line 973: The ready condition is never set, I believe that's the reason why it's absent here but it doesn't hurt :)Just wanted to give you the context here.

kubectl-fdb/cmd/k8s_client.go
```diff
@@ -393,17 +394,27 @@ func fetchPodsOnNode(kubeClient client.Client, clusterName string, namespace str
...
 // getProcessGroupsByCluster returns a map of processGroupIDs by FDB cluster that match the criteria in the provided
-// processGroupSelectionOptions.
-func getProcessGroupsByCluster(kubeClient client.Client, opts processGroupSelectionOptions) (map[*fdbv1beta2.FoundationDBCluster][]fdbv1beta2.ProcessGroupID, error) {
+// processSelectionOptions.
+func getProcessGroupsByCluster(cmd *cobra.Command, kubeClient client.Client, opts processGroupSelectionOptions) (map[*fdbv1beta2.FoundationDBCluster][]fdbv1beta2.ProcessGroupID, error) {
+	// option compatibility checks
 	if opts.clusterName == "" && opts.clusterLabel == "" {
 		return nil, errors.New("processGroups will not be selected without cluster specification")
 	}
+	if opts.clusterName == "" { // cli has a valid default for clusterLabel
```
johscheuer, line 403:
```suggestion
	if opts.clusterName == "" { // cli has a default for clusterLabel
```

I would remove the `valid` term as it depends on how the cluster where created.

johscheuer (approved):
LGTM 👍

api/v1beta2/foundationdbcluster_types.go
nicmorales9: I saw that! but it did make sense to me to have all the conditions here

Issue #1369, opened 2022-10-03
Write design for fault domain based replacements


johscheuer:
### What would you like to be added/changed?

The current automatic replacement implementation has the limitation that it only works on a per process group level. That means that the operator is not efficient at replacing multiple process groups when a host or a fault domain fails. Instead of concurrently replacing those process groups the operator will replace them one by one. We should extend the current implementation to focus on fault domains (or hosts) instead of single process groups.

PR #1943, opened 2024-02-16
Add support for replacing failed process groups in a fault domain

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1369

## Type of change

## Discussion

-

## Testing

Will add some additional unit tests and e2e tests for this feature.

## Documentation

Will be updated.

## Follow-up

-


api/v1beta2/foundationdbcluster_types.go
```diff
@@ -1215,6 +1215,12 @@ type AutomaticReplacementOptions struct {
 	// The default is false.
 	Enabled *bool `json:"enabled,omitempty"`
 
+	// FaultDomainBasedReplacements controls whether automatic replacements are targeting all failed process groups
+	// in a fault domain or only specific Process Groups. If this setting is enabled MaxConcurrentReplacements will
+	// provide the information in how many fault domains failed process groups can be replaced.
+	// The default is true.
+	FaultDomainBasedReplacements *bool `json:"faultDomainBasedReplacements,omitempty"`
```
johscheuer, line 1222: We have to run `make clean all` to generate the new docs and we have to update the according docs.

internal/replacements/replace_failed_process_groups.go
```diff
@@ -37,10 +40,30 @@ func getMaxReplacements(cluster *fdbv1beta2.FoundationDBCluster, maxReplacements
...
+}
+
+// removalAllowed will return true if the removal is allowed based on the clusters automatic replacement configuration.
+func removalAllowed(cluster *fdbv1beta2.FoundationDBCluster, maxReplacements int, faultDomainsWithReplacements map[fdbv1beta2.FaultDomain]fdbv1beta2.None, faultDomain fdbv1beta2.FaultDomain) bool {
+	if cluster.FaultDomainBasedReplacements() {
+		_, faultDomainHasReplacements := faultDomainsWithReplacements[faultDomain]
+		// The fault domain has already a replacement ongoing, so we are able to approve the replacement of the failed
+		// process group.
+		if faultDomainHasReplacements {
+			return true
```
nicmorales9, line 57: Still learning obviously, but this *feels* worrisome to me as we could end up exceeding the max concurrent replacements to a much greater degree when we allow all the processes in it go down.  
In my head I'm imagining a scenario where one processGroup in each fault domain has problems, is marked for replacement, but that leads to all fault domains getting an OK to replace all processGroups if that's somehow possible?

nicmorales9: ok that makes sense

internal/replacements/replace_failed_process_groups_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+import (
+	fdbv1beta2 "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta2"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"k8s.io/utils/pointer"
+)
+
+var _ = Describe("replace_failed_process_groups", func() {
+	DescribeTable("check if removal is allowed", func(cluster *fdbv1beta2.FoundationDBCluster, maxReplacements int, faultDomainsWithReplacements map[fdbv1beta2.FaultDomain]fdbv1beta2.None, faultDomain fdbv1beta2.FaultDomain, expected bool) {
```
nicmorales9, line 31: nit: I think this could be a lot more readable with a testCase struct - but this only really matters outside an IDE

nicmorales9 (approved):
LGTM, thanks for clarifying things!

internal/replacements/replace_failed_process_groups_test.go
johscheuer: We can change that later if we want.

PR #1941, opened 2024-02-15
use common process selection function in buggify crash-loop, buggify no-schedule, and restart commands

nicmorales9:
# Description

This PR should end up resolving https://github.com/FoundationDB/fdb-kubernetes-operator/issues/615 by having all the process-group selection commands (buggify no-schedule, buggify crash-loop, and restart) use the common function that remove-process-groups has come to use in previous PRs https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1937 and https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1940 

## Type of change

- Tech Debt / New feature (non-breaking change which adds functionality)

## Discussion

*Are there any design details that you would like to discuss further?*
Not particularly

## Testing

*Please describe the tests that you ran to verify your changes. Unit tests?
Manual testing?*
Unit tests, manually (but mostly spotchecked as there are so many possible combinations)

*Do we need to perform additional testing once this is merged, or perform in a larger testing environment?*
No

## Documentation

docstrings, help strings, and comments have been updated accordingly

## Follow-up

*Are there any follow-up issues that we should pursue in the future?*
None related to this change

*Does this introduce new defaults that we should re-evaluate in the future?*
No

PR #1944, opened 2024-02-19
add podName selection functions

nicmorales9:
# Description

This is just a followup to https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1941 which used an unsatisfying way to get podNames from processGroupIDs.  

I am not sure that this is much better, but they do seem like useful functions in the transition away from selecting by processGroupID.  

**It will be much easier to see the changes if you view by commit to avoid looking at the second commit,** which changes test pod names to include processClass in the name, which is an expectation of built-in functions (GetPodName will index-out-of-bounds).

## Type of change

- Code Cleanup

## Discussion

I wonder if there should be some explicit creation function which does not allow FDB pods to have invalid names, as I suspect that it is not only GetPodName which could have undefined behavior if expectations aren't met. 

## Testing

Manual testing?*
Unit tests, manual test of restart.

## Documentation

Docstrings have been added

## Follow-up

Arguably this could be progress towards not allowing useProcessGroupID and working more with podNames when possible, but we had trouble finding that issue before.  Perhaps we should raise one?  

No

### DIFF
--- a/kubectl-fdb/cmd/analyze_test.go
+++ b/kubectl-fdb/cmd/analyze_test.go
@@ -66,12 +66,12 @@ func getPodList(clusterName string, namespace string, status corev1.PodStatus, d
 		Items: []corev1.Pod{
 			{
 				ObjectMeta: metav1.ObjectMeta{
-					Name:      "instance-1",
+					Name:      "storage-1",
 					Namespace: namespace,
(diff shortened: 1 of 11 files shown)


johscheuer (approved):
LGTM 👍 Thanks!

Issue #1918, opened 2024-01-12
maxConcurrentReplacements causing deletion update strategy


simenl:
### What happened?

As a mitigation to storage roles being recruited to log processes [[forum post](https://forums.foundationdb.org/t/never-assign-storage-role-to-log-processes/4294)], we tried setting [maxConcurrentReplacements](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/73270bac4f6fa7faab4f7820c64c618d5a5f1b11/docs/manual/replacements_and_deletions.md?plain=1#L27) to reduce the number of concurrent exclusions.
However, this caused the [Deletion strategy](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/73270bac4f6fa7faab4f7820c64c618d5a5f1b11/docs/manual/replacements_and_deletions.md?plain=1#L27) to incorrectly be applied on the remaining processes, for updates that requires the Replacement strategy. Consequently this resulted in unschedulable pods, as we updated the node selector to an availability zone incompatible with the existing persistent volume on the process.

### What did you expect to happen?

Processes that requires replacement, should not be eligible for the delete update strategy. Even if they were not selected for replacement (yet) due to maxConcurrentReplacements.

### How can we reproduce it (as minimally and precisely as possible)?

1. Create a foundationdb cluster with `maxConcurrentReplacements` and multiple storage pods/processes:
```
spec:
  automationOptions:
    maxConcurrentReplacements: 1
```

2. Make an update to the CRD that requires a [replacement](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/dcefba04c87dbdbc4a9ca9aeacc098da1eb56a89/docs/manual/replacements_and_deletions.md?plain=1#L18). E.g. changing the the node selector.

3. Observe that some of the storage pods will be updated through the delete update strategy

### FDB Kubernetes operator

We run FoundationDB on kubernetes through the [fdb-kubernetes-operator: v1.28.1](https://github.com/foundationDB/fdb-kubernetes-operator).
FoundationDB version: [7.1.43](https://github.com/apple/foundationdb/blob/main/documentation/sphinx/source/release-notes/release-notes-710.rst#7143)

### Kubernetes version

version.Info{Major:"1", Minor:"26", GitVersion:"v1.26.7-gke.500"}

### Cloud provider

GCP

johscheuer:
Thanks for the report, do you think you are able to work on a fix and test for this?

[...]

PR #1954, opened 2024-03-03
Update pods check needs removal

simenl:
# Description

Fixes bug where delete strategy is applied on pods not marked for removal in the replaceMisconfiguredProcessGroups step, https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1918, by checking if the pod needs removal in the updatePods reconciler.

## Type of change

## Discussion

## Testing
In addition to the ginkgo tests, also tested in a kubernetes deployment, verifying that update deletion strategy will not be applied if the update requires a removal / replacement.

### DIFF
--- a/e2e/test_operator/operator_test.go
+++ b/e2e/test_operator/operator_test.go
@@ -2006,17 +2006,17 @@ var _ = Describe("Operator", Label("e2e", "pr"), func() {
 
 		When("a change that requires a replacement of all storage pods", func() {
 			var initialVolumeClaims *corev1.PersistentVolumeClaimList
-			var newCpuRequest, initialCpuRequest resource.Quantity
+			var newCPURequest, initialCPURequest resource.Quantity
 			BeforeEach(func() {
 
 				initialVolumeClaims = fdbCluster.GetVolumeClaimsForProcesses(fdbv1beta2.ProcessClassStorage)
 				spec := fdbCluster.GetCluster().Spec.DeepCopy()
-				initialCpuRequest = spec.Processes[fdbv1beta2.ProcessClassStorage].PodTemplate.Spec.Containers[0].Resources.Requests["cpu"]
-				newCpuRequest = initialCpuRequest.DeepCopy()
+				initialCPURequest = spec.Processes[fdbv1beta2.ProcessClassStorage].PodTemplate.Spec.Containers[0].Resources.Requests["cpu"]
+				newCPURequest = initialCPURequest.DeepCopy()
 
 				// An increase in request requires a replacement when ReplaceInstancesWhenResourcesChange is set to true
-				newCpuRequest.Add(resource.MustParse("1m"))
-				spec.Processes[fdbv1beta2.ProcessClassStorage].PodTemplate.Spec.Containers[0].Resources.Requests["cpu"] = newCpuRequest
+				newCPURequest.Add(resource.MustParse("1m"))
+				spec.Processes[fdbv1beta2.ProcessClassStorage].PodTemplate.Spec.Containers[0].Resources.Requests["cpu"] = newCPURequest
 
 				fdbCluster.UpdateClusterSpecWithSpec(spec)
 
@@ -2027,7 +2027,7 @@ var _ = Describe("Operator", Label("e2e", "pr"), func() {
 			AfterEach(func() {
 				// Undo the change to cpu requests
 				spec := fdbCluster.GetCluster().Spec.DeepCopy()
-				spec.Processes[fdbv1beta2.ProcessClassStorage].PodTemplate.Spec.Containers[0].Resources.Requests["cpu"] = initialCpuRequest
+				spec.Processes[fdbv1beta2.ProcessClassStorage].PodTemplate.Spec.Containers[0].Resources.Requests["cpu"] = initialCPURequest
 				fdbCluster.UpdateClusterSpecWithSpec(spec)
 
 				// Wait for the reconciliation to finish
--- a/e2e/test_operator/operator_test.go
+++ b/e2e/test_operator/operator_test.go
@@ -1976,4 +1976,71 @@ var _ = Describe("Operator", Label("e2e", "pr"), func() {
 			})
 		})
 	})
+
+	When("maxConcurrentReplacements is lower than the number of storage pods", func() {
+		var initialConcurrentReplacements *int
+		var initialPodUpdateStrategy fdbv1beta2.PodUpdateStrategy
+		var initialReplaceInstancesWhenResourcesChange *bool
+
+		BeforeEach(func() {
+			// Remember the current settings before updating the spec
+			initialConcurrentReplacements = fdbCluster.GetCluster().Spec.AutomationOptions.MaxConcurrentReplacements
+			initialPodUpdateStrategy = fdbCluster.GetClusterSpec().AutomationOptions.PodUpdateStrategy
+			initialReplaceInstancesWhenResourcesChange = fdbCluster.GetCluster().Spec.ReplaceInstancesWhenResourcesChange
+			// Allow to replacement of 3 pods concurrently, as there are 5 storage servers, there need to be at least 2 rounds of replacements to replace all.
+			spec := fdbCluster.GetCluster().Spec.DeepCopy()
+			spec.AutomationOptions.MaxConcurrentReplacements = pointer.Int(3)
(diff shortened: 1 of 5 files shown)


internal/replacements/replacements.go
```diff
@@ -94,9 +70,43 @@ func ReplaceMisconfiguredProcessGroups(ctx context.Context, podManager podmanage
...
+		return false, podErr
+	}
+
+	needsRemoval, err := processGroupNeedsRemovalForPod(cluster, pod, processGroup, log)
+
+	if err != nil {
+		return false, err
+	}
+
+	return needsRemoval, nil
```
johscheuer, line 104:
```suggestion
	return  processGroupNeedsRemovalForPod(cluster, pod, processGroup, log)
```

We can simplify the return logic here.

internal/replacements/replacements.go
```diff
@@ -94,9 +70,43 @@ func ReplaceMisconfiguredProcessGroups(ctx context.Context, podManager podmanage
...
+}
+
 func processGroupNeedsRemovalForPVC(cluster *fdbv1beta2.FoundationDBCluster, pvc corev1.PersistentVolumeClaim, log logr.Logger, processGroup *fdbv1beta2.ProcessGroupStatus) (bool, error) {
 	processGroupID := internal.GetProcessGroupIDFromMeta(cluster, pvc.ObjectMeta)
-	logger := log.WithValues("namespace", cluster.Namespace, "cluster", cluster.Name, "pvc", pvc.Name, "processGroupID", processGroupID, "reconciler", "replaceMisconfiguredProcessGroups")
+	logger := log.WithValues("namespace", cluster.Namespace, "cluster", cluster.Name, "pvc", pvc.Name, "processGroupID", processGroupID)
```
johscheuer, line 109: We probably can also remove the `namespace` and `cluster` I believe those values are already added. We probably can move the logger creation into the `ProcessGroupNeedsRemoval` method (just omitting the pvc name) and pass down that logger into the `processGroupNeedsRemovalForPVC` and ` processGroupNeedsRemovalForPod` methods.

internal/replacements/replacements.go
```diff
@@ -153,7 +163,7 @@ func processGroupNeedsRemoval(cluster *fdbv1beta2.FoundationDBCluster, pod *core
 
 	_, desiredProcessGroupID := cluster.GetProcessGroupID(processGroupStatus.ProcessClass, idNum)
 	if processGroupStatus.ProcessGroupID != desiredProcessGroupID {
-		logger.Info("Replace process group",
+		logger.Info("Process group requires removal",
```
johscheuer, line 166: What is the benefit of changing those logs? That could break some existing dashboard/alerts based on this logging statement. In general I would favour the term `Process group requires replacement`.

johscheuer (commented):
Could you try to add an additional test case for this in our e2e test framework? We should be adding this to the operator tests here: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/e2e/test_operator/operator_test.go

internal/replacements/replacements.go
simenl: The function is also used by updatePods. `"Replace process group"` is a bit misleading log message from the updatePods reconciler, which will not perform an update (deletion) in that case.
We could log it higher up in the ReplaceMisconfiguredProcessGroups function (if needsRemoval == true).

Regarding replacement vs removal. Would it make sense to refactor the methods from processGroupNeeds**Removal*** to processGroupNeeds**Replacement*** and replace (😄) other occurrences of removal with replacement?

johscheuer: > The function is also used by updatePods. "Replace process group" is a bit misleading log message from the updatePods reconciler, which will not perform an update (deletion) in that case.
We could log it higher up in the ReplaceMisconfiguredProcessGroups function (if needsRemoval == true).

It's probably fine enough to use the combination of this log message and the `ReplaceMisconfiguredProcessGroups` reconciler in the log message to figure out that those messages mean that the process group gets replaced.

> Regarding replacement vs removal. Would it make sense to refactor the methods from processGroupNeedsRemoval* to processGroupNeedsReplacement* and replace (😄) other occurrences of removal with replacement?

I think that would make sense but we should do that in a different PR.

johscheuer (commented):
Could you make sure that the tests are passing? Based on the CI output you have to run the `make fmt` command. In general it's a good idea to run `make fmt lint test` before pushing the changes.

simenl:
@johscheuer
I've adjusted the PR, addressing the comments. Note: Only removed the reconciler part from `logger.withValues`.

(I haven't added e2e tests, this requires more investment from my side. Do you have any recommendations for reducing the feedback loop when implementing a new test?)

johscheuer (commented):
> @johscheuer
I've adjusted the PR, addressing the comments. Note: Only removed the reconciler part from logger.withValues.

Thanks, I take a look today.

> (I haven't added e2e tests, this requires more investment from my side. Do you have any recommendations for reducing the feedback loop when implementing a new test?)

You can run a single e2e test with prefixing the test case with `F`, that's the normal Ginkgo feature. If you run frequent tests, it might be faster to reuse ac cluster: https://github.com/FoundationDB/fdb-kubernetes-operator/tree/main/e2e#reusing-an-existing-test-cluster. Depending on your test, you just have to make sure the cluster is in a good shape again.

johscheuer (commented):
The changes LGTM. It would be great if you could add an e2e test for this behaviour.

simenl:
Added an e2e tests which:
1. Sets `maxConcurrentReplacements=3` and `replaceInstancesWhenResourcesChange=true`
2. Change the spec for storage pods, by increasing the cpu request of the foundationdb container
3. Wait for reconciliation. 
4. Check if there is no overlap between current set of PVCs and the previous set of PVCs
a. Update by replacement will use a new PVC
b. Update by deletion will reuse the PVC

In the current version, replacement will start on 3 of the storage pods, while the remaining 2 are updated through deletion.
With this PR, replacement will start on 3 of the storage pods, and the remaining storage pods **will not** be updated through deletion, but wait for replacement. 

_For step 4, it might be more robust to compare names of the volumes referred by the PVC, and not the whole PVC object_

e2e/test_operator/operator_test.go
```diff
@@ -1976,4 +1976,71 @@ var _ = Describe("Operator", Label("e2e", "pr"), func() {
...
+			spec.ReplaceInstancesWhenResourcesChange = initialReplaceInstancesWhenResourcesChange
+			fdbCluster.UpdateClusterSpecWithSpec(spec)
+		})
+
+		When("a change that requires a replacement of all storage pods", func() {
+			var initialVolumeClaims *corev1.PersistentVolumeClaimList
+			var newCPURequest, initialCPURequest resource.Quantity
+			BeforeEach(func() {
+
+				initialVolumeClaims = fdbCluster.GetVolumeClaimsForProcesses(fdbv1beta2.ProcessClassStorage)
```
johscheuer, line 2012:
```suggestion
			BeforeEach(func() {
				initialVolumeClaims = fdbCluster.GetVolumeClaimsForProcesses(fdbv1beta2.ProcessClassStorage)
```

It's probably better to store the UID in a slice, that makes the comparison easier and also will prevent cases where the comparison is failing because of a different pointer.

e2e/test_operator/operator_test.go
```diff
@@ -1976,4 +1976,71 @@ var _ = Describe("Operator", Label("e2e", "pr"), func() {
...
+				// Wait for the reconciliation to finish
+				Expect(fdbCluster.WaitForReconciliation()).NotTo(HaveOccurred())
+			})
+
+			It("should replace all storage pods", func() {
+				// A replacement of a storage pod will create a new PVC. After reconciliation the set of PVCs should be completely changed.
+				volumeClaims := fdbCluster.GetVolumeClaimsForProcesses(fdbv1beta2.ProcessClassStorage)
+				for _, initialVolumeClaim := range initialVolumeClaims.Items {
+					Expect(volumeClaims.Items).NotTo(ContainElement(initialVolumeClaim), "PVC should not be present in the new set of PVCs")
+				}
```
johscheuer, line 2042: When we fetch the UID of the PVC this would be simply a call like this:

```
Expect(initialVolumeClaimIDs).ToNot(ContainElements(currentVolumeClaimIDs))
```

johscheuer (approved):
Changes LGTM 👍 I left a comment on the e2e test but it's only informational.

Thanks for the PR!

Issue #1950, opened 2024-02-23
Fix race condition in FDB go bindings


johscheuer:
### What happened?

See: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/1932#issuecomment-1960821984, current PR: https://github.com/apple/foundationdb/pull/11221. The error is:

```text
  goroutine 312 [syscall]:
  runtime.cgocall(0x15c3390, 0xc0009d5780)
  	/usr/local/go/src/runtime/cgocall.go:157 +0x5c fp=0xc0009d5758 sp=0xc0009d5720 pc=0x40887c
  github.com/apple/foundationdb/bindings/go/src/fdb._Cfunc_fdb_run_network()
  	_cgo_gotypes.go:378 +0x48 fp=0xc0009d5780 sp=0xc0009d5758 pc=0x139f3e8
  github.com/apple/foundationdb/bindings/go/src/fdb.startNetwork.func1()
  	/go/pkg/mod/github.com/johscheuer/foundationdb/bindings/go@v0.0.0-20240223074533-f06c55dce3a5/src/fdb/fdb.go:217 +0x1d fp=0xc0009d57e0 sp=0xc0009d5780 pc=0x13a841d
  runtime.goexit()
  	/usr/local/go/src/runtime/asm_amd64.s:1598 +0x1 fp=0xc0009d57e8 sp=0xc0009d57e0 pc=0x470141
  created by github.com/apple/foundationdb/bindings/go/src/fdb.startNetwork
  	/go/pkg/mod/github.com/johscheuer/foundationdb/bindings/go@v0.0.0-20240223074533-f06c55dce3a5/src/fdb/fdb.go:216 +0x165
```

Since we got a repeatable test setup we have to invest on how to solve that issue.

### What did you expect to happen?

FDB bindings shouldn't crash. Those are the last printed log lines:

```text
  Error determining public address.
  fatal error: unexpected signal during runtime execution
  Error determining public address.
  [signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x7f8c78ded03a]
```

### How can we reproduce it (as minimally and precisely as possible)?

See the PR above.

### Anything else we need to know?

There was an attempt in the past to fix that issue: https://github.com/apple/foundationdb/pull/11104 but that didn't solve the issue.

### FDB Kubernetes operator

<details>

latest version.

</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

[...]

PR #1956, opened 2024-03-05
Document issue with older fdb bindings and DNS support

johscheuer:
# Description

Documents the limitations of the copy approach for the primary FDB lib, see: https://github.com/apple/foundationdb/issues/11222

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1950

## Type of change

- Documentation

## Discussion

I also added `tini` to the operator image to allow the manager to generate coredumps, if the process is PID 1 it doesn't work.

## Testing

-

## Documentation

Updated.

## Follow-up

-

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,14 +54,6 @@ VOLUME /usr/lib/fdb
 
 WORKDIR /
 
-# Add tini to the operator image to allow to take coredumps.
-ENV TINI_VERSION v0.19.0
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -54,6 +54,14 @@ VOLUME /usr/lib/fdb
 
 WORKDIR /
 
+# Add tini to the operator image to allow to take coredumps.
+ENV TINI_VERSION v0.19.0
```
johscheuer, line 58: I'll remove the tini setup from this PR and create a new one.

brownleej (approved):
👍

PR #1957, opened 2024-03-05
Minor typo fix in mention of localities

gm42:
# Description

This is a minor typo fix in the documentation (technical design).

## Type of change

- Documentation

### DIFF
--- a/docs/manual/technical_design.md
+++ b/docs/manual/technical_design.md
@@ -445,10 +445,7 @@ If a process is missing in the machine-readable status the operator will issue a
 The current default for the operator is to use the Pod IP for the exclusion command, if a Pod get's deleted and recreated it could get a new IP address and the operator has to issue a new exclude command for the new IP address.
 To workaround this FoundationDB added support for locality based exclusions in 7.0 and the operator supports this by setting [useLocalitiesForExclusion](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/docs/cluster_spec.md#foundationdbclusterautomationoptions) in the FoundationDBCluster spec.
 
-**Warning** the locality-based exclusions are not well tested in our e2e test setup ye.
-locality based exclusion
-
-The operator is not able to use the `failed` option for exclusions.
+NOTE: the operator is not able to use the `failed` option for exclusions.
 
 ## Next
 


docs/manual/technical_design.md
```diff
@@ -445,8 +445,7 @@ If a process is missing in the machine-readable status the operator will issue a
 The current default for the operator is to use the Pod IP for the exclusion command, if a Pod get's deleted and recreated it could get a new IP address and the operator has to issue a new exclude command for the new IP address.
 To workaround this FoundationDB added support for locality based exclusions in 7.0 and the operator supports this by setting [useLocalitiesForExclusion](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/docs/cluster_spec.md#foundationdbclusterautomationoptions) in the FoundationDBCluster spec.
 
-**Warning** the locality-based exclusions are not well tested in our e2e test setup ye.
-locality based exclusion
+**Warning** the locality-based exclusions are not well tested in our e2e test setup yet.
```
johscheuer, line 448: Can we actually drop this statement as we have multiple tests for this feature now and we just forgot to remove it :)

johscheuer (commented):
Thanks for the PR!

docs/manual/technical_design.md
gm42: Ha! Good to know :)

I have left the second statement, is it still valid?

johscheuer (approved):
LGTM 👍

gm42:
Thanks @johscheuer!

I have a related question: I am creating a cluster using the `faultDomain` as described [here](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/0d2046a/docs/manual/fault_domains.md#option-1-single-kubernetes-replication). I copy/pasted that block in my FDB definition, even if documentation mentions that it is the default and thus should be unnecessary.

However pods are still getting scheduled on same nodes, and I do not see any topology spread / pod anti-affinity being generated in the pod YAML definitions created by the operator. I also tried to find references to this configuration in the codebase of the operator, but I can't find anything related to the default as mentioned there :thinking: instead I see plenty of checks for the key `foundationdb.org/kubernetes-cluster`.

Do you have any pointers for me to understand better how the fault domain feature works, or perhaps existing E2E tests covering this?

gm42:
Sure, I wanted to make sure that there is a problem first, by checking the relevant code. I will try to reproduce it and then open an issue.

PR #1960, opened 2024-03-08
Allow more addresses to be present in the upgrade bounce case

johscheuer:
# Description

We have seen a case in our nightly runs where one process was marked for removal but not yet fully excluded. I believe in such cases it's fine to have more addresses than the expected list, as it could take a while until a process is fully excluded,

## Type of change

## Discussion

-

## Testing

Ran unit tests and CI will run e2e tests.

## Documentation

-

## Follow-up

-

### DIFF
--- a/controllers/bounce_processes.go
+++ b/controllers/bounce_processes.go
@@ -263,15 +263,17 @@ func getProcessesReadyForRestart(logger logr.Logger, cluster *fdbv1beta2.Foundat
 		return nil, &requeue{message: "Waiting for config map to sync to all pods", delayedRequeue: true}
 	}
 
-	counts, err := cluster.GetProcessCountsWithDefaults()
-	if err != nil {
-		return nil, &requeue{
-			curError:       err,
-			delayedRequeue: true,
-		}
-	}
-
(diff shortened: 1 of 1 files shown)


controllers/bounce_processes.go
```diff
@@ -263,15 +263,17 @@ func getProcessesReadyForRestart(logger logr.Logger, cluster *fdbv1beta2.Foundat
 		return nil, &requeue{message: "Waiting for config map to sync to all pods", delayedRequeue: true}
 	}
 
-	counts, err := cluster.GetProcessCountsWithDefaults()
-	if err != nil {
-		return nil, &requeue{
-			curError:       err,
-			delayedRequeue: true,
+	// Only if the cluster is upgraded with ab incompatible version we have to make sure that all processes are ready to be restarted.
```
nicmorales9, line 266:
```suggestion
	// Only if the cluster is upgraded with an incompatible version we have to make sure that all processes are ready to be restarted.
```

nicmorales9 (approved):
LGTM

johscheuer (commented):
I merge the PR as the PR comment was only a typo fix.

PR #1962, opened 2024-03-08
Make use of unstructured when using SSA to prevent default values causing issues

johscheuer:
# Description

There have been some issues in the past when we tested SSA in the current implementation with default values and it's also a known issue with the controller-runtime, e.g.: https://github.com/kubernetes-sigs/controller-runtime/issues/1669 (only to link one issue).

## Type of change

## Discussion

This will add some more cpu cycles to the operator when pushing the patch but I think this is the simplest way to approach this issue right now. Once the controller-runtime has a better approach we can remove this workaround.

## Testing

Ran a manual e2e test.

CI will run all e2e tests with the SSA feature as I enabled it in the operator deployment.

## Documentation

-

## Follow-up

-

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -24,7 +24,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer/yaml"
 	"k8s.io/apimachinery/pkg/types"
 	"regexp"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
@@ -41,6 +44,7 @@ import (
 
 	"github.com/FoundationDB/fdb-kubernetes-operator/internal"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	sigyaml "sigs.k8s.io/yaml"
 
 	fdbv1beta2 "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta2"
 	"github.com/FoundationDB/fdb-kubernetes-operator/pkg/podclient"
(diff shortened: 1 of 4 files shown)


nicmorales9 (approved):
overall looks good if it works in testing, though it'd be good to get clarification on the forceOwnership thread.

controllers/cluster_controller.go
```diff
@@ -442,7 +448,22 @@ func (r *FoundationDBClusterReconciler) updateOrApply(ctx context.Context, clust
...
+			return err
+		}
+
+		unstructuredPatch := &unstructured.Unstructured{}
+		_, _, err = r.decodingSerializer.Decode(outBytes, nil, unstructuredPatch)
+		if err != nil {
+			return err
+		}
+
+		return r.Status().Patch(ctx, unstructuredPatch, client.Apply, client.FieldOwner("fdb-operator")) //, client.ForceOwnership)
```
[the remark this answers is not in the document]
johscheuer: This doesn't work for sub resources like the status.

e2e/fixtures/fdb_operator_client.go
```diff
@@ -466,7 +466,7 @@ spec:
           - --max-concurrent-reconciles=5
           - --zap-log-level=debug
           - --minimum-required-uptime-for-cc-bounce=60s
-          #- --server-side-apply
+          - --server-side-apply
```
johscheuer, line 469: This should be a feature that can be enabled/disabled.

nicmorales9 (approved):
looks right to me

e2e/fixtures/options.go
```diff
@@ -192,6 +193,12 @@ func (options *FactoryOptions) BindFlags(fs *flag.FlagSet) {
 		true,
 		"defines if the operator tests should print the state of the cluster and the according Pods for better debugging.",
 	)
+	fs.BoolVar(
+		&options.featureOperatorServerSideApply,
+		"feature-server-side-apply",
+		true, // TODO change to false.
```
johscheuer, line 199:
```suggestion
		false,
```

Issue #1310, opened 2022-08-03
GetProcessSettings should not return volumeClaimTemplate for stateless processes


johscheuer:
### What would you like to be added/changed?

When a user calls `spec.GetProcessSettings("stateless")` the pod template will contain a volumeClaimTemplate which will be ignored in a later state (here: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/internal/pod_models.go#L332-L345) I think it makes more sense to remove the volumeClaimTemplate directly in the `spec.GetProcessSettings` method to prevent confusion on the user side.

PR #1965, opened 2024-03-12
make GetProcessSettings not return volumeClaimTemplate for stateless …

nicmorales9:
…processes

# Description

This addresses https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1310.  
Some odd logic was discovered in the `usePvc()` function, which roughly resulted in its logic being "use PVC if a processGroup is stateful and does not have a volumeClaimTemplate (nil), or if a processGroup is stateful and has a non-zero request for storage resources".  This seemed a bit counter-intuitive, and could result in the case where someone accidentally creates storage pods with no PVC, which is dangerous for a non-test FDB cluster.
As a result, I removed the `usePvc()` function and replaced it with a `processClass.isStateful()` check.  I think this more closely matches the original intent.  (As of writing, it is valid for a stateful process to have a nil volumeClaimTemplate, see https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1966, so, as far as I can tell the only meaningful check is `isStateful()`)

## Type of change

- "Bug" fix (non-breaking change which fixes an issue)

- Other

## Discussion

TODO link forum post regarding the breaking change after build number is created

## Testing

unit tests

## Documentation

## Follow-up

### DIFF
--- a/internal/pod_models.go
+++ b/internal/pod_models.go
@@ -320,7 +320,7 @@ func configureVolumesForContainers(cluster *fdbv1beta2.FoundationDBCluster, podS
 	}
 
 	var mainVolumeSource corev1.VolumeSource
-	if usePvc(cluster, processClass) {
+	if processClass.IsStateful() {
 		var volumeClaimSourceName string
 		if volumeClaimTemplate != nil && volumeClaimTemplate.Name != "" {
 			volumeClaimSourceName = fmt.Sprintf("%s-%s", podName, volumeClaimTemplate.Name)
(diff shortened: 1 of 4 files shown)


internal/pod_models.go
```diff
@@ -733,7 +733,8 @@ func usePvc(cluster *fdbv1beta2.FoundationDBCluster, processClass fdbv1beta2.Pro
 			storage = &storageCopy
 		}
 	}
-	return processClass.IsStateful() && (storage == nil || !storage.IsZero())
+	// TODO can a stateful process ever have this be true? maybe with testing?
+	return storage == nil || !storage.IsZero()
```
johscheuer, line 737: I believe this could be the case if someone doesn't specify a `volumeClaimTemplate` in the `FoundationDBCluster` spec, e.g. if this part: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/config/tests/base/cluster.yaml#L42-L47 would be missing.

johscheuer (commented):
Changes in general look good to me, could you make sure that the tests are passing?

brownleej (approved):
👍

My recollection is that we added a deliberate mechanism to use a non-persistent volume for storage pods in the early stages of the operator development, as a workaround for issues with PVC compatibility. Those issues have since been resolved, so I agree that this option is more dangerous than helpful.

johscheuer (approved):
LGTM 👍

Issue #1958, opened 2024-03-06
Out-of-date connection strings when running three_data_hall


simenl:
### What happened?

When running with multiple `FoundationDBCluster` k8s resources, managing the same cluster, like in a [three_data_hall](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/docs/manual/fault_domains.md#three-data-hall-replication) setup, a change to the connection string does not propagate to all `FoundationDBCluster` resources, leaving the connection string in the `FoundationDBCluster` resource status and the associated ConfigMap (cluster-file) out-of-date.

Foundationdb clients use the cluster-file (connection string) mounted from the ConfigMap. When the connection string is stale,  `status json` reports  `client.database_status.healthy: false`, with the message `Cluster file contents do not match current cluster connection string. Verify the cluster file and its parent directory are writable and that the cluster file has not been overwritten externally.`.

The issue resolves whenever the reconciliation loop of the out-of-date `FoundationDBCluster` is run again. This seems to require a trigger, such that there is no bound on how long they are out-of-date.

### What did you expect to happen?

An update to the connection string should eventually (within a few minutes) propagate to all `FoundationDBCluster` resources managing the cluster, and their associated ConfigMap.

### How can we reproduce it (as minimally and precisely as possible)?

Create a cluster with multiple `FoundationDBCluster` resource. E.g. by following the [three_data_hall example](https://github.com/FoundationDB/fdb-kubernetes-operator/tree/46534301cab561439e62647b89b15ad73963d86e/config/tests/three_data_hall).
Wait for it to reconcile.

Apply a change dummy change to _one_ of the `FoundationDBCluster` resources which triggers a change of coordinators (e.g. updating the node selector). The connection strings should be out-of-sync between the `FoundationDBCluster` resources (and ConfigMaps).

### FDB Kubernetes operator

v1.33.0

### Kubernetes version

1.26

### Cloud provider

Azure, GCP

johscheuer:
That issue probably happens for clients outside of the FDB Pods? I believe the issue is that the client cannot write to the mounted cluster file from the ConfigMap. It's better to copy the cluster file to a location where the client can write to, e.g. with an init container. If the client can write to the cluster file it should automatically update the cluster file when the coordinators change.

simenl:
> That issue probably happens for clients outside of the FDB Pods? I believe the issue is that the client cannot write to the mounted cluster file from the ConfigMap.

[...]

PR #1975, opened 2024-03-19
Improve documentation

johscheuer:
# Description

Updated a few sections in our documentation to improve the wording and the expectations.

## Type of change

- Documentation

## Discussion

-

## Testing

Not needed.

## Documentation

Updated.

## Follow-up

-

### DIFF
--- a/docs/manual/getting_started.md
+++ b/docs/manual/getting_started.md
@@ -19,7 +19,7 @@ You can see logs from the operator by running `kubectl logs -f -l app=fdb-kubern
 
 The example below will cover creating a cluster. All subsequent examples will assume that you have just created this cluster, and will cover an operation on this cluster.
 
-For more information on the fields you can define on the cluster resource, see the [go docs](https://godoc.org/github.com/FoundationDB/fdb-kubernetes-operator/pkg/apis/apps/v1beta2#FoundationDBCluster).
+For more information on the fields you can define on the cluster resource, see the [cluster spec docs](../cluster_spec.md).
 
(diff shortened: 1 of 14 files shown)


docs/manual/getting_started.md
```diff
@@ -74,18 +103,21 @@ spec:
             - name: FDB_CLUSTER_FILE
               value: /mnt/config-volume/cluster-file
             volumeMounts:
-            - name: config-volume
+            - name: shared-volume
               mountPath: /mnt/config-volume
```
nicmorales9, line 107: doesn't match the name above?

nicmorales9 (approved):
check the one question, but the changes look good to me! thanks

docs/manual/getting_started.md
johscheuer: The init container mounts the `config-volume` which is a `ConfigMap` and copies the cluster file into the `shared-volume` which is an `EmptyDir`. The reason for this is to make the cluster file writable by the client.

nicmorales9: yep that makes sense. It might be helpful to have a comment for those of us who always mix these things up

johscheuer: I'll add a comment in another PR, I still have to create a note on why the cluster file should always be copied and not directly be used by the application: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1958#issuecomment-1985554885.

nicmorales9: even better!

PR #1977, opened 2024-03-20
Remove an unnecessary error return value by compiling the regex

gm42:
# Description

Move regex compilation to the `init()` function of the module, freeing up an unused error return value.

## Type of change

- Other

## Discussion

It is a trivial change to make functions a bit nicer to use.

## Testing

Pre-existing tests were updated.

## Documentation

Does not apply.

## Follow-up

No follow-up needed.


johscheuer (approved):
Changes LGTM 👍 I'll start an e2e test for the changes.

Thanks!

PR #1980, opened 2024-03-26
Documentation updates for fault domain and ADDITIONAL_ENV_FILE

gm42:
# Description

Two changes:
* remove non-working example about fictitious `spec.zoneName`
* mention that `spec.sidecarVariables` must be declared in the FoundationDB k8s object

## Type of change

- Documentation

## Discussion

None

## Testing

None; the existing test does not cover the `sidecarVariables` aspect, but that's fine.

## Documentation

This is a minor documentation update.

## Follow-up

None needed, unless the examples need rework.


gm42:
Alternatively, I could leave the non-working example and put there a question mark `?` in place of `spec.zoneName`

gm42:
I added another commit to document `ADDITIONAL_ENV_FILE`, as I thought it's useful and relevant for split images. The unified image also has a similar mechanism, so we might want to use that one when it's no more considered under development.

brownleej (approved):
👍

johscheuer (approved):

Issue #597, opened 2021-04-06
Remove instances based on a label selector


johscheuer:
It could be very useful to be able to remove all instances with a matching label selector.

PR #1981, opened 2024-04-03
Label selection for pods/processGroups

nicmorales9:
# Description

This PR adds the ability to select pods/processGroups by label selection.  This still requires a cluster name to be provided as I felt that would reduce the risk of accidental selection.
It also improves the error message of the `cordon` command for when no pods are found to be running on a node.

This solves https://github.com/FoundationDB/fdb-kubernetes-operator/issues/597

## Type of change

## Discussion

## Testing

t

## Documentation

## Follow-up


nicmorales9:
Github doesn't want me to edit the comment today but for testing:
Unit tests were added, and I tested all the process selection commands manually.  I also ensured that cordon is still working fine since it has some overlapping functions.

brownleej (approved):
👍

johscheuer (approved):
LGTM 👍 Thanks! As a next step we should remove the `useProcessGroupID` option.

Issue #1240, opened 2022-06-09
Automatically add port for backup and restore


johscheuer:
### What would you like to be added/changed?

The `fdbbackup` and `fdbrestore` command are not working correctly if the port is missing in the blob URL we should ensure on FDB side that this works like expected. In the meantime we could add the port based on a heuristic (if sc=0 we use port 80 otherwise we use port 443) if the account is missing the port.

In addition to that we should document this behaviour to make it easier for other user to debug and fix this known issue.

PR #1985, opened 2024-04-11
Automatically add port for backup and restore

nicmorales9:
# Description

This PR adds a default port, 443 unless secure_connection is set to 0 (in which case, it will default to port 80) as described here https://apple.github.io/foundationdb/backups.html#backup-urls
As described in the issue, a port is required by FDB backup/restore, so this should not be breaking.

Solves https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1240 

## Type of change
Depends who you ask

## Discussion

## Testing

Manual testing?*
Unit tests 

Unsure

## Documentation

I believe so

No

## Follow-up

I do not think they are of concern

### DIFF
--- a/api/v1beta2/foundationdbbackup_types.go
+++ b/api/v1beta2/foundationdbbackup_types.go
@@ -183,6 +183,8 @@ type BlobStoreConfiguration struct {
 	BackupName string `json:"backupName,omitempty"`
 
 	// The account name to use with the backup destination.
+	// If no port is included, it will default to 443,
+	// or 80 if secure_connection URL Parameter is set to 0.
 	// +kubebuilder:validation:MaxLength=100
 	// +kubebuilder:validation:Required
 	AccountName string `json:"accountName"`
@@ -194,6 +196,7 @@ type BlobStoreConfiguration struct {
 	Bucket string `json:"bucket,omitempty"`
 
 	// Additional URL parameters passed to the blobstore URL.
+	// See: https://apple.github.io/foundationdb/backups.html#backup-urls
 	// +kubebuilder:validation:MaxItems=100
 	URLParameters []URLParameter `json:"urlParameters,omitempty"`
 }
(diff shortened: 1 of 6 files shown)


api/v1beta2/foundationdbbackup_types.go
```diff
@@ -320,14 +323,32 @@ func (configuration *BlobStoreConfiguration) getURL(backup string, bucket string
 	if configuration.AccountName == "" {
 		return ""
 	}
-
-	var sb strings.Builder
+	var (
+		defaultPort string
+		sb          strings.Builder
+	)
+	if !strings.Contains(configuration.AccountName, ":") { // add default http port
```
johscheuer, line 330: It's probably better to do something like this:

```go
backupURL := &url.URL{Host: "account@account:931"}
if backupURL.port() == "" {
    defaultPort = ":443"
}
```

If we use the current approach IPv6 addresses will be unhappy :)

In general it probably would be good to refactor this setup to make use of the net.Url package from go instead of putting a raw string together. Nothing that you have to fix in this PR, just something I noticed looking at the PR. I'll create an issue for that follow up -> https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1987

johscheuer (changes requested):
Can we update the docs for backups: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/docs/manual/backup.md with the new behaviour?

api/v1beta2/foundationdbbackup_types.go
nicmorales9: good point! I wasn't sure how urls would handle the @ but I will dig at it

nicmorales9: it handled it fine

johscheuer (approved):
LGTM 👍 Thanks!

johscheuer:
Merging as the e2e tests are not covering the backup setup yet.

Issue #1976, opened 2024-03-20
Using a restart_delay override creates a cluster with a persistent IncorrectProcess reconciliation problem


gm42:
### What happened?

Creating a cluster with a `restart_delay` custom parameter override leads to the very first pods being created with the `IncorrectCommandline` / `IncorrectProcess` problem.

This is what the logs for this newly created cluster contain (filtered for `IncorrectProcess`):
```
{"level":"info","ts":"2024-03-20T14:16:01Z","logger":"controller","msg":"IncorrectProcess","namespace":"mynamespace","cluster":"mycluster","reconciler":"controllers.updateStatus","expected":"/usr/bin/fdbserver --class=coordinator --cluster_file=/var/fdb/data/fdb.cluster --datadir=/var/fdb/data --listen_address=10.20.180.127:4501 --locality_instance_id=coordinator-1 --locality_machineid=ip-10-20-186-38.us-east-1.compute.internal --locality_zoneid=ip-10-20-186-38.us-east-1.compute.internal --logdir=/var/log/fdb-trace-logs --loggroup=mycluster --public_address=10.20.180.127:4501 --restart_delay=15 --seed_cluster_file=/var/dynamic-conf/fdb.cluster --trace_format=json","got":"/usr/bin/fdbserver --class=coordinator --cluster_file=/var/fdb/data/fdb.cluster --datadir=/var/fdb/data --listen_address=10.20.180.127:4501 --locality_instance_id=coordinator-1 --locality_machineid=ip-10-20-186-38.us-east-1.compute.internal --locality_zoneid=ip-10-20-186-38.us-east-1.compute.internal --logdir=/var/log/fdb-trace-logs --loggroup=mycluster --public_address=10.20.180.127:4501 --seed_cluster_file=/var/dynamic-conf/fdb.cluster --trace_format=json","expectedVersion":"7.1.15","version":"7.1.15","processGroupID":"coordinator-1","emptyMonitorConf":false}
[...]

PR #1988, opened 2024-04-12
Document the limitation that only custom parameters for fdbserver can be defined

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1976 

## Type of change

## Discussion

As long as we are not supporting to set those general parameters we should make sure that the operator rejects those.

## Testing

Added unit tests.

## Documentation

Updated.

## Follow-up

We have to clarify if there is a need to set those parameters.


nicmorales9 (approved):
LGTM

Issue #1991, opened 2024-04-16
Deferring reconciliation due to unsupported clients (coordinators)


gm42:
### What happened?

I run a cluster with separate coordinator pods which is using an older FoundationDB version e.g. `7.1.17`.
Once I set the version to latest `7.3.27`, operator never performs the upgrade because of this error:
```
{"level":"info","ts":"2024-04-16T06:22:20Z","logger":"controller","msg":"Reconciliation terminated early","namespace":"skunkworks","cluster":"mycluster","reconciler":"controllers.checkClientCompatibility","requeueAfter":60,"message":"7 clients do not support version 7.3.27: 192.168.149.171:49018, 192.168.149.171:52564, 192.168.159.185:52772, 192.168.159.185:55498, 192.168.190.37:36264, 192.168.190.37:42684, 192.168.190.37:52468"}
```

However those IPs correspond to the pod IPs of 3 of the coordinators (!), not an external client (from cluster PoV).

### What did you expect to happen?

Operator carrying out the update by pod replacement, not considering connections from coordinators as blocking.

### How can we reproduce it (as minimally and precisely as possible)?

Create a cluster:
```yaml
apiVersion: apps.foundationdb.org/v1beta2
kind: FoundationDBCluster
metadata:
  name: mycluster
  namespace: skunkworks
spec:
  version: 7.1.17
  routing:
    publicIPSource: service
  coordinatorSelection:
  - priority: 100
    processClass: coordinator
  databaseConfiguration:
    logs: 3
    storage: 3
    redundancy_mode: single
    storage_engine: ssd-2
  processCounts:
    coordinator: 3
    stateless: 3
    storage: 3
    log: 3
```

Start using it using some other client, so that there is some activity on it; then change its version to `7.3.27` using `kubectl edit`.

### FDB Kubernetes operator

<details>
v1.33
</details>

### Kubernetes version

<details>

```console
$ kubectl version
Client Version: v1.29.3
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.26.14-eks-b9c9ed7
```

</details>

### Cloud provider

<details>
AWS
</details>

johscheuer:
Could you please verify if there are any other processes/tools running in our setup? I was not able to reproduce the issue. Is it possible that you ran `fdbcli` from those Pods?

johscheuer:
[...]

PR #1995, opened 2024-04-17
Add additional variation test with dedicated coordinators and service IP

johscheuer:
# Description

Test case for: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1991 

## Type of change

- Other

## Discussion

## Testing

Ran the e2e test manually.

## Documentation

-

## Follow-up

-

### DIFF
--- a/e2e/test_operator_upgrades_variations/operator_upgrades_variations_test.go
+++ b/e2e/test_operator_upgrades_variations/operator_upgrades_variations_test.go
@@ -79,11 +79,11 @@ func clusterSetupWithConfig(config testConfig) *fixtures.FdbCluster {
 	return cluster
 }
 
-func performUpgrade(config testConfig, validateFunc func(cluster *fixtures.FdbCluster)) {
+func performUpgrade(config testConfig, preUpgradeFunction func(cluster *fixtures.FdbCluster)) {
 	fdbCluster = clusterSetupWithConfig(config)
 	startTime := time.Now()
+	preUpgradeFunction(fdbCluster)
 	Expect(fdbCluster.UpgradeCluster(config.targetVersion, false)).NotTo(HaveOccurred())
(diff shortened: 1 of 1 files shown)


e2e/test_operator_upgrades_variations/operator_upgrades_variations_test.go
```diff
@@ -79,11 +79,11 @@ func clusterSetupWithConfig(config testConfig) *fixtures.FdbCluster {
 	return cluster
 }
 
-func performUpgrade(config testConfig, validateFunc func(cluster *fixtures.FdbCluster)) {
+func performUpgrade(config testConfig, preUpgradeFunction func(cluster *fixtures.FdbCluster)) {
 	fdbCluster = clusterSetupWithConfig(config)
 	startTime := time.Now()
+	preUpgradeFunction(fdbCluster)
```
nicmorales9, line 85: Looking at the usages, I am a bit confused as to how this passes? the callers seem to use it as a validation function

johscheuer: That's correct, in most test cases this was used as a pre-validation step. The validation should happen before the upgrade is triggered to make sure the conditions are met before spending time on the upgrade.

nicmorales9: ah, I was understanding cases where cluster config is set e.g. https://github.com/johscheuer/fdb-kubernetes-operator/blob/6a1ea6c10d5750ebff7d040a4c1d9c9f19cc6cb9/e2e/test_operator_upgrades_variations/operator_upgrades_variations_test.go#L197 as a new change instead of version change, that makes more sense

nicmorales9 (approved):
Looks right to me

PR #2004, opened 2024-04-23
Make sure to skip process groups that are under maintenance

johscheuer:
# Description

If a process group is under maintenance we shouldn't update the conditions. it's expected that the process group related resources will be unavailable for some time.

## Type of change

- Other

## Discussion

-
## Testing

Added a new e2e upgrade test and added an additional unit test.

## Documentation

-

## Follow-up

-


nicmorales9 (approved):
LGTM

PR #2008, opened 2024-04-26
Add safe guard to reduce risk when a large number of nodes are tainted

johscheuer:
# Description

Add safe guard to reduce risk when a large number of nodes are tainted. The idea is to prevent the operator from automatically replacing any process groups on tainted nodes if too many nodes in the cluster are tainted. The `MaxConcurrentReplacements` is still in place, so this is basically a second safe guard.

## Type of change

## Discussion

-

## Testing

Updated the unit testing setup for the taint feature and added additional tests.

## Documentation

Added in the cluster docs.

## Follow-up

-


controllers/update_status_test.go
```diff
@@ -255,21 +255,18 @@ var _ = Describe("update_status", func() {
...
+					node.Spec.Taints = nil
+					Expect(k8sClient.Update(context.TODO(), node)).NotTo(HaveOccurred())
 					globalControllerLogger.Info("Remove node taint", "Node name", pod.Name, "Node taints", node.Spec.Taints, "Now", time.Now())
 
-					err = validateProcessGroup(context.TODO(), clusterReconciler, cluster, pod, pvc, pod.ObjectMeta.Annotations[fdbv1beta2.LastConfigMapKey], processGroupStatus, cluster.IsTaintFeatureDisabled(), logger)
-					Expect(err).NotTo(HaveOccurred())
-					Expect(len(processGroupStatus.ProcessGroupConditions)).To(Equal(1))
-					Expect(processGroupStatus.GetCondition(fdbv1beta2.NodeTaintReplacing)).NotTo(Equal(nil))
+					Expect(validateProcessGroup(context.TODO(), clusterReconciler, cluster, pod, pvc, pod.ObjectMeta.Annotations[fdbv1beta2.LastConfigMapKey], processGroupStatus, cluster.IsTaintFeatureDisabled(), logger)).NotTo(HaveOccurred())
+					Expect(processGroupStatus.ProcessGroupConditions).To(BeEmpty())
```
nicmorales9, line 269: change is because we are hitting that limit now? I am not entirely clear what would cause the change other than if it is hitting the fault domain cap due to the test setup?

nicmorales9 (approved):
Did not dig at test refactor, but the changes look good to me other than the comments!

Issue #208, opened 2020-04-28
Replacing instances when changing the security context on a pod


brownleej:
Changing the security context on a pod can cause its volume to retain files with the old ownership attached to them. While we can continue to read the files, we can see strange behavior with changing file ownership in the atomic writes in FDB. As a workaround for this, we can have the operator replace any pods that have incorrect security contexts. This will ensure that we get fresh pods with clean configuration.

PR #2014, opened 2024-05-07
Replace pod for instances with changed file security context

nicmorales9:
# Description

Should solve https://github.com/FoundationDB/fdb-kubernetes-operator/issues/208 by adding a check to replacement workflow.  It also adds a feature flag (default off) and an e2e test.

## Type of change

## Discussion

## Testing

Manual testing?*
Unit test, e2e tests
e2e test output (focus tested on the new one)
```
 {"level":"info","ts":"2024-05-21T17:54:19Z","logger":"controller","msg":"Attempting to run sub-reconciler","namespace":"nicmorales","cluster":"nicdev","reconciler":"controllers.replaceMisconfiguredProcessGroups"}
  {"level":"info","ts":"2024-05-21T17:54:19Z","logger":"controller","msg":"Replace process group","namespace":"nicmorales","cluster":"nicdev","reconciler":"controllers.replaceMisconfiguredProcessGroups","reason":"either FSGroup or FSGroupChangePolicy has changed for the pod SecurityContext"}
  {"level":"info","ts":"2024-05-21T17:54:19Z","logger":"controller","msg":"Replace process group","namespace":"nicmorales","cluster":"nicdev","reconciler":"controllers.replaceMisconfiguredProcessGroups","reason":"either FSGroup or FSGroupChangePolicy has changed for the pod SecurityContext"}
  {"level":"info","ts":"2024-05-21T17:54:19Z","logger":"controller","msg":"Replace process group","namespace":"nicmorales","cluster":"nicdev","reconciler":"controllers.replaceMisconfiguredProcessGroups","reason":"either FSGroup or FSGroupChangePolicy has changed for the pod SecurityContext"}
  {"level":"info","ts":"2024-05-21T17:54:19Z","logger":"controller","msg":"Replace process group","namespace":"nicmorales","cluster":"nicdev","reconciler":"controllers.replaceMisconfiguredProcessGroups","reason":"either FSGroup or FSGroupChangePolicy has changed for the pod SecurityContext"}
  {"level":"info","ts":"2024-05-21T17:54:19Z","logger":"controller","msg":"Replace process group","namespace":"nicmorales","cluster":"nicdev","reconciler":"controllers.replaceMisconfiguredProcessGroups","reason":"either FSGroup or FSGroupChangePolicy has changed for the pod SecurityContext"}
  {"level":"info","ts":"2024-05-21T17:54:19Z","logger":"controller","msg":"Removals have been updated in the cluster status","namespace":"nicmorales","cluster":"nicdev","reconciler":"controllers.replaceMisconfiguredProcessGroups"}
  {"level":"info","ts":"2024-05-21T17:54:19Z","logger":"controller","msg":"Subreconciler finished run","namespace":"nicmorales","cluster":"nicdev","reconciler":"controllers.replaceMisconfiguredProcessGroups","duration_seconds":0.031225991}
```

Any unique kubernetes environment should be checked for securityContext server-side config changes before enabling this feature.

## Documentation

I think so! Added docstrings + section in `docs/manual/replacements_and_deletions.md`.

I think the addition in `docs/manual/replacements_and_deletions.md` covers it but let me know if not.

N/A

I think they are straightforward enough, but let me know if not

## Follow-up

### DIFF
--- a/docs/manual/replacements_and_deletions.md
+++ b/docs/manual/replacements_and_deletions.md
@@ -73,6 +73,13 @@ This feature is disabled by default, but can be enabled by setting `automationOp
 
 We use three examples below to illustrate how to set up the feature.
 
+## Automatic Replacement of Pods with SecurityContext changes
+
+Changes in SecurityContext - file ownership ones specifically - can cause problems where FDB is able to read but not write 
+files.  This can cause weird behavior with FDB's atomic writes.  If the Operator setting `ReplaceOnSecurityContextChange` 
+is set to `true`, the Operator can automatically replace pods which have changes to any of the following fields: 
+`FSGroup`, `FSGroupChangePolicy`, `RunAsGroup`, `RunAsUser`.\
+
 ### Example Setup 1
 
 The following YAML setup lets the operator detect Pods running on Nodes with taint key `example.com/maintenance`, set the ProcessGroup' condition to `NodeTaintReplacing`, if their Nodes have been tainted for 3600 seconds, and replace the Pods after 1800 seconds.
--- a/docs/manual/replacements_and_deletions.md
+++ b/docs/manual/replacements_and_deletions.md
@@ -75,10 +75,10 @@ We use three examples below to illustrate how to set up the feature.
 
 ## Automatic Replacement of Pods with SecurityContext changes
 
-Changes in SecurityContext - file ownership ones specifically - can cause problems where FDB is able to read but not write 
-files.  This can cause weird behavior with FDB's atomic writes.  If the Operator setting `ReplaceOnSecurityContextChange` 
-is set to `true`, the Operator can automatically replace pods which have changes to any of the following fields: 
-`FSGroup`, `FSGroupChangePolicy`, `RunAsGroup`, `RunAsUser`.\
+Changes in SecurityContext - file ownership ones specifically - can cause problems where FDB is not able to use (read or write) the
+files.  This can potentially lead to an outage and unavailability of the cluster.  If the Operator command line parameter `--replace-on-security-context-change`
+is set to `true`, the Operator can automatically replace pods which have changes to any of the following fields:
+`FSGroup`, `FSGroupChangePolicy`, `RunAsGroup`, `RunAsUser`.
 
 ### Example Setup 1
 
--- a/e2e/test_operator/operator_test.go
+++ b/e2e/test_operator/operator_test.go
@@ -1484,7 +1484,7 @@ var _ = Describe("Operator", Label("e2e", "pr"), func() {
 				currentPods := fdbCluster.GetPodsNames()
 				g.Expect(currentPods).NotTo(ContainElements(podsToReplace))
 				g.Expect(currentPods).To(ContainElements(podsNotToReplace))
-				// check that General template pods (Log + Stateless) are not
+				// check that General template pods (Log + Stateless) are replaced with the new security context
 				logPods := fdbCluster.GetLogPods()
 				for _, pod := range logPods.Items {
 					g.Expect(pod.Spec.SecurityContext.FSGroupChangePolicy).ToNot(BeNil())
--- a/e2e/test_operator/operator_test.go
+++ b/e2e/test_operator/operator_test.go
@@ -428,6 +428,7 @@ var _ = Describe("Operator", Label("e2e", "pr"), func() {
 				fdbCluster.EnsurePodIsDeleted(replacedPod.Name)
 			})
 		})
+
 	})
 
 	When("setting storageServersPerPod", func() {
@@ -1425,6 +1426,49 @@ var _ = Describe("Operator", Label("e2e", "pr"), func() {
 					return fdbCluster.CheckPodIsDeleted(podName)
 				}).WithTimeout(2 * time.Minute).WithPolling(15 * time.Second).Should(BeTrue())
 			})
+		})
+	})
+
+	// TODO also test container level security context change
+	When("replacing Pods due to securityContext changes", func() {
+		var initialPods *corev1.PodList
+		var originalPodSpec, modifiedPodSpec *corev1.PodSpec
+
+		BeforeEach(func() {
+			originalPodSpec = fdbCluster.GetPodTemplateSpec(fdbv1beta2.ProcessClassLog)
+			modifiedPodSpec = originalPodSpec.DeepCopy()
+			modifiedPodSpec.SecurityContext.FSGroupChangePolicy = &[]corev1.PodFSGroupChangePolicy{corev1.FSGroupChangeOnRootMismatch}[0]
+			initialPods = fdbCluster.GetLogPods()
+			Expect(fdbCluster.SetPodTemplateSpec(fdbv1beta2.ProcessClassLog, modifiedPodSpec, false)).NotTo(HaveOccurred())
+		})
+
+		AfterEach(func() {
+			Expect(fdbCluster.SetPodTemplateSpec(fdbv1beta2.ProcessClassLog, originalPodSpec, false)).NotTo(HaveOccurred())
+			// TODO maybe test that it does the same for the revert?
+		})
+
+		It("should replace all the log pods (which had the security context change)", func() {
+			lastForcedReconciliationTime := time.Now()
+			forceReconcileDuration := 4 * time.Minute
+			Eventually(func() bool {
+				// Force a reconcile if needed to make sure we speed up the reconciliation if needed.
+				if time.Since(lastForcedReconciliationTime) >= forceReconcileDuration {
+					fdbCluster.ForceReconcile()
+					lastForcedReconciliationTime = time.Now()
+				}
+
+				// check that all original pods are gone
+				pods := fdbCluster.GetLogPods()
+				Expect(pods.Items).NotTo(ContainElements(initialPods.Items))
+				// TODO maybe instead check if all logs are marked for removal? unclear how long *all* would take?
+				//for _, processGroup := range fdbCluster.GetCluster().Status.ProcessGroups {
+				//	if processGroup.IsMarkedForRemoval() && processGroup.IsExcluded() {
+				//		continue
+				//	}
+				//  <fail>
+				//}
+				return true
+			}).WithTimeout(20 * time.Minute).WithPolling(5 * time.Second).Should(BeTrue())
 		})
 	})
 
--- a/e2e/test_operator/operator_test.go
+++ b/e2e/test_operator/operator_test.go
@@ -1428,16 +1428,13 @@ var _ = Describe("Operator", Label("e2e", "pr"), func() {
 		})
 	})
 
-	// this test took 14min to run for me, so I think it is best to just have the one test
-	FWhen("replacing Pods due to securityContext changes (General spec change = log + stateless)", func() {
-		var initialLogPods, initialStatelessPods *corev1.PodList
+	// this test took ~8min to run for me, so I think it is best to just have the one test
+	When("replacing Pods due to securityContext changes (General spec change = log + stateless)", func() {
+		var podsNotToReplace, podsToReplace []string
 		var originalPodSpec, modifiedPodSpec, originalStoragePodSpec *corev1.PodSpec
 		var initialPodUpdateStrategy fdbv1beta2.PodUpdateStrategy
 
 		BeforeEach(func() {
-			if fdbCluster.GetClusterSpec().ReplaceInstancesWhenResourcesChange != nil {
-				Expect(*fdbCluster.GetClusterSpec().ReplaceInstancesWhenResourcesChange).To(BeFalse())
-			}
 			originalPodSpec = fdbCluster.GetPodTemplateSpec(fdbv1beta2.ProcessClassGeneral)
 			originalStoragePodSpec = fdbCluster.GetPodTemplateSpec(fdbv1beta2.ProcessClassStorage)
 			Expect(originalPodSpec).NotTo(BeNil())
@@ -1447,8 +1444,15 @@ var _ = Describe("Operator", Label("e2e", "pr"), func() {
 				Expect(*originalPodSpec.SecurityContext.FSGroupChangePolicy).NotTo(Equal(corev1.FSGroupChangeOnRootMismatch))
 			}
 			modifiedPodSpec.SecurityContext.FSGroupChangePolicy = &[]corev1.PodFSGroupChangePolicy{corev1.FSGroupChangeOnRootMismatch}[0]
-			initialLogPods = fdbCluster.GetLogPods()
-			initialStatelessPods = fdbCluster.GetStatelessPods()
+			for _, logPod := range fdbCluster.GetLogPods().Items {
+				podsToReplace = append(podsToReplace, logPod.Name)
+			}
+			for _, statelessPod := range fdbCluster.GetStatelessPods().Items {
+				podsToReplace = append(podsToReplace, statelessPod.Name)
+			}
+			for _, storagePod := range fdbCluster.GetStoragePods().Items {
+				podsToReplace = append(podsToReplace, storagePod.Name)
+			}
 			// if we do not set the PodUpdateStrategy to delete, then the pods will get replaced for the spec change,
 			// meaning we cannot test the security context change on non-storage pods unless we use PodUpdateStrategyDelete
 			// (and storage pod use would make the test take ~5min longer)
@@ -1476,20 +1480,22 @@ var _ = Describe("Operator", Label("e2e", "pr"), func() {
 					lastForcedReconciliationTime = time.Now()
 				}
 
-				// check that all original pods are gone (that use the General template)
+				// check that we replaced the pods we should have and did not replace others
+				currentPods := fdbCluster.GetPodsNames()
+				g.Expect(currentPods).NotTo(ContainElements(podsToReplace))
+				g.Expect(currentPods).To(ContainElements(podsNotToReplace))
+				// check that General template pods (Log + Stateless) are not
 				logPods := fdbCluster.GetLogPods()
-				g.Expect(logPods.Items).NotTo(ContainElements(initialLogPods.Items))
 				for _, pod := range logPods.Items {
 					g.Expect(pod.Spec.SecurityContext.FSGroupChangePolicy).ToNot(BeNil())
 					g.Expect(*pod.Spec.SecurityContext.FSGroupChangePolicy).To(Equal(corev1.FSGroupChangeOnRootMismatch))
 				}
 				statelessPods := fdbCluster.GetStatelessPods()
-				g.Expect(statelessPods.Items).NotTo(ContainElements(initialStatelessPods.Items))
 				for _, pod := range statelessPods.Items {
 					g.Expect(pod.Spec.SecurityContext.FSGroupChangePolicy).ToNot(BeNil())
 					g.Expect(*pod.Spec.SecurityContext.FSGroupChangePolicy).To(Equal(corev1.FSGroupChangeOnRootMismatch))
 				}
-				// ensure that we only replace pods that are expected to be replaced (i.e. don't touch storage)
+				// ensure that we only changed pods that are expected to be changed (i.e. don't touch storage)
 				storagePods := fdbCluster.GetStoragePods()
 				for _, pod := range storagePods.Items {
 					g.Expect(pod.Spec.SecurityContext).To(Equal(originalStoragePodSpec.SecurityContext))
--- a/e2e/test_operator/operator_test.go
+++ b/e2e/test_operator/operator_test.go
@@ -1429,28 +1429,32 @@ var _ = Describe("Operator", Label("e2e", "pr"), func() {
 		})
 	})
 
-	// TODO also test container level security context change
+	// this test took 14min to run for me, so I think it is best to just have the one test
 	When("replacing Pods due to securityContext changes", func() {
 		var initialPods *corev1.PodList
 		var originalPodSpec, modifiedPodSpec *corev1.PodSpec
 
 		BeforeEach(func() {
-			originalPodSpec = fdbCluster.GetPodTemplateSpec(fdbv1beta2.ProcessClassLog)
+			if fdbCluster.GetClusterSpec().ReplaceInstancesWhenResourcesChange != nil {
+				Expect(*fdbCluster.GetClusterSpec().ReplaceInstancesWhenResourcesChange).To(BeFalse())
+			}
+			originalPodSpec = fdbCluster.GetPodTemplateSpec(fdbv1beta2.ProcessClassStorage)
+			Expect(originalPodSpec).NotTo(BeNil())
 			modifiedPodSpec = originalPodSpec.DeepCopy()
+			Expect(modifiedPodSpec.SecurityContext).NotTo(BeNil())
 			modifiedPodSpec.SecurityContext.FSGroupChangePolicy = &[]corev1.PodFSGroupChangePolicy{corev1.FSGroupChangeOnRootMismatch}[0]
-			initialPods = fdbCluster.GetLogPods()
-			Expect(fdbCluster.SetPodTemplateSpec(fdbv1beta2.ProcessClassLog, modifiedPodSpec, false)).NotTo(HaveOccurred())
+			initialPods = fdbCluster.GetStoragePods()
+			Expect(fdbCluster.SetPodTemplateSpec(fdbv1beta2.ProcessClassStorage, modifiedPodSpec, false)).NotTo(HaveOccurred())
 		})
 
 		AfterEach(func() {
-			Expect(fdbCluster.SetPodTemplateSpec(fdbv1beta2.ProcessClassLog, originalPodSpec, false)).NotTo(HaveOccurred())
-			// TODO maybe test that it does the same for the revert?
+			Expect(fdbCluster.SetPodTemplateSpec(fdbv1beta2.ProcessClassStorage, originalPodSpec, true)).NotTo(HaveOccurred())
 		})
 
-		It("should replace all the log pods (which had the security context change)", func() {
+		It("should replace all the storage pods (which had the securityContext change)", func() {
 			lastForcedReconciliationTime := time.Now()
 			forceReconcileDuration := 4 * time.Minute
-			Eventually(func() bool {
+			Eventually(func(g Gomega) bool {
 				// Force a reconcile if needed to make sure we speed up the reconciliation if needed.
 				if time.Since(lastForcedReconciliationTime) >= forceReconcileDuration {
 					fdbCluster.ForceReconcile()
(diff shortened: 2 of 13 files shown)


internal/replacements/replacements_test.go
```diff
@@ -296,6 +296,23 @@ var _ = Describe("replace_misconfigured_pods", func() {
 			})
 		})
 
+		Context("when the securityContext doesn't match and the PodSpecHash doesn't match", func() {
```
nicmorales9, line 299: this is wrong as the setting is for all changes for this test case, I will add another which does not have this set to true (default is false)

internal/replacements/replacements.go
```diff
@@ -246,7 +247,14 @@ func processGroupNeedsRemovalForPod(cluster *fdbv1beta2.FoundationDBCluster, pod
 		}
 	}
 
-	return false, nil
+	desiredPod, err := internal.GetPod(cluster, processGroupStatus)
+	if err != nil {
+		return false, err
+	}
+	// TODO deprecated builtin k8s features edited securityContext automatically, and it doesn't seem outlandish that someone's cluster
```
johscheuer, line 254: We have similar cases in the statements above where some other controllers are adding additional information in the spec. That's the reason we have this line at some checks:


```go
if pod.ObjectMeta.Annotations[fdbv1beta2.LastSpecKey] != specHash {
    // replace ..
}
```

In this case the replacement will only be triggered if the applied specHash is different from the last applied spec hash. This can still result in replacements that shouldn't be a replacement, e.g. if the spec hash changes because of another field but now the operator detects the mismatch of the security context and triggers a replacement. But as far as I know, this is the best we can do right now.

internal/replacements/replacements.go
```diff
@@ -257,6 +265,51 @@ func resourcesNeedsReplacement(desired []corev1.Container, current []corev1.Cont
...
+// to the following SecurityContext (or PodSecurityContext) fields:
+// RunAsGroup, RunAsUser, FSGroup, or FSGroupChangePolicy
+// See https://github.com/FoundationDB/fdb-kubernetes-operator/issues/208 for motivation
+// only makes sense if both pods have containers with matching names
+func fileSecurityContextChanged(desired, current *corev1.Pod) bool {
+	// first check for FSGroup or FSGroupChangePolicy changes as that cannot be overridden at container level
+	// (if pod security context is identical, skip these checks)
+	if (desired.Spec.SecurityContext != nil || current.Spec.SecurityContext != nil) &&
+		!equality.Semantic.DeepEqualWithNilDifferentFromEmpty(desired.Spec.SecurityContext, current.Spec.SecurityContext) {
+		if desired.Spec.SecurityContext == nil { // check if changed non-nil -> nil
```
johscheuer, line 278:
```suggestion
		if desired.Spec.SecurityContext == nil && current.Spec.SecurityContext != nil { // check if changed non-nil -> nil
```

Otherwise the check below could cause a nil pointer exception?

internal/replacements/replacements.go
```diff
@@ -257,6 +265,51 @@ func resourcesNeedsReplacement(desired []corev1.Container, current []corev1.Cont
...
+			if current.Spec.SecurityContext.FSGroup != nil || current.Spec.SecurityContext.FSGroupChangePolicy != nil {
+				return true
+			}
+		} else if current.Spec.SecurityContext == nil { // check if changed nil -> non-nil
```
johscheuer, line 282:
```suggestion
		} else if current.Spec.SecurityContext == nil && desired.Spec.SecurityContext != nil { // check if changed nil -> non-nil
```

Same as above?

internal/replacements/replacements.go
```diff
@@ -257,6 +265,51 @@ func resourcesNeedsReplacement(desired []corev1.Container, current []corev1.Cont
...
+			}
+		} else { // both pod security contexts are defined so check they are the same
+			if !equality.Semantic.DeepEqualWithNilDifferentFromEmpty(desired.Spec.SecurityContext.FSGroup, current.Spec.SecurityContext.FSGroup) ||
+				!equality.Semantic.DeepEqualWithNilDifferentFromEmpty(desired.Spec.SecurityContext.FSGroupChangePolicy, current.Spec.SecurityContext.FSGroupChangePolicy) {
+				return true
+			}
+		}
+	}
+	// check for RunAsUser and RunAsGroup changes (have to check with container settings, since that can override pod settings)
+	for _, desiredContainer := range desired.Spec.Containers {
```
johscheuer, line 294: Since we only have a few containers this approach is fine, otherwise I would have preferred to have the desired container security context in a map, to remove the nested looping.

internal/replacements/replacements.go
```diff
@@ -257,6 +265,51 @@ func resourcesNeedsReplacement(desired []corev1.Container, current []corev1.Cont
...
+		for _, currentContainer := range current.Spec.Containers {
+			if desiredContainer.Name == currentContainer.Name {
+				desiredEffectiveSecCtx := securitycontext.DetermineEffectiveSecurityContext(desired, &desiredContainer)
+				currentEffectiveSecCtx := securitycontext.DetermineEffectiveSecurityContext(current, &currentContainer)
+				if equality.Semantic.DeepEqual(desiredEffectiveSecCtx, currentEffectiveSecCtx) {
+					continue
```
johscheuer, line 300:
```suggestion
					break
```

If we are at this code line, we can skip the further iteration of the inner loop as we already found the matching container.

internal/replacements/replacements.go
```diff
@@ -257,6 +265,51 @@ func resourcesNeedsReplacement(desired []corev1.Container, current []corev1.Cont
...
+				}
+				if !equality.Semantic.DeepEqual(desiredEffectiveSecCtx.RunAsUser, currentEffectiveSecCtx.RunAsUser) ||
```
johscheuer, line 302: What is the benefit of the check above? Either the context matches and we stop the further execution or it doesn't match and we have to perform the checks here (line 302/303) too. And if the check above matches I would expect that the check here (line 302/303) also matches?

internal/replacements/replacements.go
```diff
@@ -257,6 +265,51 @@ func resourcesNeedsReplacement(desired []corev1.Container, current []corev1.Cont
...
+					!equality.Semantic.DeepEqual(desiredEffectiveSecCtx.RunAsGroup, currentEffectiveSecCtx.RunAsGroup) {
+					// FSGroup is checked at top/pod level
+					return true
+				}
+			}
```
johscheuer, line 307:
```suggestion
				}
				
				break
			}
```

We would hit this line if the security context doesn't match but the `RunAsUser` and `RunAsGroup` settings are matching.

internal/replacements/replacements.go
```diff
@@ -257,6 +265,51 @@ func resourcesNeedsReplacement(desired []corev1.Container, current []corev1.Cont
 	return desiredCPURequests.Cmp(*currentCPURequests) == 1 || desiredMemoryRequests.Cmp(*currentMemoryRequests) == 1
 }
 
+// fileSecurityContextChanged checks for changes in the effective security context by checking that there are no changes
+// to the following SecurityContext (or PodSecurityContext) fields:
+// RunAsGroup, RunAsUser, FSGroup, or FSGroupChangePolicy
+// See https://github.com/FoundationDB/fdb-kubernetes-operator/issues/208 for motivation
+// only makes sense if both pods have containers with matching names
+func fileSecurityContextChanged(desired, current *corev1.Pod) bool {
```
johscheuer, line 273: We should add some logging here, similar to what we do in the other replacement checks to let the user know why the process group was replaced.

internal/replacements/replacements.go
nicmorales9: My feeling is that adding annotations for securityContext spec change might not be warranted, but that adding a feature flag to turn it off could be worthwhile, but I know that flag bloat is very real

internal/replacements/replacements.go
nicmorales9: yeah I figured it's small enough that converting to a map would be more troublesome

internal/replacements/replacements.go
nicmorales9: it took me several minutes to remember why so I will add a comment even if this makes sense to you, but I was thinking that if other things on the security context changed that were not the runasX settings which we care about then we do not want to trigger a replacement.  
I will add some test cases to demonstrate this behavior

go.mod
```diff
@@ -30,7 +30,10 @@ require (
 	sigs.k8s.io/yaml v1.3.0
 )
 
-require golang.org/x/sync v0.3.0
+require (
+	golang.org/x/sync v0.3.0
+	k8s.io/kubernetes v1.26.10
```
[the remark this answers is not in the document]
nicmorales9: good point, I do not know, I'll see if I can limit scope more on the import and double check if there is no option

go.mod
```diff
@@ -1,67 +1,69 @@
 module github.com/FoundationDB/fdb-kubernetes-operator
 
-go 1.22
+go 1.22.0
```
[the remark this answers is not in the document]
nicmorales9: early commit - sorry about the untidiness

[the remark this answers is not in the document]
nicmorales9: though I think I need 23 now...

go.mod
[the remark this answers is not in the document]
nicmorales9: hmm, I did not look far before finding that this is not supported https://github.com/kubernetes/kubernetes?tab=readme-ov-file#to-start-using-k8s 
> To use Kubernetes code as a library in other applications, see the [list of published components](https://git.k8s.io/kubernetes/staging/README.md). Use of the k8s.io/kubernetes module or k8s.io/kubernetes/... packages as libraries is not supported.
*sigh*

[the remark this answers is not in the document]
nicmorales9: it's not the worst to create a helper to do this since we only care for 2 fields, I'll just do that and scrap the import

internal/replacements/replacements.go
johscheuer: If we remove the line `equality.Semantic.DeepEqual(desiredEffectiveSecCtx, currentEffectiveSecCtx)` above, we will only check for the `RunAsUser` and `RunAsGroup` settings. If they match we will not trigger a replacement as we would return `false` later, so how is that different from doing the `DeepEqual` check above? Maybe I'm missing something.

nicmorales9: ah, yes I think you are correct, I guess I was thinking to skip the checks (no idea why)? this code is going away with the import but yes I agree the check above is not useful

internal/replacements/replacements.go
nicmorales9: discussed offline: added a feature flag

e2e/test_operator/operator_test.go
```diff
@@ -1428,6 +1429,50 @@ var _ = Describe("Operator", Label("e2e", "pr"), func() {
...
+				// Force a reconcile if needed to make sure we speed up the reconciliation if needed.
+				if time.Since(lastForcedReconciliationTime) >= forceReconcileDuration {
+					fdbCluster.ForceReconcile()
+					lastForcedReconciliationTime = time.Now()
+				}
+
+				// check that all original pods are gone
+				pods := fdbCluster.GetStoragePods()
+				g.Expect(pods.Items).NotTo(ContainElements(initialPods.Items))
+				for _, pod := range pods.Items {
```
nicmorales9, line 1467: would it make more sense to instead check if all original storage pods/pgs are marked for removal? or would that be incomplete

nicmorales9:
I believe the only failure was unrelated to the test as it passed when I reran locally and I believe I have seen it flake before:
```
[FAIL] Operator Upgrades one process is marked for removal and is stuck in terminating state [It] Upgrade from 7.1.57 to 7.3.33 [e2e, pr]
  /codebuild/output/src477579299/src/github.com/FoundationDB/fdb-kubernetes-operator/e2e/test_operator_upgrades/operator_upgrades_test.go:573
```
rerunning

docs/manual/replacements_and_deletions.md
```diff
@@ -73,6 +73,13 @@ This feature is disabled by default, but can be enabled by setting `automationOp
 
 We use three examples below to illustrate how to set up the feature.
 
+## Automatic Replacement of Pods with SecurityContext changes
+
+Changes in SecurityContext - file ownership ones specifically - can cause problems where FDB is able to read but not write 
+files.  This can cause weird behavior with FDB's atomic writes.  If the Operator setting `ReplaceOnSecurityContextChange` 
+is set to `true`, the Operator can automatically replace pods which have changes to any of the following fields: 
+`FSGroup`, `FSGroupChangePolicy`, `RunAsGroup`, `RunAsUser`.\
```
johscheuer, line 81:
```suggestion
`FSGroup`, `FSGroupChangePolicy`, `RunAsGroup`, `RunAsUser`.
```

docs/manual/replacements_and_deletions.md
```diff
@@ -73,6 +73,13 @@ This feature is disabled by default, but can be enabled by setting `automationOp
 
 We use three examples below to illustrate how to set up the feature.
 
+## Automatic Replacement of Pods with SecurityContext changes
+
+Changes in SecurityContext - file ownership ones specifically - can cause problems where FDB is able to read but not write 
+files.  This can cause weird behavior with FDB's atomic writes.  If the Operator setting `ReplaceOnSecurityContextChange` 
+is set to `true`, the Operator can automatically replace pods which have changes to any of the following fields:
```
johscheuer, line 80:
```suggestion
Changes in SecurityContext - file ownership ones specifically - can cause problems where FDB is not able to use (read or write) the
files.  This can potentially lead to an outage and unavailability of the cluster.  If the Operator command line parameter `--replace-on-security-context-change` 
is set to `true`, the Operator can automatically replace pods which have changes to any of the following fields: 
```

e2e/test_operator/operator_test.go
```diff
@@ -428,6 +428,7 @@ var _ = Describe("Operator", Label("e2e", "pr"), func() {
 				fdbCluster.EnsurePodIsDeleted(replacedPod.Name)
 			})
 		})
+
```
johscheuer, line 431: Can we remove that new newline?

e2e/test_operator/operator_test.go
```diff
@@ -1428,6 +1429,50 @@ var _ = Describe("Operator", Label("e2e", "pr"), func() {
...
+	// this test took 14min to run for me, so I think it is best to just have the one test
+	When("replacing Pods due to securityContext changes", func() {
+		var initialPods *corev1.PodList
+		var originalPodSpec, modifiedPodSpec *corev1.PodSpec
+
+		BeforeEach(func() {
+			if fdbCluster.GetClusterSpec().ReplaceInstancesWhenResourcesChange != nil {
+				Expect(*fdbCluster.GetClusterSpec().ReplaceInstancesWhenResourcesChange).To(BeFalse())
+			}
+			originalPodSpec = fdbCluster.GetPodTemplateSpec(fdbv1beta2.ProcessClassStorage)
```
johscheuer, line 1441: Using `General` will be faster as there will be no data movement involved.

e2e/test_operator/operator_test.go
johscheuer: Have you tested what the difference in for the test execution would be? Will be the test much faster or still take that long?

internal/replacements/replacements.go
```diff
@@ -95,7 +96,7 @@ func ProcessGroupNeedsRemoval(ctx context.Context, podManager podmanager.PodLife
 		return false, podErr
 	}
 
-	return processGroupNeedsRemovalForPod(cluster, pod, processGroup, log)
+	return processGroupNeedsRemovalForPod(cluster, pod, processGroup, log, hasPVC && replaceOnSecurityContextChange)
```
johscheuer, line 99: Why are we checking if the `hasPVC` is true? What is the implication of not doing that?

internal/replacements/replacements.go
```diff
@@ -246,6 +247,32 @@ func processGroupNeedsRemovalForPod(cluster *fdbv1beta2.FoundationDBCluster, pod
...
+	// https://kubernetes.io/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/
+	if replaceOnSecurityContextChange {
+		spec, err := internal.GetPodSpec(cluster, processGroupStatus)
+		if err != nil {
+			return false, err
+		}
+		specHash, err := internal.GetPodSpecHash(cluster, processGroupStatus, spec)
+		if err != nil {
+			return false, err
+		}
```
johscheuer, line 267: Do you think it makes sense to get the `spec` and `specHash` once in this method instead of fetching the same information in multiple if branches above?

johscheuer (changes requested):

internal/replacements/replacements.go
nicmorales9: oh, I just figured if it doesn't have a PVC then it doesn't have to worry about file security context changes

internal/replacements/replacements.go
nicmorales9: I was basically trying to avoid grabbing it for settings that don't need the check, but I can definitely do that as it would be much cleaner (could avoid with if (any of the settings), but that's also ehh)

e2e/test_operator/operator_test.go
nicmorales9: yeah I was working under the assumption that PVC should be a criteria for the change

e2e/test_operator/operator_test.go
nicmorales9: no, I wasn't sure if it was a valid way to test honestly, since the point is that the pod be replaced

nicmorales9: I'll give it a go

nicmorales9: 772.232 seconds =~13min so not much improvement (this is also with storage pods) I'll do some tests with general / log

internal/replacements/replacements.go
nicmorales9: though this does not seem to work, when I tested with General process class it replaces stateless pods 🤔

nicmorales9: discussed offline, hasPVC doesnt really matter as we'd eventually replace for spec change anyways in defaults, and having stateless pods with matching specs is desirable

e2e/test_operator/operator_test.go
```diff
@@ -1428,6 +1428,76 @@ var _ = Describe("Operator", Label("e2e", "pr"), func() {
...
+
+				// check that all original pods are gone (that use the General template)
+				logPods := fdbCluster.GetLogPods()
+				g.Expect(logPods.Items).NotTo(ContainElements(initialLogPods.Items))
+				for _, pod := range logPods.Items {
+					g.Expect(pod.Spec.SecurityContext.FSGroupChangePolicy).ToNot(BeNil())
+					g.Expect(*pod.Spec.SecurityContext.FSGroupChangePolicy).To(Equal(corev1.FSGroupChangeOnRootMismatch))
+				}
+				statelessPods := fdbCluster.GetStatelessPods()
+				g.Expect(statelessPods.Items).NotTo(ContainElements(initialStatelessPods.Items))
```
nicmorales9, line 1487: This seems to always be the case, trying to track down exact changed fields, but I am thinking checking the names / Pgids changed is more meaningful

johscheuer: The `fdbCluster.GetPodsNames()` could be used here, e.g.: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/e2e/test_operator/operator_test.go#L920-L937

internal/replacements/replacements.go
```diff
@@ -186,13 +187,17 @@ func processGroupNeedsRemovalForPod(cluster *fdbv1beta2.FoundationDBCluster, pod
...
 
+	specHash, err := internal.GetPodSpecHash(cluster, processGroupStatus, nil)
+	if err != nil {
+		return false, err
+	}
+	spec, err := internal.GetPodSpec(cluster, processGroupStatus)
+	if err != nil {
+		return false, err
+	}
+	// TODO the below logic could be consolidated further since the majority of them would be skipped if pod Spec has not changed.
```
johscheuer, line 198: Can you create an issue for that?

internal/replacements/replacements.go
```diff
@@ -228,24 +223,36 @@ func processGroupNeedsRemovalForPod(cluster *fdbv1beta2.FoundationDBCluster, pod
...
+	// to constantly be seen as having a security context change, hence we want to feature guard this
+	// and also guard on the spec hash below
+	// https://kubernetes.io/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/
+	if replaceOnSecurityContextChange {
+		if pod.ObjectMeta.Annotations[fdbv1beta2.LastSpecKey] == specHash {
+			// no changes have been made outside of server-side injected values, do not check for changes
+			// to avoid looping
+			return false, nil
+		}
+		desiredPod, err := internal.GetPod(cluster, processGroupStatus)
```
johscheuer, line 250: Is there any need to generate the pod struct? Could we just make use of the already generated `spec`?

internal/replacements/replacements_test.go
```diff
@@ -290,12 +290,69 @@ var _ = Describe("replace_misconfigured_pods", func() {
...
+				// do not change last spec key
+				// if ReplaceInstancesWhenResourcesChange is true, any spec change should result in replacement
+				cluster.Spec.ReplaceInstancesWhenResourcesChange = new(bool)
+
+				pod.Spec.SecurityContext = &corev1.PodSecurityContext{FSGroup: new(int64)}
+				needsRemoval, err := processGroupNeedsRemovalForPod(cluster, pod, processGroup, log, true)
+				Expect(needsRemoval).To(BeFalse())
+				Expect(err).NotTo(HaveOccurred())
+			})
+			It("should need a removal with ReplaceInstancesWhenResourcesChange", func() {
```
johscheuer, line 335: Wrong text in the `It` statement? How is the `ReplaceInstancesWhenResourcesChange` related to this test case?

internal/replacements/replacements_test.go
```diff
@@ -290,12 +290,69 @@ var _ = Describe("replace_misconfigured_pods", func() {
...
 					"dummy": "test",
 				}
-				needsRemoval, err := processGroupNeedsRemovalForPod(cluster, pod, processGroup, log)
+				needsRemoval, err := processGroupNeedsRemovalForPod(cluster, pod, processGroup, log, true)
 				Expect(needsRemoval).To(BeFalse())
 				Expect(err).NotTo(HaveOccurred())
 			})
 		})
 
+		Context("when the securityContext doesn't match", func() {
```
johscheuer, line 299:
```suggestion
		When("the securityContext doesn't match", func() {
```

internal/replacements/replacements_test.go
```diff
@@ -290,12 +290,69 @@ var _ = Describe("replace_misconfigured_pods", func() {
...
+			BeforeEach(func() {
+				pClass = fdbv1beta2.ProcessClassStorage
+				remove = false
+			})
+
+			It("should need a removal for RunAsUser change", func() {
+				pod.ObjectMeta.Annotations[fdbv1beta2.LastSpecKey] = "banana"
+				// if ReplaceInstancesWhenResourcesChange is true, any spec change should result in replacement
+				cluster.Spec.ReplaceInstancesWhenResourcesChange = new(bool)
```
johscheuer, line 308: This shouldn't be the case and would mean something is wrong with the test setup.

internal/replacements/replacements_test.go
```diff
@@ -290,12 +290,69 @@ var _ = Describe("replace_misconfigured_pods", func() {
...
 			})
 		})
 
+		Context("when the securityContext doesn't match", func() {
+			BeforeEach(func() {
+				pClass = fdbv1beta2.ProcessClassStorage
+				remove = false
+			})
+
+			It("should need a removal for RunAsUser change", func() {
```
johscheuer, line 305: Can we try to better group those different tests together in different `When` statements with their set of `BeforeEach` statements? e.g.


```go
		When("the securityContext doesn't match", func() {
			var needsRemoval bool
			var err error

			BeforeEach(func() {
				pClass = fdbv1beta2.ProcessClassStorage
				remove = false
			})

			JustBeforeEach(func() {
				needsRemoval, err = processGroupNeedsRemovalForPod(cluster, pod, processGroup, log, true)
			})

			When("the last-applied-spec key is updated ", func() {
				BeforeEach(func() {
					pod.ObjectMeta.Annotations[fdbv1beta2.LastSpecKey] = "banana"
				})

				When("the FSGroup is changed", func() {
					BeforeEach(func() {
						pod.Spec.SecurityContext = &corev1.PodSecurityContext{FSGroup: new(int64)}
					})

					It("should need a removal for FSGroup change", func() {
						Expect(needsRemoval).To(BeTrue())
					})
				})

				When("the RunAsUser is changed", func() {
					BeforeEach(func() {
						pod.Spec.SecurityContext = &corev1.PodSecurityContext{RunAsUser: new(int64)}
					})

					It("should need a removal for RunAsUser change", func() {
						Expect(needsRemoval).To(BeTrue())
					})
				})
			})
		})
```

e2e/test_operator/operator_test.go
```diff
@@ -1428,6 +1428,76 @@ var _ = Describe("Operator", Label("e2e", "pr"), func() {
 		})
 	})
 
+	// this test took 14min to run for me, so I think it is best to just have the one test
+	FWhen("replacing Pods due to securityContext changes (General spec change = log + stateless)", func() {
```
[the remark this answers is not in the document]
nicmorales9: oops, good catch 😓

e2e/test_operator/operator_test.go
```diff
@@ -1428,6 +1428,76 @@ var _ = Describe("Operator", Label("e2e", "pr"), func() {
...
+		var initialLogPods, initialStatelessPods *corev1.PodList
+		var originalPodSpec, modifiedPodSpec, originalStoragePodSpec *corev1.PodSpec
+		var initialPodUpdateStrategy fdbv1beta2.PodUpdateStrategy
+
+		BeforeEach(func() {
+			if fdbCluster.GetClusterSpec().ReplaceInstancesWhenResourcesChange != nil {
+				Expect(*fdbCluster.GetClusterSpec().ReplaceInstancesWhenResourcesChange).To(BeFalse())
+			}
```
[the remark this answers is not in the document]
nicmorales9: I just figured the default is off but yeah that's a better idea

internal/replacements/replacements.go
nicmorales9: yeah, I could also do it in this PR if you prefer but it feels like there are too many changes going in already

johscheuer: Creating an issue is good enough :) We can work on that in another PR.

e2e/test_operator/operator_test.go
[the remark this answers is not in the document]
nicmorales9: but honestly it shouldn't matter, I was confused on some behaviour and I think this was just a troubleshooting relic

e2e/test_operator/operator_test.go
```diff
@@ -1428,6 +1428,82 @@ var _ = Describe("Operator", Label("e2e", "pr"), func() {
...
+				if time.Since(lastForcedReconciliationTime) >= forceReconcileDuration {
+					fdbCluster.ForceReconcile()
+					lastForcedReconciliationTime = time.Now()
+				}
+
+				// check that we replaced the pods we should have and did not replace others
+				currentPods := fdbCluster.GetPodsNames()
+				g.Expect(currentPods).NotTo(ContainElements(podsToReplace))
+				g.Expect(currentPods).To(ContainElements(podsNotToReplace))
+				// check that General template pods (Log + Stateless) are not
```
johscheuer, line 1487: The comment seems to be incomplete?

johscheuer (approved):
LGTM 👍

internal/replacements/replacements_test.go
```diff
@@ -290,13 +290,83 @@ var _ = Describe("replace_misconfigured_pods", func() {
...
+
+					When("replaceOnSecurityContextChange is false", func() {
+						It("should *not* need a removal", func() {
+							needsRemoval, err := processGroupNeedsRemovalForPod(cluster, pod, processGroup, log, false)
+							Expect(needsRemoval).To(BeFalse())
+							Expect(err).NotTo(HaveOccurred())
+						})
+					})
+
+					AfterAll(func() {
```
johscheuer, line 336:
```suggestion
					AfterEach(func() {
```

AfterAll can only be used in an ordered setup: https://onsi.github.io/ginkgo/#setup-in-ordered-containers-beforeall-and-afterall, you have to use `AfterEach`.

internal/replacements/replacements_test.go
```diff
@@ -290,13 +290,83 @@ var _ = Describe("replace_misconfigured_pods", func() {
...
+						pod.Spec.SecurityContext = originalSecurityContext
+					})
+				})
+
+				AfterAll(func() {
```
johscheuer, line 341:
```suggestion
				AfterEach(func() {
```

PR #2021, opened 2024-05-14
updating max length length for connection string

09harsh:
# Description

Updating max length length for connection string

## Type of change

## Discussion

No

## Testing

N/A

## Documentation

N/A

## Follow-up

No


johscheuer (changes requested):
Why do we need this change?

Please run `make clean all` to generate the new CRDs.

PR #2023, opened 2024-05-14
Run trivy only for the operator image

johscheuer:
# Description

Only run the trivy check for the operator image.

## Type of change

## Discussion

The previous configuration had issues which caused the actions to fail.

## Testing

This PR.

## Documentation

-

## Follow-up

-


nicmorales9 (approved):
tricky!

Issue #1030, opened 2022-01-14
Supporting the new unified image in the backup pods


brownleej:
We've added support for using the unified foundationdb-kubernetes image in the main server pods. We should also allow using this image in the backup pods as well, so we only need to configure a single image.

PR #2025, opened 2024-05-14
Initial support for the unified image in the backup deployment

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1030

## Type of change

## Discussion

Adds support for using the unified image in the backup deployment. Even thought the current implementation of backups are not using any features of the monitor.

## Testing

Will be done manually.

## Documentation

-

## Follow-up

-


internal/pod_models.go
```diff
@@ -859,15 +868,37 @@ func GetBackupDeployment(backup *fdbv1beta2.FoundationDBBackup) (*appsv1.Deploym
...
+			backup.Spec.MainContainer.ImageConfigs = []fdbv1beta2.ImageConfig{
+				{
+					BaseImage: "foundationdb/foundationdb",
+				},
+			}
+		}
+	}
+
+	if len(backup.Spec.SidecarContainer.ImageConfigs) == 0 {
+		if !pointer.BoolDeref(backup.Spec.UseUnifiedImage, false) {
```
nicmorales9, line 887: if it's early enough, would it be helpful to make `false` a constant for all these similar checks that we can flip quickly later? e.g. `UseUnifiedImageDefault`? 
If it's not early, grep is probably easier haha

nicmorales9 (approved):
LGTM

internal/pod_models.go
johscheuer: We can move that into a method on the backup struct.

Issue #1033, opened 2022-01-14
Using the same monitor conf for different storage server densities on the new image


brownleej:
We currently customize the monitor conf for the unified image with different flags based on how many storage servers we are running per disk. This adds some complexity, because we have to switch between different entries in the config map when we are changing the storage server density. On the old image, we were specifying multiple command lines when running multiple storage servers, but that's not the case on the new image, so we should be able to move to using the same monitor conf regardless of storage server density.

PR #2038, opened 2024-05-23
Use the same ConfigMap entries for all different process configuration

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1033

## Type of change

## Discussion

This removes the need to maintain different ConfigMap entries for different server counts, in the case of the unified image, this will be handled by the fdb-kubernetes-monitor based on the start arguments.

## Testing

Ran manual e2e test with the unified image.

## Documentation

-

## Follow-up

-


nicmorales9 (approved):
LGTM

Issue #2034, opened 2024-05-21
Coordinator selection should prefer primary


johscheuer:
### What would you like to be added/changed?

In the case of coordinator selection we should prefer the primary dc if the coordinator count cannot be equally distributed. Otherwise we might end up with a semi-optimal distribution and preferring the primary gives a more consistent coordinator set.

PR #2045, opened 2024-06-03
Prefer processes in coordinator selection for multi-region clusters

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/2034

## Type of change

## Discussion

Most of the code changes are adjustments for the testing code, to make sure we add some randomness.

## Testing

Added additional unit tests and modified them.

## Documentation

-

## Follow-up

-

### DIFF
--- a/api/v1beta2/foundationdb_database_configuration.go
+++ b/api/v1beta2/foundationdb_database_configuration.go
@@ -24,6 +24,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"html/template"
+	"math"
 	"reflect"
 	"sort"
 	"strconv"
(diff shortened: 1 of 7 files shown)


internal/locality/locality.go
```diff
@@ -180,6 +185,29 @@ func ChooseDistributedProcesses(cluster *fdbv1beta2.FoundationDBCluster, process
 		currentLimits[field] = 1
 	}
 
+	if constraint.SelectingCoordinators {
+		// If the cluster runs in a multi-region config and has more than 3 DCs we want to make sure that the primary is
+		// preferred for the 3 coordinators.
+		// See: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/2034
+		if cluster.Spec.DatabaseConfiguration.UsableRegions > 1 && cluster.Spec.DatabaseConfiguration.CountUniqueDataCenters() > 3 {
```
johscheuer, line 192: This works for some specific setups e.g. 4 different DCs but not for some more exotic setups like having 1 primary dc with 2 different primary satellites and having 1 remote and 2 different remote satellites (that would be 6 unique dcs).

johscheuer (commented):
The new benchmark will present a result like this:


```text
• [0.035 seconds]
Localities chooseDistributedProcesses when a multi-region cluster is used with 4 dcs when measuring the performance for choose distributed processes [measurement]
/Users/jscheuermann/go/src/github.com/FoundationDB/fdb-kubernetes-operator/internal/locality/locality_test.go:630

  Report Entries >>
  Choose Distributed Processes - /Users/jscheuermann/go/src/github.com/FoundationDB/fdb-kubernetes-operator/internal/locality/locality_test.go:636 @ 06/04/24 11:52:09.548
    Choose Distributed Processes
    Name                                  | N  | Min   | Median | Mean  | StdDev | Max
    ==================================================================================
    ChooseDistributedProcesses [duration] | 50 | 300µs | 400µs  | 400µs | 100µs  | 1ms
```

internal/locality/locality_test.go
```diff
@@ -138,105 +163,260 @@ var _ = Describe("Localities", func() {
...
+									Priority: 0,
+								},
+								{
+									ID:        remoteSatelliteID,
+									Priority:  1,
+									Satellite: 1,
+								},
+								{
+									ID:        primarySatelliteID,
+									Priority:  0,
```
nicmorales9, line 354: why is primary lower than remote here? or does the priority not actually signify primary vs remote? (though I don't see a distinction made in the tests so I don't think it matters)

nicmorales9 (approved):
just a few things to double check but the core lgtm

api/v1beta2/foundationdb_database_configuration.go
```diff
@@ -562,6 +563,58 @@ func (configuration DatabaseConfiguration) GetNextConfigurationChange(finalConfi
...
+// it will only be counted as a main dc.
+func (configuration DatabaseConfiguration) GetMainDCsAndSatellites() (map[string]None, map[string]None) {
+	mainDCs := map[string]None{}
+	satellites := map[string]None{}
+
+	for _, region := range configuration.Regions {
+		for _, dataCenter := range region.DataCenters {
+			if dataCenter.Satellite == 0 {
+				mainDCs[dataCenter.ID] = None{}
+				// If the dc ID is present as satellite we will remove it.
```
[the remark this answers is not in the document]
johscheuer: That could be the case in a triplet configuration, e.g. https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/config/tests/multi_dc/final.yaml#L25-L45

internal/locality/locality_test.go
johscheuer: This is the config for the remote side and the remote side should prefer the remote satellite to store mutations and not the primary satellite. See: https://github.com/apple/foundationdb/wiki/Multi-Region-Replication#concepts

Issue #2043, opened 2024-05-28
processGroupNeedsRemovalForPod flag logic can be more nicely consolidated


nicmorales9:
### What would you like to be added/changed?

See `processGroupNeedsRemovalForPod` in internal/replacements/replacements.go; there are many flag guards which do nothing if the pod spec has not changed, so we can check earlier if it has changed and exit out early if so.

nicmorales9:
See specifically the section below the TODO https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/internal/replacements/replacements.go#L198

PR #2051, opened 2024-06-05
Unified image code clean up and make use of the same constant

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/2043

## Type of change

## Discussion

We changed the disk layout with the unified image, if only a single storage server per pod is used, then the data directory is still suffixed with `/1`. This would break the transition to the unified image for cluster with a single storage server per pod.

I also made sure we remove some internal method duplications.

## Testing

Updated the unit tests and CI will run e2e tests.

## Documentation

-

## Follow-up

-


nicmorales9 (approved):
looks good to me! thanks for the cleanup as usual!

PR #2054, opened 2024-06-06
Make sure we skip the automatic replacements if the cluster is unavailable

johscheuer:
# Description

If the cluster is not available we shouldn't perform automatic replacements as the operator is not able to make any good decision. The operator will not replace more than `MaxConcurrentReplacements` in the worst case, but I still think it's better to skip this feature at all if the cluster is unavailable.

## Type of change

## Discussion

-

## Testing

CI will run e2e tests.

## Documentation

-

## Follow-up

-

### DIFF
--- a/controllers/replace_failed_process_groups.go
+++ b/controllers/replace_failed_process_groups.go
@@ -58,6 +58,13 @@ func (c replaceFailedProcessGroups) reconcile(ctx context.Context, r *Foundation
 		}
 	}
 
+	// If the database is unavailable skip further steps as the operator is not able to make any good decision about the automatic
+	// replacements.
+	if !status.Client.DatabaseStatus.Available {
+		logger.Info("Skipping replaceFailedProcessGroups reconciler as the database is not available.")
+		return nil
+	}
+
 	// Only replace process groups without an address, if the cluster has the desired fault tolerance and is available.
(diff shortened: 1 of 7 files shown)


controllers/replace_failed_process_groups.go
```diff
@@ -58,6 +58,13 @@ func (c replaceFailedProcessGroups) reconcile(ctx context.Context, r *Foundation
 		}
 	}
 
+	// If the database is unavailable skip further steps as the operator is not able to make any good decision about the automatic
+	// replacements.
+	if !status.Client.DatabaseStatus.Available {
+		logger.Info("Skipping replaceFailedProcessGroups reconciler as the database is not available.")
+		return nil
```
nicmorales9, line 65: I am just curious, what is your decision logic for deciding whether this should requeue or not?

nicmorales9 (approved):
LGTM + understanding question!

controllers/replace_failed_process_groups.go
nicmorales9: the tradeoff being knowing what is coming next / that this is retried vs blocking other operations I suppose?

johscheuer: In general we tried to build our subreconcilers in a way that if no errors occurs we try to proceed further, so doing a delayed requeue again makes sense to make sure the operator is retrying the reconciliation once the DB is available again. Could be an intermediate issue.

nicmorales9:
Looks good with requeue as well 👍

PR #2056, opened 2024-06-07
Fix the calculation of expected processes during upgrades for multiple servers per pod

johscheuer:
# Description

The previous calculation didn't account  properly for missing processes and processes that were marked for removal in the case of multiple servers per pod.

## Type of change

## Discussion

-

## Testing

Updated the unit tests and ran a manual test.

## Documentation

-

## Follow-up

-


nicmorales9 (approved):
LGTM

PR #2057, opened 2024-06-10
Fix e2e upgrade tests with chaos for unified image

johscheuer:
# Description

Some of the e2e upgrade tests with chaos had some assumption that were only true for the split image approach.

## Type of change

## Discussion

-

## Testing

I ran those tests manually with the fix from https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2056.

## Documentation

-

## Follow-up

-

### DIFF
--- a/e2e/fixtures/fdb_cluster.go
+++ b/e2e/fixtures/fdb_cluster.go
@@ -61,6 +61,12 @@ func (fdbCluster *FdbCluster) GetFDBImage() string {
 
 // GetSidecarImageForVersion return the sidecar image used for the specified version.
 func (fdbCluster *FdbCluster) GetSidecarImageForVersion(version string) string {
+	// In the case of the unified image the sidecar will also be the main container image.
+	if fdbCluster.cluster.UseUnifiedImage() {
+		return fdbv1beta2.SelectImageConfig(fdbCluster.GetClusterSpec().MainContainer.ImageConfigs, version).
+			Image()
+	}
+
 	return fdbv1beta2.SelectImageConfig(fdbCluster.GetClusterSpec().SidecarContainer.ImageConfigs, version).
 		Image()
 }
@@ -1194,12 +1200,11 @@ func (fdbCluster *FdbCluster) Destroy() error {
 }
 
 // SetIgnoreMissingProcessesSeconds sets the IgnoreMissingProcessesSeconds setting.
-func (fdbCluster *FdbCluster) SetIgnoreMissingProcessesSeconds(duration time.Duration) error {
+func (fdbCluster *FdbCluster) SetIgnoreMissingProcessesSeconds(duration time.Duration) {
 	fdbCluster.cluster.Spec.AutomationOptions.IgnoreMissingProcessesSeconds = pointer.Int(
 		int(duration.Seconds()),
 	)
 	fdbCluster.UpdateClusterSpec()
-	return fdbCluster.WaitForReconciliation()
 }
 
 // SetKillProcesses sets the automation option to allow the operator to restart processes or not.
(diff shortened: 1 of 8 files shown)


e2e/fixtures/fdb_cluster.go
```diff
@@ -61,6 +61,12 @@ func (fdbCluster *FdbCluster) GetFDBImage() string {
 
 // GetSidecarImageForVersion return the sidecar image used for the specified version.
 func (fdbCluster *FdbCluster) GetSidecarImageForVersion(version string) string {
+	// In the case of the unified image the sidecar will also be the main container image.
+	if fdbCluster.cluster.UseUnifiedImage() {
+		return fdbv1beta2.SelectImageConfig(fdbCluster.GetClusterSpec().MainContainer.ImageConfigs, version).
```
nicmorales9, line 66: if not a huge pain, it might be clearer to just set the sidecar config in `fdbCluster.GetClusterSpec()` to what it should be for the unified image
i.e. I think it would make sense in GetSidecarContainerOverrides take a useUnifiedImage param and just set to the main container so other checks like this don't have to do this special logic

e2e/test_operator_upgrades/operator_upgrades_test.go
```diff
@@ -315,6 +315,7 @@ var _ = Describe("Operator Upgrades", Label("e2e", "pr"), func() {
 		fixtures.GenerateUpgradeTableEntries(testOptions),
 	)
 
+	// TODO double check failure!!
```
nicmorales9, line 318: is this for this PR or post-merge reviewing?

e2e/test_operator_upgrades_with_chaos/operator_upgrades_with_chaos_test.go
```diff
@@ -93,9 +94,7 @@ var _ = Describe("Operator Upgrades with chaos-mesh", Label("e2e", "pr"), func()
 			clusterSetup(beforeVersion)
 			Expect(fdbCluster.SetAutoReplacements(false, 20*time.Minute)).ToNot(HaveOccurred())
 			// Ensure the operator is not skipping the process because it's missing for to long
-			Expect(
-				fdbCluster.SetIgnoreMissingProcessesSeconds(5 * time.Minute),
-			).NotTo(HaveOccurred())
+			fdbCluster.SetIgnoreMissingProcessesSeconds(5 * time.Minute)
```
nicmorales9, line 97: why does this error for unified but not split? or is it just not meaningful to check the error?

nicmorales9 (approved):
overall looks good, some nits and questions for my understanding though

e2e/fixtures/fdb_cluster.go
johscheuer: The operator will use the `MainContainer` config when the unified image is used, so I think it's better to use the `MainContainer` config here to reflect the expected configuration.

e2e/test_operator_upgrades/operator_upgrades_test.go
johscheuer: That's an old comment I forgot to remove 🙈

e2e/test_operator_upgrades_with_chaos/operator_upgrades_with_chaos_test.go
johscheuer: I updated the method to not wait for reconciliation as everything that this method does is updating a setting in the `FoundationDBCluster` spec which doesn't have a direct affect on the running state (or will not cause any replacements). That's the reason the error return value was removed.

nicmorales9: ooh I see, thanks for the clarification!

e2e/test_operator_upgrades/operator_upgrades_test.go
johscheuer, line 318:
```suggestion
```

johscheuer:
Merging as I only removed an old comment.

Issue #245, opened 2020-06-09
Setting the log dir and log group through environment variables in our pods


brownleej:
We set the log dir and log group through parameters to the fdbserver process. This allows us to send logs to an emptyDir volume that users can attach additional containers to for log forwarding. If someone execs into a pod and runs an fdbcli command, it doesn't generate any trace logs at all. I think it would be better for us to set the log dir and log group through environment variables. This would allow fdbcli commands from the server pods to get logged to the same directory, and get picked up by log management with no additional action on the user's part.

johscheuer:
I changed the label. It probably makes sense to adjust this for the new unified image and make this the default for the unified image.

PR #2060, opened 2024-06-12
Set log dir and log group through environment variables for unified image

nicmorales9:
# Description

fixes https://github.com/FoundationDB/fdb-kubernetes-operator/issues/245

## Type of change

## Discussion

N/A

## Testing

Unit tests, 
spun up e2e tests and checked that the logs are sent to the directory specified by the env variable, and that they all included the expected log group (cluster.spec.logGroup) incl. the logs from fdbcli

Don't think so, no

## Documentation

I don't think documentation change is needed, I think the issue is that there is nothing passively allowing logging to work and that this shouldn't be intrusive.  Do let me know if that sounds wrong!

## Follow-up

N/A

### DIFF
--- a/internal/pod_models.go
+++ b/internal/pod_models.go
@@ -179,7 +179,7 @@ func getContainers(podSpec *corev1.PodSpec) (*corev1.Container, *corev1.Containe
 	return mainContainer, sidecarContainer, nil
 }
 
-func configureContainersForUnifiedImages(cluster *fdbv1beta2.FoundationDBCluster, mainContainer *corev1.Container, sidecarContainer *corev1.Container, processGroup *fdbv1beta2.ProcessGroupStatus, desiredVersion string) error {
+func configureContainersForUnifiedImages(cluster *fdbv1beta2.FoundationDBCluster, mainContainer *corev1.Container, sidecarContainer *corev1.Container, processGroup *fdbv1beta2.ProcessGroupStatus, desiredVersion string, logGroup string) error {
 	mainContainer.Args = []string{
 		"--input-dir", "/var/dynamic-conf",
 		"--log-path", "/var/log/fdb-trace-logs/monitor.log",
@@ -207,6 +207,8 @@ func configureContainersForUnifiedImages(cluster *fdbv1beta2.FoundationDBCluster
 	mainContainer.Env = append(mainContainer.Env, corev1.EnvVar{Name: "FDB_POD_NAMESPACE", ValueFrom: &corev1.EnvVarSource{
 		FieldRef: &corev1.ObjectFieldSelector{FieldPath: "metadata.namespace"},
 	}})
+	mainContainer.Env = append(mainContainer.Env, corev1.EnvVar{Name: "FDB_NETWORK_OPTION_TRACE_LOG_GROUP", Value: logGroup})
+	mainContainer.Env = append(mainContainer.Env, corev1.EnvVar{Name: "FDB_NETWORK_OPTION_TRACE_ENABLE", Value: "/var/log/fdb-trace-logs"})
 	if cluster.DefineDNSLocalityFields() {
 		mainContainer.Env = append(mainContainer.Env, corev1.EnvVar{Name: "FDB_DNS_NAME", Value: GetPodDNSName(cluster, processGroup.GetPodName(cluster))})
 	}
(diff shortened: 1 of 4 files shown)


internal/pod_models.go
```diff
@@ -225,6 +227,8 @@ func configureContainersForUnifiedImages(cluster *fdbv1beta2.FoundationDBCluster
 		"--copy-binary", "fdbcli",
 		"--log-path", "/var/log/fdb-trace-logs/monitor.log",
 	}
+	sidecarContainer.Env = append(sidecarContainer.Env, corev1.EnvVar{Name: "FDB_NETWORK_OPTION_TRACE_LOG_GROUP", Value: logGroup})
+	sidecarContainer.Env = append(sidecarContainer.Env, corev1.EnvVar{Name: "FDB_NETWORK_OPTION_TRACE_ENABLE", Value: "/var/log/fdb-trace-logs"})
```
nicmorales9, line 231: unclear if sidecare matters, I don't know of a case where someone would fdbcli the sidecar specifically myself

johscheuer: Hm, yeah we can remove it from the sidecar.

internal/pod_models.go
```diff
@@ -426,7 +430,7 @@ func GetPodSpec(cluster *fdbv1beta2.FoundationDBCluster, processGroup *fdbv1beta
 
 	podName := processGroup.GetPodName(cluster)
 	if useUnifiedImage {
-		err = configureContainersForUnifiedImages(cluster, mainContainer, sidecarContainer, processGroup, desiredVersion)
+		err = configureContainersForUnifiedImages(cluster, mainContainer, sidecarContainer, processGroup, desiredVersion, logGroup)
```
johscheuer, line 433: I would prefer if we move the method above (the one where we get the log group), into a `FoundationDBCluster` method and use that instead of adding another parameter. What are your thoughts on that?

johscheuer (commented):
Changes look good in general, just a question :)

internal/pod_models.go
nicmorales9: Like `cluster.GetLogGroup()`? yes I completely agree

nicmorales9: done

internal/pod_models.go
```diff
@@ -207,6 +207,8 @@ func configureContainersForUnifiedImages(cluster *fdbv1beta2.FoundationDBCluster
 	mainContainer.Env = append(mainContainer.Env, corev1.EnvVar{Name: "FDB_POD_NAMESPACE", ValueFrom: &corev1.EnvVarSource{
 		FieldRef: &corev1.ObjectFieldSelector{FieldPath: "metadata.namespace"},
 	}})
+	mainContainer.Env = append(mainContainer.Env, corev1.EnvVar{Name: "FDB_NETWORK_OPTION_TRACE_LOG_GROUP", Value: cluster.GetLogGroup()})
+	mainContainer.Env = append(mainContainer.Env, corev1.EnvVar{Name: "FDB_NETWORK_OPTION_TRACE_ENABLE", Value: "/var/log/fdb-trace-logs"})
```
johscheuer, line 211: Like discussed offline, we should only set those env variables if they are not already present.

johscheuer (approved):
LGTM 👍

nicmorales9:
hmm, the test failure _looks_ unrelated:
```
• [FAILED] [2699.779 seconds]
Operator Upgrades with chaos-mesh with a partitioned pod which eventually gets replaced [It] Upgrade from 7.1.57 to 7.3.33 [e2e, pr]
/codebuild/output/src1695824796/src/github.com/FoundationDB/fdb-kubernetes-operator/e2e/fixtures/upgrade_test_configuration.go:115

  [FAILED] Unexpected error:
      <*errors.errorString | 0xc00044bb10>: 
      pod pr-911-lb2hrqt3/fdb-cluster-ulu39f6e-storage-5 was not removed in the expected time
      {
          s: "pod pr-911-lb2hrqt3/fdb-cluster-ulu39f6e-storage-5 was not removed in the expected time",
      }
  occurred
  In [It] at: /codebuild/output/src1695824796/src/github.com/FoundationDB/fdb-kubernetes-operator/e2e/test_operator_upgrades_with_chaos/operator_upgrades_with_chaos_test.go:188 @ 06/13/24 16:36:42.964
```
going to try rerunning

johscheuer (commented):
Test failure is unrelated, going to merge it. In addition the new code path is not executed with the split image.

PR #2062, opened 2024-06-14
use extendEnv instead of append to container.Env

nicmorales9:
# Description

this helps avoid collisions in the case that a variable is declared in the cluster spec already (collisions will prevent Patch()s and prevent Operator from functioning normally)

## Type of change

- minor improvement

## Discussion

## Testing

Manual testing?*

## Documentation

## Follow-up

### DIFF
--- a/internal/pod_models.go
+++ b/internal/pod_models.go
@@ -184,12 +184,13 @@ func configureContainersForUnifiedImages(cluster *fdbv1beta2.FoundationDBCluster
 		"--input-dir", "/var/dynamic-conf",
 		"--log-path", "/var/log/fdb-trace-logs/monitor.log",
 	}
+	var mainContainerEnv []corev1.EnvVar
 
 	serversPerPod := cluster.GetDesiredServersPerPod(processGroup.ProcessClass)
 	if serversPerPod > 1 {
 		desiredServersPerPod := strconv.Itoa(serversPerPod)
 		mainContainer.Args = append(mainContainer.Args, "--process-count", desiredServersPerPod)
-		mainContainer.Env = append(mainContainer.Env, corev1.EnvVar{Name: processGroup.ProcessClass.GetServersPerPodEnvName(), Value: desiredServersPerPod})
+		mainContainerEnv = append(mainContainerEnv, corev1.EnvVar{Name: processGroup.ProcessClass.GetServersPerPodEnvName(), Value: desiredServersPerPod})
 	}
 
 	mainContainer.VolumeMounts = append(mainContainer.VolumeMounts,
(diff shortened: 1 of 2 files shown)


internal/pod_models_test.go
```diff
@@ -512,7 +512,7 @@ var _ = Describe("pod_models", func() {
 						"--log-path", "/var/log/fdb-trace-logs/monitor.log",
 					}))
 
-					Expect(mainContainer.Env).To(Equal([]corev1.EnvVar{
+					Expect(mainContainer.Env).To(ContainElements([]corev1.EnvVar{
```
johscheuer, line 515: What was the reasoning behind this change? I believe if we want to change that it should be `ConsistOf` to make sure that exactly those env variables are present (without caring about the order).

internal/pod_models.go
```diff
@@ -199,19 +200,22 @@ func configureContainersForUnifiedImages(cluster *fdbv1beta2.FoundationDBCluster
...
+		}},
+		corev1.EnvVar{Name: "FDB_NETWORK_OPTION_TRACE_LOG_GROUP", Value: cluster.GetLogGroup()},
+		corev1.EnvVar{Name: "FDB_NETWORK_OPTION_TRACE_ENABLE", Value: "/var/log/fdb-trace-logs"},
+	)
+	mainContainerEnv = append(mainContainerEnv, getEnvForMonitorConfigSubstitution(cluster, processGroup.ProcessGroupID)...)
 	if cluster.DefineDNSLocalityFields() {
-		mainContainer.Env = append(mainContainer.Env, corev1.EnvVar{Name: "FDB_DNS_NAME", Value: GetPodDNSName(cluster, processGroup.GetPodName(cluster))})
+		mainContainerEnv = append(mainContainerEnv, corev1.EnvVar{Name: "FDB_DNS_NAME", Value: GetPodDNSName(cluster, processGroup.GetPodName(cluster))})
 	}
+	extendEnv(mainContainer, mainContainerEnv...)
```
johscheuer, line 218: I noticed that the `extendEnv ` has a bug where a user could add the same env multiple times in the same `extendEnv` call. The fix is pretty simple, we have to add here: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/internal/pod_models.go#L810 (in the if statement) `existingVars[envVar.Name] = true`.

internal/pod_models_test.go
nicmorales9: the new method will change the cap and cause `Equal()` checks to fail, I'll swap to `ConsistOf` no problem

internal/pod_models.go
nicmorales9: ooh nice catch

johscheuer (approved):
LGTM 👍

johscheuer (approved):
LGTM 👍

johscheuer:
> * Commit ID: [2a03693](https://github.com/FoundationDB/fdb-kubernetes-operator/commit/2a03693fc0c2d3ed254c68cad884072916989a1b)
> * Duration 2:38:48
> * Result: ❌ **FAILED**
> * [Build Log](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator/fdb-kubernetes-operator-pr/6a52ca2e-69cb-4e44-b5fa-919269ba67cf.gz) terminal output (available for 30 days)
> * [Build Workspace](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator-pr/6a52ca2e-69cb-4e44-b5fa-919269ba67cf/workspace.zip) zip file of the working directory (available for 30 days)

Failure is unrelated.

PR #2063, opened 2024-06-14
Change the image type from an env variable to an annotation

johscheuer:
# Description

It's probably better and more flexible to use an annotation instead of an env variable on the Pod.

## Type of change

## Discussion

-

## Testing

I'm running the migration test suite manually (will update the result here).

## Documentation

-

## Follow-up

-

### DIFF
--- a/e2e/test_operator_migrate_image_type/operator_migate_image_type_test.go
+++ b/e2e/test_operator_migrate_image_type/operator_migate_image_type_test.go
@@ -55,7 +55,7 @@ var _ = AfterSuite(func() {
 	factory.Shutdown()
 })
 
-var _ = Describe("Operator Migrate Image Type", Label("e2e"), func() {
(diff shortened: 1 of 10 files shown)


e2e/test_operator_migrate_image_type/operator_migate_image_type_test.go
```diff
@@ -55,7 +55,7 @@ var _ = AfterSuite(func() {
 	factory.Shutdown()
 })
 
-var _ = PDescribe("Operator Migrate Image Type", Label("e2e"), func() {
+var _ = Describe("Operator Migrate Image Type", Label("e2e"), func() {
```
johscheuer, line 58: That must be `PDescribe`.

nicmorales9 (approved):
Looks sensible to me, I'd just double check for usages of the old method

PR #2064, opened 2024-06-17
Add a test case for a multi-region cluster where the majority of coordinators fail

johscheuer:
# Description

The test is in pending mode as there is no way right now to recover the loss of the majority of coordinators. The idea for this test case is to have a repeatable way to test different recovery strategies.

## Type of change

- Other

## Discussion

-

## Testing

Ran manually.

## Documentation

-

## Follow-up

-

### DIFF
--- a/e2e/test_operator_ha/operator_ha_test.go
+++ b/e2e/test_operator_ha/operator_ha_test.go
@@ -30,8 +30,11 @@ This cluster will be used for all tests.
 */
 
 import (
+	"encoding/json"
+	"fmt"
 	"log"
 	"strconv"
+	"strings"
 	"time"
 
 	fdbv1beta2 "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta2"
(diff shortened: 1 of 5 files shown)


e2e/test_operator_ha/operator_ha_test.go
```diff
@@ -207,4 +210,93 @@ var _ = Describe("Operator HA tests", Label("e2e", "pr"), func() {
...
+
+			coordinatorCmd := fmt.Sprintf(
+				"coordinators %s",
+				fdbv1beta2.ProcessAddressesString(newCoordinators, " "),
+			)
+			log.Println("Command:", coordinatorCmd)
+
+			stdout, stderr, err := remote.RunFdbCliCommandInOperatorWithoutRetry(coordinatorCmd, true, 40)
+			log.Println("stdout:\n", stdout, "\nstderr:\n", stderr)
+			Expect(err).NotTo(HaveOccurred())
```
nicmorales9, line 297: In the current state, we'd expect an error here because this is not currently possible without a majority of current coordinators? https://apple.github.io/foundationdb/configuration.html

nicmorales9 (approved):
Looks right to me

e2e/test_operator_ha/operator_ha_test.go
johscheuer: Correct

Issue #1876, opened 2023-11-08
fdb-kubernetes-monitor delay shutdown


johscheuer:
### What would you like to be added/changed?

There are cases where it could be useful for the new fdb-kubernetes-monitor to delay the shutdown of its own process but shutting down the fdbserver processes when receiving a shutdown signal. This should be optional with the option to opt-out of this configuration. Shutting down the fdbserver processes but not the monitor itself has the benefit that the Pod will be stuck in the terminating state for the defined grace period (30 seconds) but the fdbserver process(es) will be shut down, giving the FDB cluster controller the opportunity to better react on the down processes.

johscheuer:
This actually doesn't require changes in the operator and only have to be documented.

PR #2065, opened 2024-06-17
Document the delay shutdown feature and check in update Pods if pod is missing

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1876

## Type of change

- Documentation

## Discussion

-

## Testing

-

## Documentation

Updated.

## Follow-up

-

### DIFF
--- a/docs/manual/operations.md
+++ b/docs/manual/operations.md
@@ -192,11 +192,16 @@ _NOTE_: You should always set the processes under maintenance before setting the
 ## Delaying the shutdown of the Pod
 
 When using the [unified image](./customization.md#unified-vs-split-images) the `fdb-kubernetes-monitor` supports to delay the shutdown of itself.
-The idea is to shutdown the fdbserver processes immediately, but keep the `fdb-kubernetes-monitor` process running for the specified `foundationdb.org/delay-shutdown`.
+This feature can help to reduce the risk of race conditions in the different checks of the operator and FDB itself.
+Right now the operator performs a different set of checks before updating (deleting and creating) Pods, like a fault tolerance check.
+In some rare cases there could be a risk where a Pod is already deleted and the FDB cluster haven't detected the failed servers yet.
+Per default the failure detection mechanism of FDB takes 60 seconds to mark a storage server as failed.
+The idea of the delayed shutdown feature is to shutdown the fdbserver processes, but keep the `fdb-kubernetes-monitor` process running for the specified `foundationdb.org/delay-shutdown`.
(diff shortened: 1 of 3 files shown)


docs/manual/operations.md
```diff
@@ -189,6 +189,15 @@ This will make sure that the processes are proactively excluded, instead of wait
 
 _NOTE_: You should always set the processes under maintenance before setting the maintenance mode. See [Internals](#internals) for more details.
 
+## Delaying the shutdown of the Pod
+
+When using the [unified image](./customization.md#unified-vs-split-images) the `fdb-kubernetes-monitor` supports to delay the shutdown of itself.
+The idea is to shutdown the fdbserver processes immediately, but keep the `fdb-kubernetes-monitor` process running for the specified `foundationdb.org/delay-shutdown`.
```
nicmorales9, line 195: I'm a bit unclear here, this makes it sound like it will just call for the fdbserver processes to get immediately shut down in any case, but below it makes it sound like changing the delay for the monitor somehow helps it with failed fdbserver processes, and I don't understand the benefit of detecting before shutting down

johscheuer: When the operator is doing an update on all Pods, it performs different checks e.g. https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/controllers/update_pods.go#L193-L195 it checks if there are any Pods that are recently deleted, if so it will not update other Pods. In addition the operator checks the fault tolerance and will only update Pods if the fault tolerance is at the desired level.

When the operator is now updating Pods and it deletes a Pod to update it, there can a window where the Pod is already gone and the operator has not yet created the new Pod. In addition if the shutdown of the fdbserver processes is less than 60 seconds ago, the fdb cluster will report full fault tolerance (as 60 seconds is the default failure detection time in fdb). So if the operator runs the update pod subreconciler again it might take another set of pods down, as there is no pod with a deletion timestamp and the fault tolerance is up. By adding the delay shutdown, the pod will be "stuck" in deletion and the fdbserver process will be shutdown. This will reduce the risk of the race condition above.

nicmorales9: that makes total sense, I think we should add some reference to this race being a motivator for using the delayed shutdown (I don't know if there is other documentation or not, but any sort of allusion would go a long way I think(

nicmorales9 (approved):
Looks good to me, though the PR name should reflect that non-documentation changes are going in, and it would be nice to have a test showing the logic when it's been over 90s since the condition time was set

nicmorales9 (approved):
test looks good

Issue #2071, opened 2024-06-19
Process Group ID should be increasing until a upper limit is reached


johscheuer:
### What would you like to be added/changed?

The current logic will take the first free Process Group ID. The intention was that the process group ID space was used efficiently. The issue with that approach is that if a process group fails to come up, we can see cases where a new process group gets the same name as the old one, even thought those are actually different process groups. This can be confusing when debugging issues. A better approach would be to increase the process group IDs until we hit an upper limit and then start from the beginning.

PR #2072, opened 2024-06-19
Randomize process group ids

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/2071. The issue was talking about another option, but I thought a bit more about it and using a randomized approach should be the best to prevent potential conflicts.

The PR is split in two commits: 1.) The actual code changes 2.) Test changes for all tests that assumed a deterministic setup.

## Type of change

## Discussion

As described in the issue, reusing the same process group ID, can be confusing in cases where a new process group is created with a process group ID that was just recently used.

## Testing

Updated all unit tests and CI will run e2e tests.

## Documentation

I will screen the docs to see if we have to adjust them.

## Follow-up

-

### DIFF
--- a/controllers/cluster_controller_test.go
+++ b/controllers/cluster_controller_test.go
@@ -92,7 +92,7 @@ func sortPodsByName(pods *corev1.PodList) {
 	})
 }
 
-var _ = FDescribe("cluster_controller", func() {
+var _ = Describe("cluster_controller", func() {
 	var cluster *fdbv1beta2.FoundationDBCluster
 	var fakeConnectionString string
 
--- a/controllers/cluster_controller_test.go
+++ b/controllers/cluster_controller_test.go
@@ -92,7 +92,7 @@ func sortPodsByName(pods *corev1.PodList) {
 	})
 }
 
-var _ = Describe("cluster_controller", func() {
+var _ = FDescribe("cluster_controller", func() {
 	var cluster *fdbv1beta2.FoundationDBCluster
 	var fakeConnectionString string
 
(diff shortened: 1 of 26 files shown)


johscheuer:
> * Commit ID: [6fbd0a9](https://github.com/FoundationDB/fdb-kubernetes-operator/commit/6fbd0a9b9447a7b181a6c78828a1c9990d8bc4a3)
> * Duration 2:38:09
> * Result: ❌ **FAILED**
> * [Build Log](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator/fdb-kubernetes-operator-pr/16d0edda-9f75-41e9-b97e-a0134919296b.gz) terminal output (available for 30 days)
> * [Build Workspace](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator-pr/16d0edda-9f75-41e9-b97e-a0134919296b/workspace.zip) zip file of the working directory (available for 30 days)

Checking the test failure, some are "expected" as they used the older methods.

controllers/cluster_controller_test.go
```diff
@@ -92,7 +92,7 @@ func sortPodsByName(pods *corev1.PodList) {
 	})
 }
 
-var _ = Describe("cluster_controller", func() {
+var _ = FDescribe("cluster_controller", func() {
```
nicmorales9, line 95: is this focus left on?

controllers/update_status_test.go
```diff
@@ -318,7 +313,7 @@ var _ = Describe("update_status", func() {
 
 		When("process group has no Pod", func() {
 			It("should be added to the failing Pods", func() {
-				processGroupStatus := fdbv1beta2.NewProcessGroupStatus("storage-1337", fdbv1beta2.ProcessClassStorage, []string{"1.1.1.1"})
+				processGroupStatus := fdbv1beta2.NewProcessGroupStatus("storage-100000", fdbv1beta2.ProcessClassStorage, []string{"1.1.1.1"})
```
nicmorales9, line 316: I don't follow the reasoning for this change

controllers/update_status_test.go
```diff
@@ -367,69 +358,55 @@ var _ = Describe("update_status", func() {
...
 					Expect(err).NotTo(HaveOccurred())
 
 					incorrectProcesses := fdbv1beta2.FilterByCondition(cluster.Status.ProcessGroups, fdbv1beta2.IncorrectCommandLine, false)
-					Expect(incorrectProcesses).To(Equal([]fdbv1beta2.ProcessGroupID{storageOneProcessGroupID}))
-
-					Expect(len(cluster.Status.ProcessGroups)).To(BeNumerically(">", 4))
-					processGroup := cluster.Status.ProcessGroups[len(cluster.Status.ProcessGroups)-4]
-					Expect(processGroup.ProcessGroupID).To(Equal(storageOneProcessGroupID))
-					Expect(len(processGroup.ProcessGroupConditions)).To(Equal(1))
-					Expect(processGroup.IsMarkedForRemoval()).To(BeTrue())
```
nicmorales9, line 412: the check on 412 doesn't seem superfluous to me (not in the filterByCondition like previous ones)

nicmorales9 (commented):
overall looks good but some concerns

e2e/test_operator_upgrades_with_chaos/operator_upgrades_with_chaos_test.go
```diff
@@ -164,9 +164,6 @@ var _ = Describe("Operator Upgrades with chaos-mesh", Label("e2e", "pr"), func()
 				Consistently(func() bool {
 					return fdbCluster.GetCluster().Status.RunningVersion == beforeVersion
 				}).WithTimeout(2 * time.Minute).WithPolling(2 * time.Second).Should(BeTrue())
-			} else {
-				// Simulate the 2 minute wait otherwise the auto replacement might not succeed in time.
-				time.Sleep(2 * time.Minute)
```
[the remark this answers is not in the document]
johscheuer: That sleep time was useless in the case of protocol compatible upgrades as the operator would move ahead and recreate the Pods for the new desired version. In addition below we only check for Pod removal in the case of version incompatible upgrades (because in the case of the protocol compatible upgrade the upgrade process is not blocked by a missing process and if that partitioned pod is recreated the chaos-mesh partition is lost.

controllers/update_status_test.go
johscheuer: We add a new process group here and the idea was to just add a process group with the id `100000` as `99999` is the maximum generated id. The idea was to make sure we prevent conflicts where the randomness chooses the ID 1337. I updated the code with another solution I used in a different place.

nicmorales9: oooh, neat! that makes sense

johscheuer: I decided just to delete the pod in that test case, not sure why we didn't do that before 🤔

nicmorales9 (approved):
changes look great, though I would also check over for documentation that suggests process group ID selection would be in order as I seem to recall that being a thing *somewhere*

johscheuer:
> changes look great, though I would also check over for documentation that suggests process group ID selection would be in order as I seem to recall that being a thing _somewhere_

Good point, I updated the docs.

Issue #873, opened 2021-08-06
Proposal for a different way to track reconciliation progress


brownleej:
We currently track progress of reconciliation through a `generations` field, which tracks how far reconciliation has gone for different generations. This introduces some confusing behaviors when determining if reconciliation is complete. For instance, when reconciliation is complete we generally set the generation status to `{Reconciled: $gen}`, with no generations set for any of the other fields. However, if there are pods stuck in terminating, we set it to `{Reconciled: $gen, HasPendingRemoval: $gen}`. There are also cases where reconciliation can complete, and then another event will happen that leaves the cluster in a not-fully-deployed state, such as a pod getting deleted and needing to get recreated. When that happens we set the generation status to `{}`, effectively resetting the reconciled generation to 0.

As an alternative, I think we should consider tracking reconciliation through an `observedGeneration` field and a list of `conditions`. The `UpdateStatus` action will set the `observedGeneration` to the current generation from the metadata, and add in any conditions it observes that reconciliation needs to remedy. The conditions here could be an extension of the conditions we set on process groups, and we could add new conditions to be analogous to the fields in the `GenerationStatus` struct we have today. Users of the operator could track reconciliation being complete by checking if `generation==observedGeneration && conditions==[]`. For cases like the pending removal condition where people may want to ignore certain conditions that can persist a long time, we can document those conditions leave it up to each user to determine what they want to wait on. We could also include the generation, observedGeneration, and conditions in the `kubectl get` output to give a quicker view into where clusters are held up in reconciliation.

Issue #1000, opened 2021-11-28
Move ClusterGenerationStatus into conditions


johscheuer:
I think that the [ClusterGenerationStatus](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/b97e85a59dc859a60eede96340dc305031cbc8f3/api/v1beta1/foundationdbcluster_types.go#L850-L921) is a good candidate to actually move into [conitions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) with the benefit that a user can see when the `LastTransitionTime` of that condition was without looking into the logging system.

PR #2073, opened 2024-06-20
Don't reset the last reconciled generation

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/873

## Type of change

## Discussion

Before that we would reset the generation to 0 which can be confusing in many cases when the reconciliation drops from a precious number to 0.

## Testing

Ran unit tests and an e2e test suite.

## Documentation

Not needed.

## Follow-up

-

### DIFF
--- a/api/v1beta2/foundationdbcluster_types.go
+++ b/api/v1beta2/foundationdbcluster_types.go
@@ -1669,8 +1669,6 @@ func (cluster *FoundationDBCluster) CheckReconciliation(log logr.Logger) (bool, 
 
 	if reconciled {
 		cluster.Status.Generations.Reconciled = cluster.ObjectMeta.Generation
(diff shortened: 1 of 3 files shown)


nicmorales9 (approved):
Changes look good to me though it feels like it should not close the linked issue as it does not add the observedGenerations nor Conditions.  If this is intentional, an explanation on the issue would be good

johscheuer:
> Changes look good to me though it feels like it should not close the linked issue as it does not add the observedGenerations nor Conditions. If this is intentional, an explanation on the issue would be good

For the conditions we have a dedicated issue: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1000. Adding the `observedGeneration` didn't seem to bring a huge value in this context.

PR #2076, opened 2024-06-24
Add an upgrade test case where a single fdb-kubernetes-monitor is not able to connect to the Kubernetes API

johscheuer:
# Description

As expected the upgrade will be stuck until the fdb-kubernetes-monitor can communicate to the K8s API again.

## Type of change

- Other

## Discussion

-

## Testing

Ran the test manually:

```text
• [377.644 seconds]
[AfterSuite]
/Users/jscheuermann/go/src/github.com/FoundationDB/fdb-kubernetes-operator/e2e/test_operator_upgrades/operator_upgrades_test.go:55
[AfterSuite] PASSED [0.000 seconds]
[ReportAfterSuite] Autogenerated ReportAfterSuite for --junit-report
autogenerated by Ginkgo
[ReportAfterSuite] PASSED [0.004 seconds]

Ran 1 of 11 Specs in 377.644 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 10 Skipped
PASS | FOCUSED
FAIL    github.com/FoundationDB/fdb-kubernetes-operator/e2e/test_operator_upgrades      378.067s
FAIL
```

## Documentation

-

## Follow-up

-

### DIFF
--- a/e2e/test_operator_upgrades/operator_upgrades_test.go
+++ b/e2e/test_operator_upgrades/operator_upgrades_test.go
@@ -650,4 +650,75 @@ var _ = Describe("Operator Upgrades", Label("e2e", "pr"), func() {
 		EntryDescription("Upgrade from %[1]s to %[2]s"),
 		fixtures.GenerateUpgradeTableEntries(testOptions),
 	)
+
+	DescribeTable(
+		"one process cannot connect to the Kubernetes API",
+		func(beforeVersion string, targetVersion string) {
+			if !factory.ChaosTestsEnabled() {
+				Skip("Chaos tests are skipped for the operator")
+			}
+
+			if fixtures.VersionsAreProtocolCompatible(beforeVersion, targetVersion) {
+				Skip("this test only affects version incompatible upgrades")
+			}
+
+			// If we are not using the unified image, we can skip this test.
+			if !factory.UseUnifiedImage() {
+				Skip("The sidecar image doesn't require connectivity to the Kubernetes API")
+			}
+
+			clusterSetup(beforeVersion, true)
+
+			selectedPod := factory.RandomPickOnePod(fdbCluster.GetStoragePods().Items)
+
+			var kubernetesServiceHost string
+			Eventually(func(g Gomega) error {
+				std, _, err := factory.ExecuteCmdOnPod(
+					&selectedPod,
+					fdbv1beta2.MainContainerName,
+					"printenv KUBERNETES_SERVICE_HOST",
+					false,
+				)
+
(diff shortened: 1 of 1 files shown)


e2e/test_operator_upgrades/operator_upgrades_test.go
```diff
@@ -650,4 +650,75 @@ var _ = Describe("Operator Upgrades", Label("e2e", "pr"), func() {
...
+					false,
+				)
+
+				return err
+			}).WithTimeout(2 * time.Minute).WithPolling(1 * time.Second).Should(HaveOccurred())
+
+			// Update the cluster version.
+			Expect(fdbCluster.UpgradeCluster(targetVersion, false)).NotTo(HaveOccurred())
+
+			if !fixtures.VersionsAreProtocolCompatible(beforeVersion, targetVersion) {
```
nicmorales9, line 705: just for my understanding, if they are protocol compatible then what changes with the upgrade? wouldn't we still expect the same behaviour (except that the IncorrectConfig conditions are not added, and so the bounce isn't needed)? From my reading, it seems like the only thing that would change is the wait for config maps to update before the coordinated bounce is done to upgrade the cluster, but I seem to be missing something.

nicmorales9 (approved):
changes look good! just 1 question for my understanding

e2e/test_operator_upgrades/operator_upgrades_test.go
johscheuer: When the cluster is upgraded to a protocol compatible version (patch upgrade) the operator skips the staging and restart phase and directly goes to the [recreation](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/docs/manual/upgrades.md#recreation-of-pods-phase) phase as all fdbserver processes will be able to communicate, so there is no need to stage the new binaries and that's the only step that blocks. But I agree that we should test this behaviour too.

nicmorales9: ohh so since there is no blocking / waiting for configmap update, there isn't need for the k8s API just FDB cluster communication?

johscheuer: Correct 👍 And once the Pod with the K8s API connection issues is recreated the chaos injection will be lost.

nicmorales9: okay that makes sense, thanks!

johscheuer:
I looked at the test failures and those are unrelated. Will be looking at them more closely.

PR #2082, opened 2024-06-26
Minor fixes for our e2e tests to improve reliability

johscheuer:
# Description

One fix for a test case that could hit https://github.com/apple/foundationdb/issues/11222 in rare cases and an increase in the timeout time for the crash-looping setup.

## Type of change

## Discussion

-

## Testing

Ran those test manually but CI will also run them.

## Documentation

-

## Follow-up

-


nicmorales9 (approved):

johscheuer:
As the previous run went fine and the only failure was the timeout because of the conflict (which was fixed in another PR), I'm going to merge this PR.

Issue #2018, opened 2024-05-08
Consider doing exclusion and coordinator change at once


johscheuer:
### What would you like to be added/changed?

We should explore ways to do the exclusion of processes and the coordinator change in a single fdbcli call to reduce the overall recoveries. This requires some changes in the current operator code.

Related: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2017

PR #2083, opened 2024-06-26
Run the change coordinator command before excluding the coordinator

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/2018

## Type of change

## Discussion

-

## Testing

Updated the unit tests.

## Documentation

-

## Follow-up

-


controllers/exclude_processes.go
```diff
@@ -139,12 +155,42 @@ func (e excludeProcesses) reconcile(_ context.Context, r *FoundationDBClusterRec
 		}
 	}
 
+	var coordinatorExcluded bool
+	for _, excludeProcess := range fdbProcessesToExclude {
+		excludeString := excludeProcess.String()
+		_, excludedLocality := coordinatorsExclusionString[excludeString]
+		_, excludedAddress := coordinatorsAddress[excludeString]
```
nicmorales9, line 162: out of curiosity, is it possible to have one but not the other? e.g. could a coordinator ever be listed under status.ProcessGroup but not processGroup.Addresses?

controllers/exclude_processes_test.go
```diff
@@ -756,6 +757,68 @@ var _ = Describe("exclude_processes", func() {
...
+					if _, ok := coordinators[string(processGroup.ProcessGroupID)]; !ok {
+						continue
+					}
+
+					processGroup.MarkForRemoval()
+					break
+				}
+			})
+
+			FIt("blub", func() {
```
nicmorales9, line 804: focus

internal/test_helper.go
```diff
@@ -136,3 +139,54 @@ func PickProcessGroups(cluster *fdbv1beta2.FoundationDBCluster, processClass fdb
...
+			continue
+		}
+
+		processGroupIDs[processClass] = map[int]bool{}
+		for i := 1; i <= count; i++ {
+			processGroupID := cluster.GetNextRandomProcessGroupID(processClass, processGroupIDs[processClass])
+			newProcessGroup := fdbv1beta2.NewProcessGroupStatus(processGroupID, processClass, nil)
+			newProcessGroup.ProcessGroupConditions = nil
+
+			ipIdx++
```
nicmorales9, line 170: unused?

controllers/change_coordinators_test.go
```diff
@@ -22,33 +22,24 @@ package controllers
...
 var _ = Describe("Change coordinators", func() {
 	var cluster *fdbv1beta2.FoundationDBCluster
-	var adminClient *mock.AdminClient
 
 	BeforeEach(func() {
 		cluster = internal.CreateDefaultCluster()
-		disabled := false
-		cluster.Spec.LockOptions.DisableLocks = &disabled
+		//disabled := false
+		//cluster.Spec.LockOptions.DisableLocks = &disabled
```
nicmorales9, line 42: unsure if leftover from testing, but I would remove the lines or uncomment them

internal/coordinator/coordinator_test.go
```diff
@@ -0,0 +1,1043 @@
...
+
+					// Only select Storage processes since we select 3 processes and we have 4 storage processes
+					for _, candidate := range candidates {
+						Expect(candidate.ID).NotTo(Equal(removedProcess))
+						Expect(candidate.Class).To(Equal(fdbv1beta2.ProcessClassStorage))
+					}
+				})
+			})
+
+			//When("when one storage process is excluded", func() {
```
nicmorales9, line 115: again, would be good to delete or uncomment - though ideally this could be adjusted to still pass

internal/test_helper.go
```diff
@@ -136,3 +139,54 @@ func PickProcessGroups(cluster *fdbv1beta2.FoundationDBCluster, processClass fdb
 
 	return pickedProcessGroups
 }
+
+// SetupClusterForTest will generate all required resources like Pods for test cases.
+func SetupClusterForTest(cluster *fdbv1beta2.FoundationDBCluster, k8sClient ctrlClient.Client) error {
```
nicmorales9, line 144: mostly just curious, but what is the intent of having this similar function instead of the original `SetupClusterForTest` in `controllers`?

controllers/exclude_processes.go
johscheuer: > e.g. could a coordinator ever be listed under status.ProcessGroup but not processGroup.Addresses?

The addresses are fetched from `status.ProcessGroup.Addresses`. We handle there the two different exclusions modes: Exclusion by IP (and port) or locality.

internal/test_helper.go
johscheuer: cyclic dependencies :)

nicmorales9: ooh

e2e/test_operator/operator_test.go
```diff
@@ -370,13 +371,28 @@ var _ = Describe("Operator", Label("e2e", "pr"), func() {
...
+				_, isCoordinator := coordinators[string(fixtures.GetProcessGroupID(pod))]
+				if isCoordinator {
+					replacedPod = pod
+					break
+				}
+			}
+
+			// In case that none of the log processes are coordinators
+			if replacedPod.Name == "" {
+				replacedPod = factory.RandomPickOnePod(initialPods.Items)
```
nicmorales9, line 392: is the intention to prefer coordinator logs and then just pick anything else otherwise? why prefer logs that are coordinators in the first place - just to try to add some coverage on fun cases?

johscheuer: The idea is to prefer coordinators to make sure the operator handles the case with replacing a coordinator well (with the assumption that if the operator handles the coordinator replacement well it also handles the normal replacement well). If no log processes are coordinators the test suite will randomly pick a log process.

PR #2084, opened 2024-06-26
Fix a bug where a process group with the same ID could be generated

johscheuer:
# Description

Fix a bug where the same process group ID could be generated.

## Type of change

## Discussion

0
## Testing

Added an additional test case for this. Without the fix the test case failed.

## Documentation

-

## Follow-up

-


nicmorales9 (approved):
I completely missed that too 🥲

PR #2095, opened 2024-07-04
Make sure to only check the isolate process group annotation if a pod was returned

johscheuer:
# Description

In most cases this check is not harmful, only in cases where someone would implement their own pod manager and that one would return a `nil` instead of a pointer to an empty pod struct.

## Type of change

## Discussion

-

## Testing

Ran unit tests locally.

## Documentation

-

## Follow-up

-


nicmorales9 (approved):
another nice one!

PR #2097, opened 2024-07-04
Check if the namespace is created by the same test suite

johscheuer:
# Description

Prevent cases where two concurrent test suites will choose the same random namespace and cluster name. This change will make sure that only one test suite will take the namespace and the other test suite will generate a new random name. This will prevent cases where those two concurrent test suites use the same namespace, which will cause a failure of both test cases.

## Type of change

## Discussion

-

## Testing

Ran a manual test (the test suite was picking the same namespace as I defined it with the according environment variable):

```text
Operator Stress when creating and deleting a cluster multiple times should create a healthy and available cluster [e2e]
/Users/jscheuermann/go/src/github.com/FoundationDB/fdb-kubernetes-operator/e2e/test_operator_stress/operator_stress_test.go:53
  2024/07/04 12:38:08 created namespace jscheuermann
  2024/07/04 12:38:09 namespace jscheuermann already in use by test suite: operator-test, current test suite: operator-stress
  2024/07/04 12:38:10 created namespace jscheuermann
```

## Documentation

-

## Follow-up

-

### DIFF
--- a/e2e/fixtures/kubernetes_fixtures.go
+++ b/e2e/fixtures/kubernetes_fixtures.go
@@ -96,20 +96,14 @@ func (factory *Factory) createNamespace(suffix string) string {
 		}
 
 		err := factory.checkIfNamespaceIsTerminating(namespace)
-		if err != nil {
-			return err
-		}
+		g.Expect(err).NotTo(gomega.HaveOccurred())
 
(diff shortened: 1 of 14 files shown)


nicmorales9 (approved):
looks good to me

johscheuer:
The test failures above were unrelated to this change.

PR #2098, opened 2024-07-05
Try to get the running version based on the reachable coordinators during an upgrade

johscheuer:
# Description

We sometimes see that the test case `Operator HA Upgrades when no remote storage processes are restarted` fails where the coordinators are reporting in the new upgraded version but the cluster is stuck with the following message:

```json
 "messages": [
      {
        "name": "status_incomplete_timeout",
        "description": "Timed out fetching cluster status."
      }
    ]
```

The idea for this change is to get the running version based on the reachable coordinators. If a quorum of coordinators can be reached with a specific `fdbcli` version the operator will assume that the cluster is running in that version. Doing that will unblock the operator and will allow the operator to move forward. If the running version is not updated to the new desired version the reconciler `checkClientCompatibility` will block (as the operator is not able to check for incompatible clients).

## Type of change

- Bug fix (non-breaking change which fixes an issue)n

## Discussion

-

## Testing

Will be running multiple e2e tests for this change.

## Documentation

Must be updated.

## Follow-up

-

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -592,7 +592,19 @@ func (r *FoundationDBClusterReconciler) getStatusFromClusterOrDummyStatus(logger
 		return status, nil
 	}
 
+	// When we reached this part of the code the above GetStatus() called failed for some reason.
 	if cluster.Status.Configured {
+		// If the cluster is currently under a version incompatible upgrade, we try to assume the current version based
+		// on the coordinator reachability. If all (or the majority) of coordinators are reachable with a specific version
+		// of fdbcli we can assume that the cluster is running with that version and update the cluster.Status.RunningVersion.
+		// In theory we could use the go bindings if they would expose that information from the multi-version bindings.
+		if cluster.IsBeingUpgradedWithVersionIncompatibleVersion() {
+			// Set the version from the reachable coordinators, if the version points to the desired version defined
+			// in the cluster.Spec.Version, this will unblock some further steps, to allow the operator to bring the cluster
+			// back into a better state.
+			cluster.Status.RunningVersion = adminClient.GetVersionFromReachableCoordinators()
+		}
+
 		return nil, err
 	}
 
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -602,7 +602,11 @@ func (r *FoundationDBClusterReconciler) getStatusFromClusterOrDummyStatus(logger
 			// Set the version from the reachable coordinators, if the version points to the desired version defined
 			// in the cluster.Spec.Version, this will unblock some further steps, to allow the operator to bring the cluster
 			// back into a better state.
-			cluster.Status.RunningVersion = adminClient.GetVersionFromReachableCoordinators()
+			versionFromReachableCoordinators := adminClient.GetVersionFromReachableCoordinators()
+			if versionFromReachableCoordinators != cluster.Status.RunningVersion {
+				logger.Info("Update running version in cluster status from reachable coordinators", "versionFromReachableCoordinators", versionFromReachableCoordinators, "currentRunningVersion", cluster.Status.RunningVersion)
+				cluster.Status.RunningVersion = versionFromReachableCoordinators
+			}
 		}
 
 		return nil, err
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -603,7 +603,7 @@ func (r *FoundationDBClusterReconciler) getStatusFromClusterOrDummyStatus(logger
 			// in the cluster.Spec.Version, this will unblock some further steps, to allow the operator to bring the cluster
 			// back into a better state.
 			versionFromReachableCoordinators := adminClient.GetVersionFromReachableCoordinators()
-			if versionFromReachableCoordinators != cluster.Status.RunningVersion {
+			if versionFromReachableCoordinators != "" && versionFromReachableCoordinators != cluster.Status.RunningVersion {
 				logger.Info("Update running version in cluster status from reachable coordinators", "versionFromReachableCoordinators", versionFromReachableCoordinators, "currentRunningVersion", cluster.Status.RunningVersion)
 				cluster.Status.RunningVersion = versionFromReachableCoordinators
 			}
--- a/fdbclient/admin_client.go
+++ b/fdbclient/admin_client.go
@@ -275,7 +275,7 @@ func (client *cliAdminClient) runCommand(command cliCommand) (string, error) {
 }
 
 // getStatusFromCli uses the fdbcli to connect to the FDB cluster
-func (client *cliAdminClient) getStatusFromCli() (*fdbv1beta2.FoundationDBStatus, error) {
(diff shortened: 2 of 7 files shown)


controllers/cluster_controller.go
```diff
@@ -592,7 +592,23 @@ func (r *FoundationDBClusterReconciler) getStatusFromClusterOrDummyStatus(logger
 		return status, nil
 	}
 
+	// When we reached this part of the code the above GetStatus() called failed for some reason.
 	if cluster.Status.Configured {
+		// If the cluster is currently under a version incompatible upgrade, we try to assume the current version based
+		// on the coordinator reachability. If all (or the majority) of coordinators are reachable with a specific version
+		// of fdbcli we can assume that the cluster is running with that version and update the cluster.Status.RunningVersion.
+		// In theory we could use the go bindings if they would expose that information from the multi-version bindings.
```
nicmorales9, line 600: Go version of this one? https://github.com/apple/foundationdb/blob/main/bindings/c/fdb_c.cpp#L683

controllers/cluster_controller.go
```diff
@@ -592,7 +592,23 @@ func (r *FoundationDBClusterReconciler) getStatusFromClusterOrDummyStatus(logger
 		return status, nil
 	}
 
+	// When we reached this part of the code the above GetStatus() called failed for some reason.
 	if cluster.Status.Configured {
+		// If the cluster is currently under a version incompatible upgrade, we try to assume the current version based
+		// on the coordinator reachability. If all (or the majority) of coordinators are reachable with a specific version
+		// of fdbcli we can assume that the cluster is running with that version and update the cluster.Status.RunningVersion.
```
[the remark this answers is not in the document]
nicmorales9: or is it more that after we've done a bounce, if the coordinators are up then we know the cluster is running that version?

nicmorales9 (commented):
it seems legit but I want to make sure I understand things first

controllers/cluster_controller.go
johscheuer: I haven't looked into the semantics of that method, but that might work. I'm not sure if `// Returns the protocol version reported by the coordinator this client is connected to` means that the connected coordinator also represents the quorum of running coordinators. Another subtle difference is that this will return the protocol version not the FDB release version (which could be mapped).

controllers/cluster_controller.go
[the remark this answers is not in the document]
johscheuer: This case is only here to help the operator to detect the running version when the `machine-readable` status is not available, e.g. the above `GetStatus` returned an error. If the status is returned, the operator will detect the running version based on the running processes. So the intent here is to help the operator detect the running version based on the reporting coordinators (so those that are reachable in a specific version), to unblock the operator and let it move forward. The reporting coordinators will be either reachable in a quorum of the old version, the new version or none, but all those three states are distinct, e.g. if a quorum is reachable it will be either in the old or the new version but cannot be both.

johscheuer:
> * Commit ID: [b5c3674](https://github.com/FoundationDB/fdb-kubernetes-operator/commit/b5c36749daca952eb56b5cb7bab8bfccb59dd63a)
> * Duration 2:58:25
> * Result: ❌ **FAILED**
> * [Build Log](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator/fdb-kubernetes-operator-pr/27deaebe-6091-429b-8089-826cff0c32f4.gz) terminal output (available for 30 days)
> * [Build Workspace](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator-pr/27deaebe-6091-429b-8089-826cff0c32f4/workspace.zip) zip file of the working directory (available for 30 days)

```text
[Container] 2024/07/09 14:55:01.358154 Running command if $(grep -q -- "--- FAIL:" ${CODEBUILD_SRC_DIR}/logs/*.log); then echo "TESTS FAILED SEE THESE LOGS:"; echo ; grep -l -- "--- FAIL:" ${CODEBUILD_SRC_DIR}/logs/*.log; export fail_test=true; fi
TESTS FAILED SEE THESE LOGS:

/codebuild/output/src1406809119/src/github.com/FoundationDB/fdb-kubernetes-operator/logs/test_operator_upgrades.log
```

```text
• [FAILED] [448.057 seconds]
Operator Upgrades one process is marked for removal and is stuck in terminating state [It] Upgrade from 7.1.57 to 7.3.33 [e2e, pr]
/codebuild/output/src1406809119/src/github.com/FoundationDB/fdb-kubernetes-operator/e2e/fixtures/upgrade_test_configuration.go:115

  [FAILED] Timed out after 300.001s.
  Expected
      <*int64 | 0x0>: nil
  not to be nil
  In [It] at: /codebuild/output/src1406809119/src/github.com/FoundationDB/fdb-kubernetes-operator/e2e/test_operator_upgrades/operator_upgrades_test.go:569 @ 07/09/24 13:15:55.085
```

https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/e2e/test_operator_upgrades/operator_upgrades_test.go#L557-L569 is where the failure happened, unrelated to the PR changes.

fdbclient/admin_client.go
```diff
@@ -823,3 +823,40 @@ func (client *cliAdminClient) SetProcessesUnderMaintenance(processGroupIDs []fdb
...
+func (client *cliAdminClient) GetVersionFromReachableCoordinators() string {
+	// First we test to get the status from the fdbcli with the current running version defined in cluster.Status.RunningVersion.
+	status, _ := client.getStatusFromCli(false)
+	if quorumOfCoordinatorsAreReachable(status) {
+		return client.Cluster.GetRunningVersion()
+	}
+
+	// If the majority of coordinators are not reachable with the cluster.Status.RunningVersion, we try the desired version
+	// if the cluster is currently performing an version incompatible upgrade.
+	if client.Cluster.IsBeingUpgradedWithVersionIncompatibleVersion() {
```
nicmorales9, line 840: Is this changing the version the client uses (or where it tries to read the status) so that if it works to get status then we know it's updated even if it doesn't have running processes? I don't really follow

nicmorales9 (commented):
everything looks right in theory but I still am confused 😆

fdbclient/admin_client.go
nicmorales9: Discussed offline, it makes sense now but it would be good to add a comment here that setting the running version as the desired one `checkClientCompatibility` gets skipped and allows the status to be got

nicmorales9 (approved):
makes sense now, thanks for sticking with me! I think it would be good to improve the comments as per the one remaining comment but otherwise LGTM

fdbclient/admin_client.go
johscheuer: This would be the wrong place for the comment as this part will only change the running version for the client's cluster copy to force it to use the newer fdbcli version. The actual change is done in the cluster controller where we already have a comment:

https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2098/files#diff-1ca08b6f9e9610ef14189184290a644a29adbb47e6f8653a479d75fdce4a8452R602-R604

PR #2100, opened 2024-07-10
Improve team tracker checks when DD was restarted

johscheuer:
# Description

See comments in code.

## Type of change

## Discussion

-

## Testing

Ran unit tests and CI will run e2e tests.

## Documentation

-

## Follow-up

-


nicmorales9 (approved):
LGTM assuming initializing guarantees that it will come up healthy

johscheuer:
> LGTM assuming initializing guarantees that it will come up healthy

Not necessarily, at this moment the state is unknown. But if a failure would be detected after the DD was started, it would result in the according team tracker to be updated and the Eventually statement would fail (if the replicas fall below the expected threshold).

PR #2107, opened 2024-07-26
Correct the base image for the unified image and increase the memory for the FDB pods to 8GiB

johscheuer:
# Description

Correcting the default base image for the unified image.

## Type of change

## Discussion

-

## Testing

Added unit tests and e2e test will be running.

## Documentation

-

## Follow-up

-

### DIFF
--- a/internal/pod_models_test.go
+++ b/internal/pod_models_test.go
@@ -285,9 +285,9 @@ var _ = Describe("pod_models", func() {
 				}))
 
 				Expect(*mainContainer.Resources.Limits.Cpu()).To(Equal(resource.MustParse("1")))
-				Expect(*mainContainer.Resources.Limits.Memory()).To(Equal(resource.MustParse("4Gi")))
+				Expect(*mainContainer.Resources.Limits.Memory()).To(Equal(resource.MustParse("8Gi")))
 				Expect(*mainContainer.Resources.Requests.Cpu()).To(Equal(resource.MustParse("1")))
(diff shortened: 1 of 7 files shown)


internal/pod_models_test.go
```diff
@@ -560,7 +560,7 @@ var _ = Describe("pod_models", func() {
 				It("should have the sidecar container", func() {
 					sidecarContainer := spec.Containers[1]
 					Expect(sidecarContainer.Name).To(Equal(fdbv1beta2.SidecarContainerName))
-					Expect(sidecarContainer.Image).To(Equal(fmt.Sprintf("foundationdb/foundationdb-kubernetes:%s", cluster.Spec.Version)))
+					Expect(sidecarContainer.Image).To(And(HavePrefix(fdbv1beta2.FoundationDBKubernetesBaseImage), HaveSuffix(cluster.Spec.Version)))
```
nicmorales9, line 563: wouldn't this be true for all of the BaseImage variables?

nicmorales9 (approved):
LGTM but do confirm the prefix/suffix-check comment

internal/pod_models_test.go
nicmorales9: no -- I was looking at the prior variable but `FoundationDBKubernetesBaseImage = "foundationdb/fdb-kubernetes-monitor"` is not a prefix for the others like `"foundationdb/foundationdb-kubernetes"` would be

PR #2112, opened 2024-08-07
Bug fix: Added required field namespace for global mode

geoffreytran:
# Description

Bug fix: Added required field namespace for global mode

## Type of change


johscheuer (approved):
Thanks!

PR #2113, opened 2024-08-08
Make sure the e2e tests are running with different storage engines

johscheuer:
# Description

Make sure the e2e tests are running with different storage engines
## Type of change

## Discussion

-

## Testing

Ran a manual test with it.

## Documentation

-

## Follow-up

-

### DIFF
--- a/e2e/Makefile
+++ b/e2e/Makefile
@@ -10,9 +10,9 @@ TIMEOUT?=168h
 CLUSTER_NAME?=
 NAMESPACE?=
 CONTEXT?=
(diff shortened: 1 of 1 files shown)


nicmorales9 (approved):
LGTM assuming that this is logged somewhere, I didn't see the field explicitly logged but assumed that there is something that logs all of the cluster config

johscheuer:
> LGTM assuming that this is logged somewhere, I didn't see the field explicitly logged but assumed that there is something that logs all of the cluster config

It's logged in the start command of the test and the cluster configuration.

PR #2117, opened 2024-08-27
Only release the lock when the cluster is reconciled

johscheuer:
# Description

In the past we added code to release the lock when the operation was performed. Doing that increases the risk of race conditions when multiple operator instances are managing a multi-region (or three data hall) FDB cluster. In order to reduce the risk of those race conditions the operator is releasing the lock only when the cluster is reconciled (or when the lock is timed out).

## Type of change

## Discussion

The general idea of the locking mechanism is to reduce the operations executed in parallel on the FDB cluster. When the lock is directly released after the operation we have the risk that another operator instance is directly doing another operation that could be disruptive, e.g. excluding processes.

## Testing

e2e test will be running by CI.

## Documentation

Will be updated.

## Follow-up

-

### DIFF
--- a/controllers/bounce_processes.go
+++ b/controllers/bounce_processes.go
@@ -167,6 +167,9 @@ func (c bounceProcesses) reconcile(_ context.Context, r *FoundationDBClusterReco
 		return &requeue{curError: err}
 	}
 
+	// Reset the SecondsSinceLastRecovered sine the operator just restarted some processes, which will could cause a recovery.
+	status.Cluster.RecoveryState.SecondsSinceLastRecovered = 0.0
+
 	// If the cluster was upgraded we will requeue and let the update_status command set the correct version.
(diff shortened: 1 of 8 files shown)


nicmorales9 (approved):
LGTM, though I was thinking it could be nice to output the cluster.LockDuration in the TakeLock failure output for users to see how long they should expect to wait in the case of a problem (though I don't know how many people use multi operator setups)

johscheuer (commented):
> LGTM, though I was thinking it could be nice to output the cluster.LockDuration in the TakeLock failure output for users to see how long they should expect to wait in the case of a problem (though I don't know how many people use multi operator setups)

That makes sense, I'll add it 👍

johscheuer (commented):
The logger already contains the information, so there is no need to update the code: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/fdbclient/lock_client.go#L118-L147

PR #2119, opened 2024-08-30
Read the coordinator IPs from the running Pods if they are present

johscheuer:
# Description

It's better to fetch the IP address directly from the running Pod (if the pod is running), instead of using the address information in the FoundationDBCluster resource. The Pod IP will provide the most recent information, but the `FoundationDBCluster` might be outdated (it's still used as a fall back).

## Type of change

## Discussion

-

## Testing

I updated the existing test.

## Documentation

N/A

## Follow-up

-

### DIFF
--- a/kubectl-fdb/cmd/fix_coordinator_ips.go
+++ b/kubectl-fdb/cmd/fix_coordinator_ips.go
@@ -21,13 +21,16 @@
 package cmd
 
 import (
-	ctx "context"
+	"context"
 	"fmt"
 	"log"
 	"net"
 	"os"
 	"os/exec"
 	"strings"
+
+	"github.com/go-logr/logr"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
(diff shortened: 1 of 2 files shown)


kubectl-fdb/cmd/fix_coordinator_ips.go
```diff
@@ -163,44 +166,97 @@ func buildClusterFileUpdateCommands(cluster *fdbv1beta2.FoundationDBCluster, kub
...
 		if err != nil {
 			return err
 		}
-		for _, processGroup := range cluster.Status.ProcessGroups {
+
+		processGroup, ok := coordinatorProcessGroup[coordinatorAddress.MachineAddress()]
+		if !ok {
+			// Keep the old address if the coordinator process group is missing.
+			newCoordinators[coordinatorIndex] = coordinatorAddress.String()
+			cmd.Println("ProcessGroup for", coordinatorAddress.MachineAddress(), "is missing in the FoundationDBCluster status, coordinator address will be ignored")
```
nicmorales9, line 212: The message reads a bit odd to me, IIUC it's not quite ignored- it's just not updated with a correct address? Unless something in the FDB logic will cause the no-longer-existing address to get dropped?

johscheuer: It's ignored for the purpose of the command, but I agree, the wording should be improved 👍

nicmorales9 (approved):
LGTM

PR #2124, opened 2024-09-03
Add recovery plugin command

johscheuer:
# Description

This PR is based on https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2122.

This PR adds a recovery command to recover a multi-region cluster is the majority of coordinators are lost.

## Type of change

*Please select one of the options below.*

- New feature (non-breaking change which adds functionality)

## Discussion

-

## Testing

Performed manual testing and ran the new e2e test:

```text
[ReportAfterSuite] Autogenerated ReportAfterSuite for --junit-report
autogenerated by Ginkgo
[ReportAfterSuite] PASSED [0.007 seconds]
------------------------------

Ran 2 of 2 Specs in 1509.781 seconds
SUCCESS! -- 2 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestOperator (1509.79s)
PASS
ok      github.com/FoundationDB/fdb-kubernetes-operator/e2e/test_operator_plugin        1510.343s
```

## Documentation

Will be updated.

## Follow-up

-

PR #2125, opened 2024-09-04
Disable version check for e2e plugin test

johscheuer:
# Description

Fix for https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2122#issuecomment-2328109728.

## Type of change

## Discussion

The version check makes a call to the GitHub API< if the GitHub API is not reachable that test will fail.

## Testing

Ran the test manually.

## Documentation

-

## Follow-up

-


johscheuer:
Already fixed in https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2124

PR #2147, opened 2024-10-17
Fix the case where the process group gets removed without the addresses being included

johscheuer:
# Description

Fixes a bug where the process group gets removed from the `FoundationDBCluster` status without getting the addresses included. The side effect of this bug is that some addresses in the cluster are not included again. The issue will only happen if multiple process groups should be replaced across different zones.

## Type of change

## Discussion

-

## Testing

Updated the test cases and ran the e2e test for this with an extended inclusion delay.

## Documentation

-

## Follow-up

-

### DIFF
--- a/controllers/remove_process_groups.go
+++ b/controllers/remove_process_groups.go
@@ -132,9 +132,8 @@ func (u removeProcessGroups) reconcile(ctx context.Context, r *FoundationDBClust
 	removedProcessGroups := r.removeProcessGroups(ctx, logger, cluster, zoneRemovals, zonedRemovals[removals.TerminatingZone])
 	err = includeProcessGroup(ctx, logger, r, cluster, removedProcessGroups, status, adminClient)
 	if err != nil {
-		delay := time.Duration(int(r.MinimumRecoveryTimeForInclusion-status.Cluster.RecoveryState.SecondsSinceLastRecovered)) * time.Second
-		_ = delay
-		return &requeue{curError: err, delayedRequeue: true, delay: 0}
+		// If the inclusion is blocked or another issues happened we will retry in 60 seconds.
+		return &requeue{curError: err, delayedRequeue: true, delay: 60 * time.Second}
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


controllers/remove_process_groups.go
```diff
@@ -293,59 +301,63 @@ func includeProcessGroup(ctx context.Context, logger logr.Logger, r *FoundationD
 		return err
 	}
 
-	// Reset the SecondsSinceLastRecovered sine the operator just included some processes, which will cause a recovery.
+	// Reset the SecondsSinceLastRecovered since the operator just included some processes, which will cause a recovery.
 	status.Cluster.RecoveryState.SecondsSinceLastRecovered = 0.0
+	// Update the process group list and remove all removed and included process groups.
+	cluster.Status.ProcessGroups = newProcessGroups
```
johscheuer, line 307: In a follow up PR I'll create some additional checks in our e2e tests to make sure all processes are included later, but that will required some fine tuning to make sure we are not hitting timeouts with the tests.

controllers/remove_process_groups_test.go
```diff
@@ -637,11 +641,12 @@ var _ = Describe("remove_process_groups", func() {
 				})
 
 				It("should include one process", func() {
```
nicmorales9, line 643: includes 2?

controllers/remove_process_groups_test.go
```diff
@@ -560,11 +560,12 @@ var _ = Describe("remove_process_groups", func() {
 			})
 
 			When("including no process", func() {
-				It("should not include any process", func() {
-					processesToInclude, err := getProcessesToInclude(logr.Logger{}, cluster, removedProcessGroups, status)
+				FIt("should not include any process", func() {
```
johscheuer, line 563:
```suggestion
				It("should not include any process", func() {
```

nicmorales9 (approved):

controllers/remove_process_groups.go
johscheuer: Created: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/2149

PR #2148, opened 2024-10-18
Simplify the confirmRemoval code

johscheuer:
# Description

Small code clean up.

## Type of change

- Other

## Discussion

-

## Testing

Unit tests.

## Documentation

-

## Follow-up

-

### DIFF
--- a/controllers/remove_process_groups.go
+++ b/controllers/remove_process_groups.go
@@ -197,21 +197,21 @@ func removeProcessGroup(ctx context.Context, r *FoundationDBClusterReconciler, c
 	return deletionError
 }
 
-func confirmRemoval(ctx context.Context, logger logr.Logger, r *FoundationDBClusterReconciler, cluster *fdbv1beta2.FoundationDBCluster, processGroup *fdbv1beta2.ProcessGroupStatus) (bool, bool, error) {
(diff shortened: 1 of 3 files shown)


controllers/remove_process_groups.go
```diff
@@ -408,18 +409,15 @@ func (r *FoundationDBClusterReconciler) removeProcessGroups(ctx context.Context,
 	// We have to check if the currently removed process groups are completely removed.
 	// In addition, we have to check if one of the terminating process groups has been cleaned up.
 	for _, processGroup := range processGroups {
-		removed, include, err := confirmRemoval(ctx, logger, r, cluster, processGroup)
-		if err != nil {
+		include, err := confirmRemoval(ctx, logger, r, cluster, processGroup)
+		if err != nil && !internal.IsResourceNotDeleted(err) {
```
nicmorales9, line 413: shouldn't this be an or? if there is a non-resource-deleted error it won't log it (though need to ensure it's happy with a nil error in errors.As())

nicmorales9: ohh I see, yup you are correct

nicmorales9 (approved):
much better!

Issue #2153, opened 2024-10-21
Add support for DNS entries in cluster file for multi-region recovery plugin.


johscheuer:
### What would you like to be added/changed?

Currently the subcommand to recover a multi-region cluster only works if the cluster doesn't use IP addresses.

PR #2154, opened 2024-10-21
Make sure that the remove cluster is skipped too and add more debugging output for the plugin

johscheuer:
# Description

Setting the remove cluster into the skip mode, should make the tests more reliable and reduce the risk of a race condition.

## Type of change

## Discussion

-

## Testing

Ran the e2e test for the plugin manually.

## Documentation

0

## Follow-up

https://github.com/FoundationDB/fdb-kubernetes-operator/issues/2153 (will e fixed in another PR).


brownleej (approved):
👍

Issue #2155, opened 2024-10-22
When using the unified image with a custom environment variable as zone


johscheuer:
### What happened?

The cluster was stuck in reconciliation with the error message that only 1 process out of 5 could be selected. The reason for that was that the `FDB_ZONE_ID` environment variable was not set as I used a custom environment variable for the zone.

### What did you expect to happen?

The cluster should be created.

### How can we reproduce it (as minimally and precisely as possible)?

Create a cluster with a custom environment variable for the fault domain and use the unified image.

### Anything else we need to know?

-

### FDB Kubernetes operator

<details>

```console
$ kubectl fdb version
# paste output here
```

</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Cloud provider

<details>

</details>

PR #2156, opened 2024-10-22
Fix the unified image setup during bootstrap when a custom environment variable is used for the zone

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/2155

## Type of change

## Discussion

-

## Testing

Created a new test setup. I decided to add the new test case to the upgrade test as this includes creating a new cluster.

## Documentation

-

## Follow-up

-


brownleej (approved):
👍

PR #2158, opened 2024-10-25
Try to batch the exclusions of transaction system processes together

johscheuer:
# Description

The changes of this PR try to batch the exclusions of transaction system processes together, e.g. stateless and transaction processes. This should reduce the number of exclusions (and recoveries). This change will be useful as the stateless pods are normally scheduled faster than the transaction/log pods as they have to wait for the volume to be provisioned.

## Type of change

It's not a correctness bug but the operator will be issuing more exclusions than required.

## Discussion

-

## Testing

Added new unit tests and CI will run the e2e tests.

## Documentation

Added docs in the code.

## Follow-up

-


brownleej (approved):
👍

PR #2162, opened 2024-11-11
Adding a new e2e test for producing data loss

johscheuer:
# Description

Adding a new test that will cause a fail-over with data loss. The test is currently not running with CI, we might add this test later to CI. Right now the intention is to use this test case to run some additional tests later.

## Type of change

- Other

## Discussion

-

## Testing

Ran the test manually.

## Documentation

Added in the test case.

## Follow-up

-


nicmorales9 (approved):
minor typos, looks good to me aside from the focus

johscheuer:
I merged the PR as I only changed comments.

PR #2172, opened 2024-11-19
Add design for better coordination for multi operator setup

johscheuer:
# Description

Add design for better coordination for multi operator setup. This design takes a few ideas from the lock client that we use for the minor and major version upgrades when the operator has to perform a restart all fdb server processes at "the same time".

## Type of change

- Other

## Discussion

Those changes should minimize the disruption of upgrades and settings roll-outs. 
There is still a chance that more exclusions are trigger than required e.g. if some operator instances are very slow.

## Testing

-

## Documentation

Added for the design.

## Follow-up

-

### DIFF
--- a/docs/design/better_coordination_multi_operator.md
+++ b/docs/design/better_coordination_multi_operator.md
@@ -48,17 +48,13 @@ The idea is to extend the existing `AdminClient` with some similar functionality
 
 The following methods will be added:
 
-- `AddPendingForRemoval`: Adds the process group ID to a set of process groups that are marked for removal.
-- `AddPendingForExclusion`: Adds the process group ID to a set of process groups that should be excluded.
-- `AddPendingForInclusion`: Adds the process group ID to a set of process groups that should be included.
-- `AddPendingForRestart`: Adds the process group ID to a set of process groups that should be restarted.
-- `RemoveFromPendingForRemoval`: Removes the process group ID from the set of process groups that should be removed.
-- `RemoveFromPendingForExclusion`: Removes the process group ID from the set of process groups that should be excluded.
-- `RemoveFromPendingForInclusion`: Removes the process group ID from the set of process groups that should be included.
-- `RemoveFromPendingForRestart`: Removes the process group ID from the set of process groups that should be restarted.
-- `AddReadyForExclusion`: Adds the process group ID to a set of process groups that are ready to be excluded.
-- `AddReadyForInclusion`: Adds the process group ID to a set of process groups that are ready to be included.
-- `AddReadyForRestart`: Adds the process group ID to a set of process groups that are ready to be restarted.
+- `UpdatePendingForRemoval`: Updates the set of process groups that are marked for removal, an update can be eiter the addition or removal of a process group.
+- `UpdatePendingForExclusion`: Updates the set of process groups that should be excluded, an update can be eiter the addition or removal of a process group.
+- `UpdatePendingForInclusion`: Updates the set of process groups that should be included, an update can be eiter the addition or removal of a process group.
+- `UpdatePendingForRestart`: Updates the set of process groups that should be restarted, an update can be eiter the addition or removal of a process group.
+- `UpdateReadyForExclusion`: Updates the set of process groups that are ready to be excluded, an update can be eiter the addition or removal of a process group
+- `UpdateReadyForInclusion`: Updates the set of process groups that are ready to be included, an update can be eiter the addition or removal of a process group.
+- `UpdateReadyForRestart`: Updates the set of process groups that are ready to be restarted, an update can be eiter the addition or removal of a process group
 - `GetPendingForRemoval`: Gets the process group IDs for all process groups that are marked for removal.
 - `GetPendingForExclusion`: Gets the process group IDs for all process groups that should be excluded.
 - `GetPendingForInclusion`: Gets the process group IDs for all the process groups that should be included.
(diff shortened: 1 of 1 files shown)


docs/design/better_coordination_multi_operator.md
```diff
@@ -0,0 +1,129 @@
...
+## Metadata
+
+* Authors: @johscheuer
+* Created: 2024-11-19
+* Updated: 2024-11-19
+
+## Background
+
+The current way to run multi-region FoundationDB deployments with the operator is to run multiple operator instances in different namespaces and/or in different Kubernetes clusters.
+This brings a few synchronization challenges as those different operator instances are not communication with each other and they are not sharing state.
```
brownleej, line 12:
```suggestion
This brings a few synchronization challenges as those different operator instances are not communicating with each other and they are not sharing state.
```

docs/design/better_coordination_multi_operator.md
```diff
@@ -0,0 +1,129 @@
...
+We already added the `LockClient` to synchronize the upgrade step for minor and major upgrades, where all processes must be restarted at once.
+The `LockClient` is also used to get a lease that allows the operator instance to perform certain actions, like exclusions, that should only be execute by one instance at a time.
```
brownleej, line 14:
```suggestion
The `LockClient` is also used to get a lease that allows the operator instance to perform certain actions, like exclusions, that should only be executed by one instance at a time.
```

docs/design/better_coordination_multi_operator.md
```diff
@@ -0,0 +1,129 @@
...
+- `AddPendingForRemoval`: Adds the process group ID to a set of process groups that are marked for removal.
+- `AddPendingForExclusion`: Adds the process group ID to a set of process groups that should be excluded.
+- `AddPendingForInclusion`: Adds the process group ID to a set of process groups that should be included.
+- `AddPendingForRestart`: Adds the process group ID to a set of process groups that should be restarted.
+- `RemoveFromPendingForRemoval`: Removes the process group ID from the set of process groups that should be removed.
+- `RemoveFromPendingForExclusion`: Removes the process group ID from the set of process groups that should be excluded.
+- `RemoveFromPendingForInclusion`: Removes the process group ID from the set of process groups that should be included.
+- `RemoveFromPendingForRestart`: Removes the process group ID from the set of process groups that should be restarted.
+- `AddReadyForExclusion`: Adds the process group ID to a set of process groups that are ready to be excluded.
+- `AddReadyForInlusion`: Adds the process group ID to a set of process groups that are ready to be included.
```
brownleej, line 47:
```suggestion
- `AddReadyForInclusion`: Adds the process group ID to a set of process groups that are ready to be included.
```

docs/design/better_coordination_multi_operator.md
```diff
@@ -0,0 +1,129 @@
...
+- `AddReadyForRestart`: Adds the process group ID to a set of process groups that are ready to be restarted.
+- `GetPendingForRemoval`: Gets the process group IDs for all process groups that are marked for removal.
+- `GetPendingForExclusion`: Gets the process group IDs for all process groups that should be excluded.
+- `GetPendingForInclusion`: Gets the process group IDs for all the process groups that should be included.
+- `GetPendingForRestart`: Gets the process group IDs for all the process groups that should be restarted.
+- `GetReadyForExclusion`: Gets the process group IDs for all the process groups that are ready to be excluded.
+- `GetReadyForInlusion`: Gets the process group IDs for all the process groups that are ready to be included.
```
brownleej, line 54:
```suggestion
- `GetReadyForInclusion`: Gets the process group IDs for all the process groups that are ready to be included.
```

docs/design/better_coordination_multi_operator.md
```diff
@@ -0,0 +1,129 @@
...
+## Proposed Design
+
+The idea is to extend the existing `AdminClient` with some similar functionality from the `LockClient` to coordinate the following actions:
+
+- coordinator selection
+- exclusions
+- inclusions
+- bounces
+
+The following methods will be added:
```
brownleej, line 36: Can we add more detail on how these lists will be modeled in the keyspace, and what kind of transactional properties (if any) we intend to use to control concurrent updates?

docs/design/better_coordination_multi_operator.md
```diff
@@ -0,0 +1,129 @@
...
+- `removeProcessGroups`: Will add process groups that can be included to `readyForInclusion`.
+
+We will add a new setting under the `FoundationDBClusterAutomationOptions` with the name `synronizationMode`.
+The default will be `local`, which will represent the current state and the new mechanism can be enabled by setting the value to `global`.
+Those changes will add some additional load to the cluster, but the additional load should be fairly limited compared to the customer load on those clusters and the request pattern will be mostly reading the data.
+
+Pseudo-Code changes for the `excludeProcesses` sub-reconciler:
+
+```go
+func (e excludeProcesses) reconcile(ctx context.Context, r *FoundationDBClusterReconciler, cluster *fdbv1beta2.FoundationDBCluster, status *fdbv1beta2.FoundationDBStatus, logger logr.Logger) *requeue {
```
brownleej, line 76: My interpretation of this flow is that the reconciler is checking for pending/ready exclusions that another operator instance has reported, but doesn't necessarily wait for any kind of delay or convergence before performing the known exclusions itself. Is that accurate? Does that mean that we still have the possibility for each instance of the operator to perform its own exclusions without seeing the exclusions from the other instances?

brownleej: It might be good to get a more detailed flow of how the exclusion states for different operator instances will interleave, and what kinds of things we expect to see as blockers or synchronization points between `PendingForExclusion` and `ReadyForExclusion`.

brownleej (changes requested):

docs/design/better_coordination_multi_operator.md
```diff
@@ -0,0 +1,129 @@
...
+
+* Authors: @johscheuer
+* Created: 2024-11-19
+* Updated: 2024-11-19
+
+## Background
+
+The current way to run multi-region FoundationDB deployments with the operator is to run multiple operator instances in different namespaces and/or in different Kubernetes clusters.
+This brings a few synchronization challenges as those different operator instances are not communication with each other and they are not sharing state.
+We already added the `LockClient` to synchronize the upgrade step for minor and major upgrades, where all processes must be restarted at once.
```
brownleej, line 13: There's also a general consideration we should bring into this about the difference between the upgrades and the other operations. The upgrades are built around the requirement that _every_ process in the database will be upgraded, which means that we have an external source of truth for what processes need to be ready in order for the bounce action in the upgrade to proceed. For other operations, we will have a harder time determining if the omission of processes from operator instance X indicates that there are no processes from instance X that need to be acted on, or if operator instance X has not yet processed a CRD update that would tell what processes need to be acted on. One way we can think about that is that the synchronization here is more optimistic than guaranteed, in which case we may need to be clearer about the typical expectation for the delays between different steps, and the overlap in response times between different operator instances, to know how likely it is that the signals will arrive in time to be synchronized.

brownleej (approved):
👍

PR #2183, opened 2024-12-13
Add a test case to run the exclusion in a multi-region cluster for a bigger cluster with injected latencies

johscheuer:
# Description

I used the test case to test: https://github.com/apple/foundationdb/pull/11835. I don't think we should be running this test always (as it takes a bit and also requires more resources), but it would be nice to keep it for further manual tests.

## Type of change

- Other

## Discussion

As an alternative I can create a new dedicated test suite.

## Testing

Ran the test manually.

## Documentation

-

## Follow-up

-


nicmorales9 (approved):
looks good to me

PR #2191, opened 2025-01-14
docs: document procedure to migrate a live cluster to three-data-hall redundancy

gm42:
# Description

When using single/double/triple redundancy it is not possible to migrate a live cluster to `three_data_hall` redundancy without incurring into some temporary timeouts or unavailability.

This change adds the steps necessary to perform migration without incurring into these availability problems.

## Type of change

- Documentation

## Follow-up

Note: I also changed documentation of `--locality_data_hall` to say that it is actually being used now


docs/manual/fault_domains.md
```diff
@@ -188,7 +188,25 @@ spec:
 
 Once all three `FoundationDBCluster` resources are marked as reconciled the FoundationDB cluster is up and running.
 You can run this configuration in the same namespace, different namespaces or even across multiple different Kubernetes clusters.
-Operations across the different `FoundationDBCluster` resources are [coordinated](#coordinating-global-operations). 
+Operations across the different `FoundationDBCluster` resources are [coordinated](#coordinating-global-operations).
+
+## Migrating an existing cluster to Three-Data-Hall Replication
```
johscheuer, line 193:
```suggestion
### Migrating an existing cluster to Three-Data-Hall Replication
```

Can we add a note that those steps are for the `split` image setup (still the default)?. For the `unified` image setup the migration path should be simpler. I'll take a note to update the section above with the `unified` image.

docs/manual/fault_domains.md
```diff
@@ -188,7 +188,25 @@ spec:
...
+
+It is possible to gracefully migrate a cluster in single, double or triple replication to Three-Data-Hall replication by following these steps:
+
+1. create 3 exact clones of the original k8s FoundationDB cluster object (thus still with same replication) and change these fields: `metadata.name`,`spec.processGroupIDPrefix`, `spec.dataHall`, `spec.dataCenter`
+2. make sure that `skip: true` is set on their YAML definition so that operator will not attempt to configure them
+3. make sure that `seedConnectionString:` is set to to the current connection string of the original cluster
+4. run `kubectl create` for each of them
+5. set the configured state and connection string in their status subresource by using: `kubectl patch fdb my-new-cluster-a --type=merge --subresource status --patch "status: {configured: true, connectionString: \"...\" }"`; use the original cluster connection string here
+6. set `skip: false` on each of the 3 new clusters, and then wait for reconciliation to finish
+7. start a lengthy exclude procedure that will exclude all the processes of the original cluster; suggested order: `log`, `storage`, `coordinator`, `stateless`
```
johscheuer, line 203: Before doing that shouldn't we set `skip: true` on the original cluster?

docs/manual/fault_domains.md
```diff
@@ -188,7 +188,25 @@ spec:
...
+8. delete the original cluster once all exclusions are complete
+9. set `redundancyMode` to `three_data_hall` for the 3 new FoundationDB clusters, one after another
+10. patch seed connection string of 2 of the 3 new clusters to point to the third one e.g. if you have created clusters A,B,C, set the seed connection string of B and C to point to A and make sure that A has no seed connection string; this step is not crucial but practically helpful sometimes
+11. scale down clusters to use 1/3 of the original resources (each of them needs only 3 coordinators and 1/3 of the resources used for other classes)
+
+This procedure prevents temporary unavailability which may happen with sustained trffic when data distributor and master are being reallocated while redundancy mode is changed.
```
johscheuer, line 209:
```suggestion
This procedure prevents temporary unavailability which may happen with sustained traffic when data distributor and master are being reallocated while redundancy mode is changed.
```

That's not completely true, during the migration there will be multiple recoveries. Those should be quick and the clients shouldn't notice it, but the risk of a long recovery still exsits.

johscheuer (commented):
Thanks for taking the time and document the migration path.

docs/manual/fault_domains.md
gm42: From the migrations I did with this procedure I could see that data movement had a part in this: I was not using maintenance mode and perhaps that's what compounded the appearance of 1031 timeouts?

docs/manual/fault_domains.md
gm42: I experimentally found that the reconciliation at previous step 6 was necessary to mitigate 1031 timeouts; I think this is because waiting for reconciliation also waits for data movement to finish (operator cannot reconcile if data state is not healthy for a certain amount of time).

docs/manual/fault_domains.md
gm42: I reworded it to explain that the temporary unavailability is in the form of 1031 timeouts rather than cluster being unavailable.

docs/manual/fault_domains.md
gm42: Done.

johscheuer (approved):
LGTM 👍

PR #2200, opened 2025-01-27
Fix an error handling bug in tryConnectionOptions

iwalther:
# Description

If none of the connection strings work, it would previously return the last known connection string and no error. In combination with another problem, this has led to the operator rotating all coordinators without keeping up with the latest connection string and then publishing a connection string to clients that no longer works.

## Type of change

## Testing

Added a regression test (unit test).

## Documentation

n/a

## Follow-up

n/a


controllers/update_status_test.go
```diff
@@ -311,6 +312,15 @@ var _ = Describe("update_status", func() {
 			Expect(clusterReconciler.List(context.TODO(), allPvcs, internal.GetPodListOptions(cluster, "", "")...)).NotTo(HaveOccurred())
 		})
 
+		When("cannot get cluster spec", func() {
```
johscheuer, line 315:
```suggestion
		When("cannot get the connection string", func() {
```

johscheuer (commented):
Thanks for the PR.

iwalther:
> * Result: ❌ **FAILED**

Is this CI failure real? I haven't managed to successfully run the e2e tests locally on main.

controllers/update_status.go
```diff
@@ -330,7 +330,7 @@ func tryConnectionOptions(logger logr.Logger, cluster *fdbv1beta2.FoundationDBCl
 		logger.Error(err, "Error getting connection string from cluster", "connectionString", connectionString)
 	}
 
-	return originalConnectionString, nil
+	return originalConnectionString, fmt.Errorf("could not get connection string from cluster")
```
gm42, line 333: I would not return here the `originalConnectionString` for defensive programming reasons: caller should not be allowed to ignore the error and use the connection string anyways.

iwalther: Done, looks like nothing is using it in the error case.

gm42 (approved):

johscheuer:
> > ### Result of fdb-kubernetes-operator-pr on Linux CentOS 7
> > 
> > * Result: ❌ **FAILED**
> > * Error: `Error while executing command: if $fail_test; then exit 1; fi. Reason: exit status 1`
> 
> Is this CI failure real? I haven't managed to successfully run the e2e tests locally on main.

What have you tried to run those e2e test locally? The test failure is real, most of the time the test are passing: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2201#issuecomment-2620365204 (sometimes the underlying Kubernetes platform behaves a bit different in scaling).

johscheuer (commented):
> * Commit ID: [33251c9](https://github.com/FoundationDB/fdb-kubernetes-operator/commit/33251c9df7997b9e54370b3b72a7a4b0e9dc6e47)
> * Duration 4:10:19
> * Result: ❌ **FAILED**
> * [Build Log](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator/fdb-kubernetes-operator-pr/1fb2f4f6-7829-43b8-b04f-c9eb764ed7d0.gz) terminal output (available for 30 days)
> * [Build Workspace](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator-pr/1fb2f4f6-7829-43b8-b04f-c9eb764ed7d0/fdb-kubernetes-operator-pr) zip file of the working directory (available for 30 days)

```
[Container] 2025/01/28 21:10:08.337349 Running command if $(grep -q -- "--- FAIL:" ${CODEBUILD_SRC_DIR}/logs/*.log); then echo "TESTS FAILED SEE THESE LOGS:"; echo ; grep -l -- "--- FAIL:" ${CODEBUILD_SRC_DIR}/logs/*.log; export fail_test=true; fi
TESTS FAILED SEE THESE LOGS:

/codebuild/output/src4162871180/src/github.com/FoundationDB/fdb-kubernetes-operator/logs/test_operator.log
/codebuild/output/src4162871180/src/github.com/FoundationDB/fdb-kubernetes-operator/logs/test_operator_backups.log
/codebuild/output/src4162871180/src/github.com/FoundationDB/fdb-kubernetes-operator/logs/test_operator_ha.log
/codebuild/output/src4162871180/src/github.com/FoundationDB/fdb-kubernetes-operator/logs/test_operator_ha_upgrades.log
/codebuild/output/src4162871180/src/github.com/FoundationDB/fdb-kubernetes-operator/logs/test_operator_migrations.log
/codebuild/output/src4162871180/src/github.com/FoundationDB/fdb-kubernetes-operator/logs/test_operator_plugin.log
/codebuild/output/src4162871180/src/github.com/FoundationDB/fdb-kubernetes-operator/logs/test_operator_three_data_hall.log
/codebuild/output/src4162871180/src/github.com/FoundationDB/fdb-kubernetes-operator/logs/test_operator_upgrades_with_chaos.log

[Container] 2025/01/28 21:10:08.393783 Running command for f in $(grep -L -- "SUCCESS\! --" ${CODEBUILD_SRC_DIR}/logs/*.log); do if ! grep -q -- "--- FAIL:" $f; then echo "This test has timed out or had a panic: $f"; export fail_test=true; fi ;  done
This test has timed out or had a panic: /codebuild/output/src4162871180/src/github.com/FoundationDB/fdb-kubernetes-operator/logs/test_operator_upgrades.log
This test has timed out or had a panic: /codebuild/output/src4162871180/src/github.com/FoundationDB/fdb-kubernetes-operator/logs/test_operator_upgrades_variations.log
```

Seems like the new change is blocking the operator from the initial cluster reconciliation.

iwalther:
Thank you! I tried running them with "kind" as described in the readme, let me see if I can get that e2e test suite to work (going back to draft).

I was just looking into this code a bit more and I _think_ that tryConnectionOptions() doesn't work as intended. It reads the connection string through the golang FDB client library. OpenDatabase() internally keeps a singleton map keyed by the cluster file path. In the second iteration of the loop, OpenDatabase() will return the same database object. The C++ FDB client appears not to read back the file after start, except to report out-of-sync errors and to log errors in case of a write race. It won't use the updated file content I think.

Could it be that the initial reconciliation relies on this bug, returning success but having updated the cluster file to the seed connection string, so the next "fdbcli" call would pick it up and somehow get back on track? Maybe I can refactor the code to explicitly connect via seed string the first time we connect to a cluster, and then get rid of this try-which-one-works logic.

Related draft PR: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2204 and I suspect those two bugs amplify each other.

johscheuer (commented):
I looked a bit more into it and the issue is during the initial cluster file generation. After this the cluster is not yet configured and we hist some issues, so updating the `getStatusFromClusterOrDummyStatus` will solve this:

```diff
index 62de8330..708e2317 100644
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -573,7 +573,7 @@ func (r *FoundationDBClusterReconciler) getStatusFromClusterOrDummyStatus(logger
        }

        connectionString, err := tryConnectionOptions(logger, cluster, r)
-       if err != nil {
+       if err != nil && cluster.Status.Configured {
                return nil, err
        }
```

In the previous setup the initial generated cluster file would be returned.

I'm not sure about the kind test setup, but we have never really used it (we had some issues with arm on MacOS). If you have a dev cluster that could be used, that would probably the simplest approach. 

> I was just looking into this code a bit more and I think that tryConnectionOptions() doesn't work as intended. It reads the connection string through the golang FDB client library. OpenDatabase() internally keeps a singleton map keyed by the cluster file path. In the second iteration of the loop, OpenDatabase() will return the same database object. The C++ FDB client appears not to read back the file after start, except to report out-of-sync errors and to log errors in case of a write race. It won't use the updated file content I think.

We are reading the current coordinators (cluster file content) from the special key: `"\xff/coordinators"`.  But I have to look closer into this to give a better answer, we definitely did e2e tests around this and it behaved as expected during that time, but I will revisit this code next week.

> Could it be that the initial reconciliation relies on this bug, returning success but having updated the cluster file to the seed connection string, so the next "fdbcli" call would pick it up and somehow get back on track? Maybe I can refactor the code to explicitly connect via seed string the first time we connect to a cluster, and then get rid of this try-which-one-works logic.

It doesn't, the bug was in `getStatusFromClusterOrDummyStatus`, expecting that a connection string is returned by the `tryConnectionOptions`.

iwalther:
Thank you very much! I'll try it next week.

FWIW this is the singleton I'm referring to:
https://github.com/apple/foundationdb/blob/fff5439ef28c2325fec752fe45d55a1425ad2374/bindings/go/src/fdb/fdb.go#L335

Operator is using a slightly different version where this is a map protected by mutex, but conceptually the same.

johscheuer:
Will be fixed in https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2216.

Thanks for the PR.

PR #2201, opened 2025-01-28
Add a new label for running a quick upgrade test case

johscheuer:
# Description

Adding a new Ginkgo label for running a single test (right now) to perform an upgrade/downgrade test. In the future we could be adding more test cases.

## Type of change

- Other

## Discussion

-

## Testing

Manually ran the tests with an e2e test selector.

## Documentation

-

## Follow-up

-

### DIFF
--- a/e2e/test_operator_upgrades_variations/operator_upgrades_variations_test.go
+++ b/e2e/test_operator_upgrades_variations/operator_upgrades_variations_test.go
@@ -183,7 +183,7 @@ var _ = Describe("Operator Upgrades", Label("e2e", "pr"), func() {
 	// this setup allows to dynamically generate the table entries that will be executed e.g. to test different upgrades
 	// for different versions without hard coding or having multiple flags.
 	DescribeTable(
-		"upgrading a cluster without chaos", Label("quick"),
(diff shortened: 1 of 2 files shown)


e2e/test_operator_upgrades_variations/operator_upgrades_variations_test.go
```diff
@@ -183,7 +183,7 @@ var _ = Describe("Operator Upgrades", Label("e2e", "pr"), func() {
 	// this setup allows to dynamically generate the table entries that will be executed e.g. to test different upgrades
 	// for different versions without hard coding or having multiple flags.
 	DescribeTable(
-		"upgrading a cluster without chaos",
+		"upgrading a cluster without chaos", Label("quick"),
```
saintstack, line 186: nit: Maybe a more descriptive label than 'quick'? It is not self-describing.

saintstack (approved):
A nit comment only but otherwise LGTM.

e2e/test_operator_upgrades_variations/operator_upgrades_variations_test.go
johscheuer: That makes sense 👍

PR #2203, opened 2025-01-31
Ensure reconcilation is requeued for restore controller if the restore is not completed

johscheuer:
# Description

Ensure the the restore controller requeues the restore resource if the restore is not completed.

## Type of change

## Discussion

-

## Testing

Ran the backup e2e tests manually.

## Documentation

-

## Follow-up

-

### DIFF
--- a/controllers/restore_controller.go
+++ b/controllers/restore_controller.go
@@ -22,21 +22,20 @@ package controllers
 
 import (
 	"context"
-
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
-	"github.com/FoundationDB/fdb-kubernetes-operator/pkg/fdbadminclient"
-
-	"sigs.k8s.io/controller-runtime/pkg/controller"
-	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"time"
 
 	fdbv1beta2 "github.com/FoundationDB/fdb-kubernetes-operator/api/v1beta2"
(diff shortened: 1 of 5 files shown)


controllers/restore_controller.go
```diff
@@ -76,12 +75,24 @@ func (r *FoundationDBRestoreReconciler) Reconcile(ctx context.Context, request c
...
+		return processRequeue(req, subReconciler, restore, r.Recorder, restoreLog)
+	}
+
+	if restore.Status.State != fdbv1beta2.CompletedFoundationDBRestoreState {
+		restoreLog.Info("Restore has not yet completed",
+			"Status", restore.Status.State,
+			"Running", restore.Status.Running)
+
+		// Check the status of the restore every 5 minutes, otherwise check the status every minute.
+		if restore.Status.State != fdbv1beta2.RunningFoundationDBRestoreState {
```
nicmorales9, line 91: just-in-casies: is there anything to do on the other restore states? e.g. queued etc? Or is that just a "will be fleshed out later" kind of thing?

nicmorales9 (approved):

controllers/restore_controller.go
nicmorales9: (I know this whole code area is WIP)

johscheuer: Not now. We assume right now either the restore was started, pending to start or finished. This code part only ensures that the reconciler check the status regularly. We eventually will make use of the other states in the future.

Issue #2171, opened 2024-11-18
three_data_hall inconsistent connection string problem


gm42:
### What happened?

I have found a quirky issue happening when using `three_data_hall`; the issue consists in the k8s FoundationDB cluster status containing an outdated connection string which does not correspond to what the cluster is currently using:
```
A:  connectionString:     mydb:dZiqyRWzkuM8tZ6hHx4Z3xWsFdu4BV0F@10.10.2.18:4501,10.20.22.84:4501,10.20.60.215:4501,10.20.67.98:4501,10.20.81.100:4501,10.10.265.72:4501,10.20.217.63:4501,10.20.240.122:4501,10.20.243.153:4501
A:  seedConnectionString: <none>

B:  seedConnectionString: mydb:mCYmFHPkEfyKO5gl8ae5ocZPq2a0IJW4@10.10.20.248:4501,10.20.22.192:4501,10.20.67.101:4501,10.10.210.33:4501,10.10.241.115:4501,10.10.268.53:4501,10.10.277.35:4501,10.10.285.107:4501,10.20.220.3:4501
B:  connectionString: mydb:9pION4FdMvW53gB4ikdjo61cs7HQKtK3@10.20.22.84:4501,10.20.60.215:4501,10.20.67.98:4501,10.20.81.100:4501,10.10.220.14:4501,10.10.265.72:4501,10.20.200.1:4501,10.20.217.63:4501,10.20.243.153:4501

C:  seedConnectionString: mydb:mCYmFHPkEfyKO5gl8ae5ocZPq2a0IJW4@10.10.20.248:4501,10.20.22.192:4501,10.20.67.101:4501,10.10.210.33:4501,10.10.241.115:4501,10.10.268.53:4501,10.10.277.35:4501,10.10.285.107:4501,10.20.220.3:4501
C:  connectionString: mydb:dZiqyRWzkuM8tZ6hHx4Z3xWsFdu4BV0F@10.10.2.18:4501,10.20.22.84:4501,10.20.60.215:4501,10.20.67.98:4501,10.20.81.100:4501,10.10.265.72:4501,10.20.217.63:4501,10.20.240.122:4501,10.20.243.153:4501
```

When checking directly via `get \xff\xff/connection_string`, the correct connection string for the cluster appears to be the one with generation ID `9pION4FdMvW53gB4ikdjo61cs7HQKtK3`.

### What did you expect to happen?

[...]

PR #2204, opened 2025-01-31
Privateclusterfile

iwalther:
# Description

FoundationDB Operator creates cluster files in the temp directory so that fdbcli and the FoundationDB client library can connect to the cluster. The cluster file name is deterministically derived from the cluster UID.

Three different actors write to and read from this file without synchronization:

1. fdbcli whenever it learns from the cluster that the connection string has changed,
2. the FDB client library whenever it learns from the cluster that the connection string has changed,
3. FoundationDB Operator when setting up a new admin client.  tryConnectionOptions() routinely writes the outdated seed connection string into the file in case of transient errors, without actively restoring the correct one.

While the cluster controller is single-threaded, the library may be writing in the background. Since introduction of backup and restore controllers, there can be multiple active instances of admin client at the same time. The cluster controller is holding an instance of admin client while the subreconcilers are creating other instances, though I think it is inactive during reconciliation and this last point is not an actual problem.

If a race occurs, e.g. the reader finds the cluster file in a transient state where the file is truncated or temporarily overwritten with the now-invalid seed connection string, or a writer overwrites a legitimate new file), this could go wrong. For example, when changing coordinators, we read back the file after fdbcli exits, assuming it contains the new connection string. 

To prevent this class of problems, the proposed change will create a unique cluster file for each instance of admin client and delete it when done. It instantiates a singleton database instance with its private cluster file for each cluster UID and keeps it around. To be extra safe, read the new connection string from the cluster instead of from the cluster file after changing coordinators to get an authoritative result.

## Type of change

## Discussion

We experienced an incident where the operator and subsequently all clients lost contact with the cluster, and I believe this was due to such a race. We have seen connection strings running out of sync before in a less catastrophic way. Our setup is a bit unusual and maybe that exposes us more than others: Our cluster runs in three_data_hall mode, which means a "cluster"'s coordinators are occasionally changing unexpectedly because we change coordinators on another data hall - the three "clusters" are a single FDB cluster with a shared coordinator set after all. And we are running coordinators on dedicated servers, using a custom class, so the servers actually shut down after losing the coordinator role.

The singleton database may be a bit confusing, because the FDB client library internally keeps a singleton map like this, too.  They use the cluster file name as key. We need this because with the unique cluster file names, we'd now create more and more databases in the library's cache that never get cleaned up, leading to a memory leak.  Alternative approaches could be to keep a single cluster file for the library, and only use temporary files for fdbcli invocations. We could ask the cluster (via the libarary) for the current connection string each time to write a guaranteed-correct cluster file for fdbcli. These are more involved code changes, please advise.

## Testing

Unit tests pass. Still working on my local e2e test setup.

Does this change look reasonable? Then I can try it in our staging systems for a while.

## Documentation

n/a

## Follow-up

n/a

### DIFF
--- a/fdbclient/common.go
+++ b/fdbclient/common.go
@@ -25,6 +25,7 @@ import (
 	"errors"
 	"fmt"
 	"io/fs"
+	"math/rand"
 	"os"
 	"path"
 	"time"
@@ -110,23 +111,22 @@ func createClusterFile(cluster *fdbv1beta2.FoundationDBCluster) (string, error) 
 
 // ensureClusterFileIsPresent will ensure that the cluster file with the specified connection string is present.
 func ensureClusterFileIsPresent(dir string, uid string, connectionString string) (string, error) {
-	clusterFileName := path.Join(dir, uid)
+	for {
+		clusterFileName := path.Join(dir, fmt.Sprintf("%s-%d", uid, rand.Uint64()))
 
-	// Try to read the file to check if the file already exists and if so, if the content matches
-	content, err := os.ReadFile(clusterFileName)
-
-	// If the file doesn't exist we have to create it
-	if errors.Is(err, fs.ErrNotExist) {
-		return clusterFileName, os.WriteFile(clusterFileName, []byte(connectionString), 0777)
+		f, err := os.OpenFile(clusterFileName, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0777)
+		if err != nil {
+			if errors.Is(err, fs.ErrExist) {
+				continue
+			}
+			return "", err
+		}
+		_, err = f.Write([]byte(connectionString))
+		if err1 := f.Close(); err1 != nil && err == nil {
+			err = err1
+		}
+		return clusterFileName, err
 	}
-
-	// The content of the cluster file is already correct.
-	if string(content) == connectionString {
-		return clusterFileName, nil
-	}
-
-	// The content doesn't match, so we have to write the new content to the cluster file.
-	return clusterFileName, os.WriteFile(clusterFileName, []byte(connectionString), 0777)
 }
 
 // getConnectionStringFromDB gets the database's connection string directly from the system key
--- a/fdbclient/common.go
+++ b/fdbclient/common.go
@@ -25,9 +25,11 @@ import (
 	"errors"
 	"fmt"
 	"io/fs"
+	"k8s.io/apimachinery/pkg/types"
 	"math/rand"
 	"os"
 	"path"
+	"sync"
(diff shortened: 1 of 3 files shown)


iwalther:
Thought about it a bit more, here's an idea how to let the connection string flow cleanly.

General idea: the coordinators decide on the cluster file content, the authoritative data is in the coordinated state. The operator can suggest a change by running the "coordinators" command. So the operator should just distribute it. There is already proven code in the FDB client that keeps a cluster file in sync with the coordinated state. Currently, the operator overwrites the cluster file frequently.

Proposal: When we see a cluster UID for the first time, we write its seed connection string to a file, call OpenDatabase() on it immediately and never touch the file again. The FDB client library will keep it up-to-date. To run fdbcli, we make a temporary copy of the cluster file (fdb updates it atomically in the file system) and delete the copy when done. We watch the authoritative connection string and publish it as seed connection string to the clients like today.

I'll think about it a bit more and maybe update the PR next week (or make a second one for discussion), I think this is cleaner.

johscheuer (commented):
First of all thanks for taking the time to work on this! We have plans to cut a new operator release and drop the support for older FDB release, that will help us in a way that we are able to make use of the mgmt API and then a lot of the fdbcli interactions will go away.

For some historical background: We initially started with having random cluster files but there were some issues in the previous versions of the go bindings where the tracked clusters where not delete when the DB was closed which then was causing a memory leak. So we have to ensure we are closing the DB properly.

fdbclient/common.go
```diff
@@ -83,24 +90,29 @@ func parseMachineReadableStatus(logger logr.Logger, contents []byte, checkForPro
...
+		clusterFile, err := createClusterFile(cluster)
+		if err != nil {
+			return fdb.Database{}, err
+		}
 
-	database, err := fdb.OpenDatabase(clusterFile)
-	if err != nil {
-		return fdb.Database{}, err
-	}
+		database, err := fdb.OpenDatabase(clusterFile)
```
gm42, line 103: You could use `OpenWithConnectionString` to bypass the other cache on Go SDK side; was briefly discussed with @johscheuer [here](https://github.com/FoundationDB/fdb-kubernetes-operator/issues/2171#issuecomment-2552009785)

johscheuer:
Will be fixed in https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2216.

Thanks for the PR.

fdbclient/common.go
johscheuer: `OpenWithConnectionString` is not available in the 7.1 bindings, so the operator cannot make use of this method.

Issue #2192, opened 2025-01-15
Update three data hall documentation for unified image


johscheuer:
### What would you like to be added/changed?

Update the docs in `docs/manual/fault_domains.md` and mention how to use the `unified` image for three data hall as the setup will be simpler.

PR #2206, opened 2025-01-31
Update docs for three data hall and move something to implemented

johscheuer:
# Description

[*Please include a summary of the change and which issue is addressed. If this change resolves an issue, please include the issue number in the description.*](https://github.com/FoundationDB/fdb-kubernetes-operator/issues/2192)

## Type of change

- Documentation

## Discussion

-

## Testing

-

## Documentation

Updated.

## Follow-up

-


docs/manual/fault_domains.md
```diff
@@ -131,65 +131,76 @@ This strategy uses the pod name as the fault domain, which allows each process t
...
-            nodeSelector:
-              "topology.kubernetes.io/zone": "az1"
+  # The unified image supports to make use of node labels, so setting up a three data hall cluster
+  # is easier with the unified image.
+  imageType: unified
+  version: 7.1.63
+  faultDomain:
+    key: kubernetes.io/hostname
+  processCounts:
+    stateless: -1
```
nicmorales9, line 157: I assume this makes it auto scale them but I don't recall what, so it might be good to add a comment

nicmorales9 (approved):
LGTM

docs/manual/fault_domains.md
johscheuer: `-1` means no stateless processes will be created, as this is a minimal example I would keep it like this to reduce the required resources.

simenl:
I like the simplicity of this approach!

I do have some concerns about running with using `topologySpreadConstraints` in production, as the foundationdb operator does not control the placement of the pods.

1. A pod recreation may fail, as the pod _could_ be attached to a disk (PVC) in a zone which has a positive skew. This could happen for several reasons. E.g. a zone being unavailable, or lacking resources.

2. Replacement of pods can cause an uneven distribution of pods across zones. The new pods are added while old pods are removed, and they work against the same topologySpreadConstraint.

johscheuer:
> I like the simplicity of this approach!
> 
> I do have some concerns about running with using `topologySpreadConstraints` in production, as the foundationdb operator does not control the placement of the pods.
> 
> 1. A pod recreation may fail, as the pod _could_ be attached to a disk (PVC) in a zone which has a positive skew. This could happen for several reasons. E.g. a zone being unavailable, or lacking resources.
> 2. Replacement of pods can cause an uneven distribution of pods across zones. The new pods are added while old pods are removed, and they work against the same topologySpreadConstraint.

Both points are two valid concerns. We have a plan to improve the "placement" of pods inside the operator with some we called "logical fault domains": https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/docs/design/bin_pack_fault_domains.md. We target to work on this sometime this year.

Issue #1862, opened 2023-10-25
The addProcessGroups reconciler should take the exclusion list into consideration


johscheuer:
### What would you like to be added/changed?

In the current setup with locality based exclusions it could happen that a human operator excluded a locality and the operator could potentially create a process group with that locality. This can be prevented by checking the exclusion list in the addProcessGroups.

johscheuer:
I locality based exclusions are disabled we can ignore this check.

PR #2207, opened 2025-02-02
Check if the newly added process group ID is already excluded

johscheuer:
# Description

Fixes:  https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1862

## Type of change

## Discussion

This change improves the add process group subreconciler to not pick an already excluded process group ID. In general those cases should be rare but it's better to not use the process group ID if the locality was already excluded, e.g. manually, otherwise the new process group will be created and then replaced.

## Testing

Added new unit tests.

## Documentation

N/A

## Follow-up

-

### DIFF
--- a/api/v1beta2/foundationdbcluster_types.go
+++ b/api/v1beta2/foundationdbcluster_types.go
@@ -19,13 +19,14 @@ package v1beta2
 import (
 	"errors"
 	"fmt"
-	"k8s.io/apimachinery/pkg/util/intstr"
 	"math"
 	"math/rand"
 	"regexp"
 	"strconv"
 	"strings"
 	"time"
+
+	"k8s.io/apimachinery/pkg/util/intstr"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


api/v1beta2/foundationdbcluster_types.go
```diff
@@ -3004,6 +3014,23 @@ func (cluster *FoundationDBCluster) GetNextRandomProcessGroupID(processClass Pro
...
+
+	// If the randomly picked process group is part of the locality based exclusions, we shoudln't pick it.
+	// See: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1862
+	if _, ok := exclusions[processGroupID]; ok {
+		return false
+	}
+
+	return true
+}
+
```
nicmorales9, line 3033: https://www.youtube.com/watch?v=h3uBr0CCm58

controllers/add_process_groups.go
```diff
@@ -34,17 +36,46 @@ import (
...
+			if excludedServer.Locality == "" {
+				continue
+			}
+
+			processGroupID, found := strings.CutPrefix(excludedServer.Locality, prefix)
+			if !found {
+				continue
+			}
+			exclusions[fdbv1beta2.ProcessGroupID(processGroupID)] = fdbv1beta2.None{}
+		}
```
nicmorales9, line 76: Is this pattern used a lot to determine if a process is part of a locality? Is there a nice helper or maybe should we create one? (if not repeated, don't worry about it)

nicmorales9 (approved):
LGTM!

controllers/add_process_groups.go
johscheuer: I don't think we use it somewhere else right now. If we see other places where we could use this we can move this logic into a dedicated method.

PR #2208, opened 2025-02-05
Ensure that the tag suffix is only added when it is missing

johscheuer:
# Description

Ensure that the tag suffix is only added when it is missing. This should prevent cases like `sidecar-1-1`.

## Type of change

## Discussion

-

## Testing

Ran some manual e2e tests for this.

## Documentation

-

## Follow-up

-


nicmorales9 (approved):
lgtm-lgtm

PR #2210, opened 2025-02-10
Expose the multi-region recover mechanism

johscheuer:
# Description

This allows other tools to make use of this method.

## Type of change

- Other

## Discussion

-

## Testing

e2e tests will run the tests with the new plugin.

## Documentation

-

## Follow-up

-


nicmorales9 (approved):
LGTM assuming there is some "magic" spot where the Stdout/err get output with the log statements

PR #2211, opened 2025-02-13
Allow adding additional containers to a pod.

saintstack:
E.g. In tests, I want to be able to optionally add a sidecar.

* e2e/fixtures/cluster_config.go Add new AddtionalContainers member.

* e2e/fixtures/fdb_cluster_specs.go Add AddtionalContainers to the spec if present.

### DIFF
--- a/e2e/fixtures/cluster_config.go
+++ b/e2e/fixtures/cluster_config.go
@@ -121,6 +121,8 @@ type ClusterConfig struct {
 	// RedundancyMode defines the redundancy mode that should be used. If undefined the default is triple, except for
 	// the HaFourZoneDoubleSatRF4 configuration.
 	RedundancyMode fdbv1beta2.RedundancyMode
+	// AdditionalContainers allows specifying additional containers that will be added to all Pods
+	AdditionalContainers []corev1.Container
 }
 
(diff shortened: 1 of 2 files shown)


saintstack:
Need to be able to add more than just a container. Need to be able to add a (shared) volumemount to the main container and  a volume to the pod. Environment variables too.

johscheuer (changes requested):
Those changes shouldn't be required. You can already generate the default cluster spec, modify it and then create the FDB cluster from this spec. That's basically what the default `CreateFdbCluster` (see: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/e2e/fixtures/factory.go#L167-L174) is doing.

Is there any need to expose additional fields like this?

saintstack:
> Is there any need to expose additional fields like this?

Thanks for the suggestion. I can use clusterconfig to add voumemount and envvars to the main container but I don't see how I can add volumemounts to the pod and a additional container (I don't see how I can edit the pod template).  Thanks @johscheuer

saintstack:
Closing. https://github.com/FoundationDB/fdb-kubernetes-tests/pull/866 includes changes that allow adding of containers, volumeMounts, etc., over in fdb-kubernetes-test without need of changes here in operator.

PR #2212, opened 2025-02-13
Changes for the operator v2 release

johscheuer:
# Description

Changes for the operator v2 release (CRD will stay the same) with the following changes:

- Drop support for 6.3 and 6.2.
- DNS in cluster file and locality based exclusions are turned on per default.
- unified image is the default.
- Remove unused v1beta1 code (this drops support fro the operator v2 to work with v1beta1).

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1155
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1841
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1840
Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1839

## Type of change

- Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Discussion

I create another follow up PR that makes some more changes that we need for v2.

## Testing

Ran unit tests and CI will run e2e tests.

## Documentation

Will be updated in another PR.

## Follow-up

See above.

PR #2214, opened 2025-02-13
Streamline lock client usage

johscheuer:
# Description

This PR changes the lock client interface and not getting a lock will result in an error too. We could create a dedicated error type for this, but I think it's easier to work with the lock client this way instead of having two values which were checked differently.

Based on https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2212

## Type of change

## Discussion

-

## Testing

Unit tests + CI e2e

## Documentation

-

## Follow-up

-

### DIFF
--- a/fdbclient/lock_client.go
+++ b/fdbclient/lock_client.go
@@ -120,7 +120,7 @@ func (client *realLockClient) takeLockInTransaction(transaction fdb.Transaction)
 	newOwnerDenied := transaction.Get(client.getDenyListKey(ownerID)).MustGet() != nil
 	if newOwnerDenied {
 		logger.Info("Failed to get lock due to deny list")
-		return nil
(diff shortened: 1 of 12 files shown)


fdbclient/lock_client.go
```diff
@@ -123,7 +120,7 @@ func (client *realLockClient) takeLockInTransaction(transaction fdb.Transaction)
 	newOwnerDenied := transaction.Get(client.getDenyListKey(ownerID)).MustGet() != nil
 	if newOwnerDenied {
 		logger.Info("Failed to get lock due to deny list")
-		return false, nil
+		return nil
```
nicmorales9, line 123: Shouldn't we error here to communicate the lack of lock ownership?

nicmorales9 (approved):
lgtm

PR #2215, opened 2025-02-15
Ensure that the ConfigMap name for the cluster can't be modified

johscheuer:
# Description

Ensure that the ConfigMap name for the cluster can't be modified. Before this change the `ConfigMap` name could be changed but those changes had a few issues and bugs, so it's better to completely remote this option and ensure that the `ConfigMap` name is always the same. Also fixed a small bug in the `imageConfig` for backup.

## Type of change

## Discussion

Part of 2.0 release.

## Testing

Unit tests + CI will run e2e tests.

## Documentation

Will be updated in a follow up PR.

## Follow-up

-


nicmorales9 (approved):
👍

PR #2216, opened 2025-02-19
Ensure that the generation ID is used in the cluster file path to allow different cluster files to be used

johscheuer:
# Description

Follow up of: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2200
Follow up of: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2204

## Type of change

## Discussion

The idea behind the new changes is to include the unique generation ID from the cluster in the cluster file path, that way every connection string gets their own cluster file. In addition we added a reference counter to ensure the cluster file is not deleted as long as at least one client is not yet closed.

## Testing

Unit tests and CI for e2e tests.

## Documentation

I'll add some more documentation for those changes inside the code, just wanted to kick off the CI.

## Follow-up

We should update the go bindings for 7.1 to allow to close a database to prevent a potential memory leak.

### DIFF
--- a/controllers/add_pods.go
+++ b/controllers/add_pods.go
@@ -43,13 +43,15 @@ func (a addPods) reconcile(ctx context.Context, r *FoundationDBClusterReconciler
 	}
 	existingConfigMap := &corev1.ConfigMap{}
 	err = r.Get(ctx, types.NamespacedName{Namespace: configMap.Namespace, Name: configMap.Name}, existingConfigMap)
-	if err != nil && k8serrors.IsNotFound(err) {
-		logger.Info("Creating config map", "name", configMap.Name)
-		err = r.Create(ctx, configMap)
-		if err != nil {
-			return &requeue{curError: err}
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			logger.Info("Creating config map", "name", configMap.Name)
+			err = r.Create(ctx, configMap)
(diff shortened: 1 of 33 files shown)


iwalther:
Thank you very much for picking up this work, I was distracted the last weeks.

I think this change would improve the situation but not fix all edge cases. Both the go fdb binding and fdbcli will write to the file in the background if the cluster generation changes. Also if we keep closing the connection to the cluster, we will not notice if another operator instance replaces all the coordinators, and then we may be unable to reconnect to the cluster. (the other operator should update the fdb object so we get it from there, but this is certainly a detour). 

My suggestion would be to
* keep using a single, never changing cluster file for the golang fdb binding, connect once and never close the connection to the cluster. The client will keep this file updated in the background whenever the cluster layout changes, even if operator isn't actively doing anything at that time.
* copy the file above to a new temporary file each time we want to invoke fdbcli (or ask the library for the connection string and synthesize a cluster file), clean it up when fdbcli exits.

That will keep a  single connection to the cluster established at all times, but avoids the write races.

What do you think? Happy to contribute the code, I'm just still struggling to get the e2e tests running in my environment.

iwalther:
Here's my take: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2229
Sorry I still can't run the e2e tests myself, need to rely on your CI for now.

PR #2225, opened 2025-02-24
Fix PR action for docker build

johscheuer:
# Description

Before this both steps were building the data loader, which resulted in the latest image of the operator not being updated.

## Type of change

## Discussion

-

## Testing

Waiting for CI.

## Documentation

-

## Follow-up

-

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v4
         with:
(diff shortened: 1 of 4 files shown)


johscheuer (commented):
I believe some of the failures from trivy are wrong:

```text
ocker run -ti --user=root --platform='linux/amd64' docker.io/rockylinux/rockylinux:9.5-minimal rpm --query glibc

glibc-2.34-125.el9_5.1.x86_64
```

AnnigeriShambu (approved):

PR #2229, opened 2025-02-25
Single primary cluster file per cluster

iwalther:
# Description

Follow up of: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2200
Follow up of: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2204
Follow up of:https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2216

Keep a single, long-lived cluster file per cluster UID that is once written and then only managed by the FDB client library.

The problem is that multiple entities are writing the cluster file, concurrently without synchronization, leading to races:
* The FDB client library through the golang binding - the golang binding maintains a singleton per cluster file.
* Each fdbcli instance that FDB operator spawns, on notification from the cluster about a generation change,
* the kubernetes operator itself e.g. when trying connection options.

To decouple the fdbcli invocations, this change will write a temporary cluster file per fdbcli invocation, synthesized from the current connection string as indicated by the FDB client library. We discard the file immediately after use.

To remove the race with trying connection options, we adjust the lifecycle of the FDB admin client. We create a singleton instance per cluster UID, maintained by a singleton database provider. The initial connection always uses the seed connection string, afterwards we only poll the FDB client library for updates to the connection string.

With that, the FDB client library can exclusively manage the cluster file.

## Type of change

## Discussion

n/a

## Testing

Unit tests pass but I still didn't manage to run the e2e tests.

## Documentation

n/a

## Follow-up

Not sure if there are still races with the lock client accessing the same cluster file.

### DIFF
--- a/controllers/bounce_processes.go
+++ b/controllers/bounce_processes.go
@@ -52,9 +52,6 @@ func (c bounceProcesses) reconcile(_ context.Context, r *FoundationDBClusterReco
 	if err != nil {
 		return &requeue{curError: err}
 	}
-	defer func() {
-		_ = adminClient.Close()
-	}()
 
 	// If the status is not cached, we have to fetch it.
 	if status == nil {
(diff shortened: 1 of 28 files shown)


johscheuer:
Solved by https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2230

PR #2230, opened 2025-02-27
Update cluster file usage.

johscheuer:
# Description

Make use of dedicated cluster file for the fdb client lib and ensure to reuse it. The cli tooling will get a temp cluster file which will be deleted after the command was used.

## Type of change

## Discussion

## Testing

Ran e2e tests related to HA clusters.

I have to add another test in the e2e setup to see what happens when an operator is partitioned and all coordinators change (and no old coordinator is around) and how to resolve this case, in the previous setup a user could just update the `SeedConnectionString`.

## Documentation

Will be updated before merging.

## Follow-up

We should add a command in the kubectl plugin to update the connection string in the `FoundationDBCluster` status.

### DIFF
--- a/fdbclient/admin_client_test.go
+++ b/fdbclient/admin_client_test.go
@@ -29,6 +29,9 @@ import (
 	"path"
 	"time"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
 	"k8s.io/utils/pointer"
 
 	"github.com/go-logr/logr"
@@ -202,7 +205,7 @@ var _ = Describe("admin_client_test", func() {
 			GinkgoT().Setenv("FDB_NETWORK_OPTION_TRACE_FORMAT", traceFormat)
 		}
 
-		args, timeout := client.getArgsAndTimeout(command)
+		args, timeout := client.getArgsAndTimeout(command, "test")
 		Expect(timeout).To(Equal(expectedTimeout))
 		Expect(args).To(HaveExactElements(expectedArgs))
 	},
@@ -213,9 +216,8 @@ var _ = Describe("admin_client_test", func() {
 				timeout: 1 * time.Second,
 			},
 			&cliAdminClient{
-				Cluster:         nil,
-				clusterFilePath: "test",
-				log:             logr.Discard(),
+				Cluster: nil,
+				log:     logr.Discard(),
 			},
 			"",
 			"",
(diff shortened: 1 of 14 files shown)


johscheuer (commented):
In addition to those changes we want to add a kubectl plugin command to update the connection string in the `FoundationDBCluster` status  and we have to document the limitation when an operator instance is partitioned for a long time from a cluster and how to resolve this case (update the connection string in the `FoundationDBCluster` status  and restart the operator).

fdbclient/admin_client_test.go
```diff
@@ -532,10 +522,13 @@ var _ = Describe("admin_client_test", func() {
 
 		JustBeforeEach(func() {
 			cliClient := &cliAdminClient{
-				Cluster:         nil,
-				clusterFilePath: "test",
-				log:             logr.Discard(),
-				cmdRunner:       mockRunner,
+				Cluster: &fdbv1beta2.FoundationDBCluster{
```
nicmorales9, line 525: why is this one different? was it finding the cluster via the test path before and able to get the protocol version that way?

fdbclient/common.go
```diff
@@ -103,35 +104,75 @@ func getFDBDatabase(cluster *fdbv1beta2.FoundationDBCluster) (fdb.Database, erro
...
+// createClusterFileForCommandLine will create a cluster file that can be used by the fdb cli tooling, e.g. fdbcli or
+// fdbbackup. The file should be deleted after use.
+func createClusterFileForCommandLine(cluster *fdbv1beta2.FoundationDBCluster) (*os.File, error) {
+	tmpDir := path.Join(os.TempDir(), fmt.Sprintf("%s-cli", cluster.UID))
+	err := os.MkdirAll(tmpDir, 0777)
+	if err != nil {
+		return nil, err
+	}
+
+	randomFile, err := os.CreateTemp(tmpDir, "")
```
nicmorales9, line 154: why name this `randomFile`?

nicmorales9 (approved):
just some understanding questions, lgtm

fdbclient/admin_client_test.go
johscheuer: Those are only mocked methods that are called. Since we changed the logic a bit, we had to adjust how we initialize the admin client (`createClusterFileForCommandLine` takes the cluster as input).

fdbclient/common.go
johscheuer: Because the the filename will be random? I'm happy to rename the variable into tempFile if that's better.

johscheuer (commented):
CC @iwalther if you want to take a look, the main idea was not to change the admin client interface (as this would be a breaking change). And make use of the DB caching mechanism of the go bindings by using a static cluster file for the go bindings interactions and use a temp cluster file for cli related work. So similar in some implementation ideas to your PR in https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2229.

PR #2232, opened 2025-02-28
Change the ordering of the coordinator change and the exclusion

johscheuer:
# Description

Change the ordering of the coordinator change and the exclusion. It's better to first run the exclusion and then change the coordinators, this should ensure that the processes are excluded faster, e.g. in a case where the coordinator change causes a long recovery. We previously implemented with the coordinators change first because we waited for the exclusion to complete, since we use the `no_wait` option now per default, the exclude command shouldn't timeout, expect in cases where the FDB cluster is in a bad state. So doing the exclude first and then the coordinator change seems like the better choice to ensure that the processes are quickly excluded.

## Type of change

## Discussion

I labeled this as a bug but arguably this is not a bug.

## Testing

CI will run e2e tests.

## Documentation

-

## Follow-up

-

### DIFF
--- a/controllers/exclude_processes.go
+++ b/controllers/exclude_processes.go
@@ -166,6 +166,17 @@ func (e excludeProcesses) reconcile(ctx context.Context, r *FoundationDBClusterR
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


nicmorales9 (approved):
seems like a good change to me

PR #2235, opened 2025-03-04
Check the current cluster status before removing maintenance mode

johscheuer:
# Description

There might be cases where a pod (or a subset of pods) was down for maintenance for a longer time. This normally results in  a spike in the log server queue and when the storage server comes back it will be lagging until it fetched all the new mutations. The changes in this PR will account for that and will make sure that the maintenance mode is only removed when the processes have actually catched up, otherwise this could result in some side effects.

## Type of change

## Discussion

-

## Testing

Added unit tests and e2e tests will be running.

## Documentation

I'll update the documentation for the maintenance mode feature about those changes.

## Follow-up

-


pkg/fdbstatus/status_checks_test.go
```diff
@@ -2364,7 +2364,81 @@ var _ = Describe("status_checks", func() {
...
+					},
+				},
+			},
+			fmt.Errorf("worst durability lag is to high, current worst durability lag in seconds: 61.00, maximum allowed lag:  60.00"),
+		),
+		Entry("worst queue bytes for log server is to high",
+			&fdbv1beta2.FoundationDBStatus{
+				Cluster: fdbv1beta2.FoundationDBStatusClusterInfo{
+					Qos: fdbv1beta2.FoundationDBStatusQosInfo{
+						WorstQueueBytesLogServer: 500 * 1024 * 1024,
```
nicmorales9, line 2427: nit: might be nice to use the variable `maximumQueueByteSize` here and in the error messages

nicmorales9 (approved):
change looks good!

PR #2236, opened 2025-03-04
Allow to modify the default CPU and memory for pods

johscheuer:
# Description

Allow to modify the default CPU and memory for pods, e.g. to perform tests with different CPU/memory settings.

## Type of change

## Discussion

-

## Testing

Updated unit tests and CI will run e2e tests.

## Documentation

-

## Follow-up
-


saintstack (approved):
Thanks for adding.

PR #2240, opened 2025-03-06
Fix a bug in the connection string parser when a cluster has a single letter as name

johscheuer:
# Description

Fix a bug in the connection string parser when a cluster has a single letter as name. Before that the `ParseConnectionString` method would return an error. In addition we will validate the initial cluster file before creating it.

## Type of change

## Discussion

-

## Testing

## Documentation

-

## Follow-up

-

### DIFF
--- a/api/v1beta2/foundationdbcluster_types.go
+++ b/api/v1beta2/foundationdbcluster_types.go
@@ -1744,7 +1744,7 @@ func ParseConnectionString(str string) (ConnectionString, error) {
 	// The description is a logical description of the database using alphanumeric characters (a-z, A-Z, 0-9) and underscores.
 	description := firstSplit[0]
 	if !isAlphanumericWithUnderScore.MatchString(description) {
-		return ConnectionString{}, fmt.Errorf("invalid connection string: %s, only alphanumeric characters (a-z, A-Z, 0-9) and underscores are allowed", str)
+		return ConnectionString{}, fmt.Errorf("invalid connection string: %s, database description can only contain alphanumeric characters (a-z, A-Z, 0-9) and underscores", str)
 	}
 
 	secondSplit := strings.SplitN(firstSplit[1], "@", 2)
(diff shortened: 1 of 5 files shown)


api/v1beta2/foundationdbcluster_types.go
```diff
@@ -1731,32 +1727,75 @@ type ConnectionString struct {
...
 
-	coordinatorsStrings := strings.Split(components[3], ",")
-	coordinators := make([]string, len(coordinatorsStrings))
-	for idx, coordinatorsString := range coordinatorsStrings {
+	coordinatorsStrings := strings.Split(secondSplit[1], ",")
+	coordinators := make([]string, 0, len(coordinatorsStrings))
+	for _, coordinatorsString := range coordinatorsStrings {
+		// Ignore empty strings
+		if coordinatorsString == "" {
+			continue
```
nicmorales9, line 1766: it seems sensible enough to me but will everything on the FDB client side work with empty-string coordinators?

api/v1beta2/foundationdbcluster_types_test.go
```diff
@@ -811,19 +811,111 @@ var _ = Describe("[api] FoundationDBCluster", func() {
...
+			})
+		})
+
+		When("the input has an empty description", func() {
+			BeforeEach(func() {
+				input = ":abcd@127.0.0.1:4500,127.0.0.2:4500,127.0.0.3:4500"
+			})
+
+			It("should return an error", func() {
+				Expect(err).To(MatchError(ContainSubstring("invalid connection string: :abcd@127.0.0.1:4500,127.0.0.2:4500,127.0.0.3:4500, only alphanumeric characters (a-z, A-Z, 0-9) and underscores are allowed")))
```
nicmorales9, line 861: nit: the error isn't descriptive of the actual problem 
(my read is that this is an error that will not be hit often so the uglying the code for cases like this is debatable)

nicmorales9 (approved):
lgtm

api/v1beta2/foundationdbcluster_types.go
johscheuer: The coordinators cannot be an empty string. The check above skips empty strings, without that if statement an empty string could be added. There could be a case like below in the unit tests were someone enters an invalid connection string e.g. `test:test@`, the `secondSplit[1]` will always be an empty string in this case.

api/v1beta2/foundationdbcluster_types_test.go
```diff
@@ -811,19 +811,111 @@ var _ = Describe("[api] FoundationDBCluster", func() {
...
+				Expect(err).NotTo(HaveOccurred())
+				Expect(str.DatabaseName).To(Equal("s"))
+				Expect(str.GenerationID).To(Equal("abcd"))
+				Expect(str.Coordinators).To(ConsistOf(coordinatorsStr))
+			})
+		})
+
+		When("the input is valid with amultiple underscores", func() {
+			BeforeEach(func() {
+				input = "fdb_cluster_52v1bpr8:rhUbBjrtyweZBQO1U3Td81zyP9d46yEh@100.82.81.253:4500:tls,100.82.71.5:4500:tls,100.82.119.151:4500:tls,100.82.122.125:4500:tls,100.82.76.240:4500:tls"
```
johscheuer, line 900: Note for myself, we should add additional tests for IPv6 and dns names.

api/v1beta2/foundationdbcluster_types_test.go
johscheuer: That's a good call out, I'll improve the error message 👍

johscheuer:
> * Commit ID: [6e5524a](https://github.com/FoundationDB/fdb-kubernetes-operator/commit/6e5524a6aa1f86dfe4f266b521331cac29350c5d)
> * Duration 2:51:44
> * Result: ❌ **FAILED**
> * [Build Log](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator/fdb-kubernetes-operator-pr/4d9d1b4f-e66f-458c-8bcb-1efe034d2f30.gz) terminal output (available for 30 days)
> * [Build Workspace](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator-pr/4d9d1b4f-e66f-458c-8bcb-1efe034d2f30/workspace.zip) zip file of the working directory (available for 30 days)

Unrelated failures.

PR #2244, opened 2025-03-10
Ensure that the operator checks if the processes for newly created pods are up and running for the update pod config reconciler

johscheuer:
# Description

Ensure that the operator checks if the processes for newly created pods are up and running for the update pod config reconciler. This would handle the case where the operator did update a pod, which is already running with the new configuration but the process is not yet running. In the previous implementation the operator would be allowed to proceed with the updates if the fault tolerance metrics allow to proceed, in cases where the deletion and recreation of the pod would be fairly quick (less than 60 seconds) we could hit cases where the operator is taking down pods in another zone, even thought not all processes are up and running in the previous zone (and potentially never come up).

## Type of change

## Discussion

-

## Testing

Added unit tests. Adding an e2e test for this might be a bit challenging.

## Documentation

-

## Follow-up

-

### DIFF
--- a/controllers/update_pods.go
+++ b/controllers/update_pods.go
@@ -59,8 +59,8 @@ func (u updatePods) reconcile(ctx context.Context, r *FoundationDBClusterReconci
 			return &requeue{curError: err}
 		}
 	}
-	// r.InSimulation
-	updates, err := getPodsToUpdate(ctx, logger, r, cluster, getProcessesForProcessGroup(cluster, status))
+
+	updates, err := getPodsToUpdate(ctx, logger, r, cluster, getProcessesByProcessGroup(cluster, status))
 	if err != nil {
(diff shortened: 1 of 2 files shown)


controllers/update_pods.go
```diff
@@ -131,8 +132,32 @@ func getFaultDomainsWithUnavailablePods(ctx context.Context, logger logr.Logger,
 	return faultDomainsWithUnavailablePods
 }
 
+func getProcessesForProcessGroup(cluster *fdbv1beta2.FoundationDBCluster, status *fdbv1beta2.FoundationDBStatus) map[string][]fdbv1beta2.FoundationDBStatusProcessInfo {
```
nicmorales9, line 135: name kinda implies the ProcessGroup is provided, `getProcessesByProcessGroup` seems closer to me?

nicmorales9 (approved):
LGTM!

Issue #2243, opened 2025-03-10
Force replacement for changing IP families until #834 is implemented


johscheuer:
### What happened?

Until we implement https://github.com/FoundationDB/fdb-kubernetes-operator/issues/834, we should be doing a replacement for changing the IP families, otherwise we could run into issues where the rolling update is causing an issue (when not enough new processes are up with the new IP family and the operator wants to change the coordinators).

### What did you expect to happen?

Changing the IP family should replace all pods.

### How can we reproduce it (as minimally and precisely as possible)?

Change the IP family for a running cluster.

### Anything else we need to know?

-

### FDB Kubernetes operator

<details>

```console
$ kubectl fdb version
# paste output here
```

</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Cloud provider

<details>

</details>

Issue #834, opened 2021-07-08
Changing IP families without a rolling bounce


brownleej:
When we change the IP family setting, such as when converting a cluster from IPv4 to IPv6, it does a rolling bounce of the processes that will cause the cluster to repeatedly drop and regain fault tolerance. We should see if it's possible to do this without a rolling bounce by using multiple listeners, like we do during TLS conversions.

johscheuer:
We have a few IPv6/DualStack related issue do we want to group them by a label or project?

brownleej:
Why don't we start out with a label and see if we accumulate enough issues that it makes sense to create a full project?

PR #2247, opened 2025-03-11
Ensure that pods are replaced if we change the Pod IP family

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/2243

## Type of change

## Discussion

With the previous behaviour it could happen that the coordinators went unreachable because they were listening on the wrong IP family and then the operator would get stuck. This change prevents it by spinning up a new set of pods.

## Testing

Added a new e2e test to convert from `nil` family to `v4`. I run some tests in a dual-stack cluster.

## Documentation

-

## Follow-up

https://github.com/FoundationDB/fdb-kubernetes-operator/issues/834 the idea here would be to listen on both addresses for some time.

### DIFF
--- a/internal/pod_helper.go
+++ b/internal/pod_helper.go
@@ -25,72 +25,57 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	monitorapi "github.com/apple/foundationdb/fdbkubernetesmonitor/api"
 	"net"
 	"strconv"
 
-	"github.com/go-logr/logr"
-
-	"k8s.io/utils/pointer"
-
 	fdbv1beta2 "github.com/FoundationDB/fdb-kubernetes-operator/v2/api/v1beta2"
+	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // GetPublicIPsForPod returns the public IPs for a Pod
 func GetPublicIPsForPod(pod *corev1.Pod, log logr.Logger) []string {
-	var podIPFamily *int
-
 	if pod == nil {
 		return []string{}
 	}
 
-	for _, container := range pod.Spec.Containers {
-		if container.Name != fdbv1beta2.SidecarContainerName {
+	podIPFamily, err := GetIPFamilyFromPod(pod)
+	if err != nil {
+		log.Error(err, "Could not parse IP family")
+		return []string{pod.Status.PodIP}
+	}
+
+	if podIPFamily == nil {
+		return []string{pod.Status.PodIP}
+	}
+
+	podIPs := pod.Status.PodIPs
+	matchingIPs := make([]string, 0, len(podIPs))
+	for _, podIP := range podIPs {
+		ip := net.ParseIP(podIP.IP)
+		if ip == nil {
+			log.Error(nil, "Failed to parse IP from pod", "ip", podIP)
 			continue
 		}
-		for indexOfArgument, argument := range container.Args {
-			if argument == "--public-ip-family" && indexOfArgument < len(container.Args)-1 {
-				familyString := container.Args[indexOfArgument+1]
-				family, err := strconv.Atoi(familyString)
-				if err != nil {
-					log.Error(err, "Error parsing public IP family", "family", familyString)
-					return nil
-				}
-				podIPFamily = &family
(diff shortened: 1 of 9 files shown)


johscheuer:
> * Commit ID: [de04c61](https://github.com/FoundationDB/fdb-kubernetes-operator/commit/de04c61c738aae6d6554cd6b9c7770c198f01e18)
> * Duration 3:10:31
> * Result: ❌ **FAILED**
> * [Build Log](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator/fdb-kubernetes-operator-pr/716c9574-dfc8-4681-8b13-ac389c56c61f.gz) terminal output (available for 30 days)
> * [Build Workspace](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator-pr/716c9574-dfc8-4681-8b13-ac389c56c61f/workspace.zip) zip file of the working directory (available for 30 days)

The test failed because there was a pod with a `failed` status:

johscheuer:
> * Commit ID: [abfd29c](https://github.com/FoundationDB/fdb-kubernetes-operator/commit/abfd29c7bb08111ac3b18953b0cebb850e40bc4d)
> * Duration 3:37:29
> * Result: ❌ **FAILED**
> * [Build Log](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator/fdb-kubernetes-operator-pr/a012dd42-c045-4081-8344-56a7a9122197.gz) terminal output (available for 30 days)
> * [Build Workspace](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator-pr/a012dd42-c045-4081-8344-56a7a9122197/workspace.zip) zip file of the working directory (available for 30 days)

```text
[Container] 2025/03/12 14:15:42.437957 Running command if $(grep -q -- "--- FAIL:" ${CODEBUILD_SRC_DIR}/logs/*.log); then echo "TESTS FAILED SEE THESE LOGS:"; echo ; grep -l -- "--- FAIL:" ${CODEBUILD_SRC_DIR}/logs/*.log; export fail_test=true; fi
TESTS FAILED SEE THESE LOGS:

/codebuild/output/src91157654/src/github.com/FoundationDB/fdb-kubernetes-operator/logs/test_operator_ha_upgrades.log
```

internal/pod_models.go
```diff
@@ -1039,6 +1039,9 @@ func GetPodMetadata(cluster *fdbv1beta2.FoundationDBCluster, processClass fdbv1b
 	metadata.Annotations[fdbv1beta2.LastSpecKey] = specHash
 	metadata.Annotations[fdbv1beta2.PublicIPSourceAnnotation] = string(cluster.GetPublicIPSource())
 	metadata.Annotations[fdbv1beta2.ImageTypeAnnotation] = string(cluster.DesiredImageType())
+	if cluster.Spec.Routing.PodIPFamily != nil {
+		metadata.Annotations[fdbv1beta2.IPFamilyAnnotation] = strconv.Itoa(*cluster.Spec.Routing.PodIPFamily)
```
brownleej, line 1043: This looks like it will apply this annotation to all clusters with a PodIPFamily defined as soon as the operator is deployed. Do we expect this to be a fast and non-disruptive operation?

internal/pod_helper.go
```diff
@@ -237,6 +222,73 @@ func GetPublicIPSource(pod *corev1.Pod) (fdbv1beta2.PublicIPSource, error) {
...
+		}
+	}
+
+	return nil, nil
+}
+
+// GetIPFamily determines the IP family based on the annotation.
+// TODO (johscheuer): Make use of this method once we did the next release 2.2.0. This will ensure that the operator
+// can add the fdbv1beta2.IPFamilyAnnotation to all pods.
+//func GetIPFamily(pod *corev1.Pod) (*int, error) {
```
brownleej, line 274: Is the intention that this function will be used in place of `GetIPFamilyFromPod` once we apply all the necessary annotations? What will happen if a cluster does not get re-reconciled between releases, and is left without the annotations for the IP family?

internal/pod_models.go
johscheuer: This operation will be non-disruptive and fast, as the operator will only update the pod metadata.

internal/pod_helper.go
johscheuer: That was the initial intention, I think a better approach is to use the value of the annotation if present and if not fall back to the logic from `GetIPFamilyFromPod`. I make the code adjustments later.

brownleej (approved):
👍

PR #2248, opened 2025-03-12
Use patch for pod updates

acsmyth:
# Description

In a Kubernetes environment with admission controllers that mutate pod specs (for example, to enforce a particular `appArmorProfile`), if the K8s client API versions used for pod spec ser/de in the operator are too old to include the fields that the admission controllers mutated (e.g. `appArmorProfile`), then the pod spec deserialized from `GetPod` will drop those unknown fields. 

If any of these fields are immutable, then all subsequent pod update attempts will fail because the full pod spec update will implicitly attempt to set those immutable unknown fields back to `nil`.

By using a patch instead of a full update, the operator will only attempt to update the fields it has explicitly changed, avoiding this class of issues resulting from desync between K8s client and server api versions.

## Type of change

## Discussion

Alternatively, bumping the K8s client libraries to more recent versions that include `appArmorProfile` would likely resolve this particular occurrence of this issue. However, the same issue could plausibly reoccur in the future if the full update strategy continues to be used.

### DIFF
--- a/pkg/podmanager/pod_lifecycle_manager.go
+++ b/pkg/podmanager/pod_lifecycle_manager.go
@@ -168,7 +168,7 @@ func (manager StandardPodLifecycleManager) UpdateImageVersion(ctx context.Contex
 }
 
 // UpdateMetadata updates an Pod's metadata.
-func (manager StandardPodLifecycleManager) UpdateMetadata(ctx context.Context, r client.Client, _ *fdbv1beta2.FoundationDBCluster, pod *corev1.Pod, *corev1.Pod) error {
+func (manager StandardPodLifecycleManager) UpdateMetadata(ctx context.Context, r client.Client, _ *fdbv1beta2.FoundationDBCluster, pod *corev1.Pod, originalPod *corev1.Pod) error {
 	return r.Patch(ctx, pod, client.MergeFrom(originalPod))
 }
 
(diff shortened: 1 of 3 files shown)


pkg/podmanager/pod_lifecycle_manager.go
```diff
@@ -162,13 +162,14 @@ func (manager StandardPodLifecycleManager) UpdatePods(ctx context.Context, r cli
...
+	originalPod := pod.DeepCopy()
 	pod.Spec.Containers[containerIndex].Image = image
-	return r.Update(ctx, pod)
+	return r.Patch(ctx, pod, client.MergeFrom(originalPod))
 }
 
 // UpdateMetadata updates an Pod's metadata.
-func (manager StandardPodLifecycleManager) UpdateMetadata(ctx context.Context, r client.Client, _ *fdbv1beta2.FoundationDBCluster, pod *corev1.Pod) error {
-	return r.Update(ctx, pod)
+func (manager StandardPodLifecycleManager) UpdateMetadata(ctx context.Context, r client.Client, _ *fdbv1beta2.FoundationDBCluster, pod *corev1.Pod, *corev1.Pod) error {
```
johscheuer, line 171:
```suggestion
func (manager StandardPodLifecycleManager) UpdateMetadata(ctx context.Context, r client.Client, _ *fdbv1beta2.FoundationDBCluster, pod *corev1.Pod, originalPod *corev1.Pod) error {
```

Thanks for the PR, we probably want to move this behind a flag to allow users to enable the usage of patch or still make use of the update mechanism.

johscheuer (commented):
I opened a new PR with changes to support the patch method but still use the update method as the default (to ensure the existing behaviour is not changed).

PR #2253, opened 2025-03-19
Add support for using patch to update pods instead of update

johscheuer:
# Description

Implements: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2248 

## Type of change

## Discussion

-

## Testing

Ran test with the update method, will add support for the patch method in the e2e tests.

## Documentation

-

## Follow-up

-

### DIFF
--- a/pkg/podmanager/pod_lifecycle_manager.go
+++ b/pkg/podmanager/pod_lifecycle_manager.go
@@ -71,12 +71,55 @@ type PodLifecycleManager interface {
 	GetDeletionMode(*fdbv1beta2.FoundationDBCluster) fdbv1beta2.PodUpdateMode
 }
 
+// PodUpdateMethod defines the way how a Pod should be updated, the default is "update".
+type PodUpdateMethod string
+
+const (
+	// Update is the default way to update pods with the update call.
+	Update PodUpdateMethod = "update"
+	// Patch is will use the path method to update pods.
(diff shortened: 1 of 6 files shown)


pkg/podmanager/pod_lifecycle_manager.go
```diff
@@ -71,12 +71,55 @@ type PodLifecycleManager interface {
...
 }
 
+// PodUpdateMethod defines the way how a Pod should be updated, the default is "update".
+type PodUpdateMethod string
+
+const (
+	// Update is the default way to update pods with the update call.
+	Update PodUpdateMethod = "update"
+	// Patch is will use the path method to update pods.
+	Patch PodUpdateMethod = "patch"
```
acsmyth, line 81:
```suggestion
	// Patch will use the patch method to update pods.
	Patch PodUpdateMethod = "patch"
```

acsmyth (commented):
Thanks!

AnnigeriShambu (approved):

hxu:
I wasn't able to figure this out by looking at the PR -- what's the right way to configure this option? I assume that it's something we'd have to change in the Deployment resource?

hxu:
Ah is this a command line argument?  So you would pass it into the manager command in the Deployment resource like this:

```
      containers:
      - name: manager
        command:
        - /manager
        - -pod-update-method patch
 ```

PR #2259, opened 2025-04-01
Implement changes for Multi-Cluster coordination between multiple operator instances.

johscheuer:
# Description

Implements: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/docs/design/better_coordination_multi_operator.md

## Type of change

## Discussion

See design.

## Testing

Ran some e2e tests to verify things work, will be running some additional tests with the global mode.

## Documentation

Added.

## Follow-up

N/A


fdbclient/fdb_client.go
```diff
@@ -100,10 +279,61 @@ type mockFdbLibClient struct {
 	mockedError error
 	// requestedKey will be the key that was used to call getValueFromDBUsingKey.
 	requestedKey string
+	// coordinationState represents the current coordination state.
```
brownleej, line 282: There's a bit of a naming collision here because we also use "coordination state" in other contexts to refer to the state stored by the coordinators in the core database. I don't have a better alternative in mind, but we may need to limit how widely we use this combination of words.

pkg/fdbadminclient/lock_client.go
```diff
@@ -3,7 +3,7 @@
  *
  * This source file is part of the FoundationDB open source project
  *
- * Copyright 2018-2021 Apple Inc. and the FoundationDB project authors
+ * Copyright 2018-2024 Apple Inc. and the FoundationDB project authors
```
brownleej, line 6: Are these updates coming from some kind of automated update script? If so, does that script need to be updated to reflect that, if I'm not mistaken, it's currently 2025?

brownleej (approved):
I left some general notes, but nothing that should block this PR. 👍

pkg/fdbadminclient/lock_client.go
johscheuer: I started some changes in December last year and seems like I missed those. Right now we don't have any automated way to update the headers.

internal/coordination/coordination.go
```diff
@@ -0,0 +1,371 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// In case that not all processes are ready we have to check when the last time an update to the pending and
+	// ready list has happened.
+	if len(notReadyProcesses) > 0 {
+		// If there was a change in the last 5 minutes (or longer), we will return an error message and wait for the processes
+		// to get  ready.
+		if time.Since(timestampLastAdded) < max(5*waitTime, IgnoreMissingProcessDuration) {
```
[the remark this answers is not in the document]
johscheuer: That makes sense, I'll create a radar to follow up on this.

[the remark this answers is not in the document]
johscheuer: Created: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/2262

johscheuer:
Okay, that was a bit unexpected. I enabled auto-merge and it seems like there was a race condition between the auto-merge enable and GitHub action starting 🙄

PR #2260, opened 2025-04-02
Ensure that the cordon command doesn't stop when a node doesn't host any pods

johscheuer:
# Description

The previous behaviour was bad in cases where a few hosts where not hosting any pods, in this case the whole command would be aborted. I think it's better to just print a warning instead of returning an error and stop everything. In some cases when enough nodes are selected (e.g. with a node selector) the chances that a single node not hosting any pods, is fairly high.

## Type of change

## Discussion

-

## Testing

Adjusted unit tests and ran some manual tests.

## Documentation

-

## Follow-up

-


nicmorales9 (approved):
yup, good shout!

johscheuer (commented):
Merging the PR as the e2e tests are not covering those code changes.

PR #2267, opened 2025-04-22
Add additional fields to the model for the FDB database status.

brownleej:
# Description

This adds fields to the FoundationDBStatus models to reflect more of the underlying status model.

## Type of change

## Discussion

I did some refactoring of the tests for the status parsing to accommodate the new changes. The raw diff from ginkgo proved hard to work with, so I broke it down into assertions for each field in the status, and used HaveExactElements for the array types to get more item-by-item diffs.

The specific fields I chose to add here were based on what we've found helpful in other projects, but these models might be more future-proof if we had a way to synthesize the full status model based on the contents of the FoundationDB repo, or to synthesize it inside the go bindings themselves. I didn't want to take on something that heavy without more discussion, but if folks think it's a good way to go I can explore that further.

## Testing

Manual testing?*

I updated the unit tests to cover the new fields.

I think the end-to-end in the PRB should cover our needs.

## Documentation

I added documentation in the code for the new fields, which I think is sufficient for this.

## Follow-up

No.

No.


api/v1beta2/foundationdb_status.go
```diff
@@ -140,6 +140,21 @@ type FoundationDBStatusClusterInfo struct {
 
 	// BounceImpact represents the bounce_impact part of the machine-readable status.
 	BounceImpact FoundationDBBounceImpact `json:"bounce_impact,omitempty"`
+
+	// DatabaseAvailable indicates whether the database is currently available.
+	DatabaseAvailable bool `json:"database_available,omitempty"`
```
johscheuer, line 145: I would prefer using a pointer to a bool here, that makes it easier to differentiate between the value being set intentionally and the default value. That applies to the other fields too. I know that this means some additional work when using those values but in some cases it might be helpful to know if the value was set or not.

api/v1beta2/suite_test.go
```diff
@@ -27,3 +29,27 @@ func TestCmd(t *testing.T) {
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "FDB v1beta2 API")
 }
+
+// sortedMapValues gets the values from a map as a list, sorted according to the ordering of the underlying keys.
+func sortedMapValues[K cmp.Ordered, V any](data map[K]V) []V {
```
johscheuer, line 34: Do you think it makes sense to add a comment why we can't use those helper functions from the core library: https://pkg.go.dev/maps#Values?

johscheuer (commented):
In general the changes look good. Would be great to discuss the benefits of using pointers in the newly added fields.

api/v1beta2/suite_test.go
brownleej: I took a look at those functions, and I think the main benefit to this approach is the consistent ordering for the HaveExactElements matcher. However, I think I can express what we actually want to check here more cleanly with a ConsistOf matcher, so I'll go that route instead.

nicmorales9 (approved):
LGTM

PR #2272, opened 2025-05-05
Add stricter types to some of the fields in the database status.

brownleej:
# Description

When looking at how we were using the database status in some other projects, I noticed frequent type conversions in a couple of places. This updates the related fields to have more precise types, reflecting the ranges of values they can hold. 

## Type of change

## Discussion

I'm keeping this as a draft PR to start with, because I realized it presented a small breaking change, and I wanted to get some feedback on how we want to handle that in general. Is it better to leave this kind of change to a major version, and maybe add TODOs to indicate that we should follow up on these changes in the next major release?

## Testing

Manual testing?*

I updated the types in the unit tests, and ran the test suite locally.

No.

## Documentation

No.

No.

No.

No.

## Follow-up

No.

No.

### DIFF
--- a/api/v1beta2/foundationdb_status.go
+++ b/api/v1beta2/foundationdb_status.go
@@ -219,7 +219,7 @@ type FoundationDBStatusProcessMessage struct {
 // roles.
 type FoundationDBStatusProcessRoleInfo struct {
 	// Role defines the role a process currently has
-	Role string `json:"role,omitempty"`
+	Role ProcessRole `json:"role,omitempty"`
 	// StoredBytes defines the number of bytes that are currently stored for this process.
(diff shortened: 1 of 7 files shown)


nicmorales9:
I like it

johscheuer (commented):
I like the general idea of changing the types to more precise types. The only concern would be that this is a breaking change, so technically we would need a 3.x release. On the other hand we changed the `fdbclient` interface a few types in a breaking manner too,

johscheuer (commented):
Just wanted to get the discussion started here. I like the changes and I think those changes make sense. We should probably revisit all the places where we could use stricter types and make those changes before cutting a v3 release of the operator. We can chat out-of-band about the v3 release.

PR #2273, opened 2025-05-09
Fix crash in findFoundationDBClusterForNode when in global mode

dig-whois:
# Description

When the operator is executed with `cluster-label-key-for-node-trigger` flag in global mode it goes to crash loop, because the namespace is empty and in `FoundationDBClusterReconciler.findFoundationDBClusterForNode` a null pointer is passed to k8s-client Lister. We fix this issue by passing the variable only if namespace is not empty.

## Type of change

X Bug fix (non-breaking change which fixes an issue)
- Documentation
- Other

## Discussion

N/A

## Testing

We ran it in different environments. 

No

## Documentation

## Follow-up

### DIFF
--- a/controllers/cluster_controller_test.go
+++ b/controllers/cluster_controller_test.go
@@ -2691,6 +2691,104 @@ var _ = Describe("cluster_controller", func() {
 				})
 			})
 		})
+
+		Context("when finding fdb pods in a node", func() {
+			// All the nodes that are part of the cluster
+			var originalNodeList corev1.NodeList
+			// unrelatedNode is a node that does not hold a pod that is part of the cluster
+			var unrelatedNode *corev1.Node
+
+			BeforeEach(func() {
+				err := k8sClient.List(context.Background(), &originalNodeList)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(originalNodeList.Items).To(HaveLen(len(originalPods.Items)))
+
+				// update the cluster label keys of all pods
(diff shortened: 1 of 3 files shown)


johscheuer (commented):
The changes LGTM, could you add a unit tests where the reconciler namespace is unset?

dig-whois:
Sure, I can do that. To be clear about the scope, should the unit test target this case only or do we want something that is passed to all other test cases in `cluster_controlle_test`?

dig-whois:
> The changes LGTM, could you add a unit tests where the reconciler namespace is unset?

Sure, I can do that.

controllers/cluster_controller_test.go
```diff
@@ -2691,6 +2691,104 @@ var _ = Describe("cluster_controller", func() {
...
+				err = k8sClient.Create(context.Background(), unrelatedNode)
+				Expect(err).NotTo(HaveOccurred())
+
+			})
+			AfterEach(func() {
+				clusterReconciler = createTestClusterReconciler()
+				err := k8sClient.Delete(context.Background(), unrelatedNode)
+				Expect(err).NotTo(HaveOccurred())
+			})
+			When("wathcing all namespaces", func() {
```
johscheuer, line 2735:
```suggestion
			When("watching all namespaces", func() {
```

dig-whois:
@johscheuer, please let me know if the added test cases were enough. 
thanks

johscheuer (approved):
Thanks for the PR!

PR #2276, opened 2025-05-15
Add support for the operator to run on ARM

johscheuer:
# Description

Add support for the operator to run on ARM.

## Type of change

## Discussion

Right now only 7.3 and newer is supported as the previous versions don't have the client packages for arm.

## Testing

Build and ran the image locally.

## Documentation

Added some notes to the Dockerfile.

## Follow-up

-

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ ARG TAG="latest"
 
 RUN set -eux && \
     if [ "$TARGETARCH" = "amd64" ]; then \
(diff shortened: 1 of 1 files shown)


johscheuer (commented):
We have to update our pipeline to support the `TARGETARCH`.

nicmorales9 (approved):
Looks good to me

johscheuer (commented):
Image build works 👍

PR #2278, opened 2025-05-20
Allow bypassing the downgrade check with the IgnoreUpgradabilityChecks flag.

brownleej:
# Description

There are some minor version downgrades that _are_ supported by the core database, or are supported in specific paths and situations, and ideally we should be able to test those paths through manual overrides.

## Type of change

## Discussion

We may want a separate flag for this, because we could reasonably view these as separate checks, even though they are in the function. However, I think that circumstances where someone would want to override these two checks are very similar: either they are implementing their own automated upgrade checking external to the operator, or they are individually asserting that this upgrade is safe to proceed. In those circumstances, having a single flag to set seems like a more clear experience.

## Testing

Manual testing?*

I did a manual test of a downgrade with this flag.

No.

## Documentation

Yes, I've updated the documentation for this field, and the general documentation for this reconciler.

Yes.

N/A

Yes.

## Follow-up

No.

No.


johscheuer (approved):
Changes LGTM 👍 I restarted the CI pipeline to see what caused the failure.

PR #2283, opened 2025-05-23
Run the addPods reconciler before connecting to FDB in case we use DNS entries in the cluster file

johscheuer:
# Description

Otherwise this could lead to the operator crashing with:

```text
  Error determining public address.
  SIGSEGV: segmentation violation
  PC=0x7fc6699c41fa m=16 sigcode=1 addr=0x40
  signal arrived during cgo execution
```

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/2286 (seems like a bug caused by the previous behaviour).

## Type of change

## Discussion

-

## Testing

Updated the e2e test + an it locally.

## Documentation

-

## Follow-up

-

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -208,7 +208,7 @@ func (r *FoundationDBClusterReconciler) Reconcile(ctx context.Context, request c
 	// the FoundationDB cluster.
 	if cluster.UseDNSInClusterFile() {
 		req := runClusterSubReconciler(ctx, clusterLog, addPodsReconciler, r, cluster, nil)
-		if req != nil {
+		if req != nil && !req.delayedRequeue {
 			return processRequeue(req, addPodsReconciler, cluster, r.Recorder, clusterLog)
 		}
 	}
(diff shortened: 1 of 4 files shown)


controllers/cluster_controller.go
```diff
@@ -170,6 +169,17 @@ func (r *FoundationDBClusterReconciler) Reconcile(ctx context.Context, request c
...
 
+	// When using DNS entries in the cluster file, we want to make sure to create pods if required before doing any
+	// interaction with the FDB cluster. If none of the coordinator pods is running, this will cause a crash in the FDB
+	// go bindings (or rather the C client). To prevent those case we run the addPods reconciler before interacting with
+	// the FoundationDB cluster.
+	if cluster.UseDNSInClusterFile() {
+		req := runClusterSubReconciler(ctx, clusterLog, addPods{}, r, cluster, nil)
+		if req != nil {
+			clusterLog.Info("ran the initial add Pods reconciler", "message", req.message, "error", req.curError)
+		}
```
nicmorales9, line 180: if there's a failure here wont it just carry on and also cause the crash?

johscheuer: Right, the idea was that the crashing controller will be restarted by Kubernetes anyways, but I can add the logic to retry it.

nicmorales9: happy with that or a comment about intentionally allowing crash

PR #2305, opened 2025-06-25
Add support for arm based builds with the operator

johscheuer:
# Description

Add the capabilities to build the operator image for arm architecture.

## Type of change

## Discussion

-

## Testing

Ran the docker build commands locally.

## Documentation

Added a note about the experimental arm build.

## Follow-up

-

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,9 @@ RUN set -eux && \
          FDB_ARCH=amd64; \
     elif [ "$TARGETARCH" = "arm64" ]; then \
          FDB_ARCH=aarch64; \
+         if [ "${FDB_VERSION%.*}" = "7.1" ]; then \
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -16,6 +16,9 @@ RUN set -eux && \
          FDB_ARCH=amd64; \
     elif [ "$TARGETARCH" = "arm64" ]; then \
          FDB_ARCH=aarch64; \
+         if [ "${FDB_VERSION%.*}" = "7.1" ]; then \
+            FDB_VERSION="7.3.63"; \
```
nicmorales9, line 20: will this behaviour be a surprise? seems odd to swap the version only if ARM, but if it's not likely to affect anyone outside builds then 🤷

nicmorales9 (approved):

PR #2316, opened 2025-07-07
Update typos and change the rollout strategy for the operator pods

johscheuer:
# Description

Fixed a few typos and changed the way hoe the operator pods are recreated in the test suite. Instead of deleting them we add a new annotation in the template (similar to `kubectl rollout restart ..`). I also added a new log entry for the lock client when the lock is released to log how long the lock was held.

## Type of change

- Other

## Discussion

-

## Testing

Ran the e2e tests manually.

## Documentation

-

## Follow-up

-

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,4 @@ dist/
 
 # Used for testing locally build FDB library
 libfdb_c.so
+fdb-kubernetes-operator
--- a/e2e/fixtures/fdb_operator_client.go
+++ b/e2e/fixtures/fdb_operator_client.go
@@ -22,7 +22,7 @@ package fixtures
 
 import (
 	"bytes"
-	ctx "context"
+	"context"
 	"errors"
 	"html/template"
 	"io"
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -47,3 +47,4 @@ dist/
 
 # Used for testing locally build FDB library
 libfdb_c.so
+fdb-kubernetes-operator
```
nicmorales9, line 50: is this right?

e2e/fixtures/fdb_operator_client.go
```diff
@@ -748,10 +755,20 @@ func (factory *Factory) WaitUntilOperatorPodsRunning(namespace string) {
 // RecreateOperatorPods will recreate all operator Pods in the specified namespace and wait until the new Pods are
 // up and running.
 func (factory *Factory) RecreateOperatorPods(namespace string) {
-	gomega.Expect(
-		factory.GetControllerRuntimeClient().
-			DeleteAllOf(ctx.TODO(), &corev1.Pod{}, client.InNamespace(namespace), client.MatchingLabels(map[string]string{"app": operatorDeploymentName})),
-	).NotTo(gomega.HaveOccurred())
+	restartAnnotation := "foundationdb.org/restart-deployment"
```
nicmorales9, line 758: what is this magic?

PR #2320, opened 2025-07-08
Update the client compatibility check to support cases where the clients have a newer version

johscheuer:
# Description

When a client has a newer version present, this will be the max protocol version and therefore the current checks would fail. 

## Type of change

## Discussion

-

## Testing

Updated the unit tests.

## Documentation

-

## Follow-up

-

### DIFF
--- a/api/v1beta2/foundationdbcluster_types.go
+++ b/api/v1beta2/foundationdbcluster_types.go
@@ -1902,7 +1902,14 @@ func (cluster *FoundationDBCluster) CheckReconciliation(log logr.Logger) (bool, 
 		}
 	}
 
-	if reconciled {
+	if reconciled && cluster.Status.Generations.Reconciled != cluster.Generation {
+		logger.Info(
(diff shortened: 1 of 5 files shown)


nicmorales9 (approved):
LGTM

api/v1beta2/foundationdbcluster_types.go
```diff
@@ -1902,7 +1902,14 @@ func (cluster *FoundationDBCluster) CheckReconciliation(log logr.Logger) (bool,
 		}
 	}
 
-	if reconciled {
+	if reconciled && cluster.Status.Generations.Reconciled != cluster.Generation {
+		logger.Info(
+			"Update reconciled generation",
+			"previousGeneration",
+			"cluster.Status.Generations.Reconciled",
```
johscheuer, line 1909:
```suggestion
			cluster.Status.Generations.Reconciled,
```

PR #2323, opened 2025-07-10
Disable test for unified image, since the unified image doesn't have a check for this

johscheuer:
# Description

The unified image doesn't have a check if a file is present (it does in a sense that it checks the binary before starting the binary but not an external check from the operator side).

## Type of change

- Other

## Discussion

Disabling the test for now. If we don't disable this test we will see some spontaneous failures, e.g. because the upgrades moves forward and in cases where a transaction or stateless pod is picked the test could fail with:

```text
 [FAILED] Timed out after 120.000s.
  Unexpected error:
      <*errors.StatusError | 0x14000871400>:
      pods "jdev-stateless-59937" not found
      {
          ErrStatus: {
              TypeMeta: {Kind: "", APIVersion: ""},
              ListMeta: {
                  SelfLink: "",
                  ResourceVersion: "",
                  Continue: "",
                  RemainingItemCount: nil,
              },
              Status: "Failure",
              Message: "pods \"jdev-stateless-59937\" not found",
              Reason: "NotFound",
              Details: {
                  Name: "jdev-stateless-59937",
                  Group: "",
                  Kind: "pods",
                  UID: "",
                  Causes: nil,
                  RetryAfterSeconds: 0,
              },
              Code: 404,
          },
      }
  occurred
  In [It] at: /Users/jscheuermann/go/src/github.com/FoundationDB/fdb-kubernetes-operator/e2e/fixtures/fdb_cluster.go:596 @
```

## Testing

Ran the test manually.

## Documentation

Added in the test case.

## Follow-up

-

### DIFF
--- a/e2e/Makefile
+++ b/e2e/Makefile
@@ -192,12 +192,12 @@ endif
 	  --dump-operator-state=$(DUMP_OPERATOR_STATE) \
 	  --cluster-name=$(CLUSTER_NAME) \
 	  --storage-engine=$(STORAGE_ENGINE) \
-	  --fdb-version-tag-mapping=$(FDB_VERSION_TAG_MAPPING) \
+	  --fdb-version-tag-mapping="$(FDB_VERSION_TAG_MAPPING)" \
 	  --unified-fdb-image=$(UNIFIED_FDB_IMAGE) \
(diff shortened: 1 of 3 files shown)


nicmorales9 (approved):

johscheuer (commented):
Seems like the EKS cache had a rough time.

PR #2325, opened 2025-07-15
Ensure process groups are removed from the pending restart list if they are stuck in terminating or the process is missing

johscheuer:
# Description

In our e2e tests we have seen a few cases where the process groups were not removed if they are stuck in terminating. If the process is missing or stuck in terminating we should remove it from the pending restart set.

## Type of change

## Discussion

-

## Testing

Ran the unit tests, CI will run the e2e tests.

## Documentation

-

## Follow-up

-

### DIFF
--- a/e2e/test_operator/operator_test.go
+++ b/e2e/test_operator/operator_test.go
@@ -1758,7 +1758,7 @@ var _ = Describe("Operator", Label("e2e", "pr"), func() {
 						fdbCluster.GetCluster().Status.ProcessGroups,
 						processGroupID,
 					)
-				}).WithTimeout(5 * time.Minute).WithPolling(5 * time.Second).Should(BeNil())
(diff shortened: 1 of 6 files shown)


e2e/test_operator/operator_test.go
```diff
@@ -1758,7 +1758,7 @@ var _ = Describe("Operator", Label("e2e", "pr"), func() {
 						fdbCluster.GetCluster().Status.ProcessGroups,
 						processGroupID,
 					)
-				}).WithTimeout(5 * time.Minute).WithPolling(5 * time.Second).Should(BeNil())
+				}).WithTimeout(10 * time.Minute).WithPolling(5 * time.Second).Should(BeNil())
```
johscheuer, line 1761: I increased the wait time here. In cases where the global sync mode is enabled, the coordination takes a bit longer, so the actual replacement often takes a bit longer than 5 min (from the last 2 failed tests it's often around 6min)

PR #2326, opened 2025-07-15
Ensure that IP based environment variables are passed down as the correct argument type

johscheuer:
# Description

In cases where the Pod IP Family was set and the `FDB_PUBLIC_IP` was used as an environment variables, the operator created the wrong configuration, which could result in the cluster being stuck.

## Type of change

## Discussion

-

## Testing

Added unit tests and will do some additional manual testing.

## Documentation

-

## Follow-up

-

### DIFF
--- a/internal/monitor_conf.go
+++ b/internal/monitor_conf.go
@@ -350,7 +350,11 @@ func GetMonitorProcessConfiguration(
 
 		configuration.Arguments = append(configuration.Arguments, monitorapi.Argument{
 			ArgumentType: monitorapi.ConcatenateArgumentType,
-			Values:       generateMonitorArgumentFromCustomParameter(argument),
+			Values: generateMonitorArgumentFromCustomParameter(
+				argument,
+				cluster.GetPodIPFamily(),
+				imageType,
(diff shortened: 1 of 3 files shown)


internal/monitor_conf.go
```diff
@@ -414,11 +420,22 @@ func generateMonitorArgumentFromCustomParameter(
...
+		value := strings.Trim(knobValue, "$")
+		newArgument := monitorapi.Argument{
 			ArgumentType: monitorapi.EnvironmentArgumentType,
-			Source:       strings.Trim(knobValue, "$"),
+			Source:       value,
+		}
+
+		// If a user specifies the fdbv1beta2.EnvNamePodIP or fdbv1beta2.EnvNamePublicIP environment variables in the
+		// custom parameters, e.g. to add a new locality, we should pass this as the proper type.
+		if podIPFamily != fdbv1beta2.PodIPFamilyUnset && imageType == fdbv1beta2.ImageTypeUnified &&
```
nicmorales9, line 432: wouldn't imageType always be unified in this code?

PR #2328, opened 2025-07-17
Allow IPv4

qrpike:
# Description

Allow IPv4 binding. Pods will not start if the kernel has IPv6 disabled.


johscheuer:
Fixed by https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2340

PR #2329, opened 2025-07-18
Update the compatibility check for multiple newer versions

johscheuer:
# Description

Update the compatibility check for multiple newer versions

## Type of change

## Discussion

## Testing

-

## Documentation

Added in code.

## Follow-up

-

### DIFF
--- a/controllers/check_client_compatibility.go
+++ b/controllers/check_client_compatibility.go
@@ -154,6 +154,10 @@ func getUnsupportedClients(
 			continue
 		}
 
+		// If we are here, the client has a lower version than the targeted version, so the client doesn't support the
+		// targeted version and wouldn't be able to connect to the cluster. The loop collects additional information
+		// about the client, like the IP address and the log group to help a human operator to update the incompatible
+		// client.
 		for _, client := range versionInfo.MaxProtocolClients {
 			if _, ok := ignoredLogGroups[client.LogGroup]; ok {
 				continue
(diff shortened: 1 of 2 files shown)


controllers/check_client_compatibility.go
```diff
@@ -117,32 +116,45 @@ func (c checkClientCompatibility) reconcile(
...
+			return nil, err
+		}
+
+		// If the version is protocol compatible or newer than the targeted version, then the current client
+		// supports the targeted version.
+		if clientVersion.IsProtocolCompatible(version) || clientVersion.IsAtLeast(version) {
+			continue
+		}
+
+		for _, client := range versionInfo.MaxProtocolClients {
```
nicmorales9, line 157: It took me a bit so it might be nice to add a comment here noting that this loop is checking for clients that have a maximal version that is lower than the target version

nicmorales9 (approved):
lgtm

controllers/check_client_compatibility.go
johscheuer: I thought that would be explained by the comment above (the if statement), but I can add some additional comments if you think that comment is not sufficient.

nicmorales9: Somehow I was failing to make the connection between the loop which marks clients as unsupported and the comment above being more about "positive" case where the client is good. I think a comment like this would have helped me
```
// Now check for clients which are at a lower version than the target, and mark them as unsupported
```

PR #2338, opened 2025-07-31
Improve migration test case to reduce runtime

johscheuer:
# Description

Improve migration test case to reduce runtime, otherwise the migration will take a long  eventually timeout with the global sync mode.

## Type of change

## Discussion

I changed the migration test to only update log processes and not all processes, that should be good enough for what we are testing.

## Testing

Ran the test case manually.

## Documentation

-

## Follow-up

-


nicmorales9 (approved):
LGTM!

PR #2347, opened 2025-08-12
Modify Backup URL in Reconciler

millasml:
# Description

## Type of change

- Documentation
- Other

## Discussion

## Testing

Manual testing?*

## Documentation

## Follow-up


api/v1beta2/foundationdbbackup_types.go
```diff
@@ -322,6 +322,11 @@ func (backup *FoundationDBBackup) CheckReconciliation() (bool, error) {
 		reconciled = false
 	}
 
+	if isRunning && backup.BackupURL() != backup.Status.BackupDetails.URL {
```
johscheuer, line 325: I think it would be better to create a dedicated method (e.g. `NeedsConfigurationChange`) for this and combine the `SnapshotPeriodSeconds` and URL check. That way the method can be extended in the future without adding more if statements in CheckReconciliation.

pkg/fdbadminclient/admin_client.go
```diff
@@ -82,7 +82,7 @@ type AdminClient interface {
 	ResumeBackups() error
 
 	// ModifyBackup modifies the configuration of the backup.
-	ModifyBackup(int) error
+	ModifyBackup(snapshotPeriodSeconds int, url string) error
```
johscheuer, line 85: I think passing down a pointer to the `FoundationDBBackup` struct will be more flexible in the future and we need to break the interface only once.

johscheuer (changes requested):
One comment, otherwise the changes look good. I guess testing this in the e2e tests will be hard as we only have a single blobstore.

Issue #2349, opened 2025-08-14
Backup resource still defaults to the split image


hxu:
### What happened?

The [customization docs](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/docs/manual/customization.md#unified-vs-split-images) say that the default is the unified image, however the `FoundationDbBackup` resource [still defaults](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/db9e7671d4f44ca286925cf8db08707e95c9232e/api/v1beta2/foundationdbbackup_types.go#L341) to the split image.

Is it easy to update the default, or should we put a note in the docs? I'm happy to make a PR for this.

### What did you expect to happen?

Backup agents should use the unified image by default

### How can we reproduce it (as minimally and precisely as possible)?

On 7.3.69, create a backup resource with no imageType set

### FDB Kubernetes operator

2.5.0
FDB version 7.3.69

### Kubernetes version

n/a

### Cloud provider

n/a

johscheuer:
Great catch, do you want to create a PR to fix this?

PR #2351, opened 2025-08-14
Update backup agent default image type to unified

hxu:
# Description

See #2349

## Type of change

I think I would consider this a breaking change since customizations to the FoundationDbBackup resource will need to take into account the fact that the unified image has a slightly different pod spec.

## Discussion

## Testing

I ran unit tests, but am not yet sure how to get e2e tests to run.

## Documentation

## Follow-up


johscheuer (changes requested):
Could you adjust the failing test cases?

johscheuer (commented):
Just wanted to know if you have the time to finish the work?

internal/pod_models_test.go
```diff
@@ -3925,7 +3925,7 @@ var _ = Describe("pod_models", func() {
 				).To(Equal("foundationdb/foundationdb:dev"))
 				Expect(
 					deployment.Spec.Template.Spec.InitContainers[0].Image,
-				).To(Equal("foundationdb/foundationdb-kubernetes-sidecar:dev-1"))
+				).To(Equal("foundationdb/foundationdb:dev"))
```
hxu, line 3928: This test seemed a bit suspect to me -- shouldn't the init container be the fdb-kubernetes-monitor? The failure message indicated that it was actually the same image as the main container.

internal/pod_models_test.go
```diff
@@ -3963,22 +3963,22 @@ var _ = Describe("pod_models", func() {
 			})
 		})
 
-		When("using the unified image", func() {
+		When("using the split image", func() {
```
hxu, line 3966: I assumed that we still want to test the split image behavior, so test swaps with `pod_models when getting the backup deployment with a basic deployment the init container [It] should set the image and command for the container`

hxu:
Sorry! We got sidetracked with something else.

I've gone and fixed all the tests now.

One slightly unrelated question -- I noticed that the [build docker image](https://github.com/FoundationDB/fdb-kubernetes-operator/actions/runs/16969009117/job/48105965136?pr=2351) action failed, it seems like cross compilation of the arm64 image fails. We encountered the same issue in our build system and had to use an arm64 machine to do the compilation. Did you get cross compilation to work at any point?

internal/pod_models_test.go
johscheuer: In this test the image config is changed and the main image config is set to `foundationdb/foundationdb:dev`. Since the default image is now the unified image for the backup pods, the init image and the main image will be the same, in this case `foundationdb/foundationdb:dev` and the sidecar image config will be ignored.

johscheuer (commented):
> One slightly unrelated question -- I noticed that the [build docker image](https://github.com/FoundationDB/fdb-kubernetes-operator/actions/runs/16969009117/job/48105965136?pr=2351) action failed, it seems like cross compilation of the arm64 image fails. We encountered the same issue in our build system and had to use an arm64 machine to do the compilation. Did you get cross compilation to work at any point?

Most of the time the arm64 image builds fine but sometimes it is failing in the GitHub action worker, we haven't invested too much time debugging the issue as a retry usually fix the issue.

internal/pod_models_test.go
hxu: Got it. Even though this is valid from the operator's perspective, this would not actually work right, because the foundationdb image can't be used as the init container image?

johscheuer (approved):

johscheuer (commented):
> * Commit ID: [bfb1b37](https://github.com/FoundationDB/fdb-kubernetes-operator/commit/bfb1b379683ac47d6e44275a5ee6607d88b89eaa)
> * Duration 2:52:24
> * Result: ❌ **FAILED**
> * [Build Log](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator/fdb-kubernetes-operator-pr/444f18de-154b-4a7b-bcda-0f24f069981b.gz) terminal output (available for 30 days)
> * [Build Workspace](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator-pr/444f18de-154b-4a7b-bcda-0f24f069981b/fdb-kubernetes-operator-pr) zip file of the working directory (available for 30 days)

There was one unrelated failure.

Issue #2357, opened 2025-08-26
Backup Isn't Cleaned Up on Backup Resource Deletion


millasml:
### What happened?

When a backup resource is deleted, the backup agents are deleted as well but the fdbbackup status still reports as running. Combined with https://github.com/FoundationDB/fdb-kubernetes-operator/issues/2356, we can run into an issue where:

1. Backup Resource A is created. backup A is running
2. Backup Resource A is deleted. backup A still reports a running state through fdbbackup status, even if there are no backup agents
3. Backup Resource B is created.
4. The reconciler reports Backup Resource B to be running, but to the URL of what is specified in A
5. The system is in an inconsistent state where Backup Resource B is running the backup for A, instead of it's own.

### What did you expect to happen?

backup should be aborted on deletion of backup resource.

### How can we reproduce it (as minimally and precisely as possible)?

delete a backup resource while it is running.

### FDB Kubernetes operator

<details>

```console
$ kubectl fdb version
2.9.0
```

</details>

### Kubernetes version

<details>

```console
$ kubectl version
Client Version: v1.32.2
Kustomize Version: v5.5.0
Server Version: v1.31.1101```

</details>


### Cloud provider

<details>
aws
</details>

PR #2360, opened 2025-09-08
Add initial implementation of a finalizer for the backup cleanup

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/2357

## Type of change

## Discussion

This is the first implementation approach to cleanup backups when the FoundationDBBackup resource is deleted and the according deletion policy is set (default is to keep everything running).

## Testing

Ran an e2e test case with the policy.

## Documentation

Updated the documentation.

## Follow-up

-


controllers/backup_controller.go
```diff
@@ -231,3 +243,103 @@ func (r *FoundationDBBackupReconciler) updateOrApply(
 
 	return r.Status().Update(ctx, backup)
 }
+
+// updateOrApplySpec updates the spec either with server-side apply or if disabled with the normal update call.
```
nicmorales9, line 247: nit: I would leave out the server-side-apply part or put a (TODO)

controllers/backup_controller.go
```diff
@@ -231,3 +243,103 @@ func (r *FoundationDBBackupReconciler) updateOrApply(
...
+	if deletionPolicy == fdbv1beta2.BackupDeletionPolicyNoop {
+		if controllerutil.RemoveFinalizer(backup, fdbv1beta2.FoundationDBBackupFinalizerName) {
+			logger.V(1).Info("removing the finalizer for backup resource")
+			return r.updateOrApplySpec(ctx, backup)
+		}
+
+		return nil
+	}
+
+	if deletionPolicy == fdbv1beta2.BackupDeletionPolicyCleanup ||
```
nicmorales9, line 276: nit: might be nicer as a switch

docs/manual/backup.md
```diff
@@ -124,6 +124,36 @@ spec:
...
+This can be changed by setting the `spec.deletionPolicy`, valid options are:
+
+- `noop` (default): When set the backup state will not be changed. This can cause problems when the `FoundationDBBackup` resource is deleted and the cluster still receives write traffic, as the `backup_agents` are deleted and therefore the backup mutations will not be removed.
+- `stop`: Will [abort](https://apple.github.io/foundationdb/backups.html#abort) the backup when the `FoundationDBBackup` resource is deleted.
+- `cleanup`: Will [abort](https://apple.github.io/foundationdb/backups.html#abort) and [delete](https://apple.github.io/foundationdb/backups.html#delete) the backup and it's data in the blob store when the `FoundationDBBackup` resource is deleted.
+
+When the `stop` or the `cleanup` deletion policy is set the operator will add the `foundationdb.org/fdb-kubernetes-operator` finalizer.
+It is not recommended to remove this finalizer manually, as this could lead to an incomplete removal of data.
+In the current implementation the delete step is implemented in the operator, this step might block one fo routine for a longer duration (up to 10 minutes before the reconciliation is retried).
```
nicmorales9, line 138: What does `fo routine` mean here?

controllers/backup_controller.go
```diff
@@ -231,3 +243,103 @@ func (r *FoundationDBBackupReconciler) updateOrApply(
...
+				return r.updateOrApplySpec(ctx, backup)
+			}
+
+			return nil
+		}
+
+		logger.V(1).Info("deleting backup data")
+		// Delete the backup data from the blobstore, this will be a blocking operation. We eventually want to create
+		// a dedicated go routine for this to not block everything. Adding a go routine will make things a bit more
+		// complicated unless we get a good signal from the fdbbackup command if the data is deleted.
```
nicmorales9, line 324: Alternatively, could FDB backup binaries do that work of retrying? Possibly setting some system key in the database noting to retry if that process died?

controllers/backup_controller_test.go
```diff
@@ -226,7 +229,7 @@ var _ = Describe("backup_controller", func() {
 			})
```
nicmorales9, line 229: should we check that the finalizer is gone too?

nicmorales9 (approved):
some nits but LGTM

controllers/backup_controller.go
johscheuer: I took a look at this and I would keep the code as it is for now, we can refactor the code later if we think this makes it easier to read and maintain.

PR #2361, opened 2025-09-08
Move the update status into a single place

johscheuer:
# Description

Move the update status into a single place and update the status only once if required per reconciliation.

## Type of change

- Other

## Discussion

This makes it easier to write the sub-reconcilers and should reduce possible conflicts.

## Testing

Ran the operator test suite with the new changes.

## Documentation

-

## Follow-up

-


johscheuer (commented):
I still think this would/will be useful, but given that this PR is stale since September, I'm going to close it.

PR #2378, opened 2025-10-09
Improve the check for Pod spec changes

johscheuer:
# Description

In the current implementation we could hit issues if someone changes the pods outside the operator control, e.g. by manually changing the pod spec with `kubectl edit ..`. The risk is fairly minimal since the pod spec has a limited set of fields that can be changed, but in general the operator should validate those fields, e.g. in case someone changes the image of a running pod.

## Type of change

## Discussion

I discussed this offline and we probably want to add a new annotation to the operator which can be set on a per pod level to ignore pod changes. This could be useful in debugging cases where a user wants to keep the pod running, but wants to pull a container with additional debugging tooling. I create a GitHub issue for this before merging.

## Testing

Added additional unit and e2e tests.

## Documentation

-

## Follow-up

-

### DIFF
--- a/controllers/update_status.go
+++ b/controllers/update_status.go
@@ -1292,7 +1292,7 @@ func podSpecIsCorrect(ctx context.Context,
 
 	// Check if any of the mutable fields of the Pod have been changed outside the operators control, e.g.
 	// by using kubectl.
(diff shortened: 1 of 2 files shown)


johscheuer:
Closing for now.

PR #2405, opened 2025-12-15
Make use of hex string instead of FDB printable to remove test failures because of special characters

johscheuer:
# Description

We've seen a few cases in our nightly testing where the tests have failed because of a special character in the value, e.g. a semicolon (`;`) where the `fdbcli` command thought the current command is split. Using only the hex format is a better approach to prevent those failures.

## Type of change

## Discussion

-

## Testing

Ran the test manually.

## Documentation

-

## Follow-up

-

### DIFF
--- a/e2e/test_operator/operator_test.go
+++ b/e2e/test_operator/operator_test.go
@@ -1944,7 +1944,7 @@ var _ = Describe("Operator", Label("e2e", "pr"), func() {
 				cmd := fmt.Sprintf(
 					"writemode on; option on ACCESS_SYSTEM_KEYS; set %s %s",
 					fixtures.FdbPrintable([]byte(key)),
-					fixtures.FdbPrintable(timestampByteBuffer.Bytes()),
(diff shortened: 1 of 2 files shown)


e2e/test_operator/operator_test.go
```diff
@@ -1944,7 +1944,7 @@ var _ = Describe("Operator", Label("e2e", "pr"), func() {
 				cmd := fmt.Sprintf(
 					"writemode on; option on ACCESS_SYSTEM_KEYS; set %s %s",
 					fixtures.FdbPrintable([]byte(key)),
-					fixtures.FdbPrintable(timestampByteBuffer.Bytes()),
+					fixtures.FdbHexString(timestampByteBuffer.Bytes()),
```
nicmorales9: this makes sense to me to use hex, but would unix timestamps ever write ';'? Is it possible we are missing a spot we should swap the function for?

PR #2418, opened 2026-02-04
Update math/rand to math/rand/v2

johscheuer:
# Description

Forgot to push those changes, I worked on them last week when I was looking into something.

## Type of change

- Other

## Discussion

-

## Testing

Ran unit tests locally.

## Documentation

Nothing to update.

## Follow-up

-


nicmorales9 (approved):
lgtm

PR #2425, opened 2026-02-09
Allow unmanaged backups and add locality support

johscheuer:
# Description

Allow to create a `FoundationDBbackup` where the operator only manages the Deployment but not the backup itself. I also added changes from https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2422/

## Type of change

## Discussion

-

## Testing

Updated unit tests and e2e tests. 

## Documentation

Updated documentation in code.

## Follow-up

The zone locality should be configurable.

### DIFF
--- a/docs/backup_spec.md
+++ b/docs/backup_spec.md
@@ -54,17 +54,6 @@ BackupState defines the desired state of a backup
 
 BackupType defines the backup type that should be used for the backup.
 
-**Valid values:**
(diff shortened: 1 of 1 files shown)


saintstack (approved):
LGTM. Wouldn't mind a bit of high-level doc. on when unmanaged vs managed: i.e. when to use either. Thanks.

Issue #1987, opened 2024-04-11
Make use of net.URL package in getURL


johscheuer:
### What would you like to be added/changed?

In our backup code in `func (configuration *BlobStoreConfiguration) getURL(backup string, bucket string) string {` we currently put a raw string together. Instead we should consider to make use of the net.URL package: https://pkg.go.dev/net/url#URL to have stronger guarantees and checks.

PR #2426, opened 2026-02-11
Make use of url.URL for stricter parsing of the backup URL

johscheuer:
# Description

This change introduces stricter parsing of the `BackupURL`, I think its better to return a clear error instead of having the operator fail during a `fdbbackup`/`fdbrestore` command.

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1987

## Type of change

## Discussion

This touches some of the exposed methods of the `FoundationDBBackup` and therefore is a breaking change.

## Testing

Updated unit tests.

## Documentation

Nothing to update.

## Follow-up

-

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v4
         with:
-          go-version: 1.25.5
+          go-version: 1.25.7
       - name: Get dependencies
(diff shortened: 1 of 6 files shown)


nicmorales9 (approved):
seems good to me but am I correct that this is a non-backwards compatible change that prevents use of hostname not in @ format? I think the error should be clear enough but worth checking / noting if it is potentially breaking something

PR #2441, opened 2026-03-20
adding support for modify command

### LINKED ISSUE #2393: Encryption Breaks if modifyBackup sub reconciler runs
### What happened?

The underlying issue is really in https://github.com/apple/foundationdb/issues/12544, but having a modifyBackup sub reconciler that may run in each reconcile loop makes it impossible to guarantee that any backups started with encryption will have encryption throughout the lifetime of the backup resource.

I have staged https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2392 to minimize the impact, but the issue still persists if there are any updates to snapshot seconds. the modifyBackup sub reconciler should be removed all together until the underlying foundationDB bug is fixed.

any modification to the backup resource does not break encryption, as long as the encryption key path is specified in the spec.

change the snapshot seconds of a backup resource that has encryption key path specified.

_No response_

### FDB Kubernetes operator

<details>

```console
$ kubectl fdb version
foundationdb-operator: 2.14.0
kubectl-fdb: latest

```

</details>

### Kubernetes version

<details>

```console
$ kubectl version
Client Version: v1.32.2
Kustomize Version: v5.5.0
Server Version: v1.32.801

```

</details>

### Cloud provider

<details>
aws
</details>

09harsh:
# Description

Added a test to run the modify command on the encrypted backup using backup_modify reconciler.

Fixes: #2393

## Type of change

- 
## Discussion

No

## Testing

Manual testing?*

Ran the test locally. 

## Documentation

N/A

## Follow-up

N/A


johscheuer (commented):
Instead of running the `modify` command manually I would prefer to change the `FoundationDBBackup.Spec`, so that the operator performs the `modify` call.

api/v1beta2/foundationdbcluster_types_test.go
```diff
@@ -831,13 +831,15 @@ var _ = Describe("[api] FoundationDBCluster", func() {
...
 			Expect(err).NotTo(HaveOccurred())
+			UID := "3874300eea1e154e4079530b381f71c3"
 			Expect(status).To(Equal(FoundationDBLiveBackupStatus{
 				DestinationURL:          "blobstore://minio@minio-service:9000/sample-cluster?bucket=fdb-backups",
 				SnapshotIntervalSeconds: 864000,
 				Status: FoundationDBLiveBackupStatusState{
 					Running: true,
 				},
 				Restorable: ptr.To(false),
+				UID:        &UID,
```
johscheuer:
```suggestion
				UID:        ptr.To("3874300eea1e154e4079530b381f71c3"),
```

controllers/modify_backup.go
```diff
@@ -45,13 +45,6 @@ func (s modifyBackup) reconcile(
 		return nil
 	}
 
-	// Modifying backups will remove encryption from previously encrypted backups.
-	// To preserve encryption, skip backup modifications if encryption is enabled.
-	// See FoundationDB issue: https://github.com/apple/foundationdb/issues/12544
-	if backup.Spec.EncryptionKeyPath != "" {
```
johscheuer, line 51: I don't think we should remove this completely, there are still versions out where the bug exists and we should rather check here if the FDB version has the bug fix, in which case we should allow the modification. If the FDB version doesn't support the modify command with backup encryption we should return an error with a message that tells the user that this feature is only available in versions X and newer.

e2e/fixtures/fdb_backup.go
```diff
@@ -273,38 +273,57 @@ func (fdbBackup *FdbBackup) Pause() {
 	fdbBackup.setState(fdbv1beta2.BackupStatePaused)
 }
 
-// RunDescribeCommand runs the describe command on the backup pod.
-func (fdbBackup *FdbBackup) RunDescribeCommand() *fdbv1beta2.FDBBackupDescribe {
-	backupURL, err := fdbBackup.backup.BackupURL()
-	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+// SetSnapshotInterval updates the snapshot interval of the current backup
```
johscheuer:
```suggestion
// SetSnapshotInterval updates the snapshot interval of the current backup.
```

e2e/fixtures/fdb_backup.go
```diff
@@ -273,38 +273,57 @@ func (fdbBackup *FdbBackup) Pause() {
...
+func (fdbBackup *FdbBackup) SetSnapshotInterval(snapshotPeriodSeconds int) {
+	objectKey := client.ObjectKeyFromObject(fdbBackup.backup)
+	foundationDBBackup := &fdbv1beta2.FoundationDBBackup{}
+	gomega.Expect(fdbBackup.fdbCluster.factory.GetControllerRuntimeClient().
+		Get(context.Background(), objectKey, foundationDBBackup)).NotTo(gomega.HaveOccurred())
```
johscheuer:
```suggestion
		Get(context.Background(), objectKey, foundationDBBackup)).To(gomega.Succeed())
```

I think this way the expectations are more precise.

e2e/fixtures/fdb_backup.go
```diff
@@ -273,38 +273,57 @@ func (fdbBackup *FdbBackup) Pause() {
...
+
+	foundationDBBackup.Spec.SnapshotPeriodSeconds = &snapshotPeriodSeconds
+	gomega.Expect(fdbBackup.fdbCluster.factory.GetControllerRuntimeClient().
+		Update(context.Background(), foundationDBBackup)).NotTo(gomega.HaveOccurred())
```
johscheuer:
```suggestion
		Update(context.Background(), foundationDBBackup)).To(gomega.Succeed())
```

Same as above.

e2e/fixtures/fdb_backup.go
```diff
@@ -273,38 +273,57 @@ func (fdbBackup *FdbBackup) Pause() {
...
+	fdbBackup.backup = foundationDBBackup
+	fdbBackup.WaitForReconciliation()
+}
+
+// RunCommandOnBackupPod runs command on the backup pod.
```
johscheuer:
```suggestion
// RunCommandOnBackupPod runs the provided command on a randomly chosen backup pod.
```

e2e/fixtures/fdb_backup.go
```diff
@@ -273,38 +273,57 @@ func (fdbBackup *FdbBackup) Pause() {
...
 		command,
 		false)
 	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+	return out
+}
 
+// RunDescribeCommand runs the describe command on the backup pod.
+func (fdbBackup *FdbBackup) RunDescribeCommand() *fdbv1beta2.FDBBackupDescribe {
+	backupURL, err := fdbBackup.backup.BackupURL()
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
```
johscheuer:
```suggestion
	gomega.Expect(err).To(gomega.Succeed())
```

e2e/fixtures/fdb_backup.go
```diff
@@ -273,38 +273,57 @@ func (fdbBackup *FdbBackup) Pause() {
...
 	out, _, err := fdbBackup.fdbCluster.ExecuteCmdOnPod(
 		*backupPod,
 		fdbv1beta2.MainContainerName,
 		command,
 		false)
 	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+	return out
+}
 
+// RunDescribeCommand runs the describe command on the backup pod.
```
johscheuer:
```suggestion
// RunDescribeCommand runs the describe command on a randomly chosen backup pod.
```

e2e/fixtures/fdb_backup.go
```diff
@@ -273,38 +273,57 @@ func (fdbBackup *FdbBackup) Pause() {
...
+	backupURL, err := fdbBackup.backup.BackupURL()
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+	command := fmt.Sprintf("fdbbackup describe -d \"%s\" --json", backupURL)
+	out := fdbBackup.RunCommandOnBackupPod(command)
 	desc := &fdbv1beta2.FDBBackupDescribe{}
 	gomega.Expect(json.Unmarshal([]byte(out), desc)).To(gomega.Succeed())
 	return desc
 }
 
+// RunStatusCommand runs the status command on the backup pod.
```
johscheuer:
```suggestion
// RunStatusCommand runs the status command on a randomly chosen backup pod.
```

e2e/fixtures/fdb_backup.go
```diff
@@ -273,38 +273,57 @@ func (fdbBackup *FdbBackup) Pause() {
...
 	gomega.Expect(err).NotTo(gomega.HaveOccurred())
 	command := fmt.Sprintf("fdbbackup list -b \"%s\"", backupURL)
-	backupPod := fdbBackup.GetBackupPod()
-	out, _, err := fdbBackup.fdbCluster.ExecuteCmdOnPod(
-		*backupPod,
-		fdbv1beta2.MainContainerName,
-		command,
-		false)
-	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+	out := fdbBackup.RunCommandOnBackupPod(command)
```
johscheuer:
```suggestion
	return strings.Split(fdbBackup.RunCommandOnBackupPod(command), "\\n")
```

e2e/test_operator_backups/operator_backup_test.go
```diff
@@ -103,7 +103,7 @@ var _ = Describe("Operator Backup", Label("e2e", "pr", "foundationdb-nightly"),
 		})
 
 		When("the default backup system is used", func() {
-			var useRestorableVersion, skipRestore bool
+			var useRestorableVersion, skipRestore, shouldPauseBackup bool
```
johscheuer, line 110: I think we should add a description above `shouldPauseBackup`, given that the "false" case means we stop backups, this could be a bit confusing.

e2e/test_operator_backups/operator_backup_test.go
```diff
@@ -213,18 +217,38 @@ var _ = Describe("Operator Backup", Label("e2e", "pr", "foundationdb-nightly"),
...
+							// restart the backup before modifying since the backup is paused
+							modifiedSnapshotPeriod := 900000
+							backup.Start()
+							backup.SetSnapshotInterval(modifiedSnapshotPeriod)
+							// validating snapshot interval changed and the backup UID is same
+							statusCommandOutput = backup.RunStatusCommand()
+							Expect(
+								statusCommandOutput.SnapshotIntervalSeconds,
+							).To(Equal(modifiedSnapshotPeriod))
+							Expect(*statusCommandOutput.UID).To(Equal(backupUID))
```
johscheuer, line 249: Shouldn't we first check that `statusCommandOutput.UID` is not nil? Otherwise we could hit a nil-pointer exception.

johscheuer (changes requested):

controllers/modify_backup.go
```diff
@@ -45,11 +46,19 @@ func (s modifyBackup) reconcile(
...
-	// To preserve encryption, skip backup modifications if encryption is enabled.
+	// Modifying backups with encryption is only supported on versions that support backup encryption.
 	// See FoundationDB issue: https://github.com/apple/foundationdb/issues/12544
-	if backup.Spec.EncryptionKeyPath != "" {
-		return nil
+	fdbVersion, err := fdbv1beta2.ParseFdbVersion(backup.Spec.Version)
+	if err != nil {
+		return &requeue{curError: err}
+	}
+	if backup.Spec.EncryptionKeyPath != "" && !fdbVersion.SupportsBackupEncryption() {
```
johscheuer: We should be using `backup.GetEncryptionKey()` here, that will only return the encryption key if backups are supported. The main ask is to forbid the `modify` command for FDB versions that support backup encryptions but miss the modify command fix. See: https://github.com/apple/foundationdb/issues/12544

johscheuer (changes requested):

controllers/modify_backup.go
```diff
@@ -45,11 +46,19 @@ func (s modifyBackup) reconcile(
...
-	// To preserve encryption, skip backup modifications if encryption is enabled.
+	// Modifying backups with encryption is only supported on versions that support backup encryption.
 	// See FoundationDB issue: https://github.com/apple/foundationdb/issues/12544
-	if backup.Spec.EncryptionKeyPath != "" {
-		return nil
+	encryptionKeyPath, err := backup.GetEncryptionKey()
+	if err != nil {
+		return &requeue{curError: err}
+	}
+	if encryptionKeyPath != "" {
```
johscheuer, line 55: It seems like we are missing the version check here? This will return an error for all FDB versions that support encrypted backups. I thought that the most recent FDB versions that support encrypted backups also have the fix for the modify command?

johscheuer (changes requested):

controllers/modify_backup.go
09harsh: Bumped up `SupportsBackupEncryption73` to the latest version that supports encrypted backup. I will made a forum post, once the PR is merged.

controllers/modify_backup.go
```diff
@@ -45,11 +46,19 @@ func (s modifyBackup) reconcile(
 		return nil
 	}
 
-	// Modifying backups will remove encryption from previously encrypted backups.
-	// To preserve encryption, skip backup modifications if encryption is enabled.
+	// Modifying backups with encryption is only supported on versions that support backup encryption.
 	// See FoundationDB issue: https://github.com/apple/foundationdb/issues/12544
-	if backup.Spec.EncryptionKeyPath != "" {
-		return nil
+	encryptionKeyPath, err := backup.GetEncryptionKey()
```
johscheuer: Why do we need this if statement here if the `modify` command is now working as expected with encrypted backups?

johscheuer (changes requested):

johscheuer (approved):
LGTM 👍

PR #2446, opened 2026-03-24
Add ApplyConfiguration setup for FDB CRDs

johscheuer:
# Description

Add ApplyConfiguration setup for FDB CRDs. This allows clients to use the typed `ApplyConfiguration` to perform the `Apply`. I converted a few methods in the `kubectl-fdb`  package to show how it works. I would convert more/all in another PR since this one is already quite big.

## Type of change

- Other

## Discussion

-

## Testing

Ran a manual test with the e2e tests and performed manual testing with the `kubectl-fdb` plugin.

## Documentation

If we think it's worth we can document the `ApplyConfiguration`. But those behave similar as the upstream `ApplyConfiguration`.

## Follow-up

-


kubectl-fdb/cmd/buggify_crash_loop.go
```diff
@@ -211,15 +214,15 @@ func cleanCrashLoopContainerList(
...
+
+	return kubeClient.Apply(
+		ctx,
+		applyfdbv1beta2.FoundationDBCluster(clusterName, namespace).
+			WithSpec(applyfdbv1beta2.FoundationDBClusterSpec().
+				WithBuggify(applyfdbv1beta2.BuggifyConfig().
+					WithCrashLoopContainers(&applyfdbv1beta2.CrashLoopContainerObjectApplyConfiguration{}),
+				),
+			),
+		client.FieldOwner("kubectl-fdb"),
```
nicmorales9, line 226: nit: might be nice to make this a constant

nicmorales9 (approved):
just the nit, LGTM

PR #2447, opened 2026-03-24
Fix bug with leaked fdb connections

hsson:
# Description

This change fixes a resource leak issue caused by using short-lived FoundationDB connections without properly closing them when done. The previously used version of the Go bindings (from the 7.1.67 release) did not expose any way to close and destory the underlaying FoundationDB resources created from the C library. Newer versions of the bindings introduces a `Close()` that can be used to properly release resources.

The PR upgrades the Go bindings to the version from release 7.3.63, and makes sure that all connections created from `fdb.OpenDatabase(...)` are properly closed when no longer needed.

This required a small refactor of the lock client, such that it uses short-lived database connections that are created for each method call, instead of keeping a long-lived connection over the lifetime of the lock client. This approach creates a consistent behavior with how the fdb and admin clients are implemented (which also utilize short-lived connections), and makes sure that the connection can be closed safely without pushing this responsibility out to the user of the lock client.

We discovered this bug (i.e. the resource and connection leakage) in our production cluster. We only noticed it beacuase we use quite short lived mTLS certificates (shorter than the default `TLS_CERT_REFRESH_DELAY_SECONDS` in FoundationDB of 12 hours). We started seeing TLS issues in the trace logs, originating from the operator, which ultimately lead us to find the `fdb.OpenDatabase` without `Close()` in the operator. This was not a problem for operations e.g. directly calling `fdbcli`, since we had configured the `--knob-tls-cert-refresh-delay-seconds` knob in `customParameters`. 

## Type of change

## Discussion

- N/a

## Testing

Did run unit tests, but unit tests are mostly useless for this type of fix. Hopefully e2e tests in CI is better. We also use the 7.3.63 bindings with this pattern (i.e. `Close()` of the db connection) heavily, and have substantial tests on our side. Feel confident about this change.

## Documentation

- N/a

## Follow-up

As described above, we ultimately discovered this bug because of TLS issues. If implementation is ever changed in the future to have long-lived connections opened by `fdb.OpenDatabase`, it would be good if the TLS refersh knob (or any other custom parameters) could be configured. Seems like low prio right now though, given that all opened connections are now short-lived.

### DIFF
--- a/fdbclient/admin_client.go
+++ b/fdbclient/admin_client.go
@@ -494,7 +494,6 @@ func (client *cliAdminClient) ExcludeProcessesWithNoWait(
 
 			return nil
 		})
-
 		if err != nil {
 			return err
 		}
@@ -708,6 +707,7 @@ func (client *cliAdminClient) killWithManagementAPI(addresses []fdbv1beta2.Proce
 	if err != nil {
 		return err
 	}
+	defer db.Close()
 
 	return db.RebootWorker(fdbv1beta2.ProcessAddressesStringWithoutFlags(addresses, ","), false, 0)
 }
@@ -767,7 +767,6 @@ func (client *cliAdminClient) ChangeCoordinators(
 			)
 			return nil
 		})
-
 		if err != nil {
 			return "", err
 		}
@@ -964,7 +963,6 @@ func (client *cliAdminClient) GetBackupStatus() (*fdbv1beta2.FoundationDBLiveBac
 			"--json",
 		},
 	})
-
 	if err != nil {
 		return nil, err
 	}
@@ -1118,7 +1116,6 @@ func (client *cliAdminClient) GetProcessesUnderMaintenance() (map[fdbv1beta2.Pro
 		}
 		return nil
 	})
-
 	if err != nil {
 		return nil, err
 	}
--- a/fdbclient/common.go
+++ b/fdbclient/common.go
@@ -101,7 +101,8 @@ func parseMachineReadableStatus(
 	return status, nil
 }
 
-// getFDBDatabase opens an FDB database.
+// getFDBDatabase opens an FDB database. Caller should make sure to close the
+// database with [fdb.Database.Close] when done with it.
 func getFDBDatabase(cluster *fdbv1beta2.FoundationDBCluster) (fdb.Database, error) {
 	clusterFile, err := ensureClusterFileIsPresent(
 		path.Join(os.TempDir(), string(cluster.UID)),
--- a/fdbclient/fdb_client.go
+++ b/fdbclient/fdb_client.go
@@ -376,6 +376,7 @@ func (fdbClient *realFdbLibClient) executeTransaction(
 	if err != nil {
 		return err
 	}
+	defer db.Close()
 
 	_, err = db.Transact(func(tr fdb.Transaction) (interface{}, error) {
 		err = setCommonOptions(&tr, timeout)
@@ -404,6 +405,7 @@ func (fdbClient *realFdbLibClient) executeTransactionForManagementAPI(
 	if err != nil {
 		return err
 	}
+	defer db.Close()
 
 	_, err = db.Transact(func(tr fdb.Transaction) (interface{}, error) {
 		err = setCommonOptions(&tr, timeout)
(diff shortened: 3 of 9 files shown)


hsson:
I saw the changes from https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2444 btw, and perhaps the lock client (in `lock_client.go`) also deserves a similar refactor to use an `executeTransaction` helper? Could include it in this PR if it's of interest.

johscheuer (changes requested):
First thanks for the PR! Have you tried setting `TLS_CERT_REFRESH_DELAY_SECONDS` to an appropriate value for your use case in the operator deployment? This should be already possible by setting `FDB_NETWORK_OPTION_TLS_CERT_REFRESH_DELAY_SECONDS`.
What you are seeing is not a resource leak but rather a design decision to keep the connection open to the FDB clusters, since the operator uses the `FoundationDBCluster` UID from the Kubernetes resource, the cluster file path is always the same and the go bindings will cache the DB (https://github.com/apple/foundationdb/blob/release-7.1/bindings/go/src/fdb/fdb.go#L283-L290). We decided to keep the connection open, so that the bindings can update the connection string in the cluster file if the coordinators are changed outside of the operators scope, e.g. in cases where you run multiple operators. Closing the database connection directly can have some side effects that we have to validate first, e.g. https://github.com/apple/foundationdb/blob/main/bindings/go/src/fdb/database.go#L118-L120.
Since the operator supports FDB versions from 7.1+, we have to use the bindings for 7.1 and we should only update the bindings to 7.3 once we remove the support for 7.1 

I hope all of this makes sense.

hsson:
Thanks for taking the time to review!

I was not aware that it was possible to configure the refresh delay using environment variables. I'm aware of the `FDB_NETWORK_OPTION_<option>` method, but from I can tell from [the list of options](https://github.com/apple/foundationdb/blob/main/fdbclient/vexillographer/fdb.options) the TLS refresh delay is not in there. My understanding is that the TLS refresh delay is a knob (that can be set with `--knob-tls-cert-refresh-delay-seconds`), but I'm not sure if it is possible to set with an environment variable? I'm happy to be proven wrong, but it would be nice to understand _why_ that works then 😄 

Doesn't hurt to try of course, so I'll try setting `FDB_NETWORK_OPTION_TLS_CERT_REFRESH_DELAY_SECONDS` in the meantime, despite not knowing how it would work — I trust you!

Also, I was not aware of the caching of the connections, thanks for pointing that out and showing me! That also makes me wonder why our operator works at all (perhaps it doesn't work as well as it should), considering that cached database connection will be using an expired TLS certificate for a lot of the time...

And makes sense to not close the connection after the `RebootWorker` as you pointed out, missed that. Thanks a lot for the careful review.

Yet another thing that surprises me: I thought thta the bindings were backwards compatbile as long as the correct version was configured with e.g. `fdb.APIVersion(730)`.  Is this not the case? 🤔 Where is this documented?

I guess I/we can close this PR, since it does not really make sense given what you've pointed out. Again, thanks for taking the time to review. I would appreciate a lot if you could also help me understand these follow-up questions that I have.

johscheuer:
> I was not aware that it was possible to configure the refresh delay using environment variables. I'm aware of the FDB_NETWORK_OPTION_<option> method, but from I can tell from [the list of options](https://github.com/apple/foundationdb/blob/main/fdbclient/vexillographer/fdb.options) the TLS refresh delay is not in there. My understanding is that the TLS refresh delay is a knob (that can be set with --knob-tls-cert-refresh-delay-seconds), but I'm not sure if it is possible to set with an environment variable? I'm happy to be proven wrong, but it would be nice to understand why that works then 😄

I just assumed that this knob is part of the network option, I should have verified that 🙈 . In that case `FDB_NETWORK_OPTION_KNOB="tls_cert_refresh_delay_seconds=300"` should work. If things work, feel free to create a PR that documents how to set this knob.

hsson:
Ah great, I missed that you could also set knobs using environment variables! I'll try that, thanks 🙂

hsson:
Can confirm that setting the `FDB_NETWORK_OPTION_KNOB="tls_cert_refresh_delay_seconds=<value>"` environment variable on our operator has solved our issues! Thanks a lot for this 🙂 I'll see if I can find any appropriate place to put this in the docs.

hsson:
Here is the documentation update: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2449

PR #2448, opened 2026-03-25
Update Lock client to make use of FDB lib client

johscheuer:
# Description

Align the FDB usage over the lock client and the fdb client and make sure they use the same fdb lib client. This adds the same default knobs to the lock client (setting the timeout and setting the max retries).

## Type of change

- Other

## Discussion

In a follow up we could add additional testing and we should revisit the timeout handling, e.g. how we use the `DefaultTimeout` and the `MaxTimeout`. Right now the usage is quite different from what we documented.

I refactored the timeout usage, as it made sense for me to do that in this PR.

## Testing

Ran unit tests, CI will run e2e tests.

## Documentation

Nothing to update.

## Follow-up

-

### DIFF
--- a/fdbclient/fdb_client.go
+++ b/fdbclient/fdb_client.go
@@ -42,34 +42,35 @@ type fdbLibClient interface {
 	updateGlobalCoordinationKeys(
 		prefix string,
 		updates map[fdbv1beta2.ProcessGroupID]fdbv1beta2.UpdateAction,
-		timeout time.Duration,
 	) error
 	// getGlobalCoordinationKeys will return the entries under the provided prefix.
 	getGlobalCoordinationKeys(
 		prefix string,
-		timeout time.Duration,
 	) (map[fdbv1beta2.ProcessGroupID]time.Time, error)
 	// clearGlobalCoordinationKeys will remove all the ready information for the provided prefix.
-	clearGlobalCoordinationKeys(prefix string, timeout time.Duration) error
+	clearGlobalCoordinationKeys(prefix string) error
 	// updateProcessAddresses will update the provided addresses associated with the provided process group. If the address slice
 	// is empty or nil, the entry will be removed.
 	updateProcessAddresses(
 		updates map[fdbv1beta2.ProcessGroupID][]string,
-		timeout time.Duration,
 	) error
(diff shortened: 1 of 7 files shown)


fdbclient/fdb_client.go
```diff
@@ -42,34 +42,35 @@ type fdbLibClient interface {
...
 	executeTransactionForManagementAPI(
 		operation func(tr fdb.Transaction) error,
-		timeout time.Duration,
 	) error
 	// executeTransaction will run a transaction for the target cluster. This method will handle all the common options.
-	executeTransaction(operation func(tr fdb.Transaction) error, timeout time.Duration) error
+	executeTransaction(operation func(tr fdb.Transaction) error) error
+	// executeTransactionWithCustomTimeout will run a transaction for the target cluster. This method will add all the common options
+	// and allows to use a custom timeout
+	executeTransactionWithCustomTimeout(
```
nicmorales9: seems slightly clutter-y to have a separate interface function for this - is there any strong reason to not just allow this function to override max duration? Do we ever even go past max timeout?

johscheuer: I updated the code to get rid of this method.

fdbclient/fdb_client.go
```diff
@@ -36,8 +36,6 @@ import (
 
 // fdbLibClient is an interface to interact with FDB over the client libraries
 type fdbLibClient interface {
-	// getValueFromDBUsingKey returns the value of the provided key.
-	getValueFromDBUsingKey(fdbKey string, timeout time.Duration) ([]byte, error)
```
nicmorales9, line 40: Is there any particular reason to get rid of this? It does seem useful to have (or at least a helper for if not part of the interface)

nicmorales9 (approved):
LGTM

fdbclient/fdb_client.go
johscheuer: I removed it to make the usage of the `fdbLibClient` a bit more consistent. We used this method only in the getStatus and getConnectionString methods. I think using the `executeTransaction` is better and allows a bit more flexibility, e.g. the getStatus method can set the transaction timeout, because this is the only transaction with an increased timeout.

PR #2449, opened 2026-03-26
Document FDB_NETWORK_OPTION_KNOB for certificate refresh intervals

hsson:
# Description

Follow-up from the discussion in: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2447

Add documentation explaining how to configure the `tls_cert_refresh_delay_seconds` knob via the `FDB_NETWORK_OPTION_KNOB` environment variable in the operator deployment. Configuring the TLS refresh interval is neccesary when using TLS certificates that are short lived (current default refresh interval in FoundationDB is 12 hours).

## Type of change

- Documentation

## Discussion

- N/a

## Testing

Have tested that the configuration change described in the updated documentation has the desired effect.

## Documentation

- N/a

## Follow-up

- N/a

### DIFF
--- a/docs/manual/tls.md
+++ b/docs/manual/tls.md
@@ -99,6 +99,15 @@ You must always ensure that the peer verification rules allow access from the cl
 
 If you want to run any clusters with TLS, you must configure the operator to support FDB's mutual TLS. This requires setting the same environment variables that you set on the fdbserver processes: `FDB_TLS_CERTIFICATE_FILE`, `FDB_TLS_KEY_FILE`, and `FDB_TLS_CA_FILE`. You will probably want to use the same certificate mechanism that you use for the FDB certs for the operator certs as well. This certificate configuration will only be used for connections to FoundationDB and to the sidecar process.
 
+You can configure how often the operator refreshes TLS certificates by setting the `FDB_NETWORK_OPTION_KNOB` environment variable with the `tls_cert_refresh_delay_seconds` option. This controls the delay between certificate refresh checks in seconds. If not set, FoundationDB uses its default refresh interval.
+
+```yaml
+# In your operator deployment
+env:
+  - name: FDB_NETWORK_OPTION_KNOB
+    value: "tls_cert_refresh_delay_seconds=3600"
+```
+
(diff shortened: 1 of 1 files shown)


docs/manual/tls.md
```diff
@@ -99,6 +99,15 @@ You must always ensure that the peer verification rules allow access from the cl
 
 If you want to run any clusters with TLS, you must configure the operator to support FDB's mutual TLS. This requires setting the same environment variables that you set on the fdbserver processes: `FDB_TLS_CERTIFICATE_FILE`, `FDB_TLS_KEY_FILE`, and `FDB_TLS_CA_FILE`. You will probably want to use the same certificate mechanism that you use for the FDB certs for the operator certs as well. This certificate configuration will only be used for connections to FoundationDB and to the sidecar process.
 
+You can configure how often the operator refreshes TLS certificates by setting the `FDB_NETWORK_OPTION_KNOB` environment variable with the `tls_cert_refresh_delay_seconds` option. This controls the delay between certificate refresh checks in seconds. If not set, FoundationDB uses its default refresh interval.
```
johscheuer, line 102: Could we add a link to the documentation here and mention that the default is 12h?

johscheuer (commented):
Thanks for the PR, one comment otherwise the change looks good.

docs/manual/tls.md
hsson: I think part of the problem is that this isn't clearly documented, at least not as far as I'm aware. I only found out about it after you pointed it out [in the other PR](https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2447#issuecomment-4132971344). Had to dig through FoundationDB source code to verify. Same regarding the default 12h.

johscheuer: I think it's good enough to add here that the default is `12h` so others don't have to look in the source code.

hsson: Fair enough, added it now! 🙂

johscheuer (approved):
LGTM 👍

hsson:
> LGTM 👍

Thanks for the review! A bit unclear why the CI job fails though, is this common? Or a known problem? How to resolve?

https://github.com/FoundationDB/fdb-kubernetes-operator/actions/runs/24960148935/job/73344959888?pr=2449

PR #2450, opened 2026-03-26
return an error from GetMonitorProcessConfiguration

nicmorales9:
# Description

If you put a junk version this error will not be surfaced and logs may not be visible. This should resolve that problem

## Type of change

## Discussion

None in particular

## Testing

Manual testing?*
Unit tests

No

## Documentation

N/A

No

N/A

N/A

## Follow-up

Don't think so

No


johscheuer (approved):
LGTM

PR #2452, opened 2026-03-31
Allow user to overwrite fdb-trace-logs volume mount

nocetellaxai:
This extends PR #2417 (which added addVolumeIfMissing for volumes) to also cover volume mounts. Previously, even if a user defined a custom volume mount for fdb-trace-logs (e.g. with subPathExpr to create per-pod subdirectories on a hostPath), the operator would unconditionally append a second mount with the same name, overriding the user's configuration.

The new addVolumeMountIfMissing function checks whether a volume mount with the given name already exists before appending, preserving any user-defined mount options like subPathExpr.

# Description

## Type of change

- Documentation
- Other

## Discussion

## Testing

Manual testing?*

## Documentation

## Follow-up

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -592,13 +592,16 @@ func (r *FoundationDBClusterReconciler) updatePodDynamicConf(
 	var expectedConf string
 	imageType := internal.GetImageType(pod)
 	if imageType == fdbv1beta2.ImageTypeUnified {
-		config := internal.GetMonitorProcessConfiguration(
+		config, err := internal.GetMonitorProcessConfiguration(
 			cluster,
 			processClass,
 			serversPerPod,
 			imageType,
 			currentPodIPFamily,
 		)
+		if err != nil {
+			return false, err
+		}
 		configData, err := json.Marshal(config)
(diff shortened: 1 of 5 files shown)


internal/pod_models.go
```diff
@@ -241,8 +241,8 @@ func configureContainersForUnifiedImages(
 		corev1.VolumeMount{Name: "data", MountPath: "/var/fdb/data"},
 		corev1.VolumeMount{Name: "config-map", MountPath: "/var/dynamic-conf"},
 		corev1.VolumeMount{Name: "shared-binaries", MountPath: "/var/fdb/shared-binaries"},
-		corev1.VolumeMount{Name: "fdb-trace-logs", MountPath: "/var/log/fdb-trace-logs"},
 	)
+	addVolumeMountIfMissing(mainContainer, corev1.VolumeMount{Name: "fdb-trace-logs", MountPath: "/var/log/fdb-trace-logs"})
```
ArniDagur: For volumes, we don't special case `fdb-trace-logs`. I don't see a reason to do so for volume mounts. It'd be cleaner for the overwrites to be consistent.

PR #2456, opened 2026-04-16
Improve the data loader and allow to read data

johscheuer:
# Description

Improve the data loader and allow to read data.

## Type of change

## Discussion

The changes are made in a way that the original behaviour is kept.

## Testing

Ran a manual test with the newly build data loader.

## Documentation

Nothing to update.

## Follow-up

-

### DIFF
--- a/e2e/fixtures/fdb_data_loader.go
+++ b/e2e/fixtures/fdb_data_loader.go
@@ -29,11 +29,10 @@ import (
 	"text/template"
 	"time"
 
-	k8serrors "k8s.io/apimachinery/pkg/api/errors"
-
 	"github.com/onsi/gomega"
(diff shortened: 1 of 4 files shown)


e2e/fixtures/fdb_data_loader.go
```diff
@@ -299,43 +305,78 @@ type dataLoaderConfig struct {
 	// SecretName represents the Kubernetes secret that contains the certificates for communicating with the FoundationDB
 	// cluster.
 	SecretName string
+	// DataLoaderArguments defines the arguments that should be added to the
```
nicmorales9:
```suggestion
	// DataLoaderArguments defines the arguments that should be passed to the DataLoader
```

nicmorales9 (approved):
LGTM

PR #2458, opened 2026-04-17
Delete pods in a terminal state

ArniDagur:
Moves the special case pod deletion logic for `NodeAffinity` stuck pods into its own reconciler, and makes the logic more general.

### Motivation

My cluster was down because all the pods simultaneously breached the ephemeral storage quota. This caused the pods to be evicted (state=`Terminated`, reason=`Evicted`), and the database became unavailable.

The solution is similar to the pre-existing handling of `NodeAffinity` pods, which was shoved into the `update_status` reconciler. I moved the logic into its own reconciler so it would run even when the database is down (`update_status` terminates early in this case).

### Testing

I reproduced the evicted pod issue on my cluster, and verified that updating the operator to the version in this PR fixes the issue.

```
{"level":"info","ts":"2026-04-17T03:16:01Z","logger":"controller","msg":"Deleting pod that is stuck in a terminal state","namespace":"fdb-mycluster","cluster":"fdb-mycluster","traceID":"6ce19e6a-0f12-494f-837d-4133ee193465","reconciler":"controllers.deleteTerminalPods","processGroupID":"mydc-storage-85997","phase":"Failed","reason":"Evicted"}
{"level":"info","ts":"2026-04-17T03:16:01Z","logger":"controller","msg":"Deleting pod that is stuck in a terminal state","namespace":"fdb-mycluster","cluster":"fdb-mycluster","traceID":"6ce19e6a-0f12-494f-837d-4133ee193465","reconciler":"controllers.deleteTerminalPods","processGroupID":"mydc-storage-86758","phase":"Failed","reason":"Evicted"}
```


controllers/delete_terminal_pods.go
```diff
@@ -0,0 +1,103 @@
+/*
+ * delete_failed_pods.go
```
johscheuer:
```suggestion
 * delete_terminal_pods.go
```

controllers/update_status.go
```diff
@@ -883,24 +883,8 @@ func validateProcessGroup(
...
-		if pod.Status.Reason == "NodeAffinity" &&
-			pod.CreationTimestamp.Add(5*time.Minute).Before(time.Now()) {
-			logger.Info("Delete Pod that is stuck in NodeAffinity",
-				"processGroupID", processGroupStatus.ProcessGroupID)
-
-			err = r.PodLifecycleManager.DeletePod(logr.NewContext(ctx, logger), r, pod)
-			if err != nil {
-				return err
-			}
-		}
```
johscheuer, line 903: I think we can drop the whole if statement and change the `UpdateCondition` to `processGroupStatus.UpdateCondition(fdbv1beta2.PodFailing, pod.Status.Phase == corev1.PodFailed)`

johscheuer (commented):
I'll start an e2e test, thanks for the changes. The code looks fine to me.

controllers/update_status.go
ArniDagur: I don't think that would be right, because we also have this for loop which sets `failed`
```
	for _, container := range pod.Status.ContainerStatuses {
		if !container.Ready {
			failing = true
			break
		}
	}
```
That said, I have pushed a commit which simplifies the logic

ArniDagur:
I think the fix in this PR is a good step forward, but there's still an issue where operator will crash due to
```
Error determining public address.
Error determining public address.
SIGSEGV: segmentation violation
PC=0x7fe8be0ed6fa m=81 sigcode=1 addr=0x40
signal arrived during cgo execution
```
My cluster is running `7.3` and uses DNS in its cluster files.

The issue for this is https://github.com/apple/foundationdb/issues/11222, which mentions

> We observed that the fdb libraries are crashing when no pod is running and we created a PR with a fix on the operator side: https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2283

Maybe we should actually extend this fix by running the `delete_terminal_pods` at startup as well, if the clusterfile is using DNS. When I manually delete all the pods via `xargs`, operator stops crashing as expected.

johscheuer:
> I think the fix in this PR is a good step forward, but there's still an issue where operator will crash due to
> 
> ```
> Error determining public address.
> Error determining public address.
> SIGSEGV: segmentation violation
> PC=0x7fe8be0ed6fa m=81 sigcode=1 addr=0x40
> signal arrived during cgo execution
> ```
> 
> My cluster is running `7.3` and uses DNS in its cluster files.
> 
> The issue for this is [apple/foundationdb#11222](https://github.com/apple/foundationdb/issues/11222), which mentions
> 
> > We observed that the fdb libraries are crashing when no pod is running and we created a PR with a fix on the operator side: #2283
> 
> Maybe we should actually extend this fix by running the `delete_terminal_pods` at startup as well, if the clusterfile is using DNS. When I manually delete all the pods via `xargs`, operator stops crashing as expected.

In your case the pods are stuck in the terminated state, correct? We already have some special logic in the operator for DNS, e.g. where the operator checks if pods must be created: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/controllers/cluster_controller.go#L249-L258.

controllers/delete_terminal_pods.go
```diff
@@ -0,0 +1,116 @@
...
+
+		var minimumAge time.Duration
+		if phase == corev1.PodFailed && reason == "NodeAffinity" {
+			// The desired node is not available. Give it some time to be brought
+			// up rather than recreating the pod multiple times.
+			// See https://github.com/kubernetes/kubernetes/issues/92067
+			minimumAge = 5 * time.Minute
+		} else {
+			// In case of a "crash loop", this wait serves to slow down pod recreation.
+			minimumAge = 30 * time.Second
```
johscheuer: 30 seconds might be a tight time window to debug stuff (since this includes the scheduling time). I think it would be useful to have a command line argument for the operator to set this minimum age. The default could set a high minimum age, so that terminal pods wouldn't be deleted but logged, that would keep the current bevahiour and if someone wants to enable this feature they could they the minimum age to a low time duration like the proposed 30 seconds.

ArniDagur:
> 30 seconds might be a tight time window to debug stuff (since this includes the scheduling time). I think it would be useful to have a command line argument for the operator to set this minimum age. The default could set a high minimum age, so that terminal pods wouldn't be deleted but logged, that would keep the current bevahiour and if someone wants to enable this feature they could they the minimum age to a low time duration like the proposed 30 seconds.

I can set the default to something like 3 minutes. I don't think it should be more than that since the user expectation is that FDB operator does the "right thing" by default, meaning the operator should perform remediation actions without special configuration.

> In your case the pods are stuck in the terminated state, correct? We already have some special logic in the operator for DNS, e.g. where the operator checks if pods must be created: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/controllers/cluster_controller.go#L249-L258.

Yes. I was suggesting that as a follow up we could add the `delete_terminal_pods` reconciler under that if statement as well, before running `add_pods`

ArniDagur:
I also soft disagree with making it configurable at all, since extra config options add complexity. I don't think debugging is such a strong justification for delaying actions that move the cluster towards a healthier state, given we expect people to have logs and so on that can be looked at retroactively. In the worst case they can just scale the operator for that cluster to zero replicas.

ArniDagur:
I made the requested CLI argument change

controllers/update_status.go
```diff
@@ -875,34 +875,14 @@ func validateProcessGroup(
 		return nil
 	}
 
-	failing := false
+	failing := pod.Status.Phase == corev1.PodFailed
 	for _, container := range pod.Status.ContainerStatuses {
```
johscheuer, line 879: (Only a comment) We could put this for loop into an if statement and only iterate over all container statuses if `failing` is `false`. Seems like a small improvement, given that most of the time `failing` should be `false`.

setup/setup.go
```diff
@@ -355,6 +356,12 @@ func (o *Options) BindFlags(fs *flag.FlagSet) {
 		1*time.Minute,
 		"the minimum uptime for the cluster before configuration changes are allowed.",
 	)
+	fs.DurationVar(
+		&o.MinimumAgeForTerminalPodDeletion,
+		"minimum-age-for-terminal-pod-deletion",
+		3*time.Minute,
```
johscheuer: I think we can keep the proposed 30 seconds and if a user wants/needs to increase that time, they could be doing that.

johscheuer (approved):
Thanks, changes LGTM. I'll start a new e2e test run.

controllers/update_status.go
ArniDagur: Given that most pods will only have a low single digit number of containers, it would be a very minor improvement.

ArniDagur:
@johscheuer Do you think the CI failure is legitimate?

johscheuer (approved):
> @johscheuer Do you think the CI failure is legitimate?

I looked at the failure, seems unrelated.

PR #2463, opened 2026-04-22
Print data loader result before waiting for signal and make use of json trace format

johscheuer:
# Description

Print data loader result before waiting for signal and make use of json trace format

## Type of change

## Discussion

-

## Testing

Locally build the sample app, CI will run e2e tests.

## Documentation

Nothing to update.

## Follow-up

Nothing to update.


nicmorales9 (approved):
✨

PR #2464, opened 2026-04-23
Add GitHub action for stale PRs and issues

johscheuer:
# Description

Add GitHub action for stale PRs and issues

## Type of change

- Other

## Discussion

-

## Testing

No testing.

## Documentation

I think the action is enough.

## Follow-up

-

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,27 @@
+name: 'Close stale pull requests'
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


nicmorales9 (approved):
seems good as long as the version is pinned and we've vetted that one

johscheuer (commented):
> seems good as long as the version is pinned and we've vetted that one

I used the same version as the FDB repo uses, so should be fine.

PR #2470, opened 2026-04-29
Make use of Get in lock client instead of MustGet

johscheuer:
# Description

`Get()` will return an error if the asynchronous operation does not complete successfully, `MustGet()` would panic. I haven't seen any cases there we hit the panic, but I think using `Get()` is a better fit here.

## Type of change

## Discussion

-

## Testing

Ran unit tests + CI will run e2e tests.

## Documentation

-

## Follow-up

-


nicmorales9 (approved):
no panics at this disco!

PR #2471, opened 2026-04-29
Make use of ptr.Deref

johscheuer:
# Description

This PR makes use  `ptr.Deref` instead of doing the same work, we already use `ptr.Deref` in other places, so makes sense to use it consistently.

## Type of change

- Other

## Discussion

-

## Testing

Ran unit tests locally.

## Documentation

Nothing to update.

## Follow-up

-


nicmorales9 (approved):
🧽 ✨

PR #2475, opened 2026-04-30
Support Maintenance Mode for ProcessGroup DeletionMode

nocetellaxai:
# Description

Currently, when maintenance mode is enabled, we only set maintenance when performing deletions of storage pods during upgrades for the `Zone` deletion mode. 

Maintenance mode is useful during upgrades when we want to prevent unnecessary shuffling of data. When storage pods are deleted, there can be some lag before they are recreated and re-registered with the DataDistributor. By default, if they are unavailable for more than 60s, DataDistributor will begin rereplicating. This can happen regardless of the deletion mode; whether it's multiple pods from the same zone (`Zone` deletion mode), or individual pods (`ProcessGroup`) being recreated -- we want to support Maintenance Mode in both cases.

## Type of change
New feature / Bug Fix

## Discussion

## Testing
Manually tested on live cluster.

## Documentation

Yes

## Follow-up
We decided to move to `ProcessGroup` deletion mode (from `Zone`) because of a race condition in FoundationDB (more effort to fix and deploy right now). Suppose we have storage pods A and B in the same zone. They are recreated at the same time.

Before: 
A has ip 1.2.3
B has ip 4.5.6

After:
B is recreated, gets ip 7.8.9
A is recreated, gets ip 4.5.6

This can be a problem. A registers with DataDistibutor and an `interfaceChanged` event happens (new interface 4.5.6). It then runs a check to see if any other storage servers have the same interface. It can then observe that B's interface is also 4.5.6 (not updated yet by B's storage tracker), and, seeing that two servers have the same interface, mark one of them as "Undesired" -- causing the DataDistributor to being shuffling data.

We observed this happening reliably for several zones on each upgrade for a cluster of about 60 ss and 30 unique zones.


johscheuer (approved):
Changes LGTM 👍 Thanks for the PR. The issue that you're describing in the Follow-up section should like an FDB core issue, since the operator itself doesn't have control over the IP assignment in the Kubernetes cluster. One way to potentially workaround this might be to use service IP addresses if you are not already using them. It would be great if you could open an issue in the FDB repo.

nocetellaxai:
Yep ip issue is with FDB, will do, thanks!

Could you merge this please? thank you!

PR #2477, opened 2026-05-07
Support FDB knobs via CLI flags

nocetellaxai:
# Description

Add support for passing fdb knobs to operator via flags.

## Type of change

## Testing

Manual testing

## Documentation

Cli flags are self-documenting via --help

## Follow-up
n/a

### DIFF
--- a/main.go
+++ b/main.go
@@ -56,16 +56,6 @@ func main() {
 	logOpts.BindFlags(flag.CommandLine)
 	flag.Parse()
 
-	if len(operatorOpts.Knobs) > 0 {
-		fmt.Printf("Setting FDB client knobs: %v\n", operatorOpts.Knobs)
-		for _, knob := range operatorOpts.Knobs {
-			if err := fdb.Options().SetKnob(knob); err != nil {
-				fmt.Fprintf(os.Stderr, "failed to set FDB knob %q: %s\n", knob, err.Error())
-				os.Exit(1)
-			}
-		}
-	}
-
 	mgr, file := setup.StartManager(
 		scheme,
 		operatorOpts,
--- a/main.go
+++ b/main.go
@@ -17,7 +17,6 @@ package main
 
 import (
 	"flag"
-	"fmt"
 	"os"
 
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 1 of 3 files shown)


nocetellaxai:
test failed but looks unrelated, flakey?

[FAIL] update_status when refreshing process group status when a PVC with an unknown process group ID exists [It] should add the process group from the PVC

main.go
```diff
@@ -55,6 +56,16 @@ func main() {
 	logOpts.BindFlags(flag.CommandLine)
 	flag.Parse()
 
+	if len(operatorOpts.Knobs) > 0 {
```
johscheuer: I would prefer if we could move this into `StartManager`, so that the logger would already be Initialized. Might be worth to add a comment here why the knobs must be set here.

setup/setup.go
```diff
@@ -362,6 +372,11 @@ func (o *Options) BindFlags(fs *flag.FlagSet) {
 		30*time.Second,
 		"The minimum age of a terminal pod before it will be deleted.",
 	)
+	fs.Var(
+		&o.Knobs,
+		"knob",
+		"FDB client knob in name=value format (repeatable). e.g. --knob coordinator_reconnection_delay=2.0",
```
johscheuer: Probably worth to note that the knobs will only be used for the go bindings but not `fdbcli` and the other fdb binaries.

johscheuer (changes requested):
Changes in general look fine, just a few recommendations. I'll kick off the e2e tests later/tomorrow.

johscheuer (commented):
@nocetellaxai ping :) Just wanted to know if you are able to finish the changes?

nocetellaxai:
Sorry for the delay, fixed up now, thanks for the comments.

johscheuer (commented):
The changes look fine. Have you tried to set those knobs with `FDB_NETWORK_OPTION_KNOB`? That should (haven't tried it) work for the bindings and fdbcli (and other fdb binaries).

nocetellaxai:
that works as well, thanks, maybe not needed then! ty

johscheuer:
> that works as well, thanks, maybe not needed then! ty

Great! I'll document the environment variable later or if you have some time, feel free to create a PR with the documentation changes :)

johscheuer:
I created https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2498

PR #2480, opened 2026-05-18
Mark process as missing if important process information like commandline and version are missing

johscheuer:
# Description

In some cases there could be a case where a process is part of the machine-readable status (status json) but misses some important fields like `version` or `commandline`:

```text
  So a process appears in the status processes list (the CC knows about it) but has no version or command_line when:

  1. The worker is temporarily unreachable — network hiccup, packet loss, or the worker is just coming up and its network endpoint isn't ready yet.
  2. The worker's network thread is blocked — a disk I/O stall or CPU spike on the worker can prevent it from processing the eventLogRequest within 2 seconds even though it appears alive to the CC (which uses a slower failure detector). This is the most common production cause.
  3. The process just restarted — there's a brief window between when the CC registers the new process (after the old one died) and when the new process has fully initialised its network stack and can respond to requests. ProgramStart is logged very early but the endpoint may not be ready.
```

In this case we should assume the process is unreachable and mark it as missing. If the `MissingProcesses` is present for long enough the process group will be replaced. I think this is better than trying to restart the process (because in the current logic we compare an empty string for command line and version. against the expected version/command line which will always cause adding the `IncorrectCommandLine` condition and then the process will eventually be restarted.

## Type of change

## Discussion

-

## Testing

Added unit tests.

## Documentation

Nothing to update.

## Follow-up

-

### DIFF
--- a/controllers/update_status.go
+++ b/controllers/update_status.go
@@ -460,6 +460,29 @@ func checkAndSetProcessStatus(
 				continue
 			}
 
+			// If the process.CommandLine or process.Version fields are empty something is wrong with the process, so
+			// we assume the process is "missing". If the process is long enough in this state the operator will
+			// replace the faulty process.
+			if process.CommandLine == "" {
+				hasMissingProcesses = true
+				logger.Info(
+					"found process with missing commandline information",
+					"processGroupID",
+					processGroupStatus.ProcessGroupID,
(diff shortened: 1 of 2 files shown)


nicmorales9 (approved):
SGTM

PR #2481, opened 2026-05-22
Prioritize degraded process groups for removal during cluster shrinks

Ronitsabhaya75:
# Description

When we shrink a cluster (scale down process counts), the operator previously didn't differentiate between healthy and degraded/failing nodes. This meant a healthy storage node could be marked for termination while a failing or degraded node was kept in the cluster.

This PR updates `chooseRemovals.reconcile` to check for active degraded conditions on each process group using `NeedsReplacementWithConditions(...)`. If a process group is degraded, we temporarily assign it a low locality priority (`-100`). Because `ChooseDistributedProcesses` uses `sortLocalities` (preferring higher-priority nodes to remain in the cluster), degraded nodes are naturally sorted to the end and selected first for removal.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Discussion

Mutating the priority on the copy of the `locality.Info` struct retrieved from `localityMap` is entirely side-effect free and integrates cleanly with the existing locality sorting and selection logic.

## Testing

I added a Ginkgo unit test context (`"with a degraded process group"`) inside `controllers/choose_removals_test.go`:
* It sets up 4 storage process groups, marks one as degraded with a `PodFailing` condition, and shrinks the desired storage count to 3.
* The test asserts that the operator specifically selects the degraded process group for removal instead of a healthy one.

Code is formatted with `go fmt`.

## Documentation

No documentation updates are needed. This is an internal optimization to the shrink reconciler and doesn't change any public-facing CRDs, defaults, or APIs.

## Follow-up

No immediate follow-ups. Since this doesn't add any new configurable defaults, we don't need to re-evaluate default behaviors in the future.

### DIFF
--- a/controllers/choose_removals.go
+++ b/controllers/choose_removals.go
@@ -97,8 +97,9 @@ func (c chooseRemovals) reconcile(
 		excessCount := currentCounts[processClass] - desiredCount
 		processClassLocality := make([]locality.Info, 0, currentCounts[processClass])
 
-		// TODO (johscheuer): We could add a higher priority to the process groups that have a condition that requires
-		// an automatic replacement.
+		// Prioritize degraded process groups for removal. If a process group is degraded,
+		// we set its priority to a low value (-100) so that healthy process groups
+		// are prioritized to remain, and degraded ones are naturally chosen for removal.
 		for _, processGroup := range cluster.Status.ProcessGroupsByProcessClass(processClass) {
 			if processGroup.IsMarkedForRemoval() {
 				excessCount--
@@ -106,6 +107,14 @@ func (c chooseRemovals) reconcile(
 			}
 			localityInfo, present := localityMap[string(processGroup.ProcessGroupID)]
 			if present {
+				_, failureTime := processGroup.NeedsReplacementWithConditions(
+					cluster.GetFailureDetectionTimeSeconds(),
+					cluster.GetTaintReplacementTimeSeconds(),
+					cluster.GetConditionsThatNeedReplacement(),
+				)
+				if failureTime > 0 {
+					localityInfo.Priority = -100
+				}
 				processClassLocality = append(processClassLocality, localityInfo)
 			}
 		}
--- a/controllers/choose_removals_test.go
+++ b/controllers/choose_removals_test.go
@@ -121,6 +121,35 @@ var _ = Describe("choose_removals", func() {
 			})
 		})
 
+		Context("with a degraded process group", func() {
+			var degradedProcessGroupID fdbv1beta2.ProcessGroupID
+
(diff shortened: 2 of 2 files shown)


Ronitsabhaya75:
@johscheuer could you please review this

controllers/choose_removals.go
```diff
@@ -97,15 +97,24 @@ func (c chooseRemovals) reconcile(
...
 			}
 			localityInfo, present := localityMap[string(processGroup.ProcessGroupID)]
 			if present {
+				_, failureTime := processGroup.NeedsReplacementWithConditions(
+					cluster.GetFailureDetectionTimeSeconds(),
+					cluster.GetTaintReplacementTimeSeconds(),
+					cluster.GetConditionsThatNeedReplacement(),
+				)
+				if failureTime > 0 {
+					localityInfo.Priority = -100
```
johscheuer, line 116: Is there a reason `-100` was picked and not another value?

johscheuer (commented):
Could you please run `make fmt lint` to ensure the files are formatted correctly?

controllers/choose_removals.go
Ronitsabhaya75: @johscheuer I used -100 here mostly as an arbitrary negative number. Since healthy processes default to 0 or 1, any negative value makes sure the degraded ones get sorted to the bottom.

johscheuer: Would be nice if we could add that to the comment above (that health processes default to `0` or `1`).

johscheuer (commented):
Could you run `make fmt lint`? Otherwise the changes look fine from my end.

controllers/choose_removals.go
johscheuer: Could we move the `-100` into a constant and call it e.g. `degradedProcessPriority`.

Just a note here: if a process would be truly failed and doesn't report to the FDB cluster anymore, e.g. the pod is partitioned or stuck in a pending state, it wouldn't be part of the `localityMap` and therefore we wouldn't add the `-100` priority. So this mostly catches cases where the process group should be replaced and the process is still reporting to the FDB cluster.

Ronitsabhaya75:
> Could you run `make fmt lint`? Otherwise the changes look fine from my end.

yeah I'll change code with comments. Thank you so much for reviews :)

johscheuer (commented):
Ping @Ronitsabhaya75 just wanted to ask if you are able to work on this?

Ronitsabhaya75:
> Ping @Ronitsabhaya75 just wanted to ask if you are able to work on this?

I'm @johscheuer just busy with my class tests for summer class sorry :(

PR #2484, opened 2026-06-02
Add input validation for the configure command

johscheuer:
# Description

Add input validation for the configure command to prevent running additional fdbcli commands and reducing the risk of malformed input.

## Type of change

## Discussion

-

## Testing

Added a unit test and CI will run e2e tests.

## Documentation

Nothing to update.

## Follow-up

-

### DIFF
--- a/fdbclient/admin_client_test.go
+++ b/fdbclient/admin_client_test.go
@@ -1325,6 +1325,106 @@ protocol fdb00b071010000`,
 			).To(BeNumerically(">", 0), "coordinator changes counter should be incremented after a coordinator change")
 		})
 	})
+
+	When("changing the configuration", func() {
+		var cliClient *cliAdminClient
+		var mockRunner *mockCommandRunner
+		var newDatabase bool
+		var configuration *fdbv1beta2.DatabaseConfiguration
+		var configurationErr error
+
+		BeforeEach(func() {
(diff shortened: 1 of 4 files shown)


fdbclient/admin_client_test.go
```diff
@@ -1325,6 +1325,106 @@ protocol fdb00b071010000`,
...
+				BeforeEach(func() {
+					newDatabase = true
+					configuration = &fdbv1beta2.DatabaseConfiguration{
+						PerpetualStorageWiggleLocality: ptr.To("; status details\n"),
+					}
+				})
+
+				It("should not configure the database and return an error", func() {
+					Expect(configurationErr).To(HaveOccurred())
+					Expect(mockRunner.receivedArgs).To(BeEmpty())
```
nicmorales9, line 1422: the test seems extremely similar to the ones above, seems like it should also maybe have a check that the existing database is still fine?

nicmorales9 (approved):
nits on the value of the tests, but change seems good

fdbclient/admin_client_test.go
johscheuer: There is no existing database (everything is mocked) :) And yes it's the same as above, just to verify that it doesn't touch the already present database.

Issue #2451, opened 2026-03-31
Adding pods can result in topology skew even if you apply a topologySpreadConstraint


hxu:
### What would you like to be added/changed?

We've encountered a couple of instances where a three-data-hall configuration with a topologySpreadConstraint with maxSkew of 1 in our spec sometimes results in an actual skew that is greater, and the operator is unable to recruit enough coordinators in each zone.

This only seems to happen when the pod spec has changed and the operator is creating new pods and excluding old pods from the cluster.

We think this is because the operator adds pods too quickly and the k8s scheduler sometimes schedules incorrectly because it's view of the pods is eventually consistent.

Another hypothesis is that the excluded pods are interfering with the skew calculation, so the pods are overall balanced, but the non excluded pods are unbalanced (sort of like what is described here: https://github.com/kubernetes/kubernetes/issues/105661). But we are less convinced by this hypotheses because we've seen the total pod count violate the maxSkew too.

Could we modify the operator such that it can check that the pods its spinning up are distributed correctly?

Brainstorming a few ideas and looking at the add_pods.go code:

 - We could add some jitter to how quickly it adds pods, so that we don't try to spin all of them up at once and reduce the chances of the scheduling race condition
 - We could have a check to ensure that there is at most X pending pods before trying to schedule more pods, where X is the replication (e.g. three for three data hall, two for double, etc.). But this would cause this reconciler to requeue for an extended period of time if pods are pending for a long time
 - We could have an after-the-fact rebalancer that checks the distribution if the operator can't recruit enough coordinators or hits other error cases that are due to skew.

hxu:
@johscheuer we're looking at this issue again as it's still occurring and occasionally gets us into a bad spot if there's a lot of replacements happening while there is skew.

Would you be open to a contribution here?  A few things that we had tried and some ideas:

* We tried applying a spec hash for each "generation" to our pod specs, and then using that in our topology constraint. However, we found that the operator actually updates the labels on old pods. Here's a PR to exclude any metadata used in the topology constraint: https://github.com/Donouyi/fdb-kubernetes-operator/pull/1
* Could the operator inject a pod affinity for the zones into the pod spec in three_data_hall mode? e.g. it would calculate that if you have 9 storage pods, and you have 3 zones, then it would give three pods an affinity for each zone.

PR #2485, opened 2026-06-02
Add an opt-in pod-template-generation label for TopologySpreadConstraints.matchLabelKeys

Donouyi:
# Description

This PR adds an opt-in feature that has the operator stamp a per-generation content hash on every pod it creates, under the label key `foundationdb.org/pod-template-generation`. The value is the first 16 hex characters of a SHA-256 over the canonical rendered `PodSpec` for the pod's process class (`GetPodSpec` invoked with a sentinel ProcessGroupID per class, so all real pods of the same class produce the same hash). Anything that flows into pod construction automatically contributes to the hash.

The label is meant to be referenced from `TopologySpreadConstraints.matchLabelKeys` on the user's pod template, so the kube-scheduler can scope per-pod spread filtering to "pods built from the same generation" instead of comparing against the entire cluster.

The PR also extends `podMetadataCorrect` with a small carve-out so the label is preserved on running pods and only rotates when a pod is actually recreated — mirroring the existing carve-out for the `LastSpecKey` annotation.

Emission is gated by a new field `LabelConfig.IncludePodTemplateGenerationLabel` (default false). Existing clusters are unaffected unless they opt in.

Motivation
  
`TopologySpreadConstraints.matchLabelKeys` lets a pod's spread filter be scoped to other pods that share specific label values with it. The intended use is rolling replacements: while old-generation pods are being replaced, the scheduler should be allowed to schedule the new-generation pods independently — otherwise the existing-pods inflate the topological zone count and lead to skew across zones

Replaces https://github.com/Donouyi/fdb-kubernetes-operator/pull/1 as a more complete solution. Fixes issue [here](https://github.com/FoundationDB/fdb-kubernetes-operator/issues/2451)

## Type of change

- Documentation
- Other

## Discussion

## Testing

Manual testing?*

## Documentation

## Follow-up


internal/pod_helper.go
```diff
@@ -105,6 +105,38 @@ func GetPodSpecHash(
 	return GetJSONHash(spec)
 }
 
+// PodGenerationHashLength is the number of hex characters returned by
+// GetPodGenerationHash — fits under Kubernetes' 63-char label value limit
+// while giving 64 bits of entropy.
+const PodGenerationHashLength = 16
```
johscheuer, line 113: Is there a specific reason we choose `16` instead of `63`? What is the benefit of cutting it to 16 chars?

johscheuer (commented):
Hi 👋 Thanks for the PR, it's on my list for this week.

internal/pod_helper.go
Donouyi: It gives us more readability.

internal/pod_models.go
```diff
@@ -140,6 +141,29 @@ func GetPod(
 	metadata.Name = processGroup.GetPodName(cluster)
 	metadata.OwnerReferences = owner
 
+	// The generation label is only stamped at pod-creation time. Reconcile
+	// passes preserve the existing value via the carve-out in
+	// podMetadataCorrect, so there's no need to recompute it on every
+	// PodMetadataCorrect call.
+	if cluster.ShouldIncludePodTemplateGenerationLabel() {
+		hash, hashErr := GetPodGenerationHash(cluster, processGroup.ProcessClass)
+		if hashErr != nil {
```
johscheuer: Is there a reason we are not returning the error? It could be hard to debug if the operator only logs the error in cases where it was not able to compute the hash.

internal/pod_models.go
```diff
@@ -140,6 +141,29 @@ func GetPod(
...
+			// A missing label degrades TSC matchLabelKeys to a broader
+			// spread, which the cluster tolerates — log and continue
+			// rather than fail pod construction.
+			logf.Log.WithName("internal").
```
johscheuer: So far our pattern was to pass down the logger as a parameter for the method. I believe using logo.Log will remove all the additional context that we added to the logger.

internal/pod_models.go
```diff
@@ -140,6 +141,29 @@ func GetPod(
 	metadata.Name = processGroup.GetPodName(cluster)
 	metadata.OwnerReferences = owner
 
+	// The generation label is only stamped at pod-creation time. Reconcile
+	// passes preserve the existing value via the carve-out in
+	// podMetadataCorrect, so there's no need to recompute it on every
+	// PodMetadataCorrect call.
+	if cluster.ShouldIncludePodTemplateGenerationLabel() {
+		hash, hashErr := GetPodGenerationHash(cluster, processGroup.ProcessClass)
```
johscheuer: I think this should be fine, but we will compute the same hash multiple times if we are adding multiple pods for the same process class, e.g. if we replace multiple pods, during the cluster creation or when we scale up the cluster.

internal/pod_helper.go
johscheuer: I'm not sure if 16 chars of a hash is ore readable then 63 chars, but I'm fine with 16 chars.

johscheuer (commented):
In general I think those changes look fine a few general comments:

- It would be great if you could add some additional documentation for this feature in our `docs` section, e.g. about the use case and how to use it (with an example).
- Right now, based on the way hoe it is implemented it will not have am affect if a user enabled this setting until a pod will be deleted for another unrelated reason. This could be confusing and could lead to a period where some pods have the new label and some others are missing it.
- We have an e2e test suite, do you think you are able to add a new e2e test for this new feature?

Sorry for the delay and thanks again for creating the PR!

internal/pod_models.go
Donouyi: yes, I think given this path only runs in `addPods` and never in steady-state reconciliation it's quite cheap to keep it as is.

internal/pod_models.go
Donouyi: you are right, changed the behaviour.

Donouyi:
@johscheuer, in the last commit, I've added the implementation of a marked-for-removal sentinel that we found useful to protect against the case of a pod that is (re)created while its process group is already marked for removal. Without it, the pod would be stamped with the current live generation hash so the scheduler counts a doomed pod as part of the new generation and can skew its spread. Because topology spread is only evaluated at schedule time, that skew isn't corrected once the doomed pod is deleted.

Also added a few log fields. They made the behaviour much easier to trace while testing but happy to drop them if you consider them noise.

PR #2487, opened 2026-06-03
Add feature to restrict user input for pod templates

johscheuer:
# Description

Add feature to restrict user input for pod templates. This can be useful in multi-tenant setups or in a setup where a user has limited privileges. The feature is opt-in and per default everything is allowed.

## Type of change

## Discussion

## Testing

## Documentation

Updated documentation.

## Follow-up

-


johscheuer (commented):
Failure was related to the go version (1.26.3) but 1.26.4 was released.

api/v1beta2/allowed_pod_modifications.go
```diff
@@ -0,0 +1,234 @@
...
+		}
+
+		// Unknown type
+		return ""
+	}
+}
+
+// AllowedPodModifications defines the pod modification that are allowed for the user provided configuration. If a field
+// is unset in AllowedPodModifications, we allow everything to not break the current setups. In a new major release we could
+// change this and enforce that fields are only allowed to change if the according AllowedPodModifications is set.
```
nicmorales9:
```suggestion
// change this and enforce that fields are only allowed to change if the corresponding AllowedPodModifications is set.
```

api/v1beta2/allowed_pod_modifications.go
```diff
@@ -0,0 +1,234 @@
...
+			}
+		}
+
+		// Unknown type
+		return ""
+	}
+}
+
+// AllowedPodModifications defines the pod modification that are allowed for the user provided configuration. If a field
+// is unset in AllowedPodModifications, we allow everything to not break the current setups. In a new major release we could
```
nicmorales9:
```suggestion
// is unset in AllowedPodModifications, we allow everything to avoid breaking current setups. In a new major release we could
```

nicmorales9: nit

api/v1beta2/allowed_pod_modifications.go
```diff
@@ -0,0 +1,234 @@
...
+		}
+
+		// If allowedPodModifications.AllowAutomountServiceAccountToken is not set but podSpec.AutomountServiceAccountToken
+		// is set, we don't perform any validation.
+	}
+
+	// Check if the HostNetwork setting is set and if this is actually allowed.
+	if podSpec.HostNetwork {
+		if allowedPodModifications.AllowHostNetwork != nil {
+			if !ptr.Deref(allowedPodModifications.AllowHostNetwork, false) {
```
nicmorales9: isn't this defaulting to _not_ allowing changes when unset? (i.e. changing default and maybe breaking existing setups)

docs/manual/pod_template_restrictions.md
```diff
@@ -0,0 +1,108 @@
...
+
+This document describes how operator administrators can limit which pod template fields CR authors are permitted to set.
+
+## Background
+
+The `podTemplate` field in `FoundationDBCluster` and `FoundationDBBackup` resources lets users customize the pods the operator creates.
+Without restrictions, any user who can write these resources can instruct the operator to:
+
+* Mount arbitrary Kubernetes Secrets as volumes, potentially exfiltrating credentials the operator has access to.
+* Set `spec.serviceAccountName` to the operator's own service account, effectively granting that user the operator's RBAC permissions.
```
nicmorales9, line 11: isn't this the default / not problematic except in combination with the other issues?

nicmorales9 (commented):
initial look before I get distracted

api/v1beta2/allowed_pod_modifications.go
johscheuer: The `prt.Deref` is only called if `allowedPodModifications.AllowHostNetwork` is not `nil`, so the default value will never be used in this case. But I agree changing to to `true` would simplify the setup and would probably be inline with the documentation of those settings.

docs/manual/pod_template_restrictions.md
johscheuer: This is not the default. Per `default` the FDB container will use the `default` service account and the operator would use the `fdb-kubernetes-operator-controller-manager` service account (assuming a user used the sample deployment or a variation of it https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/config/samples/deployment.yaml).

api/v1beta2/allowed_pod_modifications.go
nicmorales9: ahh I see. There's another spot further down "short-circuit" where we also have the default effectively defined as permissive

nicmorales9: might be more readable without the nil check though so 🤷

johscheuer: I updated the code to remove the nil checks, I think it's easier to read now.

api/v1beta2/allowed_pod_modifications.go
```diff
@@ -0,0 +1,213 @@
...
+			}
+		}
+
+		// Unknown type
+		return ""
+	}
+}
+
+// AllowedPodModifications defines the pod modification that are allowed for the user provided configuration. If a field
+// is unset in AllowedPodModifications, we allow everything to avoid breaking the current setups. In a new major release we could
```
nicmorales9, line 65: do we track breaking changes we want to go in the next major release? 😃

api/v1beta2/allowed_pod_modifications.go
```diff
@@ -0,0 +1,213 @@
...
+						maps.Keys(allowedPodModifications.AllowedServiceAccountNames),
+					),
+					",",
+				),
+			)
+		}
+	}
+
+	// Validate all volumes and check if they are allowed.
+	if len(allowedPodModifications.AllowedAdditionalVolumeSources) > 0 {
```
nicmorales9, line 161: there is no case where that list would be defined and empty right? (in a working cluster anyways)

setup/setup.go
```diff
@@ -68,29 +69,38 @@ type Options struct {
...
-	HighRunLoopBusyThreshold           float64
-	LogFileMinAge                      time.Duration
-	GetTimeout                         time.Duration
-	PostTimeout                        time.Duration
-	MaintenanceListStaleDuration       time.Duration
-	MaintenanceListWaitDuration        time.Duration
+	// AllowAutomountServiceAccountToken defines if the spec.AutomountServiceAccountToken setting in the corev1.PodSpec can
+	// be set to true.
+	AllowAutomountServiceAccountToken bool
+	// AllowHostNetwork defines if a user is allowed to set the HostNetwork setting.
```
nicmorales9: nit: in some of the comments it is specified that it is set via podspec and sometimes not, would be good to be consistent in specifying possible method

main.go
```diff
@@ -54,11 +47,15 @@ func init() {
...
 
+	// zap.Options still needs a standard flag.FlagSet; bridge it in.
+	goFS := flag.NewFlagSet("", flag.ExitOnError)
 	logOpts := zap.Options{}
-	logOpts.BindFlags(flag.CommandLine)
-	flag.Parse()
+	logOpts.BindFlags(goFS)
+	pflag.CommandLine.AddGoFlagSet(goFS)
+
+	pflag.Parse()
```
nicmorales9, line 63: what is the purpose of this change?

api/v1beta2/allowed_pod_modifications_test.go
```diff
@@ -0,0 +1,356 @@
...
+			&corev1.PodSpec{AutomountServiceAccountToken: ptr.To(true)},
+			&AllowedPodModifications{AllowAutomountServiceAccountToken: ptr.To(false)},
+			MatchError(ContainSubstring("spec.AutomountServiceAccountToken is set to")),
+		),
+		Entry("automount=false, token not allowed — false is acceptable",
+			&corev1.PodSpec{AutomountServiceAccountToken: ptr.To(false)},
+			&AllowedPodModifications{AllowAutomountServiceAccountToken: ptr.To(false)},
+			Succeed(),
+		),
+		Entry("automount not set in podSpec, token not allowed",
```
nicmorales9:
```suggestion
		Entry("automount not set in podSpec, token not allowed because field defaults to true",
```

api/v1beta2/allowed_pod_modifications.go
```diff
@@ -0,0 +1,213 @@
...
+	podSpec *corev1.PodSpec,
+	allowedPodModifications *AllowedPodModifications,
+) error {
+	// Short-circuit, if no AllowedPodModifications are set, we can skip all further work.
+	if allowedPodModifications == nil {
+		return nil
+	}
+
+	// Check if the AutomountServiceAccountToken setting is set, if so, we have to check if the allowedPodModifications
+	// allow to set the setting. The default value of podSpec.AutomountServiceAccountToken is true.
```
nicmorales9, line 99: probably also worth checking what other defaults are as it seems like none of the others are defined (maybe intended just want to double check)

api/v1beta2/allowed_pod_modifications_test.go
```diff
@@ -0,0 +1,356 @@
...
+		Entry("hostIPC=true, explicitly allowed",
+			&corev1.PodSpec{HostIPC: true},
+			&AllowedPodModifications{AllowHostIPC: ptr.To(true)},
+			Succeed(),
+		),
+		Entry("hostIPC=true, not allowed",
+			&corev1.PodSpec{HostIPC: true},
+			&AllowedPodModifications{AllowHostIPC: ptr.To(false)},
+			MatchError(ContainSubstring("spec.HostIPC is set to true, which is forbidden")),
+		),
```
nicmorales9, line 125: at least one empty spec with defined rule might be good (since AutomountServiceAccountToken is the only one that has a default filled in)

nicmorales9 (commented):
still need to look over the types test but will drop the questions in meantime

api/v1beta2/allowed_pod_modifications.go
johscheuer: Not yet, but we could start tracking breaking changes for v3.0. But first we should come up with a plan for v3.0 :)

api/v1beta2/allowed_pod_modifications.go
johscheuer: The others are not pointers, so the defaults should be `false`: https://kubernetes.io/docs/reference/kubernetes-api/core/pod-v1/

main.go
johscheuer: We changed from  the `flag` library to the `pflag` library, which is already imported because of `cobra` (used in the kubectl-fdb plugin). `pflag` allows to use `StringArrayVar`, otherwise we would need to implement this by ourselves. Given that the library is already imported, I think it's better to just use it.

api/v1beta2/allowed_pod_modifications.go
johscheuer: There is, `pflag` should allow this with `--allowed-additional-volume-sources=""`. That would allow a user to forbid all additional volume sources.

api/v1beta2/allowed_pod_modifications.go
nicmorales9: I see, might be good to add a test case then

api/v1beta2/allowed_pod_modifications.go
nicmorales9: in that case I think if it is empty we skip the check and effectively permit everything right? because we are checking len instead of nil?

nicmorales9: discussed offline: no, the map would have an empty entry and len=1

nicmorales9 (approved):
LGTM

johscheuer:
> * Commit ID: [21ce2f4](https://github.com/FoundationDB/fdb-kubernetes-operator/commit/21ce2f4b40f2f45d580a8bc8e7e24effba0287f0)
> * Duration 1:58:09
> * Result: ❌ **FAILED**
> * [Build Log](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator/fdb-kubernetes-operator-pr/9f7d7f0b-c336-4fcc-b3a5-7ea155ac0e53.gz) terminal output (available for 30 days)
> * [Build Workspace](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator-pr/9f7d7f0b-c336-4fcc-b3a5-7ea155ac0e53/workspace.zip) zip file of the working directory (available for 30 days)

I'm looking at the failure, doesn't seem to be related to the changes.

PR #2488, opened 2026-06-03
Add tool verify headers

johscheuer:
# Description

Add tool verify headers to ensure they are all updated. Before this was a manual (and error prone) process.

## Type of change

- Other

## Discussion

-

## Testing

Ran the command locally.

## Documentation

Nothing to update, I think the Makefile should be enough.

## Follow-up

-


cmd/check-license/main.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+// isGoSource reports whether a file should be checked.
+// Generated files and the e2e chaos-mesh directory are excluded (matching the
+// golangci-lint exclusions in .golangci.yml).
+func isGoSource(path, name string) bool {
+	if !strings.HasSuffix(name, ".go") {
+		return false
+	}
+	if strings.HasPrefix(name, "zz_generated.") {
```
nicmorales9, line 133: this didn't seem to work since `api/v1beta2/zz_generated.deepcopy.go` has the license updated (I don't think it matters but it looks like this code wants to skip it so I figured I'd call it out)

nicmorales9 (approved):
one inconsequential question but LGTM

cmd/check-license/main.go
johscheuer: The header in `api/v1beta2/zz_generated.deepcopy.go` was updated because I updated  `hack/boilerplate.go.txt` and then I regenerated the files.

PR #2491, opened 2026-06-07
Run `go fix` to apply Go 1.26+ modernizers to api/, cmd/ and controllers/ packages

arnav-ag:
# Description

This is the first PR that applies Go 1.26+ fix modernizers to the codebase using `go fix`. Following PRs will apply this tool to the other packages to keep the LoC changed small/PR and easy to review.

**What changed:**
- Ran `go fix` across the `api/`, `cmd/` and `controllers/` packages to apply Go 1.26+ modernizers

**Why:**
Go 1.26 introduced [fix modernizers](https://go.dev/doc/modules/fix) that automatically update code to use newer, more idiomatic Go patterns. These fixes improve code quality and maintainability.

Main changes here are:
- Replace the `interface{}` type with `any`
- Use `slices.Contains` over manual lookup loops
- Use `slices.Sort` over `sort.Slice`
- Use the `min`/`max` builtins over manual clamping
- Use `for range n` / `for i := range n` over index-based loops
- Use `fmt.Appendf` over `[]byte(fmt.Sprintf(...))`
- Use `strings.CutSuffix` over a `strings.HasSuffix` + `strings.TrimSuffix` pair
- Use `strings.SplitSeq` over `strings.Split` when we only iterate the result
- Use `reflect.TypeFor[T]()` over `reflect.TypeOf(...)`

## Type of change

- Other

Slight refactoring

## Discussion

Perhaps this should be enforced before merging future PRs to keep the code paradigm consistent. Any thoughts on this?

## Testing

Eyeballed changes and it generally makes sense, this is a small change. All existing tests pass.

PR #2493, opened 2026-06-07
Run `go fix` to apply Go 1.26+ modernizers to e2e/ package

arnav-ag:
# Description

This is a follow up (but independent) PR to #2491 that applies Go 1.26+ fix modernizers to the codebase using `go fix`.

**What changed:**
- Ran `go fix` across the `e2e/` package to apply Go 1.26+ modernizers

**Why:**
Go 1.26 introduced [fix modernizers](https://go.dev/doc/modules/fix) that automatically update code to use newer, more idiomatic Go patterns. These fixes improve code quality and maintainability.

Main changes here are:
- Replace the `interface{}` type with `any` (mostly `map[string]interface{}` -> `map[string]any`)
- Drop `,omitempty` from struct-typed JSON tags where it has no effect (`encoding/json` ignores `omitempty` for non-pointer structs)
- Use `for range n` / `for i := range n` over index-based loops
- Use `strings.SplitSeq` over `strings.Split` when we only iterate the result
- Remove the redundant `item := item` loop-variable copy that has been unnecessary since the Go 1.22 loop-variable change

## Type of change

- Other

Slight refactoring

## Testing

Eyeballed changes and it generally makes sense, even though this is a slightly larger change. I will leave some inline comments about why certain changes are safe. All existing tests pass.

### DIFF
--- a/e2e/chaos-mesh/api/v1alpha1/awschaos_types.go
+++ b/e2e/chaos-mesh/api/v1alpha1/awschaos_types.go
@@ -27,10 +27,10 @@ import (
 // AWSChaos is the Schema for the awschaos API
 type AWSChaos struct {
 	metav1.TypeMeta   `json:",inline"`
-	metav1.ObjectMeta `json:"metadata"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
 
 	Spec   AWSChaosSpec   `json:"spec"`
-	Status AWSChaosStatus `json:"status"`
+	Status AWSChaosStatus `json:"status,omitempty"`
 }
 
 var _ InnerObjectWithSelector = (*AWSChaos)(nil)
@@ -97,8 +97,8 @@ type AWSSelector struct {
 	DeviceName *string `json:"deviceName,omitempty" webhook:"AWSDeviceName,nilable"`
 }
 
-func (obj *AWSChaos) GetSelectorSpecs() map[string]any {
-	return map[string]any{
+func (obj *AWSChaos) GetSelectorSpecs() map[string]interface{} {
+	return map[string]interface{}{
 		".": &obj.Spec.AWSSelector,
 	}
 }
--- a/e2e/chaos-mesh/api/v1alpha1/awschaos_types.go
+++ b/e2e/chaos-mesh/api/v1alpha1/awschaos_types.go
@@ -27,10 +27,10 @@ import (
 // AWSChaos is the Schema for the awschaos API
 type AWSChaos struct {
 	metav1.TypeMeta   `json:",inline"`
-	metav1.ObjectMeta `json:"metadata,omitempty"`
+	metav1.ObjectMeta `json:"metadata"`
 
 	Spec   AWSChaosSpec   `json:"spec"`
-	Status AWSChaosStatus `json:"status,omitempty"`
+	Status AWSChaosStatus `json:"status"`
 }
 
 var _ InnerObjectWithSelector = (*AWSChaos)(nil)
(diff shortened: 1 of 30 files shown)


e2e/chaos-mesh/api/v1alpha1/awschaos_types.go
```diff
@@ -27,10 +27,10 @@ import (
 // AWSChaos is the Schema for the awschaos API
 type AWSChaos struct {
 	metav1.TypeMeta   `json:",inline"`
-	metav1.ObjectMeta `json:"metadata,omitempty"`
+	metav1.ObjectMeta `json:"metadata"`
```
arnav-ag: `omitempty` doesn't do anything for nested structs by default, unless passed in as a pointer.

Example program showing this:

```go
package main

import (
	"encoding/json"
	"fmt"
)

type Inner struct {
	Name string `json:"name,omitempty"`
	Age  int    `json:"age,omitempty"`
}

type Outer struct {
	ID     int    `json:"id"`
	Inner  *Inner `json:"inner_ptr,omitempty"` // pointer: omitted when nil ✓
	Inner2 Inner  `json:"inner_val,omitempty"` // value: NOT omitted even when all fields are zero ✗
}

func main() {
	o := Outer{
		ID: 1,
		// Inner is nil (pointer), Inner2 is zero-value struct
	}

	b, _ := json.Marshal(o)
	fmt.Println(string(b))
	// Output: {"id":1,"inner_val":{}}
	//
	// inner_ptr is omitted (it's a nil pointer, which is "empty")
	// inner_val is NOT omitted - a zero-value struct is not considered
	// "empty" by encoding/json, even though all its fields would be
	// omitted individually. The struct itself still appears as {}.
}
```

e2e/chaos-mesh/api/v1alpha1/schedule_types.go
```diff
@@ -85,14 +85,13 @@ func (in *Schedule) IsPaused() bool {
...
 	metav1.TypeMeta `           json:",inline"`
-	metav1.ListMeta `           json:"metadata,omitempty"`
+	metav1.ListMeta `           json:"metadata"`
 	Items           []Schedule `json:"items"`
 }
 
 func (in *ScheduleList) GetItems() []GenericChaos {
 	var result []GenericChaos
 	for _, item := range in.Items {
-		item := item
```
arnav-ag, line 95: This mattered pre-Go 1.22, since the `for...range` loop reused a single variable `item` across every iteration. On each pass it would just copy the next element's value into the same memory location. Thus every `&item` appended to `result` would be the same pointer, containing the last element of `in.Items`.

The semantics were changed in Go 1.22 such that the loop variable is iteration-scoped, and each `item` gets a distinct address. See https://go.dev/blog/loopvar-preview for the fix details. Since our `go.mod` specifies `1.25.0`, this change should be safe.

johscheuer (commented):
Can we skip everything under `e2e/chaos-mesh/api` those files are only a copy of the upstream chaos-mesh repo: https://github.com/FoundationDB/fdb-kubernetes-operator/tree/main/e2e/chaos-mesh. Otherwise the changes look fine.

arnav-ag:
@johscheuer I've restored that path 😃 do you think it is worth putting `go fix` as part of the lint check (that excludes this path) so that we don't deviate from this code style? i can make a separate PR for that

johscheuer (approved):
> @johscheuer I've restored that path 😃 do you think it is worth putting go fix as part of the lint check (that excludes this path) so that we don't deviate from this code style? i can make a separate PR for that

I think that would make sense 👍

PR #2497, opened 2026-06-08
Enhance broken test that failed during release

bktsh:
Fix flaky refreshProcessGroupStatus tests by using non-collidable ID. The hardcoded "storage-99" ID could collide with the cluster's randomly assigned process group IDs (range 1-99999), making the tests flaky in CI. Use "storage-999999" which is outside that range.

More context: During the release process for v2.29.0, the [Testing step failed in first attempt](https://github.com/FoundationDB/fdb-kubernetes-operator/actions/runs/27045360566/job/79830099831), it succeeded in 2nd attempt.


johscheuer (approved):
LGTM 👍 Thanks!

PR #2498, opened 2026-06-09
Document the usage of FDB_NETWORK_OPTION_KNOB

johscheuer:
# Description

Follow up of https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2477

## Type of change

- Documentation

## Discussion

-

## Testing

-

## Documentation

Updated.

## Follow-up

-

### DIFF
--- a/docs/manual/operator_customization.md
+++ b/docs/manual/operator_customization.md
@@ -79,6 +79,22 @@ apiVersion: apps/v1
                value: /usr/bin/fdb/primary/lib
 ```
 
+## Setting custom FDB client knobs
+
+A user can set custom FDB client knobs by using the environment variable `FDB_NETWORK_OPTION_KNOB`, the value would specify the knobs in the following format: `knob_name=knob_value`, multiple knobs can be added by using `:` as a separator.
+The following snippet shows how to set some networking knobs:
+
+```yaml
+    # In your operator deployment under the operator container spec
+    env:
+    - name: FDB_NETWORK_OPTION_KNOB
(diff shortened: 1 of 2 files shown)


docs/manual/operator_customization.md
```diff
@@ -79,6 +79,22 @@ apiVersion: apps/v1
...
+
+```yaml
+    # In your operator deployment under the operator container spec
+    env:
+    - name: FDB_NETWORK_OPTION_KNOB
+      value: max_reconnection_time=20:reconnection_time_growth_rate=10:reconnection_reset_time=60
+```
+
+These knobs will be used by the FDB bindings and the `fdbcli` commands.
+A list of possible knobs can be found in the [FoundationDB repository](https://github.com/apple/foundationdb/blob/main/flow/Knobs.cpp#L49).
```
nicmorales9, line 95: Not really to do with this PR but more because it's obscure and I can never find it 
 RECONNECTION_TIME_GROWTH_RATE which could otherwise be specified by 
```suggestion
A list of possible knobs can be found in the [FoundationDB repository](https://github.com/apple/foundationdb/blob/main/flow/Knobs.cpp#L49). 
They might otherwise be set via env var. For example `RECONNECTION_TIME_GROWTH_RATE` would be set via `FDB_NETWORK_OPTION_RECONNECTION_TIME_GROWTH_RATE`.
```

nicmorales9 (approved):

docs/manual/operator_customization.md
nicmorales9: I was confusing knobs with client network options, disregard

nicmorales9 (approved):

PR #2499, opened 2026-06-10
Injection / Risky Sink — unescaped process address in fdbcli shell command | analyze

09harsh:
# Description

An operator running `kubectl fdb analyze --analyze-status --auto-fix` executes a shell command whose argument is built from an unescaped FDB process address sourced from remote status JSON, allowing a compromised FDB process to inject arbitrary shell commands inside the target FDB pod.

## Type of change

## Discussion

No

## Testing

Added unit tests for the same. 

## Documentation

N/A

## Follow-up

No

### DIFF
--- a/kubectl-fdb/cmd/analyze.go
+++ b/kubectl-fdb/cmd/analyze.go
@@ -665,8 +665,6 @@ func analyzeStatusInternal(
 			)
 		}
 
-		// fdbcli kill needs an IP. fdbserver always sets public_address to an IP (see
-		// internal/monitor_conf.go), so a non-IP here is either malformed or attacker-controlled — skip it.
 		if autoFix && process.Address.IPAddress == nil {
 			cmd.Printf(
 				"Skipping auto-fix for process %s: address %q is not a valid IP\n",
--- a/kubectl-fdb/cmd/analyze.go
+++ b/kubectl-fdb/cmd/analyze.go
@@ -21,6 +21,7 @@
 package cmd
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
@@ -41,6 +42,9 @@ import (
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
+
+// killSleepDuration is the pause between successive `fdbcli kill` invocations during --auto-fix.
+var killSleepDuration = 5 * time.Second
 
 func newAnalyzeCmd(streams genericiooptions.IOStreams) *cobra.Command {
 	o := newFDBOptions(streams)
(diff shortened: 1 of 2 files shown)


kubectl-fdb/cmd/analyze.go
```diff
@@ -42,6 +43,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// killSleepDuration is the pause between successive `fdbcli kill` invocations during --auto-fix.
+var killSleepDuration = 5 * time.Second
```
johscheuer: What is the benefit of moving this out into a `var` and not a `const`?

kubectl-fdb/cmd/analyze.go
```diff
@@ -672,19 +676,29 @@ func analyzeStatusInternal(
...
-				return fmt.Errorf("error killing process %s status: %s, %w", process, stderr, err)
+				return fmt.Errorf(
+					"error killing process %s status: %s, %w",
+					process,
+					stderr.String(),
+					err,
+				)
 			}
-			time.Sleep(5 * time.Second)
+			time.Sleep(killSleepDuration)
```
johscheuer:
```suggestion
			case <-cmd.Context().Done():
				return cmd.Context().Err()
			case <-time.After(killSleepDuration):
				continue
			}
```

The change here seems unrelated to the actual changes? If we touch/change the code here we should change it to allow the context cancelation to take affect without waiting for the sleep to complete.

johscheuer (changes requested):
> Added unit tests for the same.

We should also run a manual test for this, as the analyze command is not covered by e2e tests right now. Otherwise we don't know if those changes break things.
It would also be helpful to add some more context either in a code comment or in the PR description on the fix. I would prefer a code comment, why we use `kubeHelper.ExecuteCommandRaw` instead of `kubeHelper.ExecuteCommandOnPod` otherwise this might be changed back in a later PR. It might also be worth to ensure the the process address is a valid net.IP, that would also prevent those kinds of attacks + would ensure that the fdbcli command will not fail, I believe the fdbcli kill command only works with IP addresses and not with hostnames or localities.

kubectl-fdb/cmd/analyze.go
```diff
@@ -664,6 +664,18 @@ func analyzeStatusInternal(
 				errorMessage,
 			)
 		}
+
+		// fdbcli kill needs an IP. fdbserver always sets public_address to an IP (see
+		// internal/monitor_conf.go), so a non-IP here is either malformed or attacker-controlled — skip it.
```
johscheuer: I don't get the reference to the `internal/monito_conf.go`, the limitations with IP addresses comes only from the rebootWorkerActor (which is used in fdbcli and the bindings): https://github.com/apple/foundationdb/blob/main/fdbclient/NativeAPI.actor.cpp#L6649-L6697. I would rather remove the comment about the `internal/monito_conf.go`.

kubectl-fdb/cmd/analyze_test.go
```diff
@@ -1063,4 +1085,83 @@ var _ = Describe("[plugin] analyze cluster", func() {
...
+				Expect(commandArgs()).To(Equal([]string{
+					"fdbcli",
+					"--exec",
+					fmt.Sprintf("kill; kill %s; sleep 5; status", expectedKillAddr),
+				}))
+				Expect(commandArgs()).NotTo(ContainElement("/bin/bash"))
+				Expect(commandArgs()).NotTo(ContainElement("/bin/sh"))
+				Expect(commandArgs()).NotTo(ContainElement("-c"))
+			},
+			Entry("IPv4:port",
```
johscheuer, line 1145: Could we also add a test case for IPv6 to ensure nothing breaks?

johscheuer (commented):
Changes look good, two comments from my side.

kubectl-fdb/cmd/analyze_test.go
09harsh: added a new entry for the same

johscheuer (approved):
LGTM 👍

PR #2495, opened 2026-06-07
Run `go fix` to apply Go 1.26+ modernizers to remaining packages

arnav-ag:
# Description

This is a follow up (but independent) PR to #2491 that applies Go 1.26+ fix modernizers to the codebase using `go fix`.

**What changed:**
- Ran `go fix` across the `kubectl-fdb/`, `mock-kubernetes-client/` and `pkg/` packages to apply Go 1.26+ modernizers

**Why:**
Go 1.26 introduced [fix modernizers](https://go.dev/doc/modules/fix) that automatically update code to use newer, more idiomatic Go patterns. These fixes improve code quality and maintainability.

Main changes here are:
- Replace the `interface{}` type with `any` (mostly `map[string]interface{}` -> `map[string]any`)
- Use `maps.Copy` over manual map copy loops
- Use `for range n` / `for i := range n` over index-based loops

## Type of change

- Refactor

## Testing

Eyeballed changes and it generally makes sense, this is a small change. All existing tests pass.

PR #2494, opened 2026-06-07
Run `go fix` to apply Go 1.26+ modernizers to fdbclient/ and internal/ packages

arnav-ag:
# Description

This is a follow up (but independent) PR to #2491 that applies Go 1.26+ fix modernizers to the codebase using `go fix`.

**What changed:**
- Ran `go fix` across the `fdbclient/` and `internal/` packages to apply Go 1.26+ modernizers

**Why:**
Go 1.26 introduced [fix modernizers](https://go.dev/doc/modules/fix) that automatically update code to use newer, more idiomatic Go patterns. These fixes improve code quality and maintainability.

Main changes here are:
- Use `fmt.Appendf` over `[]byte(fmt.Sprintf(...))`
- Replace the `interface{}` type with `any`
- Use `maps.Copy` over manual map copy loops
- Use `slices.Contains` over manual lookup loops
- Use `for range n` / `for i := range n` over index-based loops

## Type of change

- Other

Slight refactoring

## Testing

Eyeballed changes and it generally makes sense, this is a small change. All existing tests pass.

PR #2500, opened 2026-06-10
Enforce Go 1.26+ `go fix` modernizers via `make fmt`

arnav-ag:
# Description

Follow up to the `go fix` modernizer PRs (#2491, #2493, #2494, #2495). Adds a `fix` make target and runs it from `make fmt`, so the `build` job's existing `make clean all && git diff --exit-code` enforces modernizers going forward, with no new CI step.

- `e2e/chaos-mesh/api` is excluded since it is a vendored copy of upstream chaos-mesh. `go fix` only rewrites the packages it is passed, so filtering the package list is enough.
- Also applies the one outstanding modernizer (`strings.Cut` in `cmd/check-license/main.go`).
- Note: the check runs a fresh `go fix`, so code landing around this PR may need `make fix` re-run and committed at merge time.

## Type of change

- Other

## Testing

`go fix` produces no diff on the current tree, and `e2e/chaos-mesh` and generated files are left untouched. Relying on CI for the full `make fmt` / lint / test run.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,7 +80,7 @@ endif
 
 all: deps generate fmt vet manager snapshot manifests samples documentation test_if_changed
 
-.PHONY: clean all manager samples documentation run install uninstall deploy manifests fmt vet generate container-build container-push container-push-if-remote rebuild-operator bounce lint check-license fix-license
+.PHONY: clean all manager samples documentation run install uninstall deploy manifests fmt vet generate container-build container-push container-push-if-remote rebuild-operator bounce lint check-license fix-license fix
 
 deps: $(BUILD_DEPS)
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -159,10 +159,15 @@ manifests: ${MANIFESTS}
 ${MANIFESTS}: ${CONTROLLER_GEN} ${GO_SRC}
 	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./api/..." paths="./controllers/..." applyconfiguration output:crd:artifacts:config=config/crd/bases
 
+# Run the Go fix modernizers against the code.
+fix: deps
```
johscheuer:
```suggestion
fix:
```

go fix is part of the standard tooling, right? So there shouldn't be any need to install/check the other dependencies? With the current setup we would run the `deps` step twice when running `make fmt`.

Makefile
```diff
@@ -159,10 +159,15 @@ manifests: ${MANIFESTS}
...
+	go list ./... | grep -v '/e2e/chaos-mesh/' | xargs go fix
+
 # Run go fmt against code
 fmt: deps bin/fmt_check
 
 bin/fmt_check: ${GO_ALL}
+	@$(MAKE) --no-print-directory fix
```
johscheuer: I would prefer to move the `fix` to the `fmt` target above, e.g. `fmt: deps fix bin/fmt_check`, that way we don't need to create a make subprocess.

johscheuer (changes requested):

johscheuer (approved):

PR #2504, opened 2026-06-16
Block process group removal when exclusion state is unverifiable

09harsh:
# Description

Block process group removal when exclusion state is unverifiable

  When FDB reports no processes for the cluster's DC, checkAndSetProcessStatus
  takes the early-return path in update_status.go and ExclusionTimestamp on each
  process group is whatever it was last set to — possibly stale. The
  removeProcessGroups reconciler short-circuits on IsExcluded() without
  re-verifying, which can lead to removal of a partially-excluded data-bearing
  process if the fault-tolerance gate happens to report healthy.

  Add a guard at the top of removeProcessGroups.reconcile that requeues when
  status.Cluster.Processes contains no processes sharing the cluster's DC and
  at least one process group is both marked for removal and flagged excluded.
  The existing reset logic at update_status.go:551 corrects stale timestamps
  once FDB returns, so the guard self-clears on the next cycle.

## Type of change

## Discussion

No

## Testing

Manual testing?*

Added unit test

## Documentation

N/A

## Follow-up

No


controllers/remove_process_groups.go
```diff
@@ -588,6 +600,31 @@ func getProcessesToInclude(
 	return fdbProcessesToInclude, processGroups[:idx]
 }
 
+// hasUnverifiableExcludedRemoval reports whether the cluster has at least one process group
+// marked for removal whose IsExcluded() flag cannot be re-verified against live cluster state
+// in this reconcile cycle. The flag is unverifiable when the FDB machine-readable status
+// reports no processes sharing the cluster's DC — the same condition under which
+// checkAndSetProcessStatus skips refreshing ExclusionTimestamp.
+func hasUnverifiableExcludedRemoval(
```
johscheuer: I don't think this actually solves the issue, e.g. let's say we have process A and process B in the same DC. 

- Process A is marked for exclusion and all data is removed.
- Process A gets excluded and the remove_process_groups set the exclusion timestamp (https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/controllers/remove_process_groups.go#L647).
- It updates the `FoundationDBCluster` with the exclusion timestamp.
- Before the operator can delete the pod the operator gets deleted/evicted/crashes.
- A human operator accidentally includes process A back.
- DD sends data back to the process.
- Now the process gets partitioned or taken down by something else.
- The operator is back.
- Since process A is missing in the status the exclusion timestamp is not reset in update_status https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/controllers/update_status.go#L547-L561.
- The operator checks now `hasUnverifiableExcludedRemoval`, since process B is there it skips all other checks and we would still remove Process A.

I think a better approach would be to adjust [getAddressesToValidateBeforeRemoval](https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/internal/removals/remove.go#L178-L229) and if a process is missing in the machine-readable status and should be removed and has an exclusion timestamp we should just rerun the `exclude` command for that process.

In general there will be a race condition between the operator verifying that a process is fully excluded and the operator doing the pod removal. If someone else outside of the operator's scope does an include of that process and the operator doesn't notice it in the `update_status`.

johscheuer (changes requested):

internal/removals/remove.go
```diff
@@ -228,6 +243,30 @@ func getAddressesToValidateBeforeRemoval(
 	return remainingMap, addresses
 }
 
+// processGroupIDsInStatus returns the set of process group IDs that have at least one process
+// reporting in the machine-readable status for the cluster's DC. Multi-process pods share an
+// instance ID across all their processes, so a single locality lookup per process is enough.
```
johscheuer: I think the last sentence of the comment is a bit confusing as the method iterates over all processes and will just overwrite the entry for multi-process pods multiple times. It might be worth to note that this will not catch cases where one single process is down/not reporting, as in this case the `present` map will have an entry.

internal/removals/remove.go
```diff
@@ -178,13 +178,28 @@ func GetRemainingMap(
 func getAddressesToValidateBeforeRemoval(
 	logger logr.Logger,
 	cluster *fdbv1beta2.FoundationDBCluster,
+	status *fdbv1beta2.FoundationDBStatus,
 ) (map[string]bool, []fdbv1beta2.ProcessAddress) {
+	processGroupsInStatus := processGroupIDsInStatus(cluster, status)
```
johscheuer, line 186: Might be worth to add a comment here that if the `status` is `nil` all entries will be re-validated

internal/removals/remove.go
```diff
@@ -178,13 +178,28 @@ func GetRemainingMap(
...
 	addresses := make([]fdbv1beta2.ProcessAddress, 0, len(cluster.Status.ProcessGroups))
 	for _, processGroup := range cluster.Status.ProcessGroups {
-		if !processGroup.IsMarkedForRemoval() || processGroup.IsExcluded() {
+		if !processGroup.IsMarkedForRemoval() {
 			continue
 		}
 
+		// Trust IsExcluded() only when the process is currently visible in the live status.
```
johscheuer:
```suggestion
		// Trust IsExcluded() only when at least one process of the process group is currently visible in the live status.
```

internal/removals/remove_test.go
```diff
@@ -642,4 +646,87 @@ var _ = Describe("remove", func() {
...
+							Addresses: []string{
+								"192.0.0.1",
+							},
+							RemovalTimestamp:   &metav1.Time{Time: time.Now()},
+							ExclusionTimestamp: &metav1.Time{Time: time.Now()},
+						},
+					},
+				},
+			},
+			fdbv1beta2.ProcessGroupID("storage-2"),
```
johscheuer, line 713: Why are we deleting `storage-2` here? It's not part of the cluster's ProcessGroups?

controllers/remove_process_groups.go
```diff
@@ -619,7 +639,20 @@ func (r *FoundationDBClusterReconciler) getProcessGroupsToRemove(
 		}
 
 		// ProcessGroup is already marked as excluded we can add it to the processGroupsToRemove and skip further checks.
+		// Exception: when getAddressesToValidateBeforeRemoval forced re-validation (because the process is missing
+		// from the machine-readable status), any of its addresses with remainingMap[addr] == true means the live
+		// re-exclude check could not confirm the process is fully excluded — so trust the live signal over the
```
johscheuer:
```suggestion
		// exclude check could not confirm the process is fully excluded — so trust the live signal over the
```

controllers/remove_process_groups.go
```diff
@@ -619,7 +639,20 @@ func (r *FoundationDBClusterReconciler) getProcessGroupsToRemove(
...
+		// possibly-stale ExclusionTimestamp. See https://github.com/FoundationDB/fdb-kubernetes-operator/issues/1912.
 		if processGroup.IsExcluded() {
+			if processGroupAddressesRemaining(processGroup, remainingMap) {
+				logger.Info(
+					"ExclusionTimestamp set but live re-exclude reports addresses still remaining; not removing",
```
johscheuer:
```suggestion
					"ExclusionTimestamp set but live exclude reports addresses still remaining; not removing",
```

controllers/remove_process_groups.go
```diff
@@ -588,6 +588,26 @@ func getProcessesToInclude(
 	return fdbProcessesToInclude, processGroups[:idx]
 }
 
+// processGroupAddressesRemaining returns true if any of the process group's addresses
+// (IP-based or locality-based) appear in remainingMap with a value of true, meaning the
+// live re-exclude check could not confirm them fully excluded. Returns false when none
```
johscheuer:
```suggestion
// live exclude check could not confirm them fully excluded. Returns false when none
```

johscheuer (commented):
Would be great if we could add an e2e test for this to ensure this work as expected in a real cluster, e.g. if we create a similar test case as `a process is marked for removal and was excluded` in the test operator suite, we should be able to evaluate the new changes.

johscheuer (approved):
LGTM 👍

PR #2506, opened 2026-06-16
Fix operator test case and make use of removal mode not deletion mode

johscheuer:
# Description

The `DeletionMode` is used for pod updates that require a deletion, but in this case we can't to block the removal of the pod.

## Type of change

## Discussion

-

## Testing

Ran manually.

## Documentation

Adjusted the auto-generated docs.

## Follow-up

-


nicmorales9 (approved):

johscheuer (commented):
Just merged the PR because the button changed from "enable auto-merge" to the normal merge button 🙄 Should be fine but I didn't meant to skip the test checks.

PR #2508, opened 2026-06-18
Backup reconciler will delete terminal pods

johscheuer:
# Description

Delete backup_agent pods that are in a terminal state, e.g. they are stuck in pending or they terminated.

## Type of change

## Discussion

We have the same functionality in the cluster reconciler, adding the functionality in the operator will improve the backup agent management.

## Testing

Added new e2e tests and new unit tests.

## Documentation

Nothing to update.

## Follow-up

-

### DIFF
--- a/controllers/backup_controller_test.go
+++ b/controllers/backup_controller_test.go
@@ -21,6 +21,7 @@
 package controllers
 
 import (
+	"context"
 	"fmt"
 
 	"k8s.io/utils/ptr"
@@ -32,12 +33,12 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
-	"context"
-
 	fdbv1beta2 "github.com/FoundationDB/fdb-kubernetes-operator/v2/api/v1beta2"
 	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 func reloadBackup(backup *fdbv1beta2.FoundationDBBackup) (int64, error) {
@@ -338,6 +339,36 @@ var _ = Describe("backup_controller", func() {
 			})
 		})
 
+		When("a backup agent pod is stuck in a Failed state", func() {
+			BeforeEach(func(ctx SpecContext) {
+				generationGap = 0
+				setupUnreadyDeploymentWithTerminalPod(
+					ctx,
+					backup,
+					corev1.PodFailed,
+				)
+			})
+
+			It("should delete the terminal pod", func() {
+				expectNoTerminalBackupAgentPods(backup)
+			})
+		})
+
+		When("a backup agent pod completed but was not cleaned up", func() {
+			BeforeEach(func(ctx SpecContext) {
+				generationGap = 0
+				setupUnreadyDeploymentWithTerminalPod(
+					ctx,
+					backup,
+					corev1.PodSucceeded,
+				)
+			})
+
+			It("should delete the terminal pod", func() {
+				expectNoTerminalBackupAgentPods(backup)
+			})
+		})
+
 		When("providing custom parameters", func() {
 			BeforeEach(func() {
 				backup.Spec.CustomParameters = fdbv1beta2.FoundationDBCustomParameters{
@@ -459,3 +490,18 @@ var _ = Describe("backup_controller", func() {
 		})
 	})
 })
+
+// expectNoTerminalBackupAgentPods asserts that no Pod with the backup's deployment selector labels
+// remains in the namespace after reconciliation.
+func expectNoTerminalBackupAgentPods(backup *fdbv1beta2.FoundationDBBackup) {
+	podList := &corev1.PodList{}
+	Expect(k8sClient.List(
+		context.TODO(),
+		podList,
+		client.InNamespace(backup.Namespace),
+		client.MatchingLabels(map[string]string{
+			fdbv1beta2.BackupDeploymentPodLabel: internal.GetBackupDeploymentName(backup),
+		}),
+	)).To(Succeed())
+	Expect(podList.Items).To(BeEmpty())
+}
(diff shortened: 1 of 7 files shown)


controllers/backup_controller_test.go
```diff
@@ -338,6 +339,36 @@ var _ = Describe("backup_controller", func() {
 			})
 		})
 
+		When("a backup agent pod is stuck in a Failed state", func() {
```
nicmorales9, line 342: this will fail in the reconciler / fail the test if the backup is misconfigured and listing for pods is empty right?  Otherwise expectNoTerminalBackupAgentPods will pass in cases where it is just looking in the wrong place

e2e/test_operator_backups/operator_backup_test.go
```diff
@@ -312,6 +316,87 @@ var _ = Describe("Operator Backup", Label("e2e", "pr", "foundationdb-pr"), func(
...
+						terminalPod := &corev1.Pod{
+							ObjectMeta: metav1.ObjectMeta{
+								Name:      terminalPodName,
+								Namespace: currentBackup.Namespace,
+								Labels:    deployment.Spec.Template.Labels,
+							},
+							Spec: spec,
+						}
+
+						Expect(factory.GetControllerRuntimeClient().DeleteAllOf(ctx, &corev1.Pod{},
```
nicmorales9, line 372: why do we delete all the other agents before creating and deleting the terminating pod?

e2e/fixtures/fdb_backup.go
```diff
@@ -444,3 +445,46 @@ func (fdbBackup *FdbBackup) ForceReconcile() {
...
+//	spec.AgentCount = "42" // Make your changes.
+//
+//	backup.UpdateBackupSpecWithSpec(backupSpec) // Update the spec.
+func (fdbBackup *FdbBackup) UpdateBackupSpecWithSpec(
+	desiredSpec *fdbv1beta2.FoundationDBBackupSpec,
+) {
+	fetchedBackup := &fdbv1beta2.FoundationDBBackup{}
+
+	// This is flaky. It sometimes responds with an error saying that the object has been updated.
+	// Try a few times before giving up.
```
nicmorales9:
```suggestion
	// Try every second for 10 minutes before giving up
```
slightly more frequent than a few times 😆

controllers/update_backup_agents.go
```diff
@@ -108,6 +111,62 @@ func (u updateBackupAgents) reconcile(
...
+				client.InNamespace(backup.Namespace),
+				client.MatchingLabels(existingDeployment.Spec.Selector.MatchLabels),
+			)
+			if err != nil {
+				return &requeue{curError: err}
+			}
+
+			// minRemaining keeps track of the minimum time the operator has to wait before it can delete a new pod that
+			// is stuck in a terminal state.
+			var minRemaining time.Duration
```
nicmorales9: nit: I think this might be slightly more readable if the variable was named `minTimeTillDeletion` or something along those lines

controllers/update_backup_agents.go
```diff
@@ -108,6 +111,62 @@ func (u updateBackupAgents) reconcile(
...
+					"phase", phase,
+					"reason", reason)
+				err = r.Delete(ctx, ptr.To(pod))
+				if err != nil {
+					return &requeue{curError: err}
+				}
+			}
+
+			if minRemaining > 0 {
+				return &requeue{
```
nicmorales9, line 163: Maybe I am missing something (def not as familiar with the reconcilers), but is it true that we are actually waiting until the max time remaining to delete pods in terminal state? E.g if we have 2 pods in terminal state and 1 we can delete in a minute, and the other in 10min, aren't we going to stop at this code point for the next 10 minutes anyways and therefore should wait until the maximum time? Or is there stuff done in between that is useful to poll for?

e2e/fixtures/fdb_backup.go
johscheuer: That comment was just copied from the FdbCluster fixture, it probably evolved over time. I update the comment 👍

e2e/test_operator_backups/operator_backup_test.go
johscheuer: The ReplicaSet controller in Kubernetes will keep the oldest pods and delete the youngest pods first, e.g. if you're scaling down. Since the terminal pod would have the newest creation timestamp it would be the youngest and would be removed by the ReplicaSet controller anyways, without our controller taking any action.

controllers/update_backup_agents.go
johscheuer: There could be other stuff like a change in the backup settings, e.g. pausing the backup or changing the configuration that requires a modification of the current backup. The requeue here will be a `delayedRequeue`, which means the backup reconciler will perform all the other sub-reconcilers until it finishes all of them or hit an issue. If it executed all the sub-reconcilers it will requeue the backup with the provided delay time.

controllers/backup_controller_test.go
johscheuer: Our test cases don't create pods for Deployments, so any pod that is present was created in the test setup. We could adjust our fake client implementation to create a ReplicaSet and pods, but I think that would be out of scope for this PR and could be done in another PR.

johscheuer:
> * Commit ID: [5af280e](https://github.com/FoundationDB/fdb-kubernetes-operator/commit/5af280ebf377410cfe2b75d6241b96e37c783791)
> * Duration 1:55:09
> * Result: ❌ **FAILED**
> * [Build Log](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator/fdb-kubernetes-operator-pr/bd1e7728-4750-4031-9e6a-654e63675a44.gz) terminal output (available for 30 days)
> * [Build Workspace](https://d1e0l78xbkh3xa.cloudfront.net/fdb-kubernetes-operator-pr/bd1e7728-4750-4031-9e6a-654e63675a44/workspace.zip) zip file of the working directory (available for 30 days)

Unrelated: 

```
[Container] 2026/06/26 12:48:05.911733 Running command for f in $(grep -L -- "SUCCESS\!" ${CODEBUILD_SRC_DIR}/logs/*.log); do if ! grep -q -- "--- FAIL:" $f; then echo "This test has timed out or had a panic: $f"; export fail_test=true; fi ;  done
This test has timed out or had a panic: /codebuild/output/src837784552/src/github.com/FoundationDB/fdb-kubernetes-operator/logs/test_operator_upgrades_variations.log
```

nicmorales9 (approved):

PR #2509, opened 2026-06-19
Add support for custom backup tags

### LINKED ISSUE #2390: Support for different backup tags / multiple backups

As of the current state of the operator, there is no CRD field to specify the backup tag when using the backup/restore functions. If one wants to natively stream backups to different S3 regions from the same cluster, for example, it is not possible.

Since the underlying support in FoundationDB for multiple backup tags is there, it seems like this would be a worthwhile improvement for the operator.

johscheuer:
# Description

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/2390

## Type of change

## Discussion

-

## Testing

Added unit tests and e2e test cases.

## Documentation

Added documentation for the new feature.

## Follow-up

-


api/v1beta2/foundationdbbackup_types.go
```diff
@@ -426,6 +444,11 @@ func (backup *FoundationDBBackup) GetDesiredAgentCount() int {
 
 // NeedsBackupReconfiguration determines if the backup needs to be reconfigured.
 func (backup *FoundationDBBackup) NeedsBackupReconfiguration() bool {
+	// If the backup is not running we don't have to modify it.
+	if backup.Status.BackupDetails == nil || !backup.Status.BackupDetails.Running {
```
nicmorales9, line 448: Running is never false because it failed to start right? Running is just indicating if it should be trying to backup? 
(Just thinking of a scenario where someone misconfigured the backup originally in a way that prevented it from starting and is trying to reconfigure it to get it started)

docs/manual/backup.md
```diff
@@ -84,6 +84,65 @@ spec:
...
+kind: FoundationDBBackup
+metadata:
+  name: sample-cluster-hourly
+spec:
+  version: 7.1.26
+  clusterName: sample-cluster
+  tag: hourly
+  snapshotPeriodSeconds: 3600  # 1 hour
+  blobStoreConfiguration:
+    accountName: account@object-store.example:443
```
nicmorales9, line 106:
```suggestion
    accountName: account@object-store.example:443
    backupName: sample-cluster-hourly
```

controllers/update_backup_status.go
```diff
@@ -119,7 +121,24 @@ func (s updateBackupStatus) reconcile(
...
+		status.BackupDetails.Tag = ptr.Deref(liveStatus.Tag, originalStatus.BackupDetails.Tag)
+	}
+
+	// Ensure that the tag was not changed, e.g. in a case where someone creates a new backup with the same name
+	// but a different tag.
+	if status.BackupDetails.Tag != "" &&
+		status.BackupDetails.Tag != string(backup.GetBackupTag()) {
+		return &requeue{
+			curError: fmt.Errorf(
+				"current tag: %s cannot be changed to %s, backup tags are immutable",
```
nicmorales9, line 136: nit: could be nice to reference this exact error in the backup doc so people can grep for it

nicmorales9 (approved):
LGTM

docs/manual/backup.md
johscheuer: There is no need to provide a backupName to the first backup, it will use the backup name (from the metadata) as name if `backupName` is absent.

nicmorales9: oh I just thought it was a bit clearer that it is different from the other but it doesn't really matter if you wanted to illustrate that defaulting

api/v1beta2/foundationdbbackup_types.go
johscheuer: No, that is the information from the `backup status` command and shows that the backup is running. If the backup is not running, e.g. because it completed or it was aborted, there is no need to modify the backup

PR #2511, opened 2026-06-22
Remove team tracker health check in DoStorageServerFaultDomainCheckOnStatus

johscheuer:
# Description

Remove team tracker health check in DoStorageServerFaultDomainCheckOnStatus to prevent long exclusions to block the removal of fully excluded processes

## Type of change

- Other

## Discussion

-

## Testing

Updated unit tests.

## Documentation

Updated the documentation for this decision in the code.

## Follow-up

-

### DIFF
--- a/pkg/fdbstatus/status_checks.go
+++ b/pkg/fdbstatus/status_checks.go
@@ -396,8 +396,8 @@ func DoStorageServerFaultDomainCheckOnStatus(status *fdbv1beta2.FoundationDBStat
 			region = "remote"
 		}
 
-		// We are not checking the health state here because any data-movement, except for rebalancing will mean this
-		// flag is set to false. This means the operator would always wait until all storage servers are fully excluded.
+		// We are not checking the health state here because any data-movement, except for rebalancing will mean that
+		// tracker.State.Healthy is set to false. This means the operator would always wait until all storage servers are fully excluded.
 		// This behaviour is too restrictive for the operator, especially in large scale deployments. The operator
 		// performs the exclusion check before trying to remove a pod, so this is an additional safety guard but the
 		// tracker.State.MinReplicasRemaining is a better check for this.
--- a/pkg/fdbstatus/status_checks.go
+++ b/pkg/fdbstatus/status_checks.go
@@ -355,7 +355,7 @@ func GetMinimumUptimeAndAddressMap(
 		}
 
 		addressMap[fdbv1beta2.ProcessGroupID(processGroupID)] = append(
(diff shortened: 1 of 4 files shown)


pkg/fdbstatus/status_checks.go
```diff
@@ -396,13 +396,26 @@ func DoStorageServerFaultDomainCheckOnStatus(status *fdbv1beta2.FoundationDBStat
 			region = "remote"
 		}
 
-		if !tracker.State.Healthy {
-			return fmt.Errorf("team tracker in %s is in unhealthy state", region)
-		}
-
+		// We are not checking the health state here because any data-movement, except for rebalancing will mean this
+		// flag is set to false. This means the operator would always wait until all storage servers are fully excluded.
```
nicmorales9: what is 'this flag'?

pkg/fdbstatus/status_checks.go
```diff
@@ -396,13 +396,26 @@ func DoStorageServerFaultDomainCheckOnStatus(status *fdbv1beta2.FoundationDBStat
...
+		// This behaviour is too restrictive for the operator, especially in large scale deployments. The operator
```
nicmorales9, line 401: I haven't dug deep into the path but the functions that call it specifically talk about a fault domain check, so doing check for full exclusions here is out of place for that reason as well

nicmorales9 (approved):
Seems sensible to me but I do wonder if this could cause faster replacements and have other repercussions as a result; would e2e tests should sufficiently cover this? Is there other testing in place? 
My feel is that e2e probably would catch this but good to think about

pkg/fdbstatus/status_checks.go
johscheuer: `tracker.State.Healthy`, I should have added that.

pkg/fdbstatus/status_checks.go
johscheuer: This method is invoked after a process was verified to be fully excluded, so this is another safe guard to reduce the risk of dataloss. But waiting for the state to get `healthy` again will mean we will block all removal until all exclusions are done. Usually there is no need for this and fully excluded processes can be removed if the storage min replicas are present.

johscheuer (commented):
> Seems sensible to me but I do wonder if this could cause faster replacements and have other repercussions as a result; would e2e tests should sufficiently cover this? Is there other testing in place?
My feel is that e2e probably would catch this but good to think about

e2e test cases should cover this. This shouldn't cause faster replacements, only faster deletions for fully excluded processes if doing a migration.

PR #2517, opened 2026-06-29
Fetch the management API error in the same transaction as the error occurred

### LINKED ISSUE #2514: tryReadSpecialKeysError probably does not work as it is not using same transaction as failure
### What happened?

I tried utilizing the operator code to read special keys errors and it didn't work. Based on testing, it seems that the error needs to be read in the same transaction. 
Arguably this is a problem that FDB clears the special error at the end of the transaction and/or that it seems completely undocumented.

I expected the code to work

Spin up an e2e test with < 3 processes and try to exclude one or more storage processes to get a special keys error then look and see if it outputs the error. I suspect not but have not tried this yet with the operator 

_No response_

### FDB Kubernetes operator

<details>

```console
$ kubectl fdb version
# paste output here
```

</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Cloud provider

<details>

</details>

johscheuer:
# Description

Fetch the management API error in the same transaction as the error is only kept in-memory and will be absent in another transaction.

Fixes: https://github.com/FoundationDB/fdb-kubernetes-operator/issues/2514

## Type of change

## Discussion

-

## Testing

I created a test case to reproduce the issue. That test case doesn't run in CI or per default and requires a FDB cluster to connect to.

## Documentation

We should update the upstream documentation about this special key.

## Follow-up

-

### DIFF
--- a/fdbclient/fdb_client.go
+++ b/fdbclient/fdb_client.go
@@ -29,9 +29,11 @@ import (
 	"time"
 
 	fdbv1beta2 "github.com/FoundationDB/fdb-kubernetes-operator/v2/api/v1beta2"
+	"github.com/FoundationDB/fdb-kubernetes-operator/v2/internal"
 	"github.com/apple/foundationdb/bindings/go/src/fdb"
 	"github.com/apple/foundationdb/bindings/go/src/fdb/tuple"
 	"github.com/go-logr/logr"
+	"k8s.io/utils/ptr"
 )
 
 // fdbLibClient is an interface to interact with FDB over the client libraries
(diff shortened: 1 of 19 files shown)


fdbclient/fdb_client.go
```diff
@@ -367,47 +367,41 @@ func (fdbClient *realFdbLibClient) executeTransactionForManagementAPI(
...
-			// Allow the operator to read and write from/to a locked database e.g. in cases where a restore is ongoing.
-			// This allows the operator to read and write from/to the FDB cluster.
-			optErr := tr.Options().SetLockAware()
-			if optErr != nil {
-				return nil, optErr
-			}
-
+		// The error message in \xff\xff/error_message is only kept in-memory and must be read in the same transaction, otherwise
+		// the result will be empty.
+		err = operation(tr)
```
brownleej: I don't know if this will cover the main cases where we need to look at this error message key. Looking at the places where we are calling `executeTransactionForManagementAPI`, many of the `operation` functions return no error at all, and the others only return errors from the read path. In a case where, for instance, we failed to change the list of excluded servers because of internal safety checks, that error would be returned at commit time. Commit errors aren't covered by this check, because at this stage nothing has been committed yet.

I think the only way to handle these errors correctly in a general case is to embed the check into the Transact function, as part of the transaction retry loop.

brownleej (changes requested):

fdbclient/fdb_client.go
```diff
@@ -350,41 +352,91 @@ func (fdbClient *realFdbLibClient) executeTransaction(
...
-	})
+		if opErr == nil {
+			// Committed successfully, done
+			return nil
+		}
+
+		// commit/operation failed — tr is still valid here, read the special key before reset.
+		opErr = fetchSpecialKeyErrorMessage(fdbClient.logger, tr, opErr)
+		if opErr == nil {
+			// fetchSpecialKeyErrorMessage determined this wasn't actually a failure (e.g. the coordinators
```
nicmorales9, line 409: seems like this comment got cut off?

fdbclient/fdb_client.go
```diff
@@ -350,41 +352,91 @@ func (fdbClient *realFdbLibClient) executeTransaction(
...
+		// which is not one of the codes FDB's client considers automatically retryable — so we must honor the
+		// retriable flag ourselves and reset the transaction manually before trying again.
+		var specialKeysErr fdbv1beta2.SpecialKeysAPIFailureError
+		if errors.As(opErr, &specialKeysErr) {
+			if !specialKeysErr.Retriable {
+				return opErr
+			}
+
+			tr.Reset()
+			continue
```
nicmorales9, line 425: I think the default for the transactions might be to retry indefinitely JFYI. I think we might want to set `db.Options().SetTransactionRetryLimit(int64(some_reasonable_retry_limit))` 
(I've seen some positive indications with long running clients in other testing with the Go client)

fdbclient/fdb_client.go
```diff
@@ -350,41 +352,91 @@ func (fdbClient *realFdbLibClient) executeTransaction(
...
-			// Allow the operator to read and write from/to a locked database e.g. in cases where a restore is ongoing.
-			// This allows the operator to read and write from/to the FDB cluster.
-			optErr := tr.Options().SetLockAware()
-			if optErr != nil {
-				return nil, optErr
+		// Otherwise fall back to letting FDB's client decide if the error is retryable (e.g. conflicts, timeouts).
+		var fdbErr fdb.Error
+		if errors.As(opErr, &fdbErr) {
+			if retryErr := tr.OnError(fdbErr).Get(); retryErr != nil {
+				return checkError(opErr) // not retryable, give up
```
nicmorales9, line 432: I completely forgot to make a PR earlier but I did find a problem with checkError a bit ago while debugging the issue with the special keys https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2523/changes

fdbclient/fdb_client.go
```diff
@@ -398,14 +450,42 @@ func (fdbClient *realFdbLibClient) executeTransactionForManagementAPI(
...
+			return fdbv1beta2.SpecialKeysAPIFailureError{
+				Err: fmt.Errorf(
+					"error from special key 0xff0xff/error_message is: %s, original error: %w",
+					errMessage,
+					err,
+				),
+			}
+		}
+
+		// fdbcli does the same, this error means the current coordinators where already the coordinators of the cluster
```
nicmorales9:
```suggestion
		// fdbcli does the same, this error means the current coordinators were already the coordinators of the cluster
```

nicmorales9: actually "are" might be more correct here? 😅

fdbclient/fdb_client.go
```diff
@@ -398,14 +450,42 @@ func (fdbClient *realFdbLibClient) executeTransactionForManagementAPI(
...
+		// https://github.com/apple/foundationdb/blob/main/fdbcli/CoordinatorsCommand.cpp#L163-L169
+		// NOTE(johscheuer): There might be more special cases and handling in fdbcli.
+		if ptr.Deref(mgmtAPIError.Command, "") == "coordinators" &&
+			ptr.Deref(
+				mgmtAPIError.Message,
+				"",
+			) == "No change (existing configuration satisfies request)" {
+			logger.V(1).
+				Info("coordinator command didn't change nay coordinators. The provided coordinators were already recruited.")
```
nicmorales9:
```suggestion
				Info("coordinator command didn't change any coordinators. The provided coordinators were already recruited.")
```

fdbclient/fdb_client_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+				Expect(err).To(HaveOccurred())
+				Expect(
+					err.Error(),
+				).To(ContainSubstring("error from special key 0xff0xff/error_message is: ERROR: 'bad' is not a valid network endpoint address"))
+			})
+		})
+
+		When("setting the maintenance mode", func() {
+			It("should return an error message that the address is invalid", func() {
```
nicmorales9: it looks like it is not returning an error?

nicmorales9 (approved):
looks good but a few important notes

fdbclient/fdb_client.go
nicmorales9: (seems especially important to check that the operator doesn't get stuck here)

johscheuer: The retry limit is set in the `getFDBDatabase` method: https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/fdbclient/common.go#L115-L125

brownleej (approved):
👍

brownleej (approved):
👍

PR #2520, opened 2026-07-08
e2e: Makefile compile target working

ploxiln:
creates binaries named test_NAME.test
(but test_NAME.run does not use them, it still uses "go test ./test_NAME")

Also, BASE_DIR not needed, root Makefile always run from its own dir.

--------

This helps to get all test dependencies and go build objects pre-cached by go.


johscheuer (approved):
LGTM 👍

PR #2535, opened 2026-07-28
Add duration calculation for removals

johscheuer:
# Description

Add duration calculation for removals, this could be useful to track how long removals in the operator take.

## Type of change

## Discussion

-

## Testing

Ran the unit tests and CI will run e2e tests.

## Documentation

Nothing to update.

## Follow-up

-


controllers/remove_process_groups.go
```diff
@@ -568,6 +573,23 @@ func getProcessesToInclude(
 				}
 			}
 
+			exclusionDuration := unknownDurationSeconds
+			if !processGroup.ExclusionTimestamp.IsZero() {
+				if processGroup.RemovalTimestamp.Time.Before(processGroup.ExclusionTimestamp.Time) {
+					exclusionDuration = processGroup.ExclusionTimestamp.Sub(processGroup.RemovalTimestamp.Time).
```
nicmorales9: Time.Time manipulations are always a bit hard to read, it might be helpful to add a comment. I _think_ what is happening is: if removal happened first, the exclusion effectively starts at the time of the removal, so set the exclusionDuration to be the time difference between the start of the removal and the exclusion?
But that last bit doesn't make sense to me, shouldn't it be adding the removal time?
Or perhaps I don't follow the logic at all 😓

johscheuer: This bit calculates the exclusion duration because the exclusion timestamp should always be after the removal timestamp and the exclusion duration is (assumed) to be the time between when the process group was marked for removal and when the process was fully excluded. Which is not totally true but a heuristic.

nicmorales9 (approved):
LGTM but would be nice to rename exclusionDuration to `timeToFullExclusion` or something more explicit given how overloaded the term exclusion is. Same with other variables and comments, would be good to rename to specify full-exclusion when possible

PR #2537, opened 2026-07-29
Add host information into process group status

johscheuer:
# Description

Add host information into process group status. This can be helpful in multiple cases and will be used for additional information logging.

## Type of change

## Discussion

This will cause additional growth in the FOundationDBCluster status, but it should be fine. In the future we should be splitting up the process group status into a dedicated CRD.

## Testing

Ran unit tests.

## Documentation

Nothing to update.

## Follow-up

-


api/v1beta2/applyconfiguration/api/v1beta2/processgroupstatus.go
```diff
@@ -108,3 +111,11 @@ func (b *ProcessGroupStatusApplyConfiguration) WithFaultDomain(value apiv1beta2.
 	b.FaultDomain = &value
 	return b
 }
+
+// WithHost sets the Host field in the declarative configuration to the given value
+// and returns the receiver, so that objects can be built by chaining "With" function invocations.
+// If called multiple times, the Host field is set to the value of the last call.
+func (b *ProcessGroupStatusApplyConfiguration) WithHost(value apiv1beta2.Host) *ProcessGroupStatusApplyConfiguration {
```
nicmorales9: I don't think it matters, but why `b`?

controllers/update_status.go
```diff
@@ -1250,6 +1252,7 @@ func getFaultDomainFromProcesses(processes []fdbv1beta2.FoundationDBStatusProces
 		}
 
 		for _, process := range processes {
+			// The assumption here is that if the process has the zoneid set, the machineid should also be set.
```
nicmorales9, line 1255: since zone is often rack / physical location based? What are possible other zone ID usages? could they conflict?

docs/cluster_spec.md
```diff
@@ -303,6 +303,12 @@ FoundationDBClusterStatus defines the observed state of FoundationDBCluster
 
 [Back to TOC](#table-of-contents)
 
+## Host
+
+Host represents the host of a process group
```
nicmorales9: I would use at least one different word in the definition, e.g. `Host represents the Node (machine) that the process group is running on`

nicmorales9 (approved):
Some comments, but functionality LGTM

api/v1beta2/applyconfiguration/api/v1beta2/processgroupstatus.go
johscheuer: I don't know, those are auto-generated :)

controllers/update_status.go
johscheuer: `zone` is used for the fault domain calculation in FDB and those usages shouldn't conflict. The comment was more about if the `zoneid` is present, the localities of the process are present and therefore the `machineid` should also exist.

PR #2539, opened 2026-08-11
Initial set of backup metrics in the operator

johscheuer:
# Description

Initial set of backup metrics in the operator to improve observability of backups made with the operator.

## Type of change

## Discussion

I think the initial set should be fine and we can extend this set of metrics in the future.

## Testing

Ran unit tests locally and CI will run e2e tests.

## Documentation

The new fields are documented, otherwise nothing specific to document as we use the same metrics endpoint.

## Follow-up

-


internal/metrics/metrics.go
```diff
@@ -108,6 +110,81 @@ var (
 		append(descClusterDefaultLabels, "process_class"),
 		nil,
 	)
+
+	descBackupDefaultLabels = []string{"namespace", "name"}
+
+	// DescBackupRunning stores whether the backup is currently running.
+	DescBackupRunning = prometheus.NewDesc(
```
nicmorales9, line 117: would a single state gauge metric with "Running" or "paused" labels make sense here? I don't see where these metrics are added (presumedly beforehand?)

internal/metrics/metrics.go
```diff
@@ -108,6 +110,81 @@ var (
...
+	DescBackupDesiredAgentCount = prometheus.NewDesc(
+		"fdb_operator_backup_desired_agent_count",
+		"the desired number of Fdb backup agents.",
+		descBackupDefaultLabels,
+		nil,
+	)
+
+	// descBackupLastReconciled stores the latest generation that was reconciled for the backup.
+	descBackupLastReconciled = prometheus.NewDesc(
+		"fdb_operator_backup_latest_reconciled_status",
```
nicmorales9, line 166:
```suggestion
		"fdb_operator_backup_latest_reconciled_generation",
```

PR #2541, opened 2026-08-21
Implement single cluster recovery command

johscheuer:
# Description

Implement single cluster recovery command for cases where a majority of coordinators are lost.

## Type of change

## Discussion

This is a breaking change as the recovery commands are put under one subcommand.

## Testing

Added e2e tests, but will do some additional validation.

## Documentation

Will be updated.

## Follow-up

-

### DIFF
--- a/e2e/fixtures/fdb_cluster.go
+++ b/e2e/fixtures/fdb_cluster.go
@@ -42,7 +42,7 @@ import (
 	"github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
-	kubeErrors "k8s.io/apimachinery/pkg/api/errors"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -829,6 +829,58 @@ func (fdbCluster *FdbCluster) SetPodAsUnschedulable(ctx context.Context, pod cor
 	}).WithTimeout(5*time.Minute).WithPolling(2*time.Second).MustPassRepeatedly(5).Should(gomega.BeEmpty(), "Not able to set pod as unschedulable")
 }
 
+// SetPodsAsUnschedulable sets the provided slice of Pods on the NoSchedule list of the current FoundationDBCluster. This will make
+// sure that the Pods are stuck in Pending.
+func (fdbCluster *FdbCluster) SetPodsAsUnschedulable(ctx context.Context, pods []corev1.Pod) {
+	unschedulableProcessGroups := make([]fdbv1beta2.ProcessGroupID, 0, len(pods))
+
+	for _, pod := range pods {
+		unschedulableProcessGroups = append(unschedulableProcessGroups, GetProcessGroupID(pod))
+	}
+
+	fdbCluster.SetProcessGroupsAsUnschedulable(
+		ctx,
+		unschedulableProcessGroups,
+	)
+
+	// context aware.
+	time.Sleep(5 * time.Second)
+
+	for _, pod := range pods {
+		fetchedPod := &corev1.Pod{}
+		err := fdbCluster.getClient().
+			Get(ctx, client.ObjectKeyFromObject(&pod), fetchedPod)
+		if err != nil {
+			continue
+		}
+		// check if error is absent -> case not created, just deleted
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+		// Try deleting the Pod as a workaround until the operator handles all cases.
+		if fetchedPod.Spec.NodeName != "" && fetchedPod.DeletionTimestamp.IsZero() {
+			gomega.Expect(fdbCluster.getClient().Delete(ctx, &pod)).
+				NotTo(gomega.HaveOccurred())
+		}
+	}
+
+	gomega.Eventually(func(g gomega.Gomega) {
+		for _, pod := range pods {
+			fetchedPod := &corev1.Pod{}
+			err := fdbCluster.getClient().
+				Get(ctx, client.ObjectKeyFromObject(&pod), fetchedPod)
+			g.Expect(err).NotTo(gomega.HaveOccurred())
+
+			// Try deleting the Pod as a workaround until the operator handles all cases.
+			if fetchedPod.Spec.NodeName != "" && fetchedPod.DeletionTimestamp.IsZero() {
+				g.Expect(fdbCluster.getClient().Delete(ctx, &pod)).
+					NotTo(gomega.HaveOccurred())
+			}
+
+			g.Expect(fetchedPod.Spec.NodeName).To(gomega.BeEmpty())
+		}
+	}).WithTimeout(5*time.Minute).WithPolling(2*time.Second).MustPassRepeatedly(5).Should(gomega.Succeed(), "Not able to set pods as unschedulable")
+}
+
 // SetProcessGroupsAsUnschedulable sets the provided process groups on the NoSchedule list of the current FoundationDBCluster. This will make
 // sure that the Pod is stuck in Pending.
 func (fdbCluster *FdbCluster) SetProcessGroupsAsUnschedulable(
@@ -980,7 +1032,7 @@ func (fdbCluster *FdbCluster) WaitForPodRemoval(ctx context.Context, pod *corev1
 	gomega.Eventually(func() bool {
 		err := fdbCluster.getClient().
 			Get(ctx, client.ObjectKeyFromObject(pod), fetchedPod)
-		if err != nil && kubeErrors.IsNotFound(err) {
+		if err != nil && k8serrors.IsNotFound(err) {
 			return true
 		}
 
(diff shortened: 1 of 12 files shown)


e2e/fixtures/fdb_cluster.go
```diff
@@ -829,6 +829,58 @@ func (fdbCluster *FdbCluster) SetPodAsUnschedulable(ctx context.Context, pod cor
...
+	for _, pod := range pods {
+		unschedulableProcessGroups = append(unschedulableProcessGroups, GetProcessGroupID(pod))
+	}
+
+	fdbCluster.SetProcessGroupsAsUnschedulable(
+		ctx,
+		unschedulableProcessGroups,
+	)
+
+	// context aware.
```
nicmorales9, line 846: what does this mean?

e2e/fixtures/fdb_cluster.go
```diff
@@ -829,6 +829,58 @@ func (fdbCluster *FdbCluster) SetPodAsUnschedulable(ctx context.Context, pod cor
...
+	time.Sleep(5 * time.Second)
+
+	for _, pod := range pods {
+		fetchedPod := &corev1.Pod{}
+		err := fdbCluster.getClient().
+			Get(ctx, client.ObjectKeyFromObject(&pod), fetchedPod)
+		if err != nil {
+			continue
+		}
+		// check if error is absent -> case not created, just deleted
```
nicmorales9, line 856: I don't follow what is meant to happening here or what this comment is saying; the error is checked to be nil already from the `if err != nil continue` above

e2e/fixtures/fdb_cluster.go
```diff
@@ -829,6 +829,58 @@ func (fdbCluster *FdbCluster) SetPodAsUnschedulable(ctx context.Context, pod cor
...
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+		// Try deleting the Pod as a workaround until the operator handles all cases.
+		if fetchedPod.Spec.NodeName != "" && fetchedPod.DeletionTimestamp.IsZero() {
```
nicmorales9, line 860: Might be good to note this behavior in the docstring, wouldn't guess it from the name (but I also don't have the best idea of what adding to the unschedulable list does)

e2e/test_operator_plugin/operator_plugin_test.go
```diff
@@ -277,14 +299,141 @@ var _ = Describe("Operator Plugin", Label("e2e", "pr"), func() {
...
+		var fdbCluster *fixtures.FdbCluster
 
 		AfterEach(func(ctx SpecContext) {
-			log.Println("Recreate cluster")
-			// Delete the broken cluster.
-			factory.Shutdown(ctx)
-			// Recreate the cluster to make sure  the next tests can proceed
-			factory = fixtures.CreateFactory(testOptions)
-			fdbCluster = factory.CreateFdbHaCluster(ctx, clusterConfig)
+			// Delete the cluster.
```
nicmorales9, line 309: nit: comment is not adding anything now that we delete with a more obvious function on the cluster

e2e/test_operator_plugin/operator_plugin_test.go
```diff
@@ -277,14 +299,141 @@ var _ = Describe("Operator Plugin", Label("e2e", "pr"), func() {
...
+
+				downCoordinators = append(downCoordinators, coordinator)
+			}
+
+			// Set those coordinators as unschedulable to simulate that those coordinators are down. Another option
+			// would be to create a network partition.
+			fdbCluster.SetPodsAsUnschedulable(ctx, downCoordinators)
+		})
+
+		// Default case is to run with DNS enabled. The test case with IPs enabled can run into issues when
```
nicmorales9, line 333: is there some way to check for the condition and skip instead? I'm probably missing something about the setup / teardown order though 😄

e2e/test_operator_plugin/operator_plugin_test.go
```diff
@@ -277,14 +299,141 @@ var _ = Describe("Operator Plugin", Label("e2e", "pr"), func() {
...
+				}
+			})
+		})
+
+		When("DNS names in the cluster file are used", func() {
+			BeforeEach(func(_ SpecContext) {
+				useDNS = true
+			})
+
+			It("should recover the coordinators", func(ctx SpecContext) {
```
nicmorales9, line 395: this is the same code as above right? is there some way to functionlize / table test this? (no worries if not)

kubectl-fdb/cmd/recover.go
```diff
@@ -0,0 +1,383 @@
...
+
+		time.Sleep(5 * time.Second)
+	}
+
+	// If DNS is used for the cluster file, we could hit cases where no DNS entry can be resolved, in this case we could
+	// assume that the cluster is also down. The error from the client side is the following:
+	//  Error: error getting status: Error determining public address.
+	//  ERROR: Unable to bind to network (1512)
+	if err != nil && strings.Contains(err.Error(), "Error determining public address") {
+		return err
```
nicmorales9, line 355: why is this a conditional if we return any non nil error anyways? did we mean to log something? (possibly related to confusion over return logic)

kubectl-fdb/cmd/recover.go
```diff
@@ -0,0 +1,383 @@
...
+			}
+
+			return err
+		}
+	}
+
+	return nil
+}
+
+func checkIfClusterIsUnavailableAndMajorityOfCoordinatorsAreUnreachable(
```
nicmorales9, line 324: nit: would be good to have a docstring, esp that describes the error return logic

kubectl-fdb/cmd/recover.go
```diff
@@ -0,0 +1,383 @@
...
+	return err
+}
+
+func getStatusAndCheckIfClusterShouldBeRecovered(ctx context.Context,
+	kubeClient client.Client,
+	config *rest.Config,
+	clientPod *corev1.Pod) error {
+	status, err := getStatus(ctx, kubeClient, config, clientPod)
+	if err != nil {
+		return err
```
nicmorales9, line 367: (I am aware that this code is also predating this change, but noticed this)
ok if I follow right, this function should return an error if the cluster is healthy / doesn't need recovery, but wont getStatus fail for possible other reasons? Possibly more likely in the case of network partitions / issues? separate error seems better to me

kubectl-fdb/cmd/recover_single_dc_cluster.go
```diff
@@ -0,0 +1,465 @@
...
+
+			excludedCoordinators, err := cmd.Flags().GetStringArray("exclude-coordinator")
+			if err != nil {
+				return err
+			}
+
+			if wait {
+				confirmed := confirmAction(
+					fmt.Sprintf(
+						"WARNING:\nThe cluster: %s/%s will be force recovered.\nOnly perform those steps if you are unable to recover the coordinator pods.\nPerforming this action could lead to data loss.\n At leas one coordinator must be active and running to copy the coordinator state.\n",
```
nicmorales9, line 87:
```suggestion
						"WARNING:\nThe cluster: %s/%s will be force recovered.\nOnly perform those steps if you are unable to recover the coordinator pods.\nPerforming this action could lead to data loss.\n At least one coordinator must be active and running to copy the coordinator state.\n",
```

kubectl-fdb/cmd/recover_single_dc_cluster.go
```diff
@@ -0,0 +1,465 @@
...
+		Example: `
+# Recover the single dc cluster "sample-cluster-1" in the current Namespace
+kubectl fdb recover single-dc sample-cluster-1
+
+# Recover the single-dc cluster "sample-cluster-1" in the "testing" Namespace
+kubectl fdb recover single-dc -n testing sample-cluster-1
+
+# Recover the single-dc cluster "sample-cluster-1" in the "testing" Namespace and excluding the coordinator on pod
+# sample-cluster-1-storage-42
+kubectl fdb recover single-dc -n testing sample-cluster-1 --exclude-coordinator sample-cluster-1-storage-42
```
nicmorales9, line 117: (only skimmed below so far but) we're not actually FDB "exclude"ing the coordinators, it might be worth renaming this to skip-as-coordinator / ineligible-as-coordinator or something like that to try to avoid overloaded terms
(I'm aware recovery is also overloaded 🥲 )

nicmorales9 (approved):
bag-o-nits but it looks like a good improvement and I don't have any serious concerns

e2e/fixtures/fdb_cluster.go
johscheuer: That was a reminder for me to use a switch here and watch if the context was canceled during the sleep.

e2e/test_operator_plugin/operator_plugin_test.go
johscheuer: What would be the benefit of a different logic? Right now the case for non-DNS is skipped as well (marked as pending)?

kubectl-fdb/cmd/recover.go
johscheuer: Good catch, that was a leftover from the refactoring :)

kubectl-fdb/cmd/recover.go
johscheuer: That could be the case yes, this is only a safeguard to reduce the risk of running those commands against healthy clusters. There is no guarantee that this will prevent the user running the command against a healthy cluster, e.g. there could be the case where a user starts the command and after this check the cluster gets healthy.

PR #2549, opened 2026-09-10
Three data hall fallback support in operator

matiascognite:
# Description
Add support for redundancy mode three_data_hall_fallback in operator

## Disclaimer
Note: It's a bit premature still - haven't looked at e2e tests, which is also pending the discussions below.
Also, this is my first contribution to the repo, and the first contribution in go, I appreciate patience and feedback!

## Type of change

#### MinimumFaultDomains
Should this be 2, the number of replicas of the data?
Or should we keep it at 3, the number of fault domains, but tolerate a single unavailability?

#### Coordinators
three_data_hall has 9 coordinators, 3 in each data hall. To be able to lose a data hall + a single pod without going down. How should we handle it in fallback mode?

- Maintain 9 coordinators, but allow up to 5 in a single data hall, allowing a 5/4/0 split. But if the 3rd hall comes up and the first goes down, will the operator re-balance fast enough, or do we risk an outage with only 4 coordinators remaining? And can we be sure there are 9 pods left in the cluster?
- Maintain 9 coordinators AND max 3 in a single data hall. Meaning that some of them may be unavailable and unscheduleable. Would the operator need to partially recruit new coordinators, if a data hall is down and one of the 6 available coordinators goes down?
- Fall back to the default faultTolerance+faultDomains = 3 coordinators. In which case we'll for certain have enough coordinators, but it'd be worse for fault tolerance.

### Operator constraints during unhealthy states
During manual testing, I discovered that the operator will not change the redundancy mode to three_data_hall_fallback if the cluster is already unhealthy. It requires a manual fdbcli + configure three_data_hall_fallback execution alongside the manifest update. How should we handle/allow configurations when the cluster is actively degraded?

### Transitioning back to three_data_hall
If the manifest says three_data_hall, should we
- Wait until the cluster can safely transition to three_data_hall
- Wait until the cluster is fallback healthy (current PR behaviour)

If we keep current behaviour, the user must manually check the cluster before reverting the manifest, to prevent a transition when a data hall is still missing.

### Code for single zone
three_data_hall has code that handles setup of a new cluster, with relaxed restrictions. Not sure whether we need to deal with that for three_data_hall_fallback, given that it's not designed to be used during normal operations, only as an emergency fallback.

### Unit tests
Added unit tests. Mostly copy paste of three_data_hall unit tests, modified to the constraints of three_data_hall_fallback.

### Manual testing
Manual testing on a running cluster.
#### Setup
Test setup was having a normal three_data_hall cluster, then modify the nodeselector on one of the data_halls, making the pods unschedulable, to simulate a data_hall outage.
#### Result
It mostly worked well. The only part is that changing to three_data_hall_fallback required me to first change redundancy_mode to three_data_hall_fallback, and then fdbcli/configure three_data_hall_fallback.

Doing configure only, caused the cluster to switch back after some minutes, when it considered itself healthy enough.
Doing the manifest only prevented the operator from doing the change. Relevant logs:
{"level":"info","ts":"2026-09-10T14:00:22Z","logger":"controller","msg":"Changing current configuration is not safe","namespace":"timeseries","cluster":"foundationdb-cluster","traceID":"bb1416ae-d1df-4266-87d2-a61024f4bf8a","reconciler":"controllers.updateDatabaseConfiguration","error":"data distribution is not healthy: healing","current configuration":{"redundancy_mode":"three_data_hall","storage_engine":"ssd-redwood-1","usable_regions":1,"logs":9,"commit_proxies":2,"grv_proxies":1,"resolvers":1,"log_routers":-1,"remote_logs":-1},"desired configuration":{"redundancy_mode":"three_data_hall_fallback","storage_engine":"ssd-redwood-1","usable_regions":1,"logs":9,"commit_proxies":2,"grv_proxies":1,"resolvers":1,"log_routers":-1,"remote_logs":-1}}

Once the cluster was in three_data_hall_fallback, it appeared to work well, both with respect to scaling up and down, and recovery from pods/coordinators being deleted.

### Additional testing
I think it could be nice to have some e2e tests, but I don't know how to best set them up, and what to test for.

## Documentation

No. Partially because I don't know how three_data_hall_fallback needs to behave.

I think it would be nice to do that.

N/A

N/A?

## Follow-up

No?
No


johscheuer (commented):
Thanks for the PR! I'll try to take a closer look at this PR by next week.

Issue #2459, opened 2026-04-17
operator created backup pods do not have the `foundationdb.org/fdb-cluster-name` label


Pranav2612000:
### What happened?

- deployed some backup pods using the operator
- these pods do not have the built in metadata labels for e.g. `foundationdb.org/fdb-cluster-name`, 

### What did you expect to happen?

According to https://github.com/FoundationDB/fdb-kubernetes-operator/blob/main/docs/manual/resources.md#built-in-metadata,

all resources created by the operator should have these metadata labels by default. Further the annotations mentioned above are also missing

### How can we reproduce it (as minimally and precisely as possible)?

backup.yml
```
apiVersion: apps.foundationdb.org/v1beta2
kind: FoundationDBBackup
metadata:
  name: ${FDB_CLUSTER_NAME}
  namespace: ${FDB_NAMESPACE}
spec:
  version: 7.3.69
  clusterName: ${FDB_CLUSTER_NAME}
  backupMode: oneTime
  backupType: unmanaged
  blobStoreConfiguration:
    accountName: "@s3.us-east-1.amazonaws.com"
    bucket: ${FDB_BACKUP_BUCKET_NAME}
    backupName: default-backup-name 
    urlParameters:
      - "secure_connection=0"
      - "sc=0"
  podTemplateSpec:
    spec:
      volumes:
        - name: backup-credentials
          secret:
            secretName: backup-credentials
      containers:
        - env:
            - name: FDB_BLOB_CREDENTIALS
              value: /var/backup-credentials/credentials
          name: foundationdb
          resources:
            limits:
              cpu: 250m
              memory: 128Mi
            requests:
              cpu: 250m
              memory: 128Mi
          securityContext:
            runAsGroup: 0
            runAsUser: 0
          volumeMounts:
            - mountPath: /var/backup-credentials
              name: backup-credentials
```

### FDB Kubernetes operator

<details>

```console
$ kubectl fdb version
# paste output here
```

</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Cloud provider

<details>

</details>

PR #2510, opened 2026-06-22
fix: add fdb-cluster-name label to backup deployment and pod template

AruneshDwivedi:
Backup deployments and their pod templates were missing the `foundationdb.org/fdb-cluster-name` label. According to the docs, all resources created by the operator should have this label.

Added the label to both the deployment metadata and the pod template metadata using `backup.Spec.ClusterName`.

PR #2551, opened 2026-09-19
fix: add fdb-cluster-name label to backup deployment and pod template

AruneshDwivedi:
Add the fdb-cluster-name label to backup deployments and pod templates so they are discoverable by label selectors. Update test assertions to account for the new label. Addresses #2459.

```release-notes
[BUGFIX] backup: add fdb-cluster-name label to backup deployment and pod template
```


johscheuer (commented):
How is this different from https://github.com/FoundationDB/fdb-kubernetes-operator/pull/2510?

